PR #34, opened 2014-01-26
Speed up packet processing for Matroska demuxed subs

ghost:
**Not for merge, discussion only!**

Processing packets from Matroska demuxed subs with ass_process_chunk() can be pretty slow. It uses an O(n) algorithm to check for packet duplicates. Likewise, every ass_render_frame() call must go through the whole event list.

This pull requests tries to fix this.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -54,7 +54,7 @@ void ass_free_track(ASS_Track *track)
     int i;
 
     if (track->parser_priv) {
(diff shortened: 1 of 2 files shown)


astiob:
It seems this landed in master, in a modified form, in commit de4ea4bbd1c5101bf0b572a4facd58875845f495.

PR #35, opened 2014-01-26
Special case modern ASS parsing

ghost:
**Not for merge, discussion only!**

Checking the `Format:` line introduces slowdown.

Old code path kept for now.


TheOneric:
See: https://github.com/libass/libass/pull/469#issuecomment-1181845457

Issue #46, opened 2014-02-12
Banner effect breaks newlines


astiob:
The banner effect should not break `\N`. Multiple lines are still allowed, whitespace on each line is still trimmed, and lines are still aligned. Wrapping style always defaults to 2 (no automatic wrapping), but it can still be overridden with `{\q}`.

astiob:
Fixed by #434.

PR #47, opened 2014-02-12
Trim trailing whitespace from Dialogue lines early

astiob:
**Not for merge yet (probably).**

Match VSFilter. In particular, trailing `\N` (with any non-zero number of spaces) no longer produces an empty line. VSFilter trims both trailing and leading whitespace early, but trimming leading whitespace late presents no such problem.

After #46 is fixed, I think `x->symbol == '\n'` can be removed from `IS_WHITESPACE` and the trailing whitespace trimming loop removed from `trim_whitespace`.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -354,12 +354,12 @@ static int process_event_tail(ASS_Track *track, ASS_Event *event,
     while (1) {
         NEXT(q, tname);
         if (ass_strcasecmp(tname, "Text") == 0) {
-            char *last;
             event->Text = strdup(p);
             if (*event->Text != 0) {
-                last = event->Text + strlen(event->Text) - 1;
-                if (last >= event->Text && *last == '\r')
-                    *last = 0;
+                char *end = event->Text + strlen(event->Text);
+                while (end > event->Text &&
+                       (end[-1] == '\r' || end[-1] == '\t' || end[-1] == ' '))
+                    *--end = 0;
             }
             event->Duration -= event->Start;
             free(format);


TheOneric:
@astiob: do you remember why this was never merged? The functionality seems innocuous enough to me and there are no (unresolved) concerns documented here *(maybe some were raised on IRC?)*
Though, I haven't checked VSFilter myself and am also not sure what *“After https://github.com/libass/libass/issues/46 is fixed, I think x->symbol == '\n' can be removed from IS_WHITESPACE and the trailing whitespace trimming loop removed from trim_whitespace.”* refers to; there appears to also have been no `IS_WHITESPACE` in the initial version of the patch.

astiob:
I’m afraid I don’t immediately remember why this wasn’t ready for merge; and I suspect I may have run my own local libass build with this patch applied (among others) for a few years without any problems.

But as for `IS_WHITESPACE`, I think I was referring to the pre-existing macro used in `trim_whitespace`:

https://github.com/libass/libass/blob/5b0dba4b012f8aa8231125a78fec6b8aabee8bdf/libass/ass_render.c#L1581-L1587

as in, _beyond_ or possibly _instead of_ this patch, `IS_WHITESPACE` and `trim_whitespace` should also have been adjusted once that adjustment stopped breaking something else.

TheOneric:
> as in, *beyond* or possibly *instead* of this patch, `IS_WHITESPACE` and `trim_whitespace` should also have been adjusted

I don't immediately see how dropping `\n` from `IS_WHITESPACE` could replace this patch. Can you explain, or were you thinking of further adjustments to `IS_WHITESPACE` and `trim_whitespace`?

astiob:
Just for posterity: having this applied seems good to me, thanks. 👍 If/when we remember what the remaining issues were, we’ll just tackle them then. I should’ve documented them better, but oh well. (Hm, maybe I should try digging into my IRC archives?)

> Can you explain, or were you thinking of further adjustments to `IS_WHITESPACE` and `trim_whitespace`?

Yes, possibly. I’m not sure what I actually meant in the OP, but this is one possible interpretation that I see. But you say it’s neither sufficient nor safe, so whatever.

PR #84, opened 2014-04-27
MSVC/ICL and general header fixes

astiob:
This doesn’t make our headers totally sane, but it’s a little bit better.

(This is currently open for discussion and liable to change significantly.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,7 +16,7 @@ AM_PROG_AS
 # Checks for header files.
 AC_HEADER_STDC
 AC_HEADER_STDBOOL
-AC_CHECK_HEADERS([stdint.h unistd.h iconv.h])
+AC_CHECK_HEADERS([stdint.h iconv.h])
 
 # Checks for library functions.
 AC_CHECK_FUNCS([strdup strndup])
--- a/libass/Makefile.am
+++ b/libass/Makefile.am
@@ -30,7 +30,7 @@ libass_la_SOURCES = ass.c ass_cache.c ass_font.c ass_fontselect.c ass_render.c \
                     ass_drawing.h ass_cache_template.h ass_render.h \
                     ass_parse.c ass_parse.h ass_render_api.c ass_shaper.c \
                     ass_shaper.h ass_strtod.c ass_fontconfig.c ass_fontconfig.h \
-                    ass_string.h ass_string.c
+                    ass_string.h ass_string.c ass_compat.h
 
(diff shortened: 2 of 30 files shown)


astiob:
Reopening for my take on this; I hope you don’t mind. It’s not much different from #191 and #196, but I think this is cleaner.

I also hope you don’t mind the Signed-off-by line; I’m happy to change or remove it if you do.

astiob:
Ugh. I thought GitHub retained old comments under a “click to view outdated comments” button, but I don’t see it anywhere. :( I was afraid of killing the comments but thought I wouldn’t—but it appears I did. Here’s the next best thing, a link to the old branch where the comments can be viewed separately from the pull request: https://github.com/libass/libass/compare/master...astiob:48d9a63

astiob:
Being somewhat anal about GitHub pull request metadata, I think we should do the following:

1. Pull the first two commits from here, because there seem to be no objections to them. The only thing to note is that #191 squashed the unistd.h commit into the ass_compat.h commit, but I think they should be kept separate.
2. Update #196 to be the next two commits.
3. Pull the two commits from #196.
4. Update this PR to be the last two commits. Discuss them some more, edit them, nuke them if need be.
5. If there’s anything left in this PR (both commits don’t get nuked), merge it when we’re happy with it.

torque:
I've updated #196 to be just my two commits in accordance with the plan laid out here.

grigorig:
Maybe config.h (and ass_compat.h) should actually be included in all headers and all source files. Else #ifdfef checks or macros (like __inline) cannot be resolved in headers, but that might be needed. Particularly, static inline functions come to my mind.

astiob:
But that’s why config.h and ass_compat.h are included in source files before any other header is. Of course, including them everywhere works too and I don’t really object to it, just want to point out that it’s not strictly needed. Mainly, I’d just like consistency.

Speaking of consistency, the several most recently added header files have include guard macro names starting with `ASS_` while all the rest have `LIBASS_`. They also include other headers before their include guards while all the rest do it after. This is really minor, but perhaps we should fix these discrepancies too.

By the way, we could make use of precompiled headers. Not sure it would give us much in way of build speed (we’re pretty fast to build anyway), but it might make maintaining the network of headers easier. Or it might not. But we could, say, stuff all of our headers into a single precompiled header and just include that everywhere. We could also do the same without the precompiled bit, but using it would ensure speed don’t suffers—but then we’re pretty fast to build anyway, so maybe it doesn’t matter.

grigorig:
Ah, sorry, minor brainfart here. Surely libass headers are only used in libass source files, so there is no problem. :) Go ahead then.

PR #107, opened 2014-06-01
Render events on separate threads, using a pthread pool

rcombs:
_FOR DISCUSSION ONLY_ right now, but hopefully rather close to being squash-and-merge-ready.
TODO:
- [ ] Add CPU count detection
- [ ] Valgrind the shit out of everything
- [ ] Deal with fribidi being non-thread-safe
  - Switch to minibidi?
  - Make people compile fribidi with `--enable-malloc`?
  - Put a mutex on fribidi?
- [ ] Figure out exactly how Windows support will work

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,3 @@ Makefile.in
 /missing
 /compile
 /libass.pc
-/m4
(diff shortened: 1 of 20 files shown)


grigorig:
Is this branch still considered alive?

wiiaboo:
The fribidi checkbox is probably checked since 0.19.7 always uses malloc.

rcombs:
@grigorig: it's been bit-rotting for a while, and at this point I think it'd be best to wait until we have a proper async API with refcounted bitmaps and build off of that. There's still some stuff here that's worth keeping around for reference next time we take a stab at it, though.
@wiiaboo: yup, so we can either require 0.19.7+ or gate the mutex code on `FRIBIDI_VERSION`.

wm4:
Note that MrSmile has a PR (cache cleanup) that leads towards refcounting.

rcombs:
This is extremely obsolete and due for a complete rewrite.

PR #120, opened 2014-07-19
fixed wrong frame top to clip

taxigps:
this bug make the subtitle invisible when ass_set_line_position set line_position to top(100.0) and ass_set_use_margins set as true(1)


astiob:
I think the reporter’s actual issue should now be fixed due to #385, which ensures text subtitles (the only ones line_position applies to) are clipped only against the whole screen when use_margins is true.

PR #149, opened 2014-11-22
Corrected problems reported by scan-build

rcombs:
Additionally, added ASS_WARN_UNUSED macro and checked some more mallocs.


grigorig:
Is this still relevant? I know the dead assignments are already fixed.

Issue #153, opened 2014-12-27
 render result with @fontname not right 


SuJiKiNen:
~~~
[Script Info]
; Script generated by Aegisub 3.2.1
; http://www.aegisub.org/
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
YCbCr Matrix: None
PlayResX: 1280
PlayResY: 720

[Aegisub Project Garbage]
Last Style Storage: Default
Video File: ?dummy:23.976000:40000:1280:720:47:163:254:
Video AR Value: 1.777778
Video Zoom Percent: 0.500000

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: JP2-furigana,Windows TV MaruGothic,24,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,10,0,1,0,0,2,10,10,15,1
Style: JP-furigana,@Windows TV MaruGothic,24,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,10,0,1,0,0,2,10,10,15,1
Style: JP,@Windows TV MaruGothic,48,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,10,0,1,0,0,2,10,10,15,1
Style: JP2,Windows TV MaruGothic,48,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,10,0,1,0,0,2,10,10,15,1

[Events]
Format: Layer, Start, End, Style, Name, MarginL, MarginR, MarginV, Effect, Text
Dialogue: 0,0:00:00.00,0:00:05.00,JP,,0,0,0,,你
Dialogue: 0,0:00:00.00,0:00:05.00,JP2,,0,0,0,,你
~~~

![dummy_001_0](https://cloud.githubusercontent.com/assets/6788956/5560792/a212fe90-8de3-11e4-8393-9fcd6daba221.png)

one is right the other is wrong,the have same result.

astiob:
Apparently the issue is that we don’t distinguish combined bitmaps by font verticality in the cache.

Issue #82, opened 2014-04-27
Cacheing thing, maybe?


rcombs:
![](http://puu.sh/8p5kh.png)
[D84BC791]
http://puu.sh/8p61R.ttf
```
Style: Mr. T(ea),Painted,60,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,0,0,5,10,10,10,1
```
```
Dialogue: 2,0:08:39.70,0:08:41.33,Mr. T(ea),,0,0,0,,{\alpha&HFF&\shad1.2\blur1\fnbrushtipTravis_tcn'd\b1\c&H1B3052&\4c&H7EB7F6&\fscx101\fscy101\pos(1115.17,-5.74)\4a&H00&}Amau{\fsp-2.8}f{\fsp\c&H3F505F&}a{*\c&H3F4F5D&\fnPainted}-{*\alpha&H00&\fnbrushtipTravis_tcn'd\c&H192F53&}a{\c&H192F53&}n
Dialogue: 1,0:08:39.70,0:08:41.33,Mr. T(ea),,0,0,0,,{\alpha&HFF&\shad1.2\blur1\fnbrushtipTravis_tcn'd\b1\c&H1B3052&\4c&H7EB7F6&\fscx101\fscy101\pos(1120.97,-1.74)\4a&H00&}Amau{\fsp-2.8}f{\fsp\c&H3F505F&}a{*\c&H1B3052&\fnPainted\alpha&H00&}-{*\alpha&HFF&\fnbrushtipTravis_tcn'd\c&H404F5B&}a{\c&H414F59&}n
Dialogue: 1,0:08:39.70,0:08:41.33,Mr. T(ea),,0,0,0,,{\shad1.2\blur1\fnbrushtipTravis_tcn'd\b1\c&H192F53&\4c&H7EB7F6&\fscx101\fscy101\pos(1117.97,-5.74)}Amau{\fsp-2.8}f{\fsp\c&H192F53&}a{*\c&H3F4F5D&\fnPainted\alpha&HFF&}-{*\fnbrushtipTravis_tcn'd\c&H192F53&}a{\c&H414F59&}n
```

astiob:
This seems to be the same as #153, namely, the combined bitmap cache doesn’t distinguish between any two bitmaps that appear to have all the same properties except the font.

PR #157, opened 2015-01-16
Clean up bitmap combining

MrSmile:
I thied to fix incorrect caching with different fonts but the same text line and got swallowed.
This is the result.

I hope it fixes at least #153.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -30,7 +30,7 @@
 #define MAX_GLYPHS_INITIAL 1024
 #define MAX_LINES_INITIAL 64
 #define MAX_BITMAPS_INITIAL 16
-#define MAX_STR_LENGTH_INITIAL 64
+#define MAX_SUB_BITMAPS_INITIAL 64
 #define SUBPIXEL_MASK 63
 #define SUBPIXEL_ACCURACY 7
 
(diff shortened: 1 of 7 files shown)


wm4:
That's a pretty big change. 

Hard to tell whether it's safe to apply.

I can't say much about it either....:
- Are we sure the way this hashes pointers is sane? I.e., do these pointers used as hash keys live at least as long as the corresponding cache entry? Otherwise, a new allocation (which has the same pointer, but is logically different) could match stale cache entries. 
- There are some ``xxx`` comments, which imply unfinished business.
- I'm seeing some unchecked allocations; unsure whether that is just moved code, or new code. 
- I find these ``++x`` (instead of ``x++``) utterly disgusting.

MrSmile:
I don't plan to apply it as is, that's something like working prototype.

* Pointers to cache values should be safe given right cache ordering at cleaning. libass already uses that in `OutlineBitmapHashKey` (pointer to `OutlineHashValue`) and `GlyphHashKey/GlyphMetricsHashKey` (pointer to `ASS_Font`).
* Yes, I mark places where additional thinking/discussion is needed.
* Hmm... that's `alloc/copy_bitmap()` in combining process? I'll fix that.
* I too prefer x++, but I was certain that ++x is libass codestyle.

wm4:
The changes for error handling look good IMO.

libass/ass_render.c
```diff
@@ -2203,244 +2134,192 @@ static void render_and_combine_glyphs(ASS_Renderer *render_priv,
...
-                current_info->cached = 0;
-                current_info->is_drawing = 0;
-
+                current_info->filter.be = info->be;
+                current_info->filter.blur = 2 * info->blur * render_priv->blur_scale;
+                current_info->filter.shadow.x = double_to_d6(info->shadow_x * render_priv->border_scale);
+                current_info->filter.shadow.y = double_to_d6(info->shadow_y * render_priv->border_scale);
+                current_info->filter.border_style = info->border_style;
+                current_info->filter.nonzero_border = info->border_x || info->border_y;
+                current_info->filter.show_shadow = info->shadow_x || info->shadow_y;
```
astiob: This is needed to ensure we show a shadow even if `shadow.x` or `shadow.y` underflow to zero because of the scaling, right? (And it matters because the glyphs can be translucent, making even a zero-offset shadow visible.)

Don’t you think it would be better to rename this to `nonzero_shadow`? This would add consistency with `nonzero_border` and reduce confusion with `draw_shadow`.

libass/ass_render.c
```diff
@@ -2203,244 +2134,192 @@ static void render_and_combine_glyphs(ASS_Renderer *render_priv,
...
-                min_bm_y = FFMIN(min_bm_y, bm_o_y);
-            }
+            info->pos.x = double_to_d6(device_x + d6_to_double(info->pos.x) * render_priv->font_scale_x);
+            info->pos.y = double_to_d6(device_y) + info->pos.y;
+            key->advance.x = info->pos.x & (SUBPIXEL_MASK & ~SUBPIXEL_ACCURACY);
+            key->advance.y = info->pos.y & (SUBPIXEL_MASK & ~SUBPIXEL_ACCURACY);
+            int x = info->pos.x >> 6, y = info->pos.y >> 6;
+            get_bitmap_glyph(render_priv, info);
+            if (!info->image)
+                continue;
```
astiob: This conditional should come after the following one, to ensure that we correctly split runs even when we fail to rasterize some glyphs.

libass/ass_render.h
```diff
@@ -107,48 +107,28 @@ typedef enum {
...
-    double scale_x, scale_y;
-    int border_style;
-    int has_border;
-    double border_x, border_y;
-    double hspacing;
-    unsigned italic;
-    unsigned bold;
-    int flags;
-    int shift_x, shift_y;
+    int first_pos_x;
```
astiob: It would be nice to separate this line from the preceding comment to reduce confusion, as the comment refers to `effect_timing`.

libass/ass_render.c
```diff
@@ -1826,38 +1760,24 @@ static int is_new_bm_run(GlyphInfo *info, GlyphInfo *last)
...
-        last->border_y != info->border_y ||
-        last->hspacing != info->hspacing ||
-        last->italic != info->italic ||
-        last->bold != info->bold ||
-        last->flags != info->flags){
-        return 1;
-    }
-    return 0;
+        (last->border_x || last->border_y) != (info->border_x || info->border_y) ||
+        !last->border != !info->border;
```
astiob: What’s up with all of this? `is_new_bm_run` must stay as it is for VSFilter compatibility. I think there may be one or two values where we still technically differ from VSFilter, but certainly the bulk of it matches exactly.

astiob:
Also please fix this comment in ass_cache.c:
```C
// create hash/compare functions for bitmap, outline and composite cache
```

astiob:
> Pointers to cache values should be safe given right cache ordering at cleaning.

Hmm. I’m looking at [`check_cache_limits`](https://github.com/MrSmile/libass/blob/faea30a32c9f9d648c9fa41d802f7a50c94fdc1a/libass/ass_render.c#L2599) and [`ass_cache_empty`](https://github.com/MrSmile/libass/blob/faea30a32c9f9d648c9fa41d802f7a50c94fdc1a/libass/ass_cache.c#L335)… Here’s how I’m reading the code:

1. If the bitmap cache is too large, empty it.
2. If the outline cache is too large, empty it.
3. If the composite cache is too large, empty it.

So, it can happen that the bitmap cache is emptied but the composite cache isn’t, which leaves it with dangling pointers. They aren’t dangerous per se as we never dereference them, but a newly allocated bitmap might get an address clashing with one that was previously used and is still referenced by some composite bitmap, leading to false composite cache hits.

> libass already uses that in OutlineBitmapHashKey (pointer to OutlineHashValue)

Does this mean the outline cache is already broken?

> and GlyphHashKey/GlyphMetricsHashKey (pointer to ASS_Font).

We only empty the font cache in `ass_renderer_done`.

libass/ass_render.c
```diff
@@ -2203,244 +2134,192 @@ static void render_and_combine_glyphs(ASS_Renderer *render_priv,
...
             }
-            cleanup:
-            info = info->next;
         }
     }
 
-    for (i = 0; i < nb_bitmaps; ++i) {
-        if(!combined_info[i].cached){
+    for (int i = 0; i < nb_bitmaps; ++i) {
+        if (combined_info[i].bitmaps) {  // not cached
```
astiob: Shouldn’t this be done in the same loop? Doing it in two separate loops can produce duplicates, can’t it?

astiob:
Other than this, very nice work!

libass/ass_render.c
```diff
@@ -2203,244 +2134,192 @@ static void render_and_combine_glyphs(ASS_Renderer *render_priv,
...
-            if(info->bm_o){
-                bm_o_x += info->bm_o->left;
-                bm_o_y += info->bm_o->top;
-                min_bm_x = FFMIN(min_bm_x, bm_o_x);
-                min_bm_y = FFMIN(min_bm_y, bm_o_y);
-            }
+            info->pos.x = double_to_d6(device_x + d6_to_double(info->pos.x) * render_priv->font_scale_x);
+            info->pos.y = double_to_d6(device_y) + info->pos.y;
+            key->advance.x = info->pos.x & (SUBPIXEL_MASK & ~SUBPIXEL_ACCURACY);
+            key->advance.y = info->pos.y & (SUBPIXEL_MASK & ~SUBPIXEL_ACCURACY);
```
astiob: Sweet!

astiob:
One related but probably separate thing I noticed: the cache size functions for the [bitmap](https://github.com/MrSmile/libass/blob/faea30a32c9f9d648c9fa41d802f7a50c94fdc1a/libass/ass_cache.c#L90) and [composite bitmap](https://github.com/MrSmile/libass/blob/faea30a32c9f9d648c9fa41d802f7a50c94fdc1a/libass/ass_cache.c#L138) caches seem pretty wrong.

libass/ass_render.c
```diff
@@ -2720,12 +2614,14 @@ void ass_free_images(ASS_Image *img)
 static void check_cache_limits(ASS_Renderer *priv, CacheStore *cache)
 {
     if (ass_cache_empty(cache->bitmap_cache, cache->bitmap_max_size)) {
+        ass_cache_empty(cache->composite_cache, 0);
         ass_free_images(priv->prev_images_root);
         priv->prev_images_root = 0;
         priv->cache_cleared = 1;
     }
     if (ass_cache_empty(cache->outline_cache, cache->glyph_max)) {
         ass_cache_empty(cache->bitmap_cache, 0);
```
astiob: Oh, I see! Outline cache invalidation already used to trigger a bitmap cache invalidation, so it isn’t broken!

astiob:
I approve! :+1: (Just need to squash down to a single commit before pushing.)

Might be worth checking if #153 really is fixed and adding a “Fixes #153.” or “Fixes libass#153.” paragraph to the commit message. Well, and generally some testing wouldn’t hurt, but the code seems all good now.

astiob:
We’ve established on IRC that this indeed fixes #153 and also #82.

@MrSmile I suggest the following procedure for pushing this to take full advantage of GitHub:

1. Combine the two commits and add the two “Fixes” lines (as a separate paragraph).
2. Push to your fork, so that this pull request is updated.
3. Push to libass, so that this pull request is automatically marked as merged and closed.

astiob:
Oh, we both missed/forgot about this:
> Also please fix this comment in ass_cache.c (L32):
> ```C
> // create hash/compare functions for bitmap, outline and composite cache
> ```

Issue #9, opened 2013-12-30
Implement fast floating-point(?) Gaussian blur using SIMD


rcombs:
Use xy-vsfilter's SSE2 implementation and Intel's AVX implementation when applicable. Write a NEON implementation for ARM, and possibly an SSE4-or-such implementation (but probably not :|). Probably can wait until after #5 is done.

astiob:
Huhh, this is rightfully closed because the thing in the issue title has been done, but the comments went off a tangent and are a goldmine.

I should probably repost those pictures I apparently posted. I do have them, but there’s _a lot_ and I don’t remember how I organized them when I uploaded them.

Anyway, libass’s and xy-VSFilter’s blur parameters are actually different? I sure didn’t remember that, and I do remember making it a point to preserve libass’s parameters when merging MrSmile’s blur. So the output looks different… yet oddly similar at the same time, because we also clip the kernels differently 🤯

PR #160, opened 2015-02-08
\be fixes including clipping and value range

astiob:
Of particular note: as discussed in #9,

* VSFilter has pixel values in the range from 0 to 64 inclusive, while libass uses the range from 0 to 255. Normally this doesn’t matter at all. For \blur, this merely means VSFilter’s has banding. But for \be with a large iteration count, the output is very different. This pull request makes libass use the range from 0 to 64 for all \be iterations but the last. The range conversions are correctly rounded (towards positive infinity where relevant). If the \be argument is 1 or 0, no range conversion takes place. The last \be pass is done on the full range to smooth out the output to make better use of the full range and hide whatever little banding was introduced by the reduced range.

* Blurs, both \blur and \be, cause bitmaps to expand. All renderers add some blur-dependent margins to bitmaps to accommodate this. However, the margins renderers add \be are too small. (In particular, traditional VSFilter always adds a single pixel*, presumably because \be originally did not support values bigger than 1.) libass already adds the largest margin of all renderers, but this pull request makes it always add just enough—and no more.

…Wait, my notes say there are some more edge cases (literally!) that are not handled correctly by our `be_blur`. I’ll double-check them.

\* And never writes to the said pixel during \be, only reads from it, so effectively it doesn’t add any margin at all.

### DIFF
--- a/libass/ass_bitmap.c
+++ b/libass/ass_bitmap.c
@@ -749,6 +749,35 @@ void be_blur_post(uint8_t *buf, intptr_t w, intptr_t h, intptr_t stride)
     }
 }
 
+/*
+ * To find these values, simulate blur on the border between two
+ * half-planes, one zero-filled (background) and the other filled
+ * with the maximum supported value (foreground). Keep incrementing
+ * the \be argument. The necessary padding is the distance by which
+ * the blurred foreground image extends beyond the original border
+ * and into the background. Initially it increases along with \be,
+ * but very soon it grinds to a halt. At some point, the blurred
+ * image actually reaches a stationary point and stays unchanged
+ * forever after, simply _shifting_ by one pixel for each \be
+ * step--moving in the direction of the non-zero half-plane and
+ * thus decreasing the necessary padding (although the large
+ * padding is still needed for intermediate results). In practice,
+ * images are finite rather than infinite like half-planes, but
+ * this can only decrease the required padding. Half-planes filled
+ * with extreme values are the theoretical limit of the worst case.
+ * Make sure to use the right pixel value range in the simulation!
+ */
+int be_padding(int be)
+{
+    if (be <= 3)
+        return be;
+    if (be <= 7)
+        return 4;
+    if (be <= 123)
+        return 5;
+    return FFMAX(128 - be, 0);
+}
+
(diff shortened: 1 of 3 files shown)


libass/ass_bitmap.c
```diff
@@ -679,32 +698,64 @@ void be_blur_c(uint8_t *buf, intptr_t w,
...
+ * but this can only decrease the required padding. Half-planes filled
+ * with extreme values are the theoretical limit of the worst case.
+ */
+int be_padding(int be)
+{
+    static const int limits[] = {0, 1, 2, 6, INT_MAX};
+    int i = 0;
+    while (be > limits[i])
+        i++;
+    return i;
```
astiob: If this code is deemed too complicated, we could use a constant radius that’s enough for any \be value ever. But doing so would increase our memory usage and slow down various operations dealing with bitmaps. (Potentially noticeable when there are large amounts of very small bitmaps.)

astiob:
@wm4’s feedback incorporated. Things double-checked and fixed. More things spotted and fixed. I think all that remains is updating the be_blur assembler code and possibly adding be_pre/be_post assembler code.

wm4:
Looks good to me, although I didn't look too closely.

grigorig:
LGTM, but I am not sure we really have to emulate the limited value range of VSFilter. Does that actually make a visible difference? I'd favor skipping that commit.

astiob:
Yes, it is very much visible. See the screenshots in [this](https://github.com/libass/libass/issues/9#issuecomment-37361115) comment.

grigorig:
LGTM. So let's just merge this now.

grigorig:
Turns out this was merged a bit too quickly in the end. The ASM needs to be changed according to the changes for the C version in the first commit. It would also be nice to have ASM versions of the pre/post passes.

astiob:
Sorry, that’s probably my fault for not making it clear that the ASM changes were missing.

Still, as has been said, nothing is really broken, so there is no need to revert. The only “broken” thing is that ASM \be is now slightly different from C \be because of the changes in the first commit. The ASM output is no worse than before (in fact, it’s _better_ due to the other commits), and the difference is fairly small and in most cases not easily noticeable.

PR #162, opened 2015-02-24
Make style overrides more fine grained

wm4:
Add some flags which allow the API user to control which fields exactly
are overridden, and which are taken from the script.

Sneaks in the Bold style too (probably was forgotten).

ASS_OVERRIDE_BIT_STYLE_ALL is a separate value, instead of an or-
combination of the other bits, so that no ABI change is required if
other override modes are added. (?)

All ASS_OVERRIDE_BIT_STYLE_* bits have effect only if
ASS_OVERRIDE_BIT_STYLE is set.

--------

Not really sure about this. I think it could be organized better. Thoughts/suggestions?

### DIFF
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -23,7 +23,7 @@
 #include <stdarg.h>
 #include "ass_types.h"
 
-#define LIBASS_VERSION 0x01201000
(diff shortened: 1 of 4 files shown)


wm4:
Merged.

PR #165, opened 2015-03-02
Third implementation of my cascade blur algorithm with full SSE2/AVX2 assembly

MrSmile:
Not for merge as is.

That should be faster than current algo even with disabled assembly, at least for large blur radii.
I considered squashing all into one commit but decided to leave it raw for the time being.
I'm open to suggestions about what parts worth keeping separate and to any other feedback.


haasn:
Tested this on a particularly problematic file I have ([Kaylith] Zankyou no Terror - 04 [720p][BEDCFB24].mkv) , and it is *significantly* faster than the standard libass - plays at realtime at 720p (where it was like <1 fps before), and only stutters a bit at 4K (where it would completely lock up for like a minute before)

haasn:
Any update on this? What's still missing?

rcombs:
Seems like the biggest concern about this at this point is that people would like better documentation of the algorithms used here and how they work, in plain English (probably in comments?); same for the rasterizer.

astiob:
As mentioned in the original post, another thing that needs to be done before this is pushed to master is a rebase to squash those fixup commits if nothing else.

I’ll also note that, while I haven’t personally tested this much (I probably should), we’ve received some responses on IRC and all of them have been entirely positive. (Correct me if I’m wrong.)

MrSmile:
I've squashed all commits into 3 weakly associated, so it should be more simply to review.

wm4:
First 2 commits look mostly ok to me. I made some comments. In general, the renaming and actual functional changes should probably have been separated for easier review, but maybe it's too late for that.

I mostly don't understand anything in the 3rd commit, so I guess I'm not qualified to review this. Though, I do hope the ``#if``s checking for the ``STRIPE_WIDTH`` value will go away to make the code a little bit cleaner.

PR #167, opened 2015-03-16
Margin fix

wm4:


wm4:
PS: I tried setting the evt_type to ``EVENT_POSITIONED`` for affected events, but this does not work at all. There is no position and the event still has to be placed using the standard rules.

I wonder if there's a less messy way to handle this.

Issue #154, opened 2014-12-27
Outlines are different with VSFilter


pingplug:
PlayResX and PlayResY are 848 and 480, rendered on 1280 x 720 video

Style: Default,Arial,31,&H00FFFFFF,&H00000000,&H00000000,&H00000000,-1,0,0,0,100,100,0,0,1,3,0,2,30,30,15,1
Dialogue: 0,0:00:00.00,1:00:00.00,Default,NTP,0000,0000,0000,,This is a test
Dialogue: 0,0:00:00.00,1:00:00.00,Default,NTP,0000,0000,0000,,This is a test

this is for libass
![libass](https://cloud.githubusercontent.com/assets/6712475/5561439/7a7f86de-8e13-11e4-9653-b780a6aecf94.png)
this is for VSFilter
![vsfilter](https://cloud.githubusercontent.com/assets/6712475/5561440/7e7de8ac-8e13-11e4-9457-45e631598dd3.png)
## 2015-01-03 update
1. the outlines are different(3px in VSFilter, 4.5px in libass)
   I think this is caused by the default ScaledBorderAndShow(1 in libass, 0 in VSFilter)
2. VSFilter add borders(top/bottom, = max border_y of the top/bottom line) between texts
   makes the collision text 6px(2 \* 3px border) higher than libass

CarVac:
I believe the issue may be that in libass, the border widths aren't scaling with the rendering resolution.

Playing back 480p video stretched to fit my 4k screen, the borders are way too tiny: they stay the same size even as the subs themselves change size.

grigorig:
It's actually the opposite. libass scales the borders with resolution by default, VSFilter does not.

astiob:
Let’s split this into #287 for `ScaledBorderAndShadow` and #304 for the space between events during collision detection.

PR #168, opened 2015-03-22
fix my issue

pingplug:
I tried to fix issue #154  
It seems work

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -1293,7 +1293,7 @@ ASS_Track *ass_new_track(ASS_Library *library)
     if (!track)
         return NULL;
     track->library = library;
-    track->ScaledBorderAndShadow = 1;
+    track->ScaledBorderAndShadow = 0;
     track->parser_priv = calloc(1, sizeof(ASS_ParserPriv));
     if (!track->parser_priv) {
         free(track);
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1448,18 +1448,22 @@ get_bitmap_glyph(ASS_Renderer *render_priv, GlyphInfo *info)
  *   lines[].height
  *   lines[].asc
  *   lines[].desc
+ *   lines[].top_border
+ *   lines[].bottom_border
  */
 static void measure_text(ASS_Renderer *render_priv)
 {
     TextInfo *text_info = &render_priv->text_info;
     int cur_line = 0;
-    double max_asc = 0., max_desc = 0.;
+    double max_asc = 0., max_desc = 0., max_border = 0.;
     GlyphInfo *last = NULL;
     int i;
     int empty_line = 1;
     text_info->height = 0.;
     for (i = 0; i < text_info->length + 1; ++i) {
         if ((i == text_info->length) || text_info->glyphs[i].linebreak) {
+            if (cur_line == 0)
+                text_info->top_border = max_border;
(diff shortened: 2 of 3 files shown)


wm4:
>change default ScaledBorderAndShadow to 1

I support this, even if it's not "compatible".

rcombs:
Just was looking at vsfilter source to see, and it's not, and whether we think it's a good idea in general or not, I find it weird to be regressing in a compatibility as part of a fix for a compatibility issue.

wm4:
Looking at the commit, your changes actually _disable_ ScaledBorderAndShadow, which conflicts with what the commit message says.

wm4:
Also be aware that you must call ``ass_set_storage_size()`` to the size of the video. (This can be a different value from ``ass_set_frame_size()`` if the subtitles are rendered into a video frame that has already be scaled, like for rendering in display resolution in a video player.)

pingplug:
sorry, I almost forgot this...
about the commit message... it's a typo
about the second one... in my MPC-HC, the space between texts is the same even the video is scaled (it's realy odd)

grigorig:
Yes, ScaledBorderAndShadow=1 may be a more sensible default. Can you explain the second commit a bit better, ideally in the commit message?

pingplug:
@grigorig 
If you put two collision texts with more than one line, the space between two texts is closer than the lines in the same text in libass
VSFilter add borders(top/bottom, = max border_y of the top/bottom line) between texts, when placing collision texts, the borders make a proper space between two texts

sorry for my poor english :P

pingplug:
rewrite commit message
commit 1 disable the ScaledBorderAndShadow by default like VSFilter does
dummy_000_0
before
dummy_001_0
after

commit 2 add a border between two collision texts, makes the space between two collision texts larger than space between two lines, Which is what VSFilter does.
dummy_002_0

ass file:
```
[V4+ Styles]
Style: Default,微软雅黑,31,&H00FFFFFF,&H00000000,&H00000000,&H00000000,-1,0,0,0,100,100,0,0,1,1,0,2,30,30,15,1

[Events]
Dialogue: 0,0:00:00.00,0:00:27.16,*Default,NTP,0,0,0,,Text1, line1 测试\NText1, line2 测试
Dialogue: 0,0:00:00.00,0:00:27.16,*Default,NTP,0,0,0,,Text2, line1 测试\NText2, line2 测试
```

wm4:
It's still not clear whether whatever code you use calls ``ass_set_storage_size()``.

I don't know whether changing the ``ScaledBorderAndShadow`` default is reasonable; it used not to make any sense with libass (because libass could render to screen resolution, i.e. different frame size than the video frame), but maybe the storage size thing handles this correctly now.

pingplug:
@wm4  there are may subtitles uses the default resolution 484x480, but the resolution of video is much more higher(1920x1080). These subtitles are created and previewed with VSFilter, and the border is not scaled. libass will scale the border, make the border thicker than VSFilter.

wm4:
That's yet another issue, and not what I asked about.

pingplug:
`ass_set_storage_size()` will not change the value of `ScaledBorderAndShadow`, so I don't think this function can help

TheOneric:
The two issues this pr attempted to resolve have since been addressed by #390 and #395 .

PR #181, opened 2015-06-07
Fix some issues reported by clang scan-build

grigorig:
I just ran clang's scan-build analysis on libass and it found some minor issues, most importantly a memory leak. These commits fix the problems reported.


wm4:
Looks ok.

PR #184, opened 2015-06-25
Remove 'xno' cargo culting from configure.ac

rcombs:


wm4:
+1

wm4:
Apparently this was rejected after much discussion?

grigorig:
Yes, there are apparently some legitimate reasons for doing this. Various edge cases related to how these comparison functions in shells work. So let's close this, then?

PR #187, opened 2015-07-24
Gradual cache clearing + ASS_Image refcounting

MrSmile:
That patch solves problem with periodical glitches due to full cache clearing.
Also I've introduced refcounting into caching system so refcounting of ASS_Images can be relatively simply implemented on top of it.


wm4:
What does ass_cache_cancel and ass_cache_commit do? (I'd expect there'd be a simple function to add an item to the cache.)

To be useful for future refcount stuff, increasing/decreasing the refcount would have to be thread-safe.

Ass for the code itself, I'm again check_flag and increment/decrement operators on the left side (``++cache->items;``).

MrSmile:
To make refcounting work there should be a way to get `CacheItem` from any cached value. For that I've made `CacheItem`, value and key adjacent in memory. Due to that, `ass_cache_get` (in case of cache miss) returns pointer to allocated but uninitialized value. `ass_cache_commit` signalizes successful initialization of that value, `ass_cache_cancel` deallocates memory and aborts addition to cache.

Thread safety is another large problem. It's not sufficient to make only inc/dec thread-safe, the whole cache subsystem should be thread-safe. I can try to work on it, but we should decide on which tools is possible to use: pthread, atomic, etc...

Without `check_flag` it's quite hard to catch incorrect usage of cache functions so it's essential for debugging.

wm4:
>Due to that, ass_cache_get (in case of cache miss) returns pointer to allocated but uninitialized value.

So it's just a trick to avoid a second lookup? Or to avoid copying ``CacheItem``?

Can't it just cache the lookup itself, so a second call does no work?

>Without check_flag it's quite hard to catch incorrect usage of cache functions so it's essential for debugging.

It'll be hard with check_flag too, then. Can it be replaced by a memory allocation? Slightly wasteful, but there are a lot of good tools to track correct use of them.

MrSmile:
> So it's just a trick to avoid a second lookup? Or to avoid copying CacheItem?

It's a trick to avoid copying value (and second lookup, of course).

> It'll be hard with check_flag too, then. Can it be replaced by a memory allocation?

By incorrect usage I mean calling `ass_cache_inc_ref` or `ass_cache_dec_ref` with uncached value or some garbage.

wm4:
>It's a trick to avoid copying value (and second lookup, of course).

If we get much more awkward code for a micro-optimization that doesn't even have a measurable effect (or did you measure it?), then I'd say simplification is preferable. Simpler code wins in the long term, even if it's slightly slower than the fastest possible implementation today.

>By incorrect usage I mean calling ass_cache_inc_ref or ass_cache_dec_ref with uncached value or some garbage.

Valgrind would catch this.

MrSmile:
> If we get much more awkward code for a micro-optimization that doesn't even have a measurable effect (or did you measure it?), then I'd say simplification is preferable.

Code is essentially the same so I don't see any problem here.

> Valgrind would catch this.

Not in all cases. Also, I don't want to support that code without `check_flag`.

grigorig:
What are the glitches again? Just frametime hiccups?

Like others have noted, the function naming and mode of operation with commit/cancel seems odd. I don't like it at all, even if it might be a bit faster. It's especially bad because it's completely undocumented.

Apart from that, nice approach, nice work.

MrSmile:
I've rebased that PR onto master after merging of the fonts branch. And now I have an idea how to make cache system thread-safe using pthreads (and get/commit scheme is essential for that). Also in multithreaded adaptation debug flag `check_flag` will transform into useful variable.

MrSmile:
New update: I've tracked all cached objects' lifetimes and ensured that each such object have positive `ref_count` during that.

There is only one step left for `ASS_Image` refcounting: design API for that. Currently all clearing work is done in `ass_free_images()` and can be moved to arbitrary moment of time. It's prohibited to cross `ass_reconfigure()` boundary though due to cache clearing (is there any need for that?).

Also there's one step for thread-safe cache: enable pthreads & atomics and add synchronization in cache functions, but that's for another PR, possible with threaded rendering.

wm4:
>There is only one step left for ASS_Image refcounting: design API for that. 

Should be trivial.

>Currently all clearing work is done in ass_free_images() and can be moved to arbitrary moment of time. It's prohibited to cross ass_reconfigure() boundary though due to cache clearing (is there any need for that?).

Yes, there is a need. The ASS_Images should survive ASS_Renderer being destroyed too.

MrSmile:
Then caches should be made global. I suppose ASS_Library is the good place for them. I'll try that.

wm4:
It should survive ASS_Library dying too. Also, we considered getting rid of ASS_Library completely.

Making ASS_Images able to survive this shouldn't be that hard though. E.g. this is the buffer pool implementation, which is used for refcounting in ffmpeg: https://git.libav.org/?p=libav.git;a=blob;f=libavutil/buffer.c (note how the pool can be destroyed while the buffers can live on).

wm4:
To be honest, I have no idea how to review this. Personally I'd say this should be applied, and then we test it for a while.

Some cosmetic changes would probably have been worth moving to separate commits.

The API should probably be only exposed once refcounting "just works": e.g. you can unref from other threads, and you can keep references to image data even if the renderer/library have been destroyed before.

I still don't like the debug magic values - I think they should be removed, or if you absolutely insist on keeping them, at least the magic values should be replaced with defines to make their function more apparent.

MrSmile:
Magic values will live only until proper threading, but I think about how to integrate debug functionality into other variables.

As for API, is it enough to just remove references of `ass_frame_[un]ref()` from `ass.h` for the time being?

wm4:
Yeah, and from the .sym file.

MrSmile:
I've removed magic numbers and postponed API change.

PR #189, opened 2015-08-31
coretext: implement GetFallbackFunc

pigoz:
I used the font family contained in ASS_FontProviderMetaData to create the font we are falling back from. I can switch to an hardcoded Arial if needed.

### DIFF
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -262,6 +262,30 @@ static void match_fonts(ASS_Library *lib, ASS_FontProvider *provider,
 }
 #endif
 
+static char *get_fallback(void *priv, ASS_FontProviderMetaData *meta,
+                          uint32_t codepoint)
+{
+    char *failed = meta->families[0];
+    CFStringRef name = CFStringCreateWithBytes(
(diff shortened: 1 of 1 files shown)


grigorig:
LGTM, although I may want to kill/replace the "meta" parameter. Does this work with the testcase I sent earlier?

wm4:
I can do this tomorrow.

I'll also merge this tomorrow.

pigoz:
@grigorig forgot to mention there are two missing glyphs with the test case you sent
test

grigorig:
It also seems strange that it uses the Comic font for the "Test" string. What is the debug output for this test case?

pigoz:
That's because I changed the font family in the test script to Comic Sans
MS to see if it was working correctly.

grigorig:
Maybe the missing glyphs are actually okay, possibly you are just lacking a font with those glyphs, or they are only available as bitmaps.

pigoz:
Well the last one is rendered as as a bitmap emoji in other applications.. the middle one is not a bitmap but looks strange even in other applications.

wm4:
Sounds good enough to me.

wm4:
Merged.

We still have doubts about that CFStringCreateWithBytes call, though.

PR #190, opened 2015-09-01
Substitutions and some fixes

grigorig:
Untested on OS X. Please test.


grigorig:
OK, agreed, this isn't so nice as an inline function.

grigorig:
Branch updated. Please test again.

grigorig:
Updated again, with some minor things. pigoz or wm4, can you please give it a shot with the test script?

grigorig:
Seems to work fine on OS X after feedback by @wm4. So let's merge it.

PR #191, opened 2015-09-07
MSVC compatibility, take two

grigorig:
The old branch, revived with some changes. I'm not sure if I really like this.


astiob:
Ah, sorry. :( I really should’ve done this. I’ll be able to check and do more useful things in another couple of days.

astiob:
Subsumed by a revived #84 and #196.

PR #192, opened 2015-09-07
NIH: add locale-independent string functions

grigorig:
OS or platform-specific locale independent functions are painful to
use and/or not available, so roll our own. Not great but the least
painful and least intrusive.


grigorig:
The table-based approach to convert to lowercase might look a bit unusual, but I actually benchmarked this: it's much faster than simple conditional bit operations on Intel Haswell for me, on Cortex A7 on other other hand, both methods perform about the same.

wm4:
Good to merge IMHO.

PR #196, opened 2015-09-13
MSVC fixes.

torque:
The fontselect merge introduced some MSVC compilation problems. This PR doesn't attempt to manage all of them because some fixes require more invasive changes.

Here's a list of things that are __not__ addressed in this PR:

- `ass_fontselect` uses iconv to convert UTF-16 to UTF-8. This makes iconv a hard dependency. `WideCharToMultiByte` works on Windows, but some abstraction is necessary for good cross-platform support.

- ~~`ass_fontselect` uses `PATH_MAX` which is undefined on Windows.~~ Fixed enough in master.

- ~~`ass_fontselect` uses `snprintf` which doesn't exist on windows.~~ This is fine with MSVC14.

### DIFF
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -650,7 +650,7 @@ static ASS_FontProviderFuncs directwrite_callbacks = {
     .get_fallback       = get_fallback,
 };
 
-typedef HRESULT WINAPI (*DWriteCreateFactoryFn)(
+typedef HRESULT (WINAPI *DWriteCreateFactoryFn)(
   _In_  DWRITE_FACTORY_TYPE factoryType,
   _In_  REFIID              iid,
   _Out_ IUnknown            **factory
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -148,8 +148,8 @@ static HRESULT STDMETHODCALLTYPE FallbackLogTextRenderer_DrawInlineObject(
     FLOAT originX,
     FLOAT originY,
     IDWriteInlineObject *inlineObject,
-    WINBOOL isSideways,
-    WINBOOL isRightToLeft,
+    BOOL isSideways,
+    BOOL isRightToLeft,
     IUnknown *clientDrawingEffect
     )
 {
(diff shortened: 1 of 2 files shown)


wm4:
What MSVC version are you using? I'm pretty sure MSVC 2015 has snprintf.

torque:
I'm still on 2013 because updating is hard.

wm4:
Then update.

wm4:
Anyway, not against the 2 commits, but someone should check if it still works with gcc (I can't do that right now).

torque:
With the above patches and an [additional patch](https://gist.githubusercontent.com/torque/7d8496bc8169dd5d7cb6/raw/dff2160fc4fdeaba042b8a1ae8376158a16d5600/libass-no-iconv.patch) to purge iconv from fontselect, libass now compiles with MSVS 2015 Community (MSVC14) using a minimal `config.h`:

```C
#define CONFIG_DIRECTWRITE 1
#define CONFIG_RASTERIZER 1
```

4re:
[This patch](https://patch-diff.githubusercontent.com/raw/libass/libass/pull/196.patch) is fine, but [the one linked before](https://gist.githubusercontent.com/torque/7d8496bc8169dd5d7cb6/raw/dff2160fc4fdeaba042b8a1ae8376158a16d5600/libass-no-iconv.patch) makes gcc complain about 
```
/var/tmp/portage/media-libs/libass-9999-r2/work/libass-9999/libass/ass_fontselect.c:39:21: fatal error: Windows.h: No such file or directory
compilation terminated.
Makefile:551: recipe for target 'ass_fontselect.lo' failed
```

torque:
The linked patch for removing iconv is not portable to non-windows which is why it has not been applied to the pull request. It is only for building with MSVC in Visual Studio. I'm not sure how to properly do cross-platform charset conversion, yet.

grigorig:
LGTM (and still works with msys/mingw, I just tested it), but I'm still unsure whether I like the ass_compat.h thing or not.

PR #197, opened 2015-09-23
RFC: Remove ENCA support

grigorig:
There is no real value in it. ENCA only works in some cases and I
can't find any information about software that integrates libass
which relies on it or users that rely on ENCA support in case the
interface is exposed directly (e.g. in MPlayer).

If there is still a wish to have charset detection integrated into
libass (it is out of scope, IMHO), a better library like uchardet
should be used for that.


wm4:
+1

But honestly, I think all external file loading should be removed. Instead, ass_read_memory() could be used. If we ever do a complete API overhaul, stupid things like file loading by path could be dropped.

grigorig:
No complaints, so I guess it's fine.

astiob:
I did previously voice my concerns on IRC (I’m −½ on removing code for the sake of literally only removing code), and I didn’t expect this to be closed so soon, especially with a title saying RFC, but whatever.

PR #200, opened 2015-10-10
configure: Set the correct x32 machine option for yasm

luke-jr:


astiob:
Thanks! I’m afraid none of us have a x32 setup to test this (anyone correct me if I’m wrong), but if your patch does genuinely fix the build for you, I’ll accept it as is.

I have to ask though: have you tested the resulting binary to make sure it not only compiles but also actually works? I have some doubts regarding the interaction between the C pointer and integer types and the assembler code.

luke-jr:
Best I was able to do so far was use the included test program with an image and ASS file. That did not crash, and reported something like 2 overlays merged (indicating it used assembly code, I think?). Anything else I could test with easily (ie, without needing a bunch of other codecs that don't work on x32 yet..)?

astiob:
I’ll whip up a test ASS file and a comparison image in a minute.

astiob:
Try this ASS file: https://gist.github.com/astiob/b4da312a8ec4d66d66fd
The output PNG should look like this: http://i.minus.com/ik71bos8rTP3.png

luke-jr:
Not quite. I ended up with http://imgur.com/8QRprDw

luke-jr:
Note: I get that same result on my regular 32-bit system too...

astiob:
Oh, I messed up the resolution. I’m fixing the gist, but it looks like it’s probably working.

astiob:
Gist updated, and new expected image: http://i.minus.com/ibeU8qGj4JtMn3.png

astiob:
Sorry for the trouble and thanks again! One last thing: would you mind rebasing your patch on top of our master branch?

luke-jr:
Perfect match.

Rebased.

astiob:
Thanks! Merged.

PR #203, opened 2015-10-21
Find fonts with PostScript outlines by PostScript name (+ minor fontselect fixes)

astiob:
I haven’t fully tested this. (I’ve tested the memory font provider with a TTF and OTF pair provided by line0.)

Also I’m slightly worried about creating DirectWrite font face objects just to extract a property and destroy them. We could immediately store them in the FontPrivate struct and keep them alive until init_font_private runs. However, judging from the documentation this may be a cheap operation, in which case there’s no point. I guess some performance testing would help here.


astiob:
The Linux build is failing because of an old Fontconfig. Looking at Debian and Ubuntu LTS releases, Debian Wheezy and Ubuntu Precise have Fontconfig that’s too old, but both have new LTS releases with new Fontconfig: Debian Jessie was released in April this year, and Ubuntu Trusty was released in April last year. For some reason Travis uses Trusty only if it’s explicitly chosen in the config file.

I can #ifdef the PostScript name extraction in ass_fontconfig, but then unsuspecting Linux users (who would otherwise build or find a new Fontconfig) can get PostScript names silently failing to be used. This per se isn’t worse than now, when they aren’t used on any platform, but it will make the behaviour of libass inconsistent, dependent on third-party libraries and not fixable by installing new binaries of the relevant third-party libraries, which isn’t nice. But maybe it’s not too bad.

grigorig:
LGTM in general, I added some comments.

astiob:
FontForge won’t let me add leading spaces to the family name and neither leading nor trailing spaces to the PostScript name, but with trailing spaces in the family name and with either leading or trailing spaces in the full name, VSFilter does **not** find the font. Hence I think we should replace *all* uses of strdup_trimmed with strdup. (And I just noticed this: we still have trim_space declared in ass_utils.h despite the implementation having been deleted in favour of strdup_trimmed.)

Further testing shows that VSFilter does not find fonts with PostScript outlines by fullname! Perhaps GDI internally substitutes the PostScript name for the fullname or something. I’ll add a commit for that.

Finally, I seem to recall we used to search for combinations of family name and style name; am I remembering correctly? We don’t do it now. And that’s good! Because VSFilter doesn’t either.

astiob:
> Also, I'd prefer to simply require an up to date fontconfig.

OK. I’ll add the appropriate flag to make Travis CI use Trusty.

grigorig:
> Finally, I seem to recall we used to search for combinations of family name and style name; is this right? We don’t do it now. And that’s good! Because VSFilter doesn’t either.

In the old fontconfig based font code? That was just a workaround for some restrictions of fontconfig. You can't sort fonts by fullname.

astiob:
I added a few commits last night as discussed before; could you please review them too? I also changed “fontconfig: read PostScript names” 018cfe7 to update the Fontconfig check in configure.ac and “font: check FT_Get_Postscript_Name return value for NULL” to avoid discarding const from a pointer.

There’s [another two const warnings](https://travis-ci.org/libass/libass/jobs/86937942#L2366) remaining because ASS_FontProviderMetaData contains non-const char pointers. I’ll see if I can fix this in a sensible way.

> In the old fontconfig based font code? That was just a workaround for some restrictions of fontconfig.

Yeah. Either way, it’s good that we don’t (need to) do it any more.

wm4:
I guess I'm fine with these changes, as far as memory management etc. goes.

grigorig:
LGTM, but the warnings are ugly and should be fixed.

astiob:
The warnings were removed by adding casts in bf8524fe5c8eb20c03f0732c3c782a330e97594e.

PR #213, opened 2016-02-24
Add heuristic for style names to style override

Akemi:
i made this modification for myself since this is how i expect the style override to work. it should leave signs and lyrics as they are even if they are not caught by the tag heuristic.

anyway others might want the same behaviour, even though it's a bit special, so i opened a pull request to discuss it.


Akemi:
added some more cases to the heuristic. things i encountered on some files since the original commit.

Akemi:
rebased and updated for the recent changes.

PR #214, opened 2016-03-10
Forwarded patches from VLC contrib collection

SeanMcG:
Hi,

Attached are several patches that are currently applied to the VLC contrib version of libass.

Would appreciate if some or all of them could be applied upstream.


wm4:
Although I don't know what they do (at least the iconv one), these seem rather inoffensive, so I'm ok with them.

grigorig:
AFAIR, there's a problem with the pkg-config control file on platforms that don't include iconv functionality in libc. The lack of libiconv in the linker flags simply causes linking of applications that use libass to fail in some cases. This might have come up on IRC quite some time ago.

Can't say anything about the Solaris patch, but it's not going to break anything major.

Overall, LGTM.

PR #215, opened 2016-03-13
configure: use proper quoting on iconv test

SeanMcG:
Unquoted strings cause an error when the test variable is empty.


SeanMcG:
Sincere apologies for not testing this properly before my earlier commit.

wm4:
Merged, thanks.

PR #223, opened 2016-05-23
font: fix NULL pointer dereference

ssbssa:


astiob:
Seems easy enough. Could you provide an example of when this occurs?

ssbssa:
Well, it's this bug in VLC: https://trac.videolan.org/vlc/ticket/16652
I see this warning earlier on line 70: `ass_msg(library, MSGL_WARN, "Font face with no charmaps");`
Example file would take me a while to get.

astiob:
So I found the font from that VLC bug report thanks to the attached vlc_log.txt.

It’s a somehow botched-up copy of HermesCTT Regular. Compared to a valid copy I found on the net, almost 99 % of the file contents match. And yet the 1 % difference is enough to break the file so badly that neither OS X nor Windows agree to even open the file. In fact, trying to open it directly in my OS X crashes Fonts.app—but the Validate File function reports that the file fails to have “font basic parsability”. Windows says “The requested file […]\HermesCTT Regular_0.ttf was not a valid font file.”

FontForge is able to open the file but reports ﻿the following:

> Table 'cmap' extends beyond end of file and must be ignored.
> The following table(s) in the font have been ignored by FontForge
>   Ignoring 'hdmx' horizontal device metrics table
> Windows will reject fonts with an OS/2 version number of 0

VSFilter doesn’t crash or complain and simply falls back to Arial. I guess we’ll follow suit.

ssbssa:
Great, thanks.

PR #227, opened 2016-06-24
bitmap: use calloc instead of malloc/memset

rcombs:
This can improve perf significantly with large bitmaps

### DIFF
--- a/libass/ass_utils.c
+++ b/libass/ass_utils.c
@@ -79,12 +79,13 @@ char *ass_strndup(const char *s, size_t n)
 }
 #endif
 
-void *ass_aligned_alloc(size_t alignment, size_t size)
+void *ass_aligned_alloc(size_t alignment, size_t size, bool zero)
 {
     assert(!(alignment & (alignment - 1))); // alignment must be power of 2
     if (size >= SIZE_MAX - alignment - sizeof(void *))
         return NULL;
-    char *allocation = malloc(size + sizeof(void *) + alignment - 1);
+    char *allocation = zero ? calloc(size + sizeof(void *) + alignment - 1, 1)
+                            : malloc(size + sizeof(void *) + alignment - 1);
     if (!allocation)
         return NULL;
     char *ptr = allocation + sizeof(void *);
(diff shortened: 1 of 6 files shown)


grigorig:
Can you provide numbers for the performance improvements?

rcombs:
I'm seeing improvements in the range of 5% on the particularly heavy (on allocation size) lines in [this sample](http://puu.sh/pE9Uc/2315b54f8b.ass), so I suppose it's "significant" in the statistical sense. It ended up appearing more dramatic during playback on my machine, since we were already very close to realtime, so a minor performance improvement ended up getting the per-frame render time down low enough that I went from 12 dropped frames to 1 pretty reliably.

The point is to avoid the `memset()` call in cases where `calloc()` is allocating (kernel-zeroed) pages directly. `calloc()` acts like `malloc()` + `memset()` in other cases, so it should be no slower than the current code there.

libass/ass_utils.c
```diff
@@ -95,6 +95,22 @@ void *ass_aligned_alloc(size_t alignment, size_t size)
...
+    if (!allocation)
+        return NULL;
+    char *ptr = allocation + sizeof(void *);
+    unsigned int misalign = (uintptr_t)ptr & (alignment - 1);
+    if (misalign)
+        ptr += alignment - misalign;
+    *((void **)ptr - 1) = allocation;
+    return ptr;
+}
+
```
wm4: Seems good overall. I wonder if you could reduce code duplication here by a bit by moving the aligning to a common function?

grigorig:
Sounds good, LGTM if you address the code duplication.

MrSmile:
For consistency, I think `ass_aligned_alloc()` and `ass_aligned_calloc()` should be combined (with similar `bool zero` parameter). Also `alloc_bitmap()` and `alloc_bitmap_raw()`.

rcombs:
Yeah, that's what I'll end up doing. This is basically how malloc and calloc work internally (on OS X) as well.

grigorig:
LGTM.

grigorig:
Merged.

PR #229, opened 2016-07-06
render: Fix overflow causing out of bounds memory access

ajaska:
I don't know if this breaks something else, but...

The following .ass file breaks libass:

```
[V4+ Styles]
Style:Default,DejaVuSans,5,,,,,,,,,100,,7680000
[Events]
Dialogue:1,0:00:00.0,1:00:0.0,Default,,,,,,lol ass
```

Also, I removed the `if (...) break` because it is pointless code (the for-loop will terminate immediately, regardless of if it is there).

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1575,7 +1575,7 @@ wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
         s1 = s2 = 0;
         for (i = 0; i <= text_info->length; ++i) {
             cur = text_info->glyphs + i;
-            if ((i == text_info->length) || cur->linebreak) {
+            if (cur->linebreak) {
                 s1 = s2;
                 s2 = s3;
                 s3 = cur;
@@ -1614,8 +1614,6 @@ wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
                     }
                 }
             }
-            if (i == text_info->length)
-                break;
         }
 
     }


wm4:
Just looking at the code, I'd be afraid that it could break something else.

What exactly causes the crash?

ajaska:
Oh, I was mistaken on the cause - this is an overflow.

``` 
(gdb) b ass_render.c:1610
[...]
Breakpoint 1, wrap_lines_smart [...]
(gdb) print l2_new
$1 = -19017695.796875
(gdb) print (s3 - 1)->bbox.xMax
$2 = 325
(gdb) print (s3 - 1)->pos.x
$3 = 7372801801
(gdb) print w->bbox.xMin
$4 = 65
(gdb) print w->pos.x
$5 = 0
```

```c
int main()
{
    signed long a, b, c, d;
    a = 325;
    b = 7372801801;
    c = 65;
    d = 0;
    printf("%ld %ld %ld %ld\n", a, b, c, d);
    printf("%ld \n", (a+b)-(c+d));
    printf("%f\n", d6_to_double((a+b)-(c+d)));
}
```

```
$ ./a.out
325 7372801801 65 0
7372802061
-19017695.796875
```

ajaska:
(The overflow causes the routine to attempt to think it should split the lines at the very beginning of the line, causing it to attempt to set the linebreak of a GlyphInfo at position `text_info->glyph_info - 1` which results in a segfault.)

ajaska:
Well, I used the wrong word-ish ... "bad conversion" instead of "overflow" maybe?

This was introduced in 0703d5f.

`d6_to_double` is currently only used in `ass_render.c` and is called with an `int` on lines 1416, 1417, 1428, 1429, 1430,1431. It is called with an `signed long` on 686, 688, 693, 694, 1532, 1533, 1599, 1601, 1603, 1606, 1635, 1651, 2100, 2177.

My guess is that this conversion has the potential to break on all 14 locations that call it with a `signed long`. I have no idea what `d6_to_double` is used for or trying to do, but I don't think it's working properly atm.

ajaska:
Well now I'm horribly confused.

FT D6 (26.6) types are supposed to be `signed long`s. Why are they being stored as `int`s all over the place?

grigorig:
> FT D6 (26.6) types are supposed to be signed longs. Why are they being stored as ints all over the place?

Good question. The FreeType definition for this type needlessly wastes space on most 64 bit ABIs, though, where long is usually 64 bit size. So int isn't exactly a clean solution, but it should always work when int is at least 32 bit size.

wm4:
Even if `d6_to_double` can have a "bad conversion" (what does that mean anyway), I don't think the line breaking code should be able to crash. So there are multiple bugs here at least.

ajaska:
Well... this is what it looks like with the right types (I think). On my (64-bit) machine, a quick and un-scientific memory profile showed that this actually had a lower memory usage.

https://github.com/libass/libass/compare/master...ajaska:fix-ft-types

ajaska:
And re: the conversion...

```c
#include <stdio.h>

typedef signed long FT_Pos;
typedef signed long FT_F26Dot6;

static double d6_to_double(int x)
{
	return x / 64.;
}

static double d6_to_double2(FT_F26Dot6 x)
{
	return x / 64.;
}

static int double_to_d6(double x)
{
    return (int) (x * 64);
}

static FT_F26Dot6 double_to_d62(double x)
{
    return (FT_F26Dot6) (x * 64);
}

int main(int argc, char *argv[])
{
    FT_Pos a, b, c, d;
    a = 325;
    b = 7372801801;
    c = 65;
    d = 0;
    printf("a=%ld b=%ld c=%ld d=%ld\n", a, b, c, d);
    printf("no conversion: %ld \n", (a+b)-(c+d));
	printf("\n");

    printf("d6_to_double: %f\n", d6_to_double((a+b)-(c+d)));
    printf("converted twice: %d\n", double_to_d6(d6_to_double((a+b)-(c+d))));
	printf("\n");

    printf("d6_to_double2: %f\n", d6_to_double2((a+b)-(c+d)));
    printf("converted twice2: %ld\n", double_to_d62(d6_to_double2((a+b)-(c+d))));
	printf("\n");

	int e = 64;
    printf("%f\n", d6_to_double(e));
    printf("%f\n", d6_to_double2(e));

	printf("Double: %ld bytes\n", sizeof(double));
	printf("Signed Long: %ld bytes\n", sizeof(signed long));
	printf("Int: %ld bytes\n", sizeof(int));

    return 0;
}
```

```
a=325 b=7372801801 c=65 d=0
no conversion: 7372802061 

d6_to_double: -19017695.796875
converted twice: -1217132531

d6_to_double2: 115200032.203125
converted twice2: 7372802061

1.000000
1.000000
Double: 8 bytes
Signed Long: 8 bytes
Int: 4 bytes
```

So these should be stored as FT types (not `int`s), or there needs to be conversion to an int happening everywhere FT values are being ingested?

grigorig:
Hold on, this might not be completely fixed yet.

astiob:
The crash should be fixed, but the overflow probably isn’t—assuming the overflow is an actual problem.

@ajaska Your examples use values that don’t fit in 26.6=32 bits. Are such values valid in the first place?

ajaska:
Hi -- been a while since I took a look at this. I'm going to say that the values are valid, because it's possible to define them in an ass script. (If they're not valid, throw an error before rendering/ignore the line/something else, but definitely don't segfault.)

grigorig:
@ajaska FWIW I like your patch to use the proper fixed point types. Can submit a proper PR?

PR #232, opened 2016-07-10
font load from dir: use MSGL_INFO instead of MSGL_WARN

### LINKED ISSUE #231: "Loading font file ..." is MSGL_WARN, should probably be MSGL_INFO or less
Here https://github.com/libass/libass/blob/master/libass/ass_fontselect.c#L178

As far as I can tell this is a legitimate setup where libass loads fonts from a non system dir, and if the fonts are there then it's highly likely that the user put them there exactly such that libass can load them, at which case I don't quite see why the message about loading them should be flagged as a warning.

This ends up with very verbose output of application which pass libass messages above certain level - such as mpv. With mpv, if the user wants to load fonts from a non default dir then the console is cluttered with libass messages since libass classifies them as warning - which mpv finds important enough to display to the user by default.

I think It should probably be either MSGL_INFO or even MSGL_V, since as far as I can tell this is a normal and expected course of action when loading those fonts.

I can also send a PR, just tell me which level you prefer.

avih:
This is a normal course of action and should not generate a warning,
especially for applications which use libass and might notify the user
on such "warnings", while in fact it should be info or even verbose.

Fixes #231

### DIFF
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -175,7 +175,7 @@ static void load_fonts_from_dir(ASS_Library *library, const char *dir)
         char fullname[4096];
         snprintf(fullname, sizeof(fullname), "%s/%s", dir, entry->d_name);
         size_t bufsize = 0;
-        ass_msg(library, MSGL_WARN, "Loading font file '%s'", fullname);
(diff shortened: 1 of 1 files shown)


grigorig:
I don't see any objections, so let's merge this.

PR #236, opened 2016-08-14
New stroker algorithm

MrSmile:
Not for merge as is.

This PR basically consists of 3 things:
- Replacement of the FreeType stroker, that is much faster and can output two border outlines (inside and outside) simultaneously.
- Adaptation of the rasterizer to work with both generated outlines. That should help with large glyphs and drawings, but for small there is x2 slowdown compared to one border outline.
- Replacement of one border outline with two throughout rendering pipeline.

That solves problems with broken fonts with reversed contour directions and self-intersecting drawings (at cost of twofold rasterization). Also, my stroker is much faster and more stable than FreeType's and correctly works with non-uniform border sizes.


rcombs:
Does this handle strokes on drawings with more than 2^15 points?
Would be good to see some benchmarks on the savings from faster stroking vs the cost of 2x rasterization.
I wonder how much of the difference is due to FreeType's slow fixed-point trigonometry.

MrSmile:
Yes, old 2^15 limit isn't matter anymore.
My insane test case is 2.5 times faster than FreeType and don't have broken outlines.

MrSmile:
That's rebased version after some cleanup.
Also I fixed problem with random dots due to small invisible contours in some fonts.

wm4:
"rasterizer: refactoring" is quite intrasparent... it does a lot of things, including cosmetics, C99 conversions, and semantic changes. But I guess they're ok.

libass/ass_rasterizer.c
```diff
@@ -258,13 +264,19 @@ static bool add_cubic(RasterizerData *rst, const OutlinePoint *pt)
...
+                            const ASS_Outline *path, bool extra)
 {
     enum Status {
         S_ON, S_Q, S_C1, S_C2
     };
 
-    rst->size[0] = 0;
+    if (!extra) {
+        rst->x_min = rst->y_min = 0x7FFFFFFF;
+        rst->x_max = rst->y_max = 0x80000000;
```
wm4: Why not INT_MIN/INT_MAX?

libass/ass_rasterizer.c
```diff
@@ -553,80 +565,99 @@ static inline int segment_check_bottom(const struct segment *line, int32_t y)
...
+                                int winding[2], int32_t y)
 {
-    int winding = 0;
-    const struct segment *end = src + n_src;
+    const struct segment *cmp = src + n_src[0];
+    const struct segment *end = cmp + n_src[1];
+    n_dst0[0] = n_dst0[1] = 0;
+    n_dst1[0] = n_dst1[1] = 0;
     for (; src != end; src++) {
+        int group = src < cmp ? 0 : 1;
```
wm4: Could be `int group = src >= cmp;`, but maybe above is clearer.

libass/ass_rasterizer.c
```diff
@@ -683,6 +714,33 @@ static inline void rasterizer_fill_halfplane(const BitmapEngine *engine,
...
+    if (((line->flags & test) != test) == !(line->flags & SEGFLAG_DN))
+        winding++;
+
+    int flags = 0;
+    if (winding) flags ^= 1;
+    if (winding - 1) flags ^= 3;
+    static const int result[] = {
+        0, FLAG_COMPLEX, FLAG_SOLID, FLAG_COMPLEX | FLAG_REVERSE
+    };
+    return result[flags];
```
wm4: Is there really any point to those bitops and they array, instead of using the FLAG_s directly?

libass/ass_rasterizer.c
```diff
@@ -694,72 +752,92 @@ static inline void rasterizer_fill_halfplane(const BitmapEngine *engine,
...
+            engine->fill_generic(buf, stride, line, n_lines[0], winding[0]);
+        else
+            engine->fill_halfplane(buf, stride, line->a, line->b, line->c,
+                                   flags0 & FLAG_REVERSE ? -line->scale : line->scale);
+        if (flags1 & FLAG_GENERIC)
+            engine->fill_generic(rst->tile, width, line1, n_lines[1], winding[1]);
+        else
+            engine->fill_halfplane(rst->tile, width, line1->a, line1->b, line1->c,
+                                   flags1 & FLAG_REVERSE ? -line1->scale : line1->scale);
+        // XXX: should be max instead of add
```
wm4: Should that XXX be taken care of before merge?

libass/ass_outline.c
```diff
@@ -133,143 +172,1008 @@ void outline_get_cbox(const ASS_Outline *outline, FT_BBox *cbox)
...
+    if (result1)
+        result1->n_contours = result1->n_points = 0;
+
+    int rad = FFMAX(xbord, ybord);
+    if (rad < eps) {
+        assert(result->max_contours >= path->n_contours);
+        assert(result->max_points >= path->n_points);
+        memcpy(result->contours, path->contours, sizeof(size_t) * path->n_contours);
+        memcpy(result->points, path->points, sizeof(FT_Vector) * path->n_points);
+        memcpy(result->tags, path->tags, path->n_points);
```
wm4: So tags is 1 byte per element?

In all those cases it might be slightly more robust to use `sizeof(result->bla[0])` instead of the direct type. (I wonder whether an "array copy" macro that checks types would be good for robustness...)

wm4:
OK to "renderer: remove legacy FreeType rasterizer". This was for "comparison", and having it really made things more complicated without much gain.

wm4:
"cosmetic: fix codestyle" adds spaces after casts. Is that really our coding style?

libass/ass_outline.c
```diff
@@ -59,12 +59,20 @@ bool outline_convert(ASS_Outline *outline, const FT_Outline *source)
...
+        // skip degenerate 2-point contours from broken fonts
+        if (n >= 3) {
+            memcpy(outline->points + outline->n_points, source->points + start, sizeof(FT_Vector) * n);
+            memcpy(outline->tags + outline->n_points, source->tags + start, n);
+
+            outline->n_points += n;
+            outline->contours[outline->n_contours++] = outline->n_points - 1;
+        }
+        start = source->contours[i] + 1;
+    }
```
wm4: Maybe you could mention an example of such a glyph in the commit message.

wm4:
So my review is a bit worthless. There are tons of intricate code changes in code I already didn't know well, and I have not the slightest clue about stroker algorithms or the algorithm used here. So I can only comment on mostly cosmetic issues. But in general, this looks good to me.

libass/ass_rasterizer.c
MrSmile: Do you mean INT32_MIN/INT32_MAX?
Well, those lines is copy-paste but indeed, using compiler constants make sense.

wm4: Right.

libass/ass_rasterizer.c
MrSmile: Using nested ifs? I don't think that will be more clearer anyway but I consider it.

libass/ass_rasterizer.c
MrSmile: That lead to really miniscule difference that exist only at rare edge cases.
And correct implementation require new assembly function so it's not cost-effective.
Also I'm planning to rewrite that part for tiled-rendering anyway.

libass/ass_rasterizer.c
wm4: Yeah, the goal should be clarity - if that makes it worse, the code is fine.

libass/ass_outline.c
MrSmile: That's from https://forum.doom9.org/showthread.php?p=1799844#post1799844
Somewhere in that thread is link to source file but such links isn't stable.

MrSmile:
Sadly, unlike blur, stroker algorithm is not straightforward. There are ton of different edge cases and complex logic. I've designed it mostly through trial and error and extensive experimentation.

libass/ass_outline.c
Blitzker: Screenshot with such a problem: 
aa1
(Japanese karaoke at top right)

MrSmile:
Rebased version with fixed aforementioned issues.
Beware of strange commit order here on github though.

grigorig:
LGTM, will merge soon if no one objects.

grigorig:
Merged.

PR #240, opened 2016-10-04
Fix issues found by fuzzing

grigorig:
This should fix the more severe issues reported by @brandonprry's fuzzing. The problems aren't that bad overall, so we decided to do this in public.

This addresses the test cases except for "id:000248,sig:11,src:004326,op:havoc,rep:16" which is a pure DoS issue that occurs simply because lots of huge bitmaps need to be processed. I'm not sure if we should add any safeguards for these kinds of things, as they tend to break.

The overall plan is to get this reviewed and merged fast, so that we can spin a new release quickly. So please review!

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -590,6 +590,7 @@ static int process_styles_line(ASS_Track *track, char *str)
     if (!strncmp(str, "Format:", 7)) {
         char *p = str + 7;
         skip_spaces(&p);
+        free(track->style_format);
         track->style_format = strdup(p);
         ass_msg(track->library, MSGL_DBG2, "Style format: %s",
                track->style_format);
--- a/libass/ass_blur.c
+++ b/libass/ass_blur.c
@@ -744,7 +744,7 @@ static void calc_coeff(double mu[4], const int index[4], int prefilter, double r
         (  17 -  126 * mul +  273 * mul2 -  164 * mul3) / 12096,
     };
 
-    double mat_freq[13];
+    double mat_freq[14];
     memcpy(mat_freq, kernel, sizeof(kernel));
     memset(mat_freq + 4, 0, sizeof(mat_freq) - sizeof(kernel));
     int n = 6;
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1607,7 +1607,9 @@ wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
                         ((s3 - 1)->bbox.xMax + (s3 - 1)->pos.x) -
                         (w->bbox.xMin + w->pos.x));
 
-                    if (DIFF(l1_new, l2_new) < DIFF(l1, l2)) {
+                    if (DIFF(l1_new, l2_new) < DIFF(l1, l2) && w > text_info->glyphs) {
+                        if (w->linebreak)
+                            text_info->n_lines--;
                         w->linebreak = 1;
                         s2->linebreak = 0;
                         exit = 0;
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1518,7 +1518,6 @@ wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
     double pen_shift_x;
     double pen_shift_y;
     int cur_line;
-    int run_offset;
     TextInfo *text_info = &render_priv->text_info;
 
     last_space = -1;
@@ -1628,7 +1627,6 @@ wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
     trim_whitespace(render_priv);
 
     cur_line = 1;
-    run_offset = 0;
 
     i = 0;
     cur = text_info->glyphs + i;
@@ -1649,7 +1647,6 @@ wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
                 text_info->lines[cur_line - 1].offset;
             text_info->lines[cur_line].offset = i;
             cur_line++;
-            run_offset++;
             pen_shift_x = d6_to_double(-cur->pos.x);
             pen_shift_y += height + render_priv->settings.line_spacing;
         }
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -100,6 +100,7 @@ static bool check_allocations(ASS_Shaper *shaper, size_t new_size)
             !ASS_REALLOC_ARRAY(shaper->emblevels, new_size) ||
             !ASS_REALLOC_ARRAY(shaper->cmap, new_size))
             return false;
+        shaper->n_glyphs = new_size;
     }
     return true;
 }
--- a/test/test.c
+++ b/test/test.c
@@ -186,6 +186,7 @@ static void print_font_providers(ASS_Library *ass_library)
         printf("%s'%s'", separator,  font_provider_labels[providers[i]]);
     }
     printf(".\n");
+    free(providers);
 }
 
 int main(int argc, char *argv[])


libass/ass.c
```diff
@@ -590,6 +590,7 @@ static int process_styles_line(ASS_Track *track, char *str)
     if (!strncmp(str, "Format:", 7)) {
         char *p = str + 7;
         skip_spaces(&p);
+        free(track->style_format);
         track->style_format = strdup(p);
         ass_msg(track->library, MSGL_DBG2, "Style format: %s",
                track->style_format);
```
wm4: Unrelated to your patch, but if the strdup fails, this will pass NULL as string to printf, which is undefined behavior (although glibc as a feature prints "(null)".

brandonprry:
Does this also fix the usage of fribidi or is that testcase something I should pass on to the fribidi devs?

libass/ass_blur.c
```diff
@@ -744,7 +744,7 @@ static void calc_coeff(double mu[4], const int index[4], int prefilter, double r
         (  17 -  126 * mul +  273 * mul2 -  164 * mul3) / 12096,
     };
 
-    double mat_freq[13];
+    double mat_freq[14];
```
wm4: Wow, this function is so extremely hard to follow. Is it right that vec_freq below still has only 13 items?

grigorig:
@brandonprry Can you explain what kind of misuse you mean? The illegal memory accesses should be fixed.

libass/ass_render.c
```diff
@@ -1607,7 +1607,9 @@ wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
                         ((s3 - 1)->bbox.xMax + (s3 - 1)->pos.x) -
                         (w->bbox.xMin + w->pos.x));
 
-                    if (DIFF(l1_new, l2_new) < DIFF(l1, l2)) {
+                    if (DIFF(l1_new, l2_new) < DIFF(l1, l2) && w > text_info->glyphs) {
```
wm4: Not sure if I follow. is that added check the same as "w != text_info->glyphs"?

grigorig: It's an array of glyph structs and I want to make sure that w isn't the first one. This is the a) case in the commit description. Should be equivalent to your idea.

brandonprry:
In the provided files, sample id:000051 causes errors in fribidi_reorder_line, which is consumed by libass. I wasn't sure if this was something caused by misuse of the fribidi API or something that the devs should be aware of.

grigorig:
@brandonprry Yes, that should be fixed by commit aa54e0b.

libass/ass_blur.c
MrSmile: I've tried to document algorithm [here](https://github.com/MrSmile/CascadeBlur/releases/download/v0.5-alpha/cascade-blur.pdf), but it's too math-heavy. Basically `mat_freq` and `vec_freq` used in calculation of coefficients for least squares problem and any sufficient size should work. I'd tried to determine minimal required size, but forgot that `coeff_blur121` requires additional zero element to work.

brandonprry:
BTW I believe https://github.com/libass/libass/pull/229 is related to this PR.

libass/ass_render.c
astiob: Shouldn’t this rather remove the first line (and join it with the second line) if w is the first glyph? Roughly like this:

```c
if (DIFF(l1_new, l2_new) < DIFF(l1, l2)) {
    if (w == text_info->glyphs || w->linebreak)
        text_info->n_lines--;
    if (w != text_info->glyphs)
        w->linebreak = 1;
    s2->linebreak = 0;
    exit = 0;
}
```

Regarding the commit message: in addition to producing an unwanted empty line at the start, I think setting `text_info->glyphs->linebreak` also causes a [buffer overread](https://github.com/grigorig/libass/blob/b72b283b936a600c730e00875d7d067bded3fc26/libass/ass_render.c#L1473-L1477) in `trim_whitespace`—probably the cause of #229.

grigorig: Sounds good. I'll see how that behaves.

grigorig: That seems to work well. If there are no objections that means I'll merge this PR.

grigorig:
Merged.

brandonprry:
Thanks a bunch

astiob:
Thank you!

astiob:
@brandonprry In your [CVE request](http://www.openwall.com/lists/oss-security/2016/10/04/9), you say:

> The third is a huge memory allocation leading to a crash that wasn’t fixed because a good solution is unavailable at the moment.

Does it lead to a crash? The fuzzing report seemed to suggest that the script had lots of giant glyphs, which naturally take up a lot of memory. We should be able to handle such cases without crashing, unless we get killed by Linux’s OOM killer. Is that what happened in your test?

It isn’t clear how exactly we can prevent that: scripts will always be able to request lots of memory, just like, say, you can always request your Web browser to open an arbitrarily large image. Of course, we can and should try to reduce our memory usage, but it’s not clear if we can ever be absolutely safe against this.

brandonprry:
Using ASAN it leads to a crash. Probably poor wording on my part. It likely does crash on systems or processes with limited memory. 

==20320==ERROR: AddressSanitizer failed to allocate 0x6644a000 (1715773440) bytes of LargeMmapAllocator (errno: 12)

wm4:
Normally we should handle memory allocation failures gracefully, but there's probably still a lot of code which just crashes. On the other hand, if overcommit is enabled on Linux, a program allocating too much memory can be randomly killed by the Linux OOM killer, which we of course can't deal with.

wm4:
PS: what I'm trying to say is, which case does happen here?

astiob:
> ==20320==ERROR: AddressSanitizer failed to allocate 0x6644a000 (1715773440) bytes of LargeMmapAllocator (errno: 12)

But isn’t that a crash in ASan rather than libass? :) I haven’t worked with ASan myself, so I’m not sure how to read that.

brandonprry:
I will need to do more testing but I see your point.

AFL however wouldn't have labelled it as a crash if it hadn't crashed (and I don't fuzz with ASAN enabled). Looking at the valgrind results, it looks like it didn't actually crash there. AFL fuzzes with 50mb of RAM for the process to use which could easily lead to an OOM.

brandonprry:
Are you concerned that the issue shouldn't have a CVE?

grigorig:
I wasn't able to make libass crash while running that test case without ASAN. It looks like libass handles OOM situations OK here.

grigorig:
The test case runs very slowly if you have a lot of memory. So I limited memory with ```ulimit -v``` to 1024 MB  (I also tested 512 MB) and ran the test. No crash, although libass obviously hit some of the error paths as it executed quite quickly. valgrind didn't report any problems either, I only saw these warnings:

    ==8390== Warning: set address range perms: large range [0xac92028, 0x2725f28f) (noaccess)

brandonprry:
Yep, those are the symptoms I see as well. 

If you are making an argument that this issue should not be allocated a CVE, I can reply back to the email thread and clarify and ask for the CVE for that issue be revoked.

grigorig:
I don't have a strong opinion about the CVE.

brandonprry:
Sorry about the miscommunication about the results of the testcase. Let me know if you want anything done. I do believe it should be behaviour on the record at the very least.

grigorig:
Here are the samples, for archival purposes.

[samples.zip](https://github.com/libass/libass/files/512382/samples.zip)

PR #241, opened 2016-10-05
Add text justification

DanOscarsson:
Subtitle recommendations often include that multi line
subtitles should be left justified as this is easier for
the eyes. This is also the standard used by several television
companies.

This add the possibility to define how subtitles are to
be justified, independently of where they are aligned.
The most common way could be to set justify to left, and have
alignment to center. But you can, for example, have alignment
to left and justify to center, giving subtitles to the left but
justifed on the center (instead of normal left justified).
Using justify right and alignment of center, might be good
choice for Arabic.

If justify is not defined, all works like before.
If justify is defined, subtitles are aligned as defined
by alignment and justified as defined by justify.

ASS is not extended by this, justify can only be defined
by setting Justify to wanted justification.


libass/ass_types.h
```diff
@@ -67,6 +70,7 @@ typedef struct ass_style {
     double Outline;
     double Shadow;
     int Alignment;
+    int Justify;
```
wm4: You'll have to add this to the end of the struct, or you'll break ABI.

wm4:
> On that note, Blur is not currently handled by handle_selective_style_overrides. @wm4, is this intentional? If so, Justify probably shouldn’t be handled by it either?

Well, it probably _could_ be put under one of those overrides, though I don't see a strong immediate need either.

DanOscarsson:
Moved Justify to the end and added a constant for the value 0.

As the text lines are left justified from beginning, to make them right justified (which is the default with right alignment) to have to know the max width to calculate the shift needed for each separate line. It is also needed for some of the other cases.

astiob:
@wm4 Well, it just seems to me that the behaviour regarding `Blur` and `Justify` should be the same: both are parameters that are not part of the SSA/ASS style definition and can only be overridden by the user. Essentially we just need to pick one behaviour, any behaviour, and stick to it.

> As the text lines are left justified from beginning, to make them right justified (which is the default with right alignment) to have to know the max width to calculate the shift needed for each separate line. It is also needed for some of the other cases.

Not exactly: as far as this function is concerned, the coordinate of the anchor point (whether the right margin or the centre point) is completely arbitrary, and could be, for example, zero. The event needs to be repositioned (later) relative to the anchor point to get screen coordinates anyway. …Equivalently, you could say zero corresponds to an arbitrary position. Why does zero have to mean the left margin? The margin needs to be added later anyway (just a few lines after the call to `align_lines`, in fact!), so why not add the anchor point instead of the margin?

Looking at the code again, I’m not entirely sure it would be a net gain in code simplicity, but with the added justification logic in `align_lines` it might become especially worthwhile, as the `align_lines` would get vastly simpler. The code in `ass_render_event` already handles vertical alignment anchor points, so handling horizontal alignment in the same place might also make the code more easy to understand.

DanOscarsson:
I can move Justify to _reset_render_context_ just like Blur. You who know the code better have to decide.
I tested to set it same way as Blur in _reset_render_context_ and it looks like it works for Justify too, the way I use it from mpv.

Should we pull this in first before trying to simplify alignment/justification code?

DanOscarsson:
I have not heard what you prefer but now I have changed so Justify is applied in _reset_render_context_ just like Blur.

libass/ass_render.c
```diff
@@ -863,6 +863,7 @@ void reset_render_context(ASS_Renderer *render_priv, ASS_Style *style)
     render_priv->state.hspacing = style->Spacing;
     render_priv->state.be = 0;
     render_priv->state.blur = style->Blur;
+    render_priv->state.justify = style->Justify;
```
astiob: I’ve just looked again at where this code is and realized that we don’t actually want to reset `justify` on every `\r`, i. e. in `reset_render_context`. It’s a property of the whole event rather than of individual glyphs, so it doesn’t make sense to change it mid-event, and I think we should set it once per event in `init_render_context` like we set `alignment`. I should’ve spotted this the first time ’round, sorry.

astiob:
Oh, the move away from `handle_selective_style_overrides` means that the user can no longer set the justification through the (single) selective override style. Instead, every style has to be modified separately. (Regardless of whether justification is reset in `reset_render_context` or `init_render_context`.) The only easy way to do this is by using `ass_set_style_overrides` & `ass_process_force_style`, so we really want to add this to `ass_process_force_style`.

> Should we pull this in first before trying to simplify alignment/justification code?

Certainly.

astiob:
Thinking more about it, having it in selective overrides was probably better after all, but conditional on an appropriate bit. The reason is that overriding `Justify` can break typeset signs, so it shouldn’t be overridden for typeset signs, and the selective override mechanism handles this. We actually have selective `Alignment` overrides with `ASS_OVERRIDE_BIT_ALIGNMENT`, so perhaps `Justify` could use the same bit.

wm4:
Agreed.

DanOscarsson:
Better now? Or do you want a separate override bit for justify?

libass/ass_types.h
```diff
@@ -28,6 +28,10 @@
 #define HALIGN_LEFT 1
 #define HALIGN_CENTER 2
 #define HALIGN_RIGHT 3
+#define JUSTIFY_AUTO 0
+#define JUSTIFY_LEFT 1
+#define JUSTIFY_CENTER 2
+#define JUSTIFY_RIGHT 3
```
wm4: As a last change, could you prefix those with ASS_?

I know the other defines don't do this, but it's a good good convention to follow (since these are public symbols) and I'd argue the other defines should be switched to this convention at some point.

wm4:
A separate bit might be a good idea (I don't know), but it can always be added later. Until we have a reason to do so it's probably better not to do that.

Requested a last change, if that's done then I'll merge.

astiob:
I think whether to have a separate bit is up to you, @DanOscarsson, right now. You’re going to use `Justify` yourself, right? What use do you have in mind? Are you happy to force all subtitles to a specific alignment and justification, or would you rather left-justify all subtitles but keep the alignment as specified in the original script? If the former, a single bit is enough, but if the latter, you’ll need separate bits.

> it can always be added later

Yes, but won’t it break backwards compatibility then? Clients setting the alignment bit will expect justification to happen too, but it won’t. Of course, we can work around this by keeping this bit’s meaning as alignment-and-justification and adding two new bits for alignment-only and justification-only, but that’s ugly. I’m not saying we should set it in stone now, but perhaps we should put a little bit of thought in this decision.

wm4:
>Yes, but won’t it break backwards compatibility then? 

Maybe so, but style overriding is naturally a very fuzzy thing full of heuristics. But yeah you could be right.

DanOscarsson:
I have very few videos with ass subtitles. I had one anime saved for testing other things that used ass and different alignment. Tested it with mpv and at least for that one it looked better by not overriding alignment. So I have added a separate override bit. I also added ASS_ prefix for the constants.

DanOscarsson:
Hmm, I thought colse and comment meant close the comment.

wm4:
Looks good to me now. Good to merge @DanOscarsson @astiob ?

wm4:
Merged via e54c123d5a08b6212533ddcced2cb1a50fa3d2b2.

wm4:
(And sorry for the delays.)

astiob:
Thanks, @DanOscarsson and @wm4!

PR #242, opened 2016-10-06
Fix FT types

ajaska:
This patch is a few months old, some parts might have changed?

Dunno if this is a good thing, haven't done real profiling, might have missed a few spots. But fixes the error from #229 and improves the type semantics.

@grigorig

### DIFF
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -430,8 +430,8 @@ static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
     metrics->hash_key.font = info->font;
     metrics->hash_key.face_index = info->face_index;
     metrics->hash_key.size = info->font_size;
-    metrics->hash_key.scale_x = double_to_d6(info->scale_x);
-    metrics->hash_key.scale_y = double_to_d6(info->scale_y);
+    metrics->hash_key.scale_x = double_to_d16(info->scale_x);
+    metrics->hash_key.scale_y = double_to_d16(info->scale_y);
 
     return hb_fonts[info->face_index];
 }
--- a/libass/ass_cache.h
+++ b/libass/ass_cache.h
@@ -44,7 +44,7 @@ typedef struct {
     ASS_Outline *border;
     FT_BBox bbox_scaled;        // bbox after scaling, but before rotation
     FT_Vector advance;          // 26.6, advance distance to the next outline in line
-    int asc, desc;              // ascender/descender
+    FT_Long asc, desc;          // ascender/descender
 } OutlineHashValue;
 
 typedef struct {
--- a/libass/ass_cache.h
+++ b/libass/ass_cache.h
@@ -44,7 +44,7 @@ typedef struct {
     ASS_Outline *border;
     FT_BBox bbox_scaled;        // bbox after scaling, but before rotation
     FT_Vector advance;          // 26.6, advance distance to the next outline in line
-    FT_Long asc, desc;          // ascender/descender
+    FT_F26Dot6 asc, desc;       // ascender/descender
 } OutlineHashValue;
 
 typedef struct {
--- a/libass/ass_cache_template.h
+++ b/libass/ass_cache_template.h
@@ -65,8 +65,8 @@ START(outline_bitmap, outline_bitmap_hash_key)
     GENERIC(FT_Long, frx) // signed 10.22
     GENERIC(FT_Long, fry) // signed 10.22
     GENERIC(FT_Long, frz) // signed 10.22
-    GENERIC(FT_Long, fax) // signed 16.16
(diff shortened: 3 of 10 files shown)


grigorig:
You are replacing a lot (all?) of these ```FT_Long```s. Might as well squash both patches?

wm4:
I see absolutely no point in this. It only makes our code more dependent on freetype. The typedefs obscure the real types and make it harder to see what is going on. If freetype ever changes the typedefs, everything will probably break.

ajaska:
I see where you're coming from, it's nice to know exactly what types you're working with. And I'm not incredibly attached to this patch or anything, I just noticed weirdness while tracking down that other bug.

On the other hand, this improves the semantic meaning of the variables and can make it more clear why something is being done. (e.g. why random ints are being multiplied by 64.) Remember that many potential contributors are less familiar with the codebase than you are.

I think it's also unlikely that freetype is going to introduce a massively breaking change like that, without bumping the major version and making it big news. Further, the code is already dependent on FT's internal definitions of their types, and I would even go as far as to say that by using the actual FT types and definitions, you do more to protect yourself against breaking changes.

Finally, this does fix actual arithmetic errors.

MrSmile:
Personally, I'm also against this. `FT_Long` is typedef for `long` which can have size of 4 or 8 bytes and that can lead to problems (and #229 is one example). Inconsistent range under 32 and 64-bit architectures is unnecessary complication and bug source. I want to get rid of FT types altogether and replace its with fixed size types from `stdint.h`, especially for 16.16 and such.

wm4:
Having our own typedefs that indicate semantic meaning (like those fixed point formats) sounds like a good idea.

grigorig:
I do think more descriptive types are good (that's why I like this general idea). I also see that more dependencies on FreeType are unnecessary. So I guess a good approach for now might be to introduce types like ```ASS_16d16```, ```ASS_26d6``` for abstraction. These can then map to the FreeType types, to ```int32_t``` or whatever we want. What do you think?

libass/ass_shaper.c
```diff
@@ -430,8 +430,8 @@ static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
     metrics->hash_key.font = info->font;
     metrics->hash_key.face_index = info->face_index;
     metrics->hash_key.size = info->font_size;
-    metrics->hash_key.scale_x = double_to_d6(info->scale_x);
-    metrics->hash_key.scale_y = double_to_d6(info->scale_y);
+    metrics->hash_key.scale_x = double_to_d16(info->scale_x);
+    metrics->hash_key.scale_y = double_to_d16(info->scale_y);
```
astiob: What’s the reason for this change?

astiob:
We do use FreeType, so in many cases, using the FreeType types would be the right thing to do. This would ensure things don’t break if FreeType ever changes anything, not the other way round. And this would fix things that are already broken because of us currently using different types. Maybe this isn’t universally applicable, but then I’d prefer to see each instance carefully reviewed.

In any case, I agree with the general idea that type names more descriptive than `int` and `unsigned` might be useful.

I also agree that it’s somewhat annoying that FreeType’s has a variable width. On this note, I’d like to reiterate what I said in #229: we should (and don’t—I don’t at least) have a very clear and unambiguous understanding of whether values longer than 32 bits are actually valid, even if only in 64-bit builds. If they are, we _need to handle them_. If they are not, we need to _ensure we never produce them_: #229 shows that we currently do.

> FT_Long is typedef for long which can have size of 4 or 8 bytes and that can lead to problems (and #229 is one example).

No, the problem is caused by using two different types for the same thing at the same time. The size difference is not an inherent bug source. If we actually made sure to use the same type and range everywhere, we wouldn’t have the bug. There are plenty of types beside `FT_Long` that have variable widths: they aren’t inherently bug sources; misusing them is.

astiob:
If this does go forward with FreeType types (as much as this doesn’t seem very likely at the moment), it would be nice to know how and why @ajaska chose which of the various FreeType types to use for any particular variable: I see `FT_Pos`, `FT_F26Dot6`, `FT_Long` and `FT_Fixed`.

ajaska:
Yeah, FWIW I just went through the [FreeType docs](https://www.freetype.org/freetype2/docs/reference/ft2-toc.html), and matched the variable types to the source FT functions.

- `FT_Pos`, returned by FreeType, can be an integer, 16.16, or 26.6 based on context (:unamused:).
- `FT_F26Dot6` is the generic 26.6 type
- `FT_Fixed` is the generic 16.16 type
- To be honest, I'm not sure where `FT_Long` came from, but it's only used in `double_to_d22` which is only used once, directly below it.
- `FT_Fixed` on line 424, 425 of `add_line` should be `FT_F26Dot6`

PR #246, opened 2016-10-15
Fix overflow in strtod function

grigorig:
Just fix this for now. We may drop ass_strtod completely, but this is still being discussed.


astiob:
I’d prefer detecting and avoiding the overflow in the first place, as it’s undefined behaviour.

rcombs:
Yeah, a particularly clever compiler could optimize this check out.

astiob:
Fixed by #247.

Issue #244, opened 2016-10-13
Global buffer overflow in ass_strtod


eugenis:
I don' recall why this code is used at all instead of strtod(), which is C89(!). Windows?

ass_strtod fails on "14E0888888888888888888888888888880000000000000000000000000000".
ass_strtod.c:235  reads powersOf10[43].

```
for (d = (double *) powersOf10; exp != 0; exp >>= 1, d += 1) {
    if (exp & 01) {
        dblExp *= *d;  // <--- this line
    }
}
```

Test case attached.

ERROR: AddressSanitizer: global-buffer-overflow on address 0x0000005b74d8 at pc 0x000000571d09 bp 0x7fffba929010 sp 0x7fffba929008
READ of size 8 at 0x0000005b74d8 thread T0
    #0 0x571d08 in ass_strtod /code/libass/libass/ass_strtod.c:235:23
    #1 0x4ff04b in process_style /code/libass/libass/ass.c:544:13
    #2 0x4fd20d in process_styles_line /code/libass/libass/ass.c:600:9
    #3 0x4fbce1 in process_line /code/libass/libass/ass.c:809:13
    #4 0x4f648f in process_text /code/libass/libass/ass.c:843:9
    #5 0x4f9b09 in parse_memory /code/libass/libass/ass.c:1141:5
    #6 0x4f9e03 in ass_read_file /code/libass/libass/ass.c:1241:13
    #7 0x4f1e1a in main /code/libass/test/test.c:208:24

[crash.ass.txt](https://github.com/libass/libass/files/528354/crash.ass.txt)

astiob:
`strtod` isn’t used because it’s locale-dependent.

eugenis:
Posix has uselocale() and afaik windows also has something similar.

rcombs:
Do we even need to handle scientific notation here?

rcombs:
(I checked and the answer is yes; VSFilter uses wcstod, which does)

astiob:
[...]

PR #247, opened 2016-10-30
ass_strtod fixes

astiob:
Several fixes for `ass_strtod` including the buffer overrun #244.

Hm, now that I think about it, #244 is fixed by two of these commits in tandem. The crash should be fixed by the very first commit, and the wrong output value is fixed by 8df6e2f. How can I reword the commit messages to make this clear?

In the first commit, I assume that `size_t` is enough to hold any substring size, although I’ve been unable to find a guarantee of this in the C standard. I avoid `ptrdiff_t` because it definitely can overflow.

I’ve kept the code C90 because it’s rather self-contained and can be usefully reused in other projects.

With these changes, `ass_strtod` almost conforms to C90 assuming the `"C"` locale and IEEE double (or any other double that has appropriate mantissa and exponent limits). It can produce results that are off by an ulp and maybe more, but C doesn’t actually mandate the precision of decimal string-to-double conversions. The only other problem is that `errno` is not always set on overflow or underflow. For libass, this doesn’t matter, as we never check `errno` after a `ass_strtod` call anyway. As for C99 conformance, `ass_strtod` doesn’t even come close, as C99 added syntax for NaN, infinities and hexadecimal numbers to `strtod`, none of which is implemented in `ass_strtod`. I believe it _shouldn’t_ be, either, as ASS doesn’t support these syntaxes to the best of my knowledge.

### DIFF
--- a/libass/ass_strtod.c
+++ b/libass/ass_strtod.c
@@ -17,6 +17,7 @@
 #include "ass_compat.h"
 
 #include <stdlib.h>
+#include <float.h>
 #include <errno.h>
 #include "ass_string.h"
 
@@ -38,6 +39,19 @@ const double powersOf10[] = {   /* Table giving binary powers of 10.  Entry */
     1.0e64,
     1.0e128,
     1.0e256
+};
+
+static
+const double negPowOf10[] = {   /* Table giving negative binary powers */
+    0.1,                        /* of 10.  Entry is 10^-2^i. */
+    0.01,                       /* Used to convert decimal exponents */
+    1.0e-4,                     /* into floating-point numbers. */
+    1.0e-8,
+    1.0e-16,
+    1.0e-32,
+    1.0e-64,
+    1.0e-128,
+    1.0e-256
 };
 
 /*
@@ -175,10 +189,11 @@ ass_strtod(
         p = string;
         goto done;
     } else {
-        int frac1, frac2;
+        int frac1, frac2, m;
         mantSize -= leadZeros;
+        m = mantSize;
         frac1 = 0;
-        for ( ; mantSize > 9; mantSize -= 1)
+        for ( ; m > 9; m -= 1)
         {
             c = *p;
             p += 1;
@@ -189,7 +204,7 @@ ass_strtod(
             frac1 = 10*frac1 + (c - '0');
         }
         frac2 = 0;
-        for (; mantSize > 0; mantSize -= 1)
+        for (; m > 0; m -= 1)
         {
             c = *p;
             p += 1;
@@ -274,8 +289,32 @@ expOverflow:
             errno = ERANGE;
         }
     }
+    /* Prefer positive powers of 10 for increased precision, especially
+     * for small powers that are represented exactly in floating-point. */
+    if ((exp <= DBL_MAX_10_EXP) || !expSign) {
+        d = powersOf10;
+    } else {
+        /* The floating-point format supports more negative exponents
+         * than positive, or perhaps the result is a subnormal number. */
+        if (exp > -DBL_MIN_10_EXP) {
+            /* The result might be a valid subnormal number, but the
+             * exponent underflows.  Tweak fraction so that it is below
+             * 1.0 first, so that if the exponent still underflows after
+             * that, the result is sure to underflow as well. */
+            exp -= mantSize;
+            dblExp = 1.0;
+            for (d = powersOf10; mantSize != 0; mantSize >>= 1, d += 1) {
+                if (mantSize & 01) {
+                    dblExp *= *d;
+                }
+            }
+            fraction /= dblExp;
+        }
+        d = negPowOf10;
+        expSign = 0;
+    }
     dblExp = 1.0;
-    for (d = powersOf10; exp != 0; exp >>= 1, d += 1) {
+    for (; exp != 0; exp >>= 1, d += 1) {
         if (exp & 01) {
             dblExp *= *d;
         }
--- a/libass/ass_strtod.c
+++ b/libass/ass_strtod.c
@@ -95,6 +95,7 @@ ass_strtod(
     size_t mantSize;    /* Number of digits in mantissa. */
     size_t decPt;       /* Number of mantissa digits BEFORE decimal
                          * point. */
+    size_t leadZeros;   /* Number of leading zeros in mantissa. */
     const char *pExp;       /* Temporarily holds location of exponent
                              * in string. */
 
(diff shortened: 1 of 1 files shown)


grigorig:
Thanks. Sorry, this got sidetracked on my side.

Do huge exponents with non-zero coefficient result in ±```Inf``` as the result with these applied? That was still an issue with my (broken) fix, and I don't see any code here to address that. Not sure if it should be considered an issue at all.

astiob:
Huge positive exponents give ±infinity and huge negative exponents give ±zero, as expected. Was this broken (other than the exponent overflow)? This happens automatically when multiplying together the `powersOf10` (assuming IEEE double). The `maxExponent` in the code is 511, but actually any exponent beyond +308 or −323 will give infinity or zero already.

Oh, I’ve just realized there is in fact a problem with very low exponents. Two problems even.

One is that denormal numbers are almost never produced, e. g. `1e-323` gives zero because `ass_strtod` tries to compute `dblExp=1e+323` and then take its reciprocal, but `dblExp` overflows to infinity. Still, technically, this behaviour conforms to the spec, as this qualifies as underflow and `strtod` is free to return a denormal number or zero, or even the smallest normal number.

The other is that e. g. while `1e-308` is converted correctly, `1.0e-308` gives zero, as it moves the decimal point and adjusts the exponent to `10e-309`, and `dblExp=1e+309` overflows. Even if denormals are fixed, this will still be a problem; it will just affect smaller numbers, e. g. `1e-323` and `1.0e-323`. I’m not sure how this can be fixed.

astiob:
Huge positive exponents give ±infinity and huge negative exponents give ±zero, as expected. Was this broken (other than the exponent overflow)? This happens automatically when multiplying together the `powersOf10` (assuming IEEE double). The `maxExponent` in the code is 511, but actually any exponent beyond +308 or −323 will give infinity or zero already.

Oh, I’ve just realized there is in fact a problem with very low exponents. Two problems even.

One is that denormal numbers are almost never produced, e. g. `1e-323` gives zero because `ass_strtod` tries to compute `dblExp=1e+323` and then take its reciprocal, but `dblExp` overflows to infinity. Still, technically, this behaviour conforms to the spec, as this qualifies as underflow and `strtod` is free to return a denormal number or zero, or even the smallest normal number.

The other is that e. g. while `1e-308` is converted correctly, `1.0e-308` gives zero, as it moves the decimal point and adjusts the exponent to `10e-309`, and `dblExp=1e+309` overflows. Even if denormals are fixed, this will still be a problem; it will just affect smaller numbers, e. g. `1e-323` and `1.0e-323`. I’m not sure how this can be fixed.

(Really, no Markdown in email replies? Wasn’t Markdown designed based on email conventions?)

astiob:
For reference, compared to my solution with `expWraparound`, glibc (which has a completely different implementation of `strtod`, but this part is similar) simply [makes](http://repo.or.cz/glibc.git/commitdiff/d6e70f4368533224e66d10b7f2126b899a3fd5e4) `exp` an `intmax_t` and asserts (at runtime!) that it doesn’t overflow, which implies that `intmax_t` has to be larger than `size_t`. I chose not to make this assumption.

astiob:
> Huge positive exponents give ±infinity and huge negative exponents give ±zero, as expected. Was this broken (other than the exponent overflow)?

And possibly to answer my own question, old `ass_strtod` would produce NaN for a huge exponent and a mantissa with many leading zeros, e. g. for `0.000000000000000001e600`. With my changes, this gives +infinity (or a finite number if there’s really many leading zeros).

Meanwhile, I’ve discovered that the `strtod` implementation in [gdtoa](http://www.netlib.org/fp/), which is used by [FreeBSD](https://github.com/freebsd/freebsd/blob/master@{2016-11-02}/contrib/gdtoa/), [macOS](http://opensource.apple.com/source/Libc/Libc-1082.50.1/gdtoa/) and [Ruby](https://github.com/ruby/ruby/blob/trunk@{2016-11-02}/util.c#L544), fails on long input strings and on exponents larger in magnitude than 19999. For example, the string generated by the Python/Ruby code `'1' + '0'*20000 + 'e-20000'` clearly denotes the number 1, but `gdtoa` gives 10. On very long strings, length counters overflow `int` much like here and in the [glibc bug](https://sourceware.org/bugzilla/show_bug.cgi?id=14459) whose fix I mentioned above, and I don’t even want to imagine what happens then.

grigorig:
So these kinds of pathological cases are generally problematic, it seems. I'd say, don't care about these. As long as there's nothing going really wrong, like invalid reads, the code is probably fine as-is. If the result of conversions from pathological cases is correct, even better. So, LGTM.

astiob:
Those cases are actually already handled by my changes in this PR, but it does show that every implementation has made the same mistakes and that the system `strtod` is not entirely reliable.

What’s left to fix in `ass_strtod` is `1.0e-323` (which gdtoa and glibc convert correctly), and I’ve been thinking about how best to do it but haven’t come to a decision yet. If you want, this can be merged without waiting for that, as it will be a separate commit anyway. The only change I might want to make to the commits already in this PR is to move the “fixes #244” line to the first commit or remove it entirely, but I’m still unsure which way is better.

astiob:
I’ve added a commit to fix `1.0e-323` and slightly edited the old commits to make it clear in the commit messages how this relates to #244 and to better fit the style of the existing code (it parenthesizes `||` operands, but I forgot to do this).

I was reluctant about this fix for `1.0e-323`, but actually it doesn’t appear to be any less precise than the existing code. The worst error I’ve gotten in my tests is still ±1 ulp, as before. For reference, `1e-306` converts to 1 ulp less than the correctly rounded value, and `100000000000000000e-323` (the same number) converts to 1 ulp more than the correctly rounded value.

It’s a bit annoying that e. g. `1.0e-307` and `1.00e-307` convert to different numbers, but then so do `1e-307` and `1.0e-307` even before this commit (and before this PR).

grigorig:
This PR goes far beyond just fixing the overflow issue, so I just merged it now. Evgeniy rightfully asked why this hasn't been fixed in mainline yet.

astiob:
I’ve just realized that for 858196e931e0076c61bc9c816b032a2f4991661c (which was 6e8ec146b66ed19177839b4ac4db1a64611b136d in the PR), an even simpler test case is `0000000000000000001`. Old `ass_strtod` converted this to zero.

(Also, 86f65f88d805372fa53d64cb4d08e98180bd735f refers to a commit hash that exists in this PR but not in master, alas.)

PR #249, opened 2016-11-06
Fix ass_set_cache_limits() to work with composite cache

MrSmile:
I've not tested this (does mpv have setting for libass cache size?) so people who knows software that relies on `ass_set_cache_limits` are welcomed.


astiob:
Seems sensible.

libass/ass_render_api.c
```diff
@@ -180,8 +180,18 @@ void ass_set_cache_limits(ASS_Renderer *render_priv, int glyph_max,
                           int bitmap_max)
 {
     render_priv->cache.glyph_max = glyph_max ? glyph_max : GLYPH_CACHE_MAX;
-    render_priv->cache.bitmap_max_size = bitmap_max ? 1048576 * bitmap_max :
-                                         BITMAP_CACHE_MAX_SIZE;
+
+    size_t bitmap_cache, composite_cache;
+    if (bitmap_max) {
+        bitmap_cache = 1048576 * (size_t) bitmap_max;
```
astiob: It’s been asked on IRC that 1048576 should be made a macro here, perhaps named something like `BITMAP_CACHE_SIZE_UNIT`. I realize this commit doesn’t introduce this constant, but it seems like a good opportunity to name it and make it just a bit more obvious how the cache size is configured.

astiob:
(Merged in 3ba18f666e87321c28d4a7281f808947c0712465.)

PR #251, opened 2016-12-07
Only enable DirectWrite if the Dwrite.h header is found

sdroege:
Older Windows versions don't have this.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -190,7 +190,8 @@ if test x$enable_directwrite != xno; then
 AC_MSG_CHECKING([for DIRECTWRITE])
 AC_LINK_IFELSE([
   AC_LANG_PROGRAM(
-    [[#include <windows.h>]],
+    [[#include <windows.h>
+      #include <Dwrite.h>]],
     [[;]],)
   ], [
     AC_DEFINE(CONFIG_DIRECTWRITE, 1, [found DirectWrite])


astiob:
Yes, but you can build libass on older Windows and use it on newer Windows. DirectWrite is detected and enabled/disabled at runtime.

wm4:
This makes no sense. We don't ever include dwrite.h. We have our own header, dwrite_stripped.h, because the mingw headers are hopelessly broken.

sdroege:
Without this, the build fails spectacularly like:

    ass_directwrite.c:69:5: error: unknown type name '_In_opt_'
    ass_directwrite.c:70:5: error: unknown type name '_Out_'
    ass_directwrite.c:79:5: error: unknown type name '_In_opt_'
    [...]

astiob:
For reference, what are you building on?

sdroege:
mingw64 cross compiler on Linux, targetting >= Windows XP

wm4:
The libass code is designed in such a way that it can be compiled for newer Windows versions, but run on XP. It will work on the latter because the header only imports compile-time definitions, and the actual dwrite DLL is loaded dynamically. I'm not sure why you need to do this targetting business (maybe for other software?).

Also I was wrong, the stripped header we use is apparently  libass/dwrite_c.h.

astiob:
Is that mingw-w64? As far as I can tell, it should have <sal.h> with definitions for `_In_opt_` etc., which is included from <specstrings.h>, which is included from a bunch of places including <windef.h>, from <unknown.h>, from "libass/dwrite_c.h".

Do you have the definitions at all? Are they completely absent, or hidden because of the targeting, or is the include chain broken so ass_directwrite.c never sees them?

We can either add another include if that fixes it or simply remove the annotations from ass_directwrite.c.

astiob:
Actually, I’m not so sure any more that it was supposed to build DirectWrite on Windows XP. I remember we had it load DirectWrite dynamically to improve compatibility across different Windows versions, but did it go both ways? (It would be best if it went both ways, but I don’t remember if we made it so.)

feliwir:
I think the problem here is that these aren't defined when not including dwrite.h: https://msdn.microsoft.com/en-us/library/hh916382.aspx .We should maybe define those in our dwrite header if not available

sdroege:
_In_opt_ is not defined anywhere in my mingw headers, there is also no sal.h

feliwir:
Yeah easiest would be to define those ourselves

astiob:
Or just not to use them. They’re only used in our own ass_directwrite.c, not in dwrite_c.h. I don’t know if they’re used in any of the other headers that we include and care about, but, as far as I understand, it should be fine in any case to omit them or to define them empty.

@sdroege It seems the original MinGW may not include sal.h with these macros. I think we’ll probably fix this in libass, but wouldn’t you be better off using MinGW-w64 regardless?

wm4:
> but wouldn’t you be better off using MinGW-w64 regardless?

Or just dropping XP.

astiob:
> Or just dropping XP.

Doesn’t matter. Dropping XP won’t fix MinGW.

And a brief search shows XP is still the third most used OS worldwide, losing only to Windows 10 and Windows 7 and certainly coming miles ahead of many other OSs that we nevertheless happily support, and possibly still the most used one in some countries. Supporting XP doesn’t take all that much effort most of the time. You’re wasting more effort arguing how XP support should be dropped than supporting it.

wm4:
> Doesn’t matter. Dropping XP won’t fix MinGW.

And MinGW as opposed to MinGW-w64.

> And a brief search shows XP is still the third most used OS worldwide, 

Even Chrome is dropping XP support. It's over.

sdroege:
@astiob I'm using mingw64, it's just apparently an older version. We're planning to update some time in the future though.

astiob:
@sdroege Can you please test the [commit](https://github.com/astiob/libass/commit/d4b9a932f2990b5b5524a70e8d4a6fbd8943e7a7) I just pushed to my fork?

sdroege:
Almost:

    Creating library file: .libs/libass.dll.a
    .libs/ass_directwrite.o: In function `FallbackLogTextRenderer_QueryInterface':
    libass-0.13.4/libass/ass_directwrite.c:193: undefined reference to `IID_IUnknown'
    collect2: error: ld returned 1 exit status

astiob:
It appears that your MinGW-w64 is older than [this commit](https://sourceforge.net/p/mingw-w64/mingw-w64/ci/81c3ea715bda3b1cef65740613d227515d963ba7/). Right?

This symbol should be defined by <Unknwn.h> when it is included from dwrite_c.h after <InitGuid.h> has already been included (from ass_directwrite.c). But MinGW-w64 failed to define it before the above commit.

Our options:
* Link against Uuid.Lib, which contains an external definition of this symbol.
* Check this in configure and then link against Uuid.Lib if the definition is missing.
* Somehow (how?) try to detect whether there is a definition in dwrite_c.h and define our own copy if it’s missing.
* Always define our own copy in dwrite_c.h.
* Provide our own copy of unknwn.h.
* Ignore this, annoy @sdroege and hope he switches to a newer MinGW-w64.

sdroege:
Yes it's older than that. Feel free to just ignore me, I'm fine with keeping a patch that disables DirectWrite for the time being until we switched to a newer toolchain

astiob:
I don’t know what your use case is, but if your build is going to be used on Vista or newer (rather than only on XP), it would be better to keep DirectWrite enabled. (If it’s going to be XP only, then you might want to disable DirectWrite even if it did build successfully, as it’s completely useless on XP.) Try to add this line just after `#include <unknwn.h>` in dwrite_c.h in addition to my commit:

    DEFINE_GUID(IID_IUnknown, 0x00000000, 0x0000, 0x0000, 0xc0,0x00, 0x00,0x00,0x00,0x00,0x00,0x46);

sdroege:
That works, yes. Including unknown.h (with the o) should also work as it defines the GUID exactly like that.

astiob:
I’ve just pushed the commit that removes SAL annotations from our code. I take it that we don’t want to work around the undefined `IID_IUnknown` issue and that you’re happy using one of the workarounds on your end until you upgrade to a newer mingw-w64, so I’m closing this issue.

PR #257, opened 2017-01-13
render_api: do not discard old images on reconfiguration

wm4:
I noticed that when resizing the mpv window while playback is ongoing
and with subtitles, that subtitles could sometimes get "stuck" on the
screen. The stuck subtitle would remain until the next subtitle event,
or until seeking to a position that has subtitles again.

It turned out that this was a libass change detection bug. The following
steps should reproduce the problem:

1. call ass_render_frame() with a time that has subtitles
2. call ass_set_frame_size() with a different size
3. call ass_render_frame() with a time that has no subtitles

The previous call will return with *detect_change==0.

To make this worse, libass will deallocate image data before the next
ass_render_frame() or ass_renderer_done(), which violates the API and
could possibly make some API users crash. (That the user can rely on
this is not documented though.)

There are two possible solutions:

1. Set a flag in ass_reconfigure(), that makes the next
   ass_render_frame() call always return *detect_change==2.
2. Do not discard the previous subtitles (images_root), so change
   detection can work reliably.

This commit implements 2. - I prefer this in part because it doesn't
clobber the previously returned image list before the next
ass_render_frame() call. (As pointed out above, this might be unexpected
behavior to the API user.)

This is a regression and was possibly broken by commit dd06ca and later.
I did not check whether it actually behaved sanely before that change,
but it probably did to a degree.

----

CC @MrSmile 

I'm not sure whether this really works. Are there any bitmap allocations that are not properly refcounted? What about the weird `priv->state.has_clips` case in `ass_detect_change`?

### DIFF
--- a/libass/ass_render_api.c
+++ b/libass/ass_render_api.c
@@ -27,11 +27,9 @@ static void ass_reconfigure(ASS_Renderer *priv)
     ASS_Settings *settings = &priv->settings;
 
     priv->render_id++;
-    ass_frame_unref(priv->images_root);
     ass_cache_empty(priv->cache.composite_cache);
     ass_cache_empty(priv->cache.bitmap_cache);
     ass_cache_empty(priv->cache.outline_cache);
-    priv->images_root = NULL;
 
     priv->width = settings->frame_width;
     priv->height = settings->frame_height;


astiob:
Just for reference, I do think this was broken by commit dd06ca30ea79ce50116a43cc5521d4eaf60a017e, although I haven’t tested. `ass_reconfigure` used to delete `prev_images_root` (which, actually, shouldn’t have had any effect at all, as `prev_images_root` is deleted at the end of `ass_render_frame`), but this commit made it delete `images_root` instead.

Incidentally, change detection after `ass_reconfigure` had always been broken, but slightly differently:

* `ass_reconfigure` clears the caches, but before #187 this meant freeing the memory referenced by ASS_Images in the most recent frame. So in fact, libass has always done this and this part wasn’t even a regression.

* The next problem occurred when change detection was run using this list of images whose bitmaps had been freed. Imagine reconfiguring libass between two very similar nonempty frames. Bitmaps allocated for the second frame could have the exact same addresses as the (now freed) bitmaps in the first frame, so change detection could incorrectly think there were no changes. I guess this is less likely when we clear all caches at once, so we never saw this happen in practice. Since 20f83b10241aa9c67c3a7e0c1a7e4d986fe199f1 there was a `cache_cleared` flag that `ass_reconfigure` should have set, but it never did.

wm4:
Yeah, I came to the same conclusion, though I was never sure whether the previous image data was really released, or if there was something that prevented it.

Issue #52, opened 2014-02-25
change detection bug with vector clips


ghost:
The second event doesn't have the change flag set:

```
Dialogue: 0,0:00:00.00,0:00:02.00,,,0,0,0,,{\fs50\pos(250,250)\iclip(m 0 0 l 400 0 l 400 220 l0 220)}Hello
Dialogue: 0,0:00:02.00,0:10:00.00,,,0,0,0,,{\fs50\pos(250,250)\iclip(m 0 0 l 400 0 l 400 230 l0 230)}Hello
```

(Complete test script: http://sprunge.us/jRWb)

The output ass_detect_change sees is the same. The vector clip merely makes the existing pixels transparent. Thus the change detection fails. This causes problems with applications which use this flag, such as mplayer and mpv.

The hard solution to this would be caching vector clips.

PR #258, opened 2017-01-13
render: remove redundant has_clips

MrSmile:
That variable was workaround for case when new frame images
reuse memory from previous frame images. Due to that reuse
physically different images could be indistinguishable in
ass_detect_change(). After cache PR such memory aliasing is
no longer possible and workaround is redundant.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -406,25 +406,17 @@ render_glyph(ASS_Renderer *render_priv, Bitmap *bm, int dst_x, int dst_y,
     b_y1 = bm->h;
 
     tmp = dst_x - clip_x0;
-    if (tmp < 0) {
+    if (tmp < 0)
         b_x0 = -tmp;
-        render_priv->state.has_clips = 1;
-    }
     tmp = dst_y - clip_y0;
-    if (tmp < 0) {
+    if (tmp < 0)
         b_y0 = -tmp;
-        render_priv->state.has_clips = 1;
-    }
     tmp = clip_x1 - dst_x - bm->w;
-    if (tmp < 0) {
+    if (tmp < 0)
         b_x1 = bm->w + tmp;
-        render_priv->state.has_clips = 1;
-    }
     tmp = clip_y1 - dst_y - bm->h;
-    if (tmp < 0) {
+    if (tmp < 0)
         b_y1 = bm->h + tmp;
-        render_priv->state.has_clips = 1;
-    }
 
     if ((b_y0 >= b_y1) || (b_x0 >= b_x1))
         return tail;
@@ -522,8 +514,6 @@ static void blend_vector_clip(ASS_Renderer *render_priv,
         int bx, by, bw, bh, bs;
         int aleft, atop, bleft, btop;
         unsigned char *abuffer, *bbuffer, *nbuffer;
-
-        render_priv->state.has_clips = 1;
 
         abuffer = cur->bitmap;
         bbuffer = clip_bm->buffer;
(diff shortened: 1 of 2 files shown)


wm4:
OK by me. CC @astiob maybe.

astiob:
TL;DR at the end of this post.

---

Some history: #52, #61, #104. The only example we’ve seen for #104 (on IRC) is:

    {\q2\fscx286\fscy286\move(1061,244,280,244,21,2982)}A Warning Has Been Issued

I don’t know what the screen or font size for this is supposed to be, but it was said that the text was wider than the screen. As you can see, it’s just moving horizontally across the screen. Presumably, there are multiple frames where the text covers the whole screen width and is clipped on both ends, and the content within this clipped rectangle is changing, but `ass_detect_change` for some reason didn’t detect this.

On IRC, we came to the conclusion that change detection failed because a bitmap was getting allocated at the same memory address that had been previously occupied by a different bitmap. I don’t know why we left the commit message in #104 like that, as it doesn’t seem to make sense. `render_glyph` is almost the last step in the whole rendering pipeline, and it merely stores pointers to full, unclipped bitmaps generated by earlier stages.

Moreover, I understand how address reuse happened in #52 (vector clips get rendered into fresh new bitmaps on every frame, which used to be freed at the start of the next frame, before change detection got to run—but since dd06ca30ea79ce50116a43cc5521d4eaf60a017e/#187 are freed at the end of the next frame, after change detection; so there `has_clips` should certainly safe to remove now), but I still don’t understand how it could happen in #104. `render_glyphs` doesn’t create new bitmaps, and the bitmaps it uses are stored in the cache, and bitmaps are never mutated once they land in the cache. How can the same address point to something different on the next frame? In fact, for the sample I gave above, you’d expect that the next frame either uses the same bitmap with a different offset or uses a different bitmap. In both cases, the pointer in the `ASS_Image` should be different.

Also, #104 didn’t modify `render_glyph_i`, which seems odd.

As I wrote this, I got one idea. The only way I can see the memory reuse in #104 happening is if the bitmap cache decided it was too large and cleared its contents at the start of the second frame.

Oh, but even then you might expect the new bitmap to be allocated at the same address, but the offset after clipping will be different! So somehow, `old_bitmap_pointer == new_bitmap_pointer + scrolling_speed_in_pixels_per_frame`. This seems quite unlikely.

---

All in all, it _seems_ that this PR should be safe, but it worries me that I don’t see why #104, which this is reverting, was needed or how it helped. Yet it did apparently solve a real problem. If we didn’t see the problem in the code back then, it may be that the problem still exists (and will be exposed again if this PR is merged) and we simply still haven’t spotted it.

astiob:
While #104 was being investigated, it came up that we had a `restride_bitmap` function. After a discussion on IRC, we concluded that we didn’t need this function. So @wm4 went ahead and removed it in 014837c4a138ae37addeaf72f90c131fcb3ea1fc, while #104 was still open. Then he rebased and pulled in #104.

You know what? #104 didn’t fix #104. 014837c4a138ae37addeaf72f90c131fcb3ea1fc did! The commit in #104 itself, as it is, works around the issue where a _restrided_ bitmap gets allocated in one frame, added to the free list (just like vector-clipped bitmaps in #52/#61), freed at the start of the next frame and then reallocated at the same address with different content. But after this commit was rebased on top of 014837c4a138ae37addeaf72f90c131fcb3ea1fc, it fixed nothing: restriding didn’t exist any more, so this issue didn’t occur.

I believe it is safe to remove `has_clips`.

wm4:
Woah. Nice archeology here. I guess I'm glad that the mess became much cleaner.

PR #259, opened 2017-01-30
Reflect border_scale in outline cache keys

astiob:
An actual example of breakage that this PR should fix was reported on IRC: when `ScaledBorderAndShadow` is changed in Aegisub, no difference is visible because the old outlines are fetched from the cache.

### DIFF
--- a/libass/ass_cache_template.h
+++ b/libass/ass_cache_template.h
@@ -90,7 +90,7 @@ START(glyph, glyph_hash_key)
     GENERIC(int, italic)
     GENERIC(unsigned, scale_x) // 16.16
     GENERIC(unsigned, scale_y) // 16.16
-    FTVECTOR(outline) // border width, 16.16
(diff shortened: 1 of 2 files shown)


libass/ass_cache_template.h
```diff
@@ -90,7 +90,7 @@ START(glyph, glyph_hash_key)
     GENERIC(int, italic)
     GENERIC(unsigned, scale_x) // 16.16
     GENERIC(unsigned, scale_y) // 16.16
-    FTVECTOR(outline) // border width, 16.16
+    FTVECTOR(outline) // border width, 26.6
```
wm4: Due to using _d6 instead of _d16? I guess I'm not sure why this needs a precision change, but it's somehow mandated due to border_scale changing the value range? We could also just avoid this "risk" and use a separate key.

wm4: Whoops, this was explained in the commit message, never mind.

PR #261, opened 2017-02-04
Integer overflow fixes

astiob:
A bunch of integer overflow fixes plus a few other fixes that I made along the way.

Unresolved issues:

  * Event timestamps can overflow `long long`. What should we do then? Saturate to LLONG_MIN/LLONG_MAX? This would be fine for the purposes of deciding which event to show when but could break animations that depend on the event duration. Then again, VSFilter doesn’t even support `long long` timestamps like we do. (It uses `int`. Which, of course, practically means `int32_t`.)

  * When `PlayResY` is set (to a positive value) but `PlayResX` isn’t, it is computed as ⁴⁄₃ × `PlayResY`. This value may not fit in `int`. What should we do then? VSFilter does `PlayResY * 4 / 3` (all in `LONG` aka `int32_t`) and doesn’t care. I forget if I’ve tested what actually happens then.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -1498,7 +1498,7 @@ void ass_lazy_track_init(ASS_Library *lib, ASS_Track *track)
             ass_msg(lib, MSGL_WARN,
                    "PlayResY undefined, setting to %d", track->PlayResY);
         } else if (track->PlayResY <= 0) {
-            track->PlayResY = FFMAX(1, track->PlayResX * 3 / 4);
+            track->PlayResY = FFMAX(1, track->PlayResX * 3LL / 4);
             ass_msg(lib, MSGL_WARN,
                    "PlayResY undefined, setting to %d", track->PlayResY);
         } else if (track->PlayResX <= 0 && track->PlayResY == 1024) {
@@ -1506,7 +1506,7 @@ void ass_lazy_track_init(ASS_Library *lib, ASS_Track *track)
             ass_msg(lib, MSGL_WARN,
                    "PlayResX undefined, setting to %d", track->PlayResX);
         } else if (track->PlayResX <= 0) {
-            track->PlayResX = FFMAX(1, track->PlayResY * 4 / 3);
+            track->PlayResX = FFMAX(1, track->PlayResY * 4LL / 3);
             ass_msg(lib, MSGL_WARN,
                    "PlayResX undefined, setting to %d", track->PlayResX);
         }
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -176,11 +176,12 @@ inline void change_alpha(uint32_t *var, int32_t new, double pwr)
  * \param a first value
  * \param b second value
  * \return result of multiplication
- * Parameters and result are limited by 0xFF.
+ * At least one of the parameters must be less than or equal to 0xFF.
+ * The result is less than or equal to max(a, b, 0xFF).
  */
 inline uint32_t mult_alpha(uint32_t a, uint32_t b)
 {
-    return 0xFF - (0xFF - a) * (0xFF - b) / 0xFF;
+    return a - (uint64_t) a * b / 0xFF + b;
 }
 
 /**
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -42,10 +42,10 @@ static inline int argtoi(struct arg arg)
     return value;
 }
 
-static inline long long argtoll(struct arg arg)
+static inline int32_t argtoi32(struct arg arg)
 {
-    long long value;
-    mystrtoll(&arg.start, &value);
+    int32_t value;
+    mystrtoi32(&arg.start, 10, &value);
     return value;
 }
 
@@ -188,8 +188,8 @@ inline uint32_t mult_alpha(uint32_t a, uint32_t b)
  * Used for \fad, \fade implementation.
  */
 static int
-interpolate_alpha(long long now, long long t1, long long t2, long long t3,
-                  long long t4, int a1, int a2, int a3)
+interpolate_alpha(long long now, int32_t t1, int32_t t2, int32_t t3,
+                  int32_t t4, int a1, int a2, int a3)
 {
     int a;
     double cf;
@@ -197,12 +197,14 @@ interpolate_alpha(long long now, long long t1, long long t2, long long t3,
     if (now < t1) {
         a = a1;
     } else if (now < t2) {
-        cf = ((double) (now - t1)) / (t2 - t1);
+        cf = ((double) (int32_t) ((uint32_t) now - t1)) /
+                (int32_t) ((uint32_t) t2 - t1);
         a = a1 * (1 - cf) + a2 * cf;
     } else if (now < t3) {
         a = a2;
     } else if (now < t4) {
-        cf = ((double) (now - t3)) / (t4 - t3);
+        cf = ((double) (int32_t) ((uint32_t) now - t3)) /
+                (int32_t) ((uint32_t) t4 - t3);
         a = a2 * (1 - cf) + a3 * cf;
     } else {                    // now >= t4
         a = a3;
@@ -443,7 +445,7 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
             render_priv->state.border_y = yval;
         } else if (complex_tag("move")) {
             double x1, x2, y1, y2;
-            long long t1, t2, delta_t, t;
+            int32_t t1, t2, delta_t, t;
             double x, y;
             double k;
             if (nargs == 4 || nargs == 6) {
@@ -453,8 +455,8 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
                 y2 = argtod(args[3]);
                 t1 = t2 = 0;
                 if (nargs == 6) {
-                    t1 = argtoll(args[4]);
-                    t2 = argtoll(args[5]);
+                    t1 = argtoi32(args[4]);
+                    t2 = argtoi32(args[5]);
                     if (t1 > t2) {
                         long long tmp = t2;
                         t2 = t1;
@@ -467,14 +469,14 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
                 t1 = 0;
                 t2 = render_priv->state.event->Duration;
             }
-            delta_t = t2 - t1;
+            delta_t = (uint32_t) t2 - t1;
             t = render_priv->time - render_priv->state.event->Start;
             if (t <= t1)
                 k = 0.;
             else if (t >= t2)
                 k = 1.;
             else
-                k = ((double) (t - t1)) / delta_t;
+                k = ((double) (int32_t) ((uint32_t) t - t1)) / delta_t;
             x = k * (x2 - x1) + x1;
             y = k * (y2 - y1) + y1;
             if (render_priv->state.evt_type != EVENT_POSITIONED) {
@@ -579,36 +581,31 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
             }
         } else if (complex_tag("fade") || complex_tag("fad")) {
             int a1, a2, a3;
-            long long t1, t2, t3, t4;
+            int32_t t1, t2, t3, t4;
             if (nargs == 2) {
                 // 2-argument version (\fad, according to specs)
                 a1 = 0xFF;
                 a2 = 0;
                 a3 = 0xFF;
                 t1 = -1;
-                t2 = argtoll(args[0]);
-                t3 = argtoll(args[1]);
+                t2 = argtoi32(args[0]);
+                t3 = argtoi32(args[1]);
                 t4 = -1;
             } else if (nargs == 7) {
                 // 7-argument version (\fade)
                 a1 = argtoi(args[0]);
                 a2 = argtoi(args[1]);
                 a3 = argtoi(args[2]);
-                t1 = argtoll(args[3]);
-                t2 = argtoll(args[4]);
-                t3 = argtoll(args[5]);
-                t4 = argtoll(args[6]);
+                t1 = argtoi32(args[3]);
+                t2 = argtoi32(args[4]);
+                t3 = argtoi32(args[5]);
+                t4 = argtoi32(args[6]);
             } else
                 continue;
             if (t1 == -1 && t4 == -1) {
                 t1 = 0;
                 t4 = render_priv->state.event->Duration;
-                // The value we parsed in t3 is an offset from the event end.
-                // What we really want in t3 is an offset from the event start.
-                // To this end, set t3 to (event duration - parsed value).
-                // If t3 >= t4, the exact value of t3 will not matter,
-                // so clamp it to avoid overflow in the subtraction.
-                t3 = t4 - FFMAX(t3, 0);
+                t3 = (uint32_t) t4 - t3;
             }
             if ((render_priv->state.parsed_tags & PARSED_FADE) == 0) {
                 render_priv->state.fade =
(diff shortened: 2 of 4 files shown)


libass/ass_parse.c
```diff
@@ -194,11 +194,12 @@ inline void change_alpha(uint32_t *var, int32_t new, double pwr)
...
  * \param b second value
  * \return result of multiplication
- * Parameters and result are limited by 0xFF.
+ * At least one of the parameters must be less than or equal to 0xFF.
+ * The result is less than or equal to max(a, b, 0xFF).
  */
 inline uint32_t mult_alpha(uint32_t a, uint32_t b)
 {
-    return 0xFF - (0xFF - a) * (0xFF - b) / 0xFF;
+    return a - (uint64_t) a * b / 0xFF + b;
```
astiob: Forgot to say: I realize that this expression isn’t exactly immediately obvious (especially how it avoids overflow and how it’s dangerous to reorder the operations), so I’m open to suggestions for improving it. This is the simplest expression I’ve found though; but maybe simpler is not strictly better in this case.

astiob: Forgot to say: I realize that this expression isn’t exactly immediately obvious (especially how it avoids overflow and how it’s dangerous to reorder the operations), so I’m open to suggestions for improving it. This is the simplest expression I’ve found though; but maybe simpler is not strictly better in this case.

astiob:
Oh, there’s one more VSFilter compatibility question: VSFilter—well, Unicode builds of it, but I believe those are the only builds anyone makes—reads “uuencoded” fonts `WCHAR` by `WCHAR`, whereas we do it `char` by `char`. Since the read characters are not validated, VSFilter can decode a font differently than we do if the encoded data contains non-ASCII bytes. I think our `char`s are UTF-8 bytes while VSFilter’s `WCHAR`s are UTF-16 code units. To make matters worse, when reading UTF-8 files, only XySubFilter and relatively recent xy-VSFilter support reading non-BMP characters, and the handling of invalid bytes and non-BMP characters differs between different VSFilter versions too.

Do we care?

libass/ass.c
```diff
@@ -1356,15 +1356,15 @@ void ass_lazy_track_init(ASS_Library *lib, ASS_Track *track)
...
+            track->PlayResY = track->PlayResX * 3LL / 4;
             ass_msg(lib, MSGL_WARN,
                    "PlayResY undefined, setting to %d", track->PlayResY);
         } else if (!track->PlayResX && track->PlayResY == 1024) {
             track->PlayResX = 1280;
             ass_msg(lib, MSGL_WARN,
                    "PlayResX undefined, setting to %d", track->PlayResX);
         } else if (!track->PlayResX) {
-            track->PlayResX = track->PlayResY * 4 / 3;
+            track->PlayResX = track->PlayResY * 4LL / 3;
```
astiob: Note that this does not fix the `int` overflow on assignment if the value exceeds `INT_MAX`. As I noted in the pull request comment, we need to decide what to do in that case.

libass/ass_parse.c
```diff
@@ -194,11 +195,28 @@ inline void change_alpha(uint32_t *var, int32_t new, double pwr)
...
  * \param b second value
  * \return result of multiplication
- * Parameters and result are limited by 0xFF.
+ * At least one of the parameters must be less than or equal to 0xFF.
+ * The result is less than or equal to max(a, b, 0xFF).
  */
 inline uint32_t mult_alpha(uint32_t a, uint32_t b)
 {
-    return 0xFF - (0xFF - a) * (0xFF - b) / 0xFF;
+    return a - ((uint64_t) a * b + 0x7F) / 0xFF + b;
```
astiob: Forgot to say: I realize that this expression isn’t exactly immediately obvious (especially how it avoids overflow and how it’s dangerous to reorder the operations), so I’m open to suggestions for improving it. This is the simplest expression I’ve found though; but maybe simpler is not strictly better in this case.

libass/ass.c
```diff
@@ -1357,28 +1333,28 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
                    "PlayResY undefined, setting to %d", track->PlayResY);
-        } else if (!track->PlayResX && track->PlayResY == 1024) {
+        } else if (track->PlayResX <= 0 && track->PlayResY == 1024) {
             track->PlayResX = 1280;
             ass_msg(lib, MSGL_WARN,
                    "PlayResX undefined, setting to %d", track->PlayResX);
-        } else if (!track->PlayResX) {
-            track->PlayResX = track->PlayResY * 4 / 3;
+        } else if (track->PlayResX <= 0) {
+            track->PlayResX = track->PlayResY * 4LL / 3;
```
astiob: Note that this does not fix the int overflow on assignment if the value exceeds INT_MAX. As I noted in the pull request comment, we need to decide what to do in that case.

libass/ass.c
```diff
@@ -232,24 +232,10 @@ static long long string2timecode(ASS_Library *library, char *p)
         ass_msg(library, MSGL_WARN, "Bad timestamp");
         return 0;
     }
-    tm = ((h * 60LL + m) * 60 + s) * 1000 + ms * 10;
+    tm = ((h * 60LL + m) * 60 + s) * 1000 + ms * 10LL;
```
astiob: Note that this does not fix the `long long` overflow on assignment if the value exceeds `LLONG_MAX` or `LLONG_MIN`. As I noted in the pull request comment, we need to decide what to do in that case.

astiob:
By the way, I would be quite happy if we made some assumptions about the platform and actually documented them. For example, I'm pretty sure that we assume that int has… maybe not 32 bits, but certainly more than 16. These assumptions should also be verified in configure or in C11 static asserts, and libass should outright refuse to build on platforms that violate them, rather than potentially build and have undefined or unexpected behaviour at runtime. Of course, I'd like these assumptions to be as few as possible. Once we have a list, it would be acceptable/easier to write code that is not completely portable because it relies on these assumptions. (Although, of course, the fewer assumptions a particular piece of code makes, the more reusable it is.)

libass/ass_parse.c
```diff
@@ -194,11 +195,28 @@ inline void change_alpha(uint32_t *var, int32_t new, double pwr)
...
+}
+
+static double subtract(long long minuend, long long subtrahend)
+{
+    assert(minuend >= subtrahend);
+#if ULLONG_MAX > LLONG_MAX
```
wm4: Uh, what? When is the signed max higher or equal than the unsigned max?

libass/ass_parse.c
```diff
@@ -632,7 +644,7 @@ char *parse_tag(ASS_Renderer *render_priv, char *p, char *end, double pwr)
         if (t1 == -1 && t4 == -1) {
             t1 = 0;
             t4 = render_priv->state.event->Duration;
-            t3 = t4 - t3;
+            t3 = t4 - FFMAX(t3, 0);  // avoid overflow
```
wm4: Misleading comment, only the FFMAX here avoids overflow, not the entire expression.

libass/ass_parse.c
astiob: The sign bit in the signed type can be a padding bit in the unsigned type, rendering their maxes equal. It can’t be lower though.

libass/ass_parse.c
astiob: Hmm, I think I see what you mean. I guess this needs a bigger comment that explains what’s going on?

libass/ass_utils.c
```diff
@@ -184,15 +185,15 @@ int mystrtoi32(char **p, int base, int32_t *res)
...
 {
     char *p = *str;
     char *start = p;
     uint32_t val = 0;
 
     while (1) {
-        int digit;
-        if (*p >= '0' && *p < base + '0')
+        unsigned digit;
+        if (*p >= '0' && *p < FFMIN(base, 10) + '0')
```
wm4: Can base really go out of range?

astiob: Well, base can be 16.

libass/ass_parse.c
wm4: Yeah, I wouldn't care about such platforms.

libass/ass_parse.c
```diff
@@ -194,11 +195,28 @@ inline void change_alpha(uint32_t *var, int32_t new, double pwr)
...
 {
-    return 0xFF - (0xFF - a) * (0xFF - b) / 0xFF;
+    return a - ((uint64_t) a * b + 0x7F) / 0xFF + b;
+}
+
+static double subtract(long long minuend, long long subtrahend)
+{
+    assert(minuend >= subtrahend);
+#if ULLONG_MAX > LLONG_MAX
+    return (unsigned long long) minuend - subtrahend;
```
astiob: Note that the `unsigned long long` branch is technically redundant, and the `uintmax_t` branch would take care of it; but this may help the compiler produce better code.

libass/ass_parse.c
astiob: To be clear, this expression is not an exact copy of VSFilter’s, but I’ve verified that it produces the same value. I’ve run exhaustive tests on all `a` from 0 to 255 and `b` in ranges around zero, `INT32_MAX / 255` and `INT32_MAX`.

libass/ass_parse.c
MrSmile: I think good solution will be to use `int64_t` for timestamps which is guaranteed to be 2's complement.

astiob: Yes, that would be good. Unfortunately, we take `long long` in the API function `ass_process_chunk`.

astiob: > Yeah, I wouldn't care about such platforms.

How much not care? `#else #error`, or add a comment, or ignore entirely? Personally, I don’t like the last one.

astiob:
Fixed a bug in the `decode_font` commit.

I think I’m going to cherry-pick the more straightforward commits that don’t deal with weird platforms into master and keep the rest here for further discussion. Sound good?

wm4:
Sure, and maybe take them out of the PR branch once you've pushed them.

libass/ass_parse.c
```diff
@@ -194,11 +195,21 @@ inline void change_alpha(uint32_t *var, int32_t new, double pwr)
...
-    return 0xFF - (0xFF - a) * (0xFF - b) / 0xFF;
+    return a - ((uint64_t) a * b + 0x7F) / 0xFF + b;
+}
+
+static inline double subtract(long long minuend, long long subtrahend)
+{
+#if ULLONG_MAX > LLONG_MAX
+    return (unsigned long long) minuend - subtrahend;
+#else
+#error unsigned long long cannot hold the difference of two long longs
```
astiob: @wm4 How about this?

libass/ass_parse.c
```diff
@@ -627,7 +637,12 @@ char *parse_tag(ASS_Renderer *render_priv, char *p, char *end, double pwr)
         if (t1 == -1 && t4 == -1) {
             t1 = 0;
             t4 = render_priv->state.event->Duration;
-            t3 = t4 - t3;
+            // If we were to subtract a large negative t3, this would
+            // overflow. However, the mathematical result would be
+            // beyond the end of the event anyway (corresponding to no
+            // fade-out), so we may as well just set it to the exact end
+            // of the event without any change in functionality.
+            t3 = t4 - FFMAX(t3, 0);
```
astiob: @wm4 Is this comment comprehensible and sufficient?

libass/ass_parse.c
```diff
@@ -199,7 +199,7 @@ inline void change_alpha(uint32_t *var, int32_t new, double pwr)
  */
 inline uint32_t mult_alpha(uint32_t a, uint32_t b)
 {
-    return a - (uint64_t) a * b / 0xFF + b;
+    return a - ((uint64_t) a * b + 0x7F) / 0xFF + b;
```
MrSmile: I prefer to see that squashed with previous commit, it's the same line anyway.

libass/ass_parse.c
```diff
@@ -501,14 +501,13 @@ char *parse_tag(ASS_Renderer *render_priv, char *p, char *end, double pwr)
...
         }
-        delta_t = t2 - t1;
         t = render_priv->time - render_priv->state.event->Start;
         if (t <= t1)
             k = 0.;
         else if (t >= t2)
             k = 1.;
         else
-            k = ((double) (t - t1)) / delta_t;
+            k = subtract(t, t1) / subtract(t2, t1);
```
MrSmile: That's essentially continuation of the previous commit, so I think that should be squashed too.

libass/ass_parse.c
wm4: If you must...

I wonder if this could just be checked in configure or so.

libass/ass_parse.c
wm4: Sure. Maybe a bit bloated (noise vs. importance). I actually wish we could leave "unimportant" overflows (that can't cause memory errors) just untouched. Not sure if that's a reasonable opinion.

libass/ass_parse.c
astiob: It’s the same line, but it’s a very different change in my opinion. Squashing this would make it harder to understand the combined commit later.

libass/ass_parse.c
astiob: OK, I’ve now removed this check entirely.

It certainly could be done in `configure`, but (1) it isn’t immediately obvious to me what the best way of implementing that is, (2) we don’t currently check any of the other assumptions we make about the platform, and, while it may not mean it’s good to add another unchecked assumption, this one is rather unlikely to break and I’d like to see all assumptions get checked simultaneously, and (3) as my commit still isolates this code in a function, the assumption is at least obvious when looking at the source code.

I’ve kept the `subtract` function (rather than inline the code) to (sort of) document that it’s doing something tricky, and I think the code that calls it actually looks quite nice too: `subtract(a, b) / subtract(c, d)` vs `(double) ((unsigned long long) a - b) / ((unsigned long long) c - d)`. (Our old code even had an extra pair of parentheses for extra parsing fun.)

libass/ass_parse.c
```diff
@@ -627,7 +633,9 @@ char *parse_tag(ASS_Renderer *render_priv, char *p, char *end, double pwr)
         if (t1 == -1 && t4 == -1) {
             t1 = 0;
             t4 = render_priv->state.event->Duration;
-            t3 = t4 - t3;
+            // Clamp t3 to 0 to avoid overflow in the subtraction.
+            // This does not affect functionality.
+            t3 = t4 - FFMAX(t3, 0);
```
astiob: @wm4 Third time’s the charm?

wm4: Sure...

astiob:
Oh, we never merged this. So _that’s_ why OSS-Fuzz kept complaining about so many unfixed overflows…

samoht0:
I understand this patches were not merged before 0.14.0, because there might be regressions with wonky scripts, but why are they still not?
I guess, hardly anybody will test the patches with scripts / muxed files for regressions, when there're just some PR patches, that need to be ported to current master by hand.
If I'm correct, there's no release expected for quiet some time. So plenty of time to catch feedback and maybe revert some patch.
Or am I missing something?

astiob:
Now with a generous extra dose of VSFilter compatibility in override tag timestamp calculations, as mentioned in #337.

The commit “Fix integer overflow while parsing \fad(arg, large negative number)” is now entirely redundant, as the bigger override tag timestamp commit overwrites its changes. I’ve kept it for future reference in case it is ever decided that this kind of VSFilter compatibility is no longer desired here but the overflow is still unwanted.

TheOneric (approved):

libass/ass_parse.c
rcombs: Fun comment here re: rounding: https://github.com/Cyberbeing/xy-VSFilter/blob/ff77719517bc0c1144a75c7b2c1e6268d582614a/src/subtitles/RTS.cpp#L1257

Pretty sure this rounding itself is otherwise correct, given that vsfilter's using this, which is documented to round half up (for positive results, which are all that matter): https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-muldiv

rcombs (approved):
One fun subtle vsfilter thing to point out (which might need fixing at some point) but LGTM.

libass/ass_parse.c
astiob: Somehow I knew what I’d see before I clicked the link 😃

I may have forgotten about that when I wrote this commit. But despite the VSFilter comment, I think there is actually no difference between its two computations: the divisor is always `0xff` which is an odd number, so the quotient simply can’t be exactly halfway two integers and hence is always rounded to the unique closest integer. A quick, _non_-exhaustive practical test failed to find any counterexamples, too.

PR #263, opened 2017-02-05
Travis CI improvements

astiob:
* Disable Fontconfig on OS X to speed up builds, as Homebrew insists on building Fontconfig’s cache when it’s installed.
  * Enable HarfBuzz.
  * Build with Clang on Linux in addition to GCC. (I assume that OS X builds will still use only Clang, but in any case it shouldn’t get any worse. There is an example in the docs that uses this configuration, so I hope it actually works.)
  * Clean up the command that installs dependencies.
  * Run Coverity Scan on every master build. We’re allowed “[up to 12 builds per week, with a maximum of 3 builds per day](https://scan.coverity.com/faq#frequency)” (assuming we still have less than 100,000 lines of code). Looking at [our build history](https://travis-ci.org/libass/libass/builds), we _may_ have reached 3 builds per day occasionally? but rarely.

~If this looks fine, I’ll try pushing this to branch `ci`. If all goes well, I will then push to `master`, but if not, I’d like to rewrite these commits via an interactive rebase and force-push `ci`. Is this OK?~

Oh wait, Travis will build this as a pull request anyway. No need to push to `ci` then.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ branches:
 
 before_install:
   - (sudo apt-get update || brew update)
-  - (sudo apt-get install -y libfontconfig1-dev libfribidi-dev yasm || (brew uninstall libtool; brew install freetype fribidi libtool yasm)) && ./autogen.sh && ./configure
+  - (sudo apt-get install -y libfontconfig1-dev libfreetype6-dev libfribidi-dev yasm || (brew uninstall libtool; brew install freetype fribidi libtool yasm)) && ./autogen.sh && ./configure
 script:
   - make -j4
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ branches:
 
 before_install:
   - (sudo apt-get update || brew update)
-  - (sudo apt-get install -y libfontconfig1-dev libfreetype6-dev libfribidi-dev yasm || brew install freetype fribidi yasm) && ./autogen.sh && ./configure
+  - (sudo apt-get install -y libfontconfig1-dev libfreetype6-dev libfribidi-dev libharfbuzz-dev yasm || brew install freetype fribidi harfbuzz yasm --without-gobject-introspection --without-icu4c) && ./autogen.sh && ./configure
 script:
   - make -j4
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ branches:
 
 before_install:
   - (sudo apt-get update || brew update)
-  - (sudo apt-get install -y libfontconfig1-dev libfreetype6-dev libfribidi-dev yasm || (brew uninstall libtool; brew install freetype fribidi libtool yasm)) && ./autogen.sh && ./configure
+  - (sudo apt-get install -y libfontconfig1-dev libfreetype6-dev libfribidi-dev yasm || brew install freetype fribidi yasm) && ./autogen.sh && ./configure
 script:
   - make -j4
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ branches:
 
 before_install:
   - (sudo apt-get update || brew update)
-  - (sudo apt-get install -y libfontconfig1-dev libfreetype6-dev libfribidi-dev libharfbuzz-dev yasm || (sed -i '' /fc-cache/d "$(brew --repository)/Library/Taps/homebrew/homebrew-core/Formula/fontconfig.rb"; brew install fontconfig freetype fribidi harfbuzz yasm)) && ./autogen.sh && ./configure
+  - (sudo apt-get install -y libfontconfig1-dev libfreetype6-dev libfribidi-dev libharfbuzz-dev yasm || brew install freetype fribidi yasm) && ./autogen.sh && ./configure
 script:
   - make -j4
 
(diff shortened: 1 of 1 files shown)


astiob:
Bah, HarfBuzz depends on Fontconfig in Homebrew. …What.

Ah, it’s a transitive dependency through GObject Introspection and then Cairo. We can ask Homebrew to build HarfBuzz from source without GObject Introspection, which will probably be faster than building the Fontconfig cache.

astiob:
`brew update` takes pretty long to run. Can we do anything to improve that, or can we drop it entirely?

A source build of HarfBuzz takes longer than I expected. But it’s still about 90 seconds faster than using a prebuild HarfBuzz and allowing Homebrew to build the Fontconfig cache!

.travis.yml
```diff
@@ -40,4 +49,4 @@ addons:
     notification_email: none@example.com
     build_command_prepend: "./configure"
     build_command:   "make -j4"
-    branch_pattern: coverity_scan
+    branch_pattern: master|coverity_scan
```
astiob: The docs say this value is a regular expression, so I _hope_ this works. I guess we’ll only find out when we push this to `master` or `coverity_scan`.

.travis.yml
```diff
@@ -20,7 +29,7 @@ branches:
 
 before_install:
   - (sudo apt-get update || brew update)
-  - (sudo apt-get install -y fontconfig libfribidi-dev yasm || (brew uninstall libtool; brew install fontconfig freetype fribidi libtool yasm)) && ./autogen.sh && ./configure
+  - (sudo apt-get install -y libfontconfig1-dev libfreetype6-dev libfribidi-dev libharfbuzz-dev yasm || brew install freetype fribidi harfbuzz yasm --without-gobject-introspection --without-icu4c) && ./autogen.sh && ./configure
```
astiob: The Fontconfig and FreeType package names in Ubuntu contain ABI version numbers (?). We don’t really care about the ABI (only the API), so it seems that this might needlessly break in the future when a newer version is released but we keep requesting the old one. If we want, we can remove these two from this list, as both are (currently) preinstalled on Travis CI.

astiob: Running the whole `brew` command with the `--without-` options produces some benign warnings about formulæ other than `harfbuzz` not recognizing them. If this is undesired, I can split the command into two.

jtanx:
I actually requested Homebrew at one point for a method of disabling fc-cache during the install of fontconfig, but they weren't willing to do that.

It's a bit hacky, but I do the following in my Travis builds to disable fc-cache during install:

```bash
 sed -i.bak '/fc-cache/d' "$(brew --prefix)/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/fontconfig.rb"
```

`brew update` has to have been run before this command.

wm4:
PR seems fine.

Once again, I don't find any reason to build with fontconfig on OSX.

astiob:
Shall I try the hack suggested by @jtanx? It might give us an OS X build that both includes Fontconfig and concludes faster.

wm4:
Again, I don't understand why you waste your time with that.

astiob:
I just want faster builds.

wm4:
So disable fontconfig on OSX?

astiob:
But that’s what I’m already doing.

It’s still brought in by the prebuilt HarfBuzz, so I need to build HarfBuzz from source to avoid Fontconfig. That’s slow (but faster than installing Fontconfig). It would probably be faster to install the prebuilt HarfBuzz but force Fontconfig to not build its cache.

astiob:
So it appears that installing the bottled HarfBuzz takes as much time as building one from source (with Fontconfig, ICU and GObject Introspection disabled). I can only speculate that this is because the bottle installs more dependencies. It might also have been a fluke, but we probably don’t want to run multiple builds just to gather statistics.

So, should I keep the last commit?

* Pros: we get our Fontconfig font provider built on OS X, even if this is not terribly valuable (and we avoid doing a dependency source build, but I can’t immediately think of any downsides to that).
* Cons: we have to patch a Homebrew formula during the build.
* Neutral: the build time doesn’t seem to be affected either way.

wm4:
I don't see any reason to specifically build with HarfBuzz on OSX either just to run the CI (for users it'd be different of course). Or if you really want it, it sounds faster and easier to build it on the CI ourselves.

astiob:
Right, I guess we could build with HarfBuzz (and Fontconfig) only on Linux. It’d be good to have it built with _somewhere,_ but it doesn’t have to include OS X.

> Or if you really want it, it sounds faster and easier to build it on the CI ourselves.

Like, download the source and run `./configure && make -j4 install`? I doubt that it will be significantly faster than asking Homebrew to build it from source (which will happen if I just remove the last commit here), although it might be possible. Asking Homebrew to do it is certainly easier though.

I think I’ll remove HarfBuzz and Fontconfig from the OS X build.

astiob:
The time it takes for an OS X build to even _start_ on Travis CI has been pretty high lately. So maybe this is all meaningless… But let’s hope they improve on that eventually.

Shall I push as-is or squash the Fontconfig/HarfBuzz changes?

wm4:
Probably good to keep the separate commits, so we could try with harfbuzz enabled too if we really want to.

PR #264, opened 2017-02-05
Speedup handling of multiple \r

MrSmile:
That should help with DoS of events with huge number of \r.
I've not extensively tested that so I appreciate if anyone can test heavy override cases and such.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -691,8 +691,7 @@ static ASS_Style *handle_selective_style_overrides(ASS_Renderer *render_priv,
     // The user style was set with ass_set_selective_style_override().
     ASS_Style *user = &render_priv->user_override_style;
     ASS_Style *new = &render_priv->state.override_style_temp_storage;
-    int explicit = event_has_hard_overrides(render_priv->state.event->Text) ||
-                   render_priv->state.evt_type != EVENT_NORMAL;
+    int explicit = render_priv->state.explicit;
     int requested = render_priv->settings.selective_style_overrides;
     double scale;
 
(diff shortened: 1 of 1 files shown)


wm4:
Probably ok, I can't tell if there are possibly some complicated interactions which I might be missing.

libass/ass_render.c
```diff
@@ -898,6 +892,12 @@ init_render_context(ASS_Renderer *render_priv, ASS_Event *event)
     render_priv->state.effect_skip_timing = 0;
 
     apply_transition_effects(render_priv, event);
+    render_priv->state.explicit = render_priv->state.evt_type != EVENT_NORMAL ||
+                                  event_has_hard_overrides(event->Text);
+
+    reset_render_context(render_priv, NULL);
+    render_priv->state.alignment = render_priv->state.style->Alignment;
+    render_priv->state.justify = render_priv->state.style->Justify;
```
astiob: Explanation for anyone who might wonder as I did why so much code is moved around:

* `clip_*` must be initialized before `apply_transition_effects` is called because `apply_transition_effects` overrides the initial values.
* `apply_transition_effects` must be called before `state.explicit` is computed because the computation uses `state.evt_type`, which is set by `apply_transition_effects`.
* `state.explicit` must be computed before `reset_render_context` is called because `state.explicit` is used inside `handle_selective_style_overrides`, which is called from `reset_render_context`.
* `reset_render_context` must be called before `state.alignment` and `state.justify` are set because they are taken from the style, which is overridden by `handle_selective_style_overrides`, which is called from `reset_render_context`.

PR #265, opened 2017-02-12
Remove arbitrary bitmap size limit

### LINKED ISSUE #260: "Glyph bounding box too large" threshold is too low
I'm seeing "Glyph bounding box too large: 4042x2308px" on some large clips, where "too large" is based on the wacky constant of 8000000 pixels defined in caf7e86eef8df95434798dec9b675da68fbef1ff. That check is now in `outline_to_bitmap` in ass_render.c (and also in `outline_to_bitmap_ft`, with slightly different logic). I assume this check was meant to protect against bitmap sizes causing integer overflow, but that's checked in `alloc_bitmap_buffer` now, so I think we should be fine? Removing the check fixes this case for me.
This particular case is clearly a bitmap being rendered at a size far larger than it needs to be (my monitor is 2880x1800, so this is well past its extents), but a fullscreen bitmap at 4K will be over 8M pixels, so there'll be reasonably-legitimate cases this affects.

rcombs:
We're not aware of any specific reason for this number to be used,
and actual potentially-dangerous cases should be handled by the
other limits.

Fixes #260

### DIFF
--- a/libass/ass_bitmap.c
+++ b/libass/ass_bitmap.c
@@ -219,7 +219,7 @@ Bitmap *outline_to_bitmap(ASS_Renderer *render_priv,
 
     int mask = (1 << render_priv->engine->tile_order) - 1;
 
-    if (w < 0 || h < 0 || w > 8000000 / FFMAX(h, 1) ||
+    if (w < 0 || h < 0 ||
         w > INT_MAX - (2 * bord + mask) || h > INT_MAX - (2 * bord + mask)) {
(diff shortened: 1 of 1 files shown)


wm4:
Fine with me. Maybe there should still be a w*h overflow check - not sure if that's taken care of elsewhere.

rcombs:
`alloc_bitmap_buffer` checks for `(s > (INT_MAX - 32) / FFMAX(h, 1))`.

MrSmile:
`w * h` overflow check is done in subsequent `alloc_bitmap` so that's not s problem.
Another question is what to do with possible DoS with extra large bitmaps and OOM.

PR #271, opened 2017-04-26
Enable setting size of background in BorderStyle: 4 using shadow offset

wiiaboo:


wm4:
LGTM. Commit message could be improved (at least the subject line), and I'm not sure if there should be overflow checks, but good enough.

grigorig:
Hm. Is there any chance that this could cause issues with fancy scripts? LGTM otherwise.

wiiaboo:
Reworked commit message a bit.

wiiaboo:
and added explanation of why shadow offset is used.

torque:
Not to bikeshed this, but, while borderstyle 4 is libass only, my understanding is that it's simply a fix for some of the quirks of borderstyle 3, which has its width specified by border size not shadow size. The justification for using shadow size makes no sense to me, regardless of whether shadow is rendered or not. Why not use border and be consistent with the existing behavior? Does borderstyle 4 render borders as well as the bounding box?

astiob:
IIRC, borderstyle 4 does render borders as well as the bounding box. I do agree that using the shadow size feels odd, but then it already uses the shadow colour, so in some sense it would be _more_ consistent than using the border size.

This is also a bit of a bikeshed, but why is it only using the x size? Wouldn’t it be (slightly) better to use `shadow_x` and `shadow_y` in the horizontal and vertical direction respectively?

wiiaboo:
> Does borderstyle 4 render borders as well as the bounding box?

Yes.
image
Maybe it also renders shadow but it's covered by the background box? I didn't read the whole code of libass, nor am I very C-fluent.

> This is also a bit of a bikeshed, but why is it only using the x size? Wouldn’t it be (slightly) better to use shadow_x and shadow_y in the horizontal and vertical direction respectively?

Reasoning being that shadow_x and shadow_y came from Shadow as specified in the script Styles. They're only different from each other if you used overrides, so I went for the simpler code.

wiiaboo:
Screenshots of [this test .ass](https://i.fsbn.eu/jqVd.txt) with [VSFilter](https://i.fsbn.eu/MKr8.png), [mpv with libass master](https://i.fsbn.eu/bh4j.png) and [mpv with this PR](https://i.fsbn.eu/nZVw.png).

astiob:
Thanks; looks good. I’ll point out that the screenshots illustrate that the normal shadow is in fact not rendered when border style 4 is in use.

> Reasoning being that shadow_x and shadow_y came from Shadow as specified in the script Styles. They're only different from each other if you used overrides, so I went for the simpler code.

I won’t insist, but personally, I’d prefer (what seems to me like) more resilient, unsurprising and self-explanatory code here. If someone does set the shadow offset to different values for the two axes while having border style 4, I think they’ll be surprised that the x has an effect while y doesn’t.

I think there are two intuitive options here: either use both `shadow_x` and `shadow_y` as appropriate, or use neither and instead take the `Shadow` value from the style and ignore `\shad` overrides completely. (In the latter case, another question is which style to use: the original line style or the one specified by the most recent `\r` override? The box colour is currently taken from the most recently overridden value, so it’s probably better to do the same for the box size.)

wiiaboo:
Since the point of BorderStyle 4 is to generate a background for the text after knowing the size of the whole subtitle and there's no shadow generated, it could make sense to use the original shadow size and color only, instead of being overridable by tags.

@wm4 any opinion on either behaviors, since you made BorderStyle 4 (afaict)

wm4:
Actually who says there's no shadow generated? Is that excluded by BorderStyle?

astiob:
Yes, shadows proper are explicitly disabled when BorderStyle is 4. You chose to do this yourself in https://github.com/libass/libass/pull/105#issuecomment-44780059. :-) This is also shown in the screenshots posted above.

wiiaboo:
Maybe in [this line](https://github.com/libass/libass/blob/f2832a0134a953dd7cef02e7c741bcaffdde4974/libass/ass_render.c#L609)? The background uses the same color as shadow so it also doesn't make much sense to even render it.

wm4:
OK, then I don't really see a problem with using the shadow params for this.

wiiaboo:
More screenshots with [added shadow color override](https://i.fsbn.eu/H1ek.txt): [master](https://i.fsbn.eu/KgCq.png) vs [this PR with just shadow_x](https://i.fsbn.eu/hb2v.png) vs [this PR with shadow_x and shadow_y](https://i.fsbn.eu/eTiX.png).

wm4:
OK with me...

wiiaboo:
Rebased and realigned the operators vertically.

wm4:
Still ok. Those `(`/`)` and the `.` on the `0` are strictly speaking not needed.

wiiaboo:
Anything still blocking this?

grigorig:
Merged.

PR #276, opened 2017-06-01
directwrite: fix font collections

grigorig:
DirectWrite's FontFileStream does not actually use the data of a specific
font in a collection, which was an expectation of the existing code. It
simply returns a stream to the underlying file, collection or not. So we
need to get the index of the font. This needs to be done lazily as this
information is only available in a FontFace, which is expensive to
initialize.

Add a new optional font provider function for lazy initialization of the
index and use it. This is similar to the check_postscript callback.

Fixes libass#275.

### DIFF
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -352,6 +352,19 @@ static bool check_postscript(void *data)
 }
 
 /*
+ * Lazily return index of font. It requires the FontFace to be present, which is expensive to initialize.
+ */
+static unsigned get_font_index(void *data)
+{
+    FontPrivate *priv = (FontPrivate *)data;
+
+    if (!init_font_private_face(priv))
+        return 0;
+
+    return IDWriteFontFace_GetIndex(priv->face);
+}
+
+/*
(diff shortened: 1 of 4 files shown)


libass/ass_directwrite.c
```diff
@@ -352,6 +352,19 @@ static bool check_postscript(void *data)
...
 
 /*
+ * Lazily return index of font. It requires the FontFace to be present, which is expensive to initialize.
+ */
+static unsigned get_font_index(void *data)
+{
+    FontPrivate *priv = (FontPrivate *)data;
+
+    if (!init_font_private_face(priv))
+        return false;
```
wm4: Maybe that should be 0 for clarity. Since the function doesn't actually return a bool, but an unsigned index.

wm4:
Other than my minor comment, LGTM.

libass/ass_directwrite.c
grigorig: Thanks! Good point. I'm a victim of copy and paste.

grigorig:
Merged.

PR #278, opened 2017-06-21
Switch from yasm to nasm

rcombs:
Our friends at x264, libav, and ffmpeg have all moved (or are moving) in this direction. Nasm is maintained, and has better debugging and macro diagnostics.

### DIFF
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -50,6 +50,11 @@
 #define ABS(x) ((x) < 0 ? -(x) : (x))
 #define MAX_FULLNAME 100
 
+// workaround for gcc and clang bugs
+// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80454
+// https://bugs.llvm.org/show_bug.cgi?id=21689
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
+
 // internal font database element
 // all strings are utf-8
 struct font_info {
--- a/libass/x86/x86inc.asm
+++ b/libass/x86/x86inc.asm
@@ -1,11 +1,11 @@
 ;*****************************************************************************
 ;* x86inc.asm: x264asm abstraction layer
 ;*****************************************************************************
-;* Copyright (C) 2005-2013 x264 project
+;* Copyright (C) 2005-2016 x264 project
 ;*
 ;* Authors: Loren Merritt <lorenm@u.washington.edu>
 ;*          Anton Mitrofanov <BugMaster@narod.ru>
-;*          Jason Garrett-Glaser <darkshikari@gmail.com>
+;*          Fiona Glaser <fiona@x264.com>
 ;*          Henrik Gramner <henrik@gramner.com>
 ;*
 ;* Permission to use, copy, modify, and/or distribute this software for any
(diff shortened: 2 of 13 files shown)


libass/ass_fontselect.c
```diff
@@ -625,7 +625,7 @@ static char *select_font(ASS_FontSelector *priv, ASS_Library *library,
                          ASS_FontStream *stream, uint32_t code)
 {
     ASS_FontProvider *default_provider = priv->default_provider;
-    ASS_FontProviderMetaData meta = {0};
+    ASS_FontProviderMetaData meta = { .families = 0 };
```
wm4: `{0}` is the universal initializer in C, and a compiler shouldn't warn against it. If it does, the warning should be disabled.

wm4:
In general LGTM.

grigorig:
Good idea, the last time I had to install yasm on Windows I was already wondering if the project had died completely. LGTM in general as well, although maybe it'd good to get this tested by people on some of the more unusual platforms.

astiob:
Just wondering: why was Yasm preferred originally? Isn’t it in some sense derived from NASM? It feels a bit weird that the world is now moving “back” to NASM. Is it just that Yasm was in some sense better originally but NASM has improved so much since then that it’s better now?

libass/ass_fontselect.c
astiob: Relevant GCC bug reports: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=36750 (fixed ages ago), https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119 (the fix broke, so refixed in 5.0?), https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64709 (the fix had a bug, fixed in 5.0?), https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80454 (open).

Relevant Clang bug report: https://bugs.llvm.org/show_bug.cgi?id=21689 (open)

libass/x86/x86inc.asm
```diff
@@ -35,13 +35,24 @@
 ; to x264-devel@videolan.org .
 
 %ifndef private_prefix
-    %define private_prefix ass
+    %define private_prefix x264
```
astiob: This doesn’t seem right.

wm4:
At first there was nasm. Then yasm was created (AFAIK new code base by different people), and either before or after that, nasm declined. For a while yasm was the active project that had all new instructions etc., while nasm seemed dead. But then the situation changed - yasm slowed down, and nasm got alive again. Now it's the other way around, and yasm is dead, while nasm has the new stuff. (I know all this from hearsay only, might be not correct.)

libass/x86/x86inc.asm
rcombs: I imported the latest version outright, and decided to define `private_prefix` in `configure.ac` this time, so we don't have to modify x86inc.asm from "upstream".

astiob: Ah, I missed the fact that it was defined in `configure.ac`. OK then!

rcombs:
Merged the relevant pieces; reopening a new PR for the other stuff in here

PR #279, opened 2017-08-05
Update ouline internal representation

MrSmile:
This PR mainly do the following things:

- replaces `FT_Vector`, `FT_BBox`, `FT_Pos` with fixed size types based on `int32_t`;
- uses libass native coordinates in outlines instead of FreeType's reversed Y axis;
- simplifies outline internal format for more straightforward processing;
- consolidates similar types and functions from different files.

The main goal of this PR is to simplify code and there shouldn't be any difference in rendering results.

### DIFF
--- a/libass/ass_drawing.c
+++ b/libass/ass_drawing.c
@@ -174,28 +174,15 @@ static void drawing_free_tokens(ASS_DrawingToken *token)
 }
 
 /*
- * \brief Update drawing cbox
- */
-static inline void update_cbox(ASS_Drawing *drawing, ASS_Vector *point)
-{
-    ASS_Rect *box = &drawing->cbox;
-
-    box->x_min = FFMIN(box->x_min, point->x);
-    box->x_max = FFMAX(box->x_max, point->x);
-    box->y_min = FFMIN(box->y_min, point->y);
-    box->y_max = FFMAX(box->y_max, point->y);
-}
-
-/*
  * \brief Translate and scale a point coordinate according to baseline
  * offset and scale.
  */
 static inline void translate_point(ASS_Drawing *drawing, ASS_Vector *point)
 {
-    point->x = drawing->point_scale_x * point->x;
-    point->y = drawing->point_scale_y * point->y;
+    point->x = lrint(drawing->point_scale_x * point->x);
+    point->y = lrint(drawing->point_scale_y * point->y);
 
-    update_cbox(drawing, point);
+    rectangle_update(&drawing->cbox, point->x, point->y, point->x, point->y);
 }
 
 /*
@@ -244,8 +231,7 @@ ASS_Drawing *ass_drawing_new(ASS_Library *lib)
     ASS_Drawing *drawing = calloc(1, sizeof(*drawing));
     if (!drawing)
         return NULL;
-    drawing->cbox.x_min = drawing->cbox.y_min = INT32_MAX;
-    drawing->cbox.x_max = drawing->cbox.y_max = INT32_MIN;
+    rectangle_reset(&drawing->cbox);
     drawing->library = lib;
     drawing->scale_x = 1.;
     drawing->scale_y = 1.;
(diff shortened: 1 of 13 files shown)


grigorig:
What's the motivation for flipping the coordinate system?

MrSmile:
I find use of two coordinate systems in rendering confusing. Also there were several places where inherently symmetric for X & Y expressions have different sign due to switch of coordinate system. Well, It's mostly to get rid of `p.x = src.x; p.y = -src.y;` inside rasterizer.

wm4:
Generally I like it.

libass/ass_outline.h
```diff
@@ -38,20 +38,31 @@ typedef struct
...
 
-typedef struct ass_outline {
-    size_t n_contours, max_contours;
-    size_t *contours;
+enum {
+    OUTLINE_LINE_SEGMENT     = 1,
+    OUTLINE_QUADRATIC_SPLINE = 2,
+    OUTLINE_CUBIC_SPLINE     = 3,
+    OUTLINE_COUNT_MASK       = 3,
+    OUTLINE_CONTOUR_END      = 4
```
astiob: It would perhaps be useful to add a comment here to explain what the values of these constants mean. As far as I can tell:

* `OUTLINE_LINE_SEGMENT`, `OUTLINE_QUADRATIC_SPLINE` and `OUTLINE_CUBIC_SPLINE` show how many points (minus the starting point) define the respective segment/spline;
* a `segment` value actually consists of two bit fields: one is the number of points as above, and it can be extracted using `OUTLINE_COUNT_MASK`; the other is a single-bit flag showing whether this is the last segment of a contour, `OUTLINE_CONTOUR_END`.

I’m not sure whether it’s worth noting that this enum must not contain zero values as zero is used to mean “don’t add any segment” in `outline_add_point` calls, or whether it’s worth making that explicit by giving zero a name and using it in those calls.

libass/ass_outline.c
```diff
@@ -113,19 +233,33 @@ void outline_free(ASS_Outline *outline)
 /*
  * \brief Add a single point to a contour.
```
astiob: Perhaps add a comment here noting that a segment is added (if and) only if the `segment` argument is nonzero.

libass/ass_drawing.c
```diff
@@ -95,9 +90,9 @@ static int token_check_values(ASS_DrawingToken *token, int i, int type)
 static ASS_DrawingToken *drawing_tokenize(char *str)
 {
     char *p = str;
-    int i, type = -1, is_set = 0;
+    int  type = -1, is_set = 0;
```
astiob: This line has a double space after `int`, which is probably unintended.

libass/ass_drawing.c
```diff
@@ -204,15 +199,12 @@ static inline void translate_point(ASS_Drawing *drawing, FT_Vector *point)
 }
 
 /*
- * \brief Evaluate a curve into lines
- * This curve evaluator is also used in VSFilter (RTS.cpp); it's a simple
- * implementation of the De Casteljau algorithm.
```
astiob: Is this comment no longer true?

libass/ass_outline.h
```diff
@@ -22,13 +22,27 @@
...
+typedef struct {
+    int32_t x, y;
+} ASS_Vector;
+
+typedef struct {
+    double x, y;
+} ASS_DVector;
+
+typedef struct
+{
```
astiob: This brace should be on the previous line.

libass/ass_outline.h
MrSmile: I dislike having named constant for zero here due to zero being prohibited value in `segments`. I guess I mention zero in comment before `outline_add_point()`.

libass/ass_drawing.c
MrSmile: Yes, that's no longer true and for quite long time at that.

MrSmile:
Fixed and rebased.

grigorig:
Merged.

PR #281, opened 2017-09-06
Fix crash due to stale pointers in font cache after ass_set_fonts

rcombs:
I'm kinda confused as to why this doesn't cause more problems.

Also some warning fixes.

### DIFF
--- a/libass/ass_render_api.c
+++ b/libass/ass_render_api.c
@@ -146,6 +146,10 @@ void ass_set_fonts(ASS_Renderer *priv, const char *default_font,
 
     ass_reconfigure(priv);
 
+    ass_cache_empty(priv->cache.font_cache);
+    if (priv->shaper)
+        ass_shaper_empty_cache(priv->shaper);
+
     if (priv->fontselect)
         ass_fontselect_free(priv->fontselect);
     priv->fontselect = ass_fontselect_init(priv->library, priv->ftlibrary,
--- a/libass/Makefile.am
+++ b/libass/Makefile.am
@@ -1,6 +1,7 @@
 AM_CFLAGS = -std=gnu99 -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter \
             -Werror-implicit-function-declaration -Wstrict-prototypes        \
-            -Wpointer-arith -Wredundant-decls -D_GNU_SOURCE
+            -Wpointer-arith -Wredundant-decls -Wno-missing-field-initializers\
+            -D_GNU_SOURCE
 
 LIBASS_LT_CURRENT = 9
 LIBASS_LT_REVISION = 1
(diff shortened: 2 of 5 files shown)


libass/ass_fontselect.c
```diff
@@ -50,6 +50,11 @@
 #define ABS(x) ((x) < 0 ? -(x) : (x))
 #define MAX_FULLNAME 100
 
+// workaround for gcc and clang bugs
+// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80454
+// https://bugs.llvm.org/show_bug.cgi?id=21689
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
```
wm4: I'd rather just put it as normal flag into the configure script.

libass/ass_render_api.c
```diff
@@ -150,6 +150,8 @@ void ass_set_fonts(ASS_Renderer *priv, const char *default_font,
         ass_fontselect_free(priv->fontselect);
     priv->fontselect = ass_fontselect_init(priv->library, priv->ftlibrary,
             default_family, default_font, config, dfp);
+
+    ass_cache_empty(priv->cache.font_cache);
```
wm4: Maybe better call it before the dependent objects are destroyed?

MrSmile:
Actually, `ass_cache_empty()` does not delete all cached `ASS_Font` objects. Such objects can be kept alive by references in keys of other caches. So it's necessary to either drop all caches altogether or somehow circumvent orphan fonts.

rcombs:
Hmmm, the outline cache and the glyph metrics cache, I guess?

MrSmile:
Also bitmap and composite due to chain dependency (i.e. literally all caches). But all except glyph metrics are already handled in `ass_reconfigure()`, so only one left.

wm4:
This is complicated by the fact that we probably want to eventually allow the user to keep bitmaps for a longer time via refcounting.

MrSmile:
Indeed. It's necessary to split cache key and value destruction in the future. That way alive bitmaps won't keep all related cache hierarchy from clearing.

rcombs:
Am I still missing anything else that needs clearing here? And are we good with the other 2 changes?

wm4:
LGTM I guess.

astiob:
I guess it may be worth noting in the commit message that it fixes a crash, but generally LGTM.

astiob:
I see the cache-related commit is now in master as 68f25e2c26de2c5c3624bf0fe6d12a0e9c35e861, so shall we close this now or keep it open for the warning-related commit?

In the warning-related commit, it may be good to note in the commit message that the warning is being disabled specifically because Clang gives it for the universal initializer `={0}`. I presume we could revert this change whenever Clang stops doing that.

astiob:
I’ve pushed the `={0}` warning fix. I believe the `ASS_Outline` warning was fixed by @MrSmile’s last merged PR, so there’s nothing else left here.

Issue #282, opened 2017-09-07
Building with media-autobuild_suite fails with libass


BlackScreen:
See these logs:
[logs.zip] (https://0x0.st/RvZ.zip)
Thanks for looking into it!

astiob:
Relevant error:

```
  NASM     x86/rasterizer.lo
../../libass/x86/rasterizer.asm:21: fatal: unable to open include file `x86/utils.asm'
```

@rcombs

rcombs:
I'd need to see a `make V=1` to know what's going on there; this looks like it's probably a bug in whatever Makefiles you're using that wrap around libass's.

wiiaboo:
```
/bin/sh ../libtool  --tag=CC --mode=compile nasm -DARCH_X86_64=1 -DPIC -f win64 -DHAVE_ALIGNED_STACK=1 -DHAVE_CPUNOP=0 -Dprivate_prefix=ass -o x86/rasterizer.lo ../../libass/x86/rasterizer.asm -prefer-non-pic
libtool: compile:  nasm -DARCH_X86_64=1 -DPIC -f win64 -DHAVE_ALIGNED_STACK=1 -DHAVE_CPUNOP=0 -Dprivate_prefix=ass ../../libass/x86/rasterizer.asm -o x86/rasterizer.o
../../libass/x86/rasterizer.asm:21: fatal: unable to open include file `x86/utils.asm'
```

This is an out-of-root build (`mkdir stuff && ../configure && make`), btw, might be the issue.

BlackScreen:
@rcombs: Any chance of fixing the problem? Thanks!

PR #283, opened 2017-09-11
x86: revert include parent directory

wiiaboo:
Fixes out-of-root builds.

See https://github.com/libass/libass/issues/282


wiiaboo:
Nevermind, this isn't the right fix.

PR #288, opened 2017-10-26
Fix CoreText font provider in GCC and on old Mac OS X

astiob:
I remembered today that I had this lying around almost finished for two years. So I finally polished it to my liking, and well, here it is.

I know I’m probably the only one around here who cares about old OSs (and over the last two years this has gotten even less relevant than it once was), but hey, this change is not terribly complicated, is it?

I’ve been using this on Mac OS X 10.6 Snow Leopard for just over two years without any trouble.

The configure script could use `AC_CHECK_DECL`, and in fact my original version of this commit did exactly that, but then (a) behdad/harfbuzz#342 must be minded and (b) the message `./configure` prints on screen is `checking whether kCTFontURLAttribute is declared` (plus a second similar message on iOS), which makes it far from obvious that it’s checking for CoreText: this is why I changed it to an explicit `AC_COMPILE_IFELSE`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -163,7 +163,7 @@ OLDLIBS="$LIBS"
 # Linking to CoreText directly only works from Mountain Lion and iOS6. In
 # earlier OS releases CoreText was part of the ApplicationServices umbrella
 # framework.
-LIBS="$LIBS -framework CoreText -framework CoreFoundation -framework CoreGraphics"
+LIBS="$LIBS -framework CoreText -framework CoreFoundation"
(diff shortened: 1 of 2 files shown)


astiob:
A note about the `TARGET_OS_IPHONE` bit: on iOS, only CoreText.h works, but on macOS, both CoreText.h and ApplicationServices.h work. On old Mac OS X in particular, only ApplicationServices.h works. So I must conditionally include both headers, and in the condition I can either check for iOS or for old Mac OS X. Checking for iOS is easier and is what HarfBuzz does, so I stole the code from them.

PR #296, opened 2018-01-04
Reduce recursion and improve VSFilter compatibility for nested \t

astiob:
This prevents [stack overflow](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=4892) on deeply nested `\t` by essentially performing tail call optimization on it.

The first commit indents the whole of `parse_tag`, so for viewing the diff, you probably want to use `git show -w` or `git log -p -w` on the command line rather than GitHub’s Web interface.

As we currently parse tags in a VSFilter 2.37–compatible way, parentheses are not parsed in nested pairs, but rather the very first closing parenthesis terminates the whole tag. So if you have a `\t()` inside another `\t()`, the first closing parenthesis terminates the outer `\t` and the inner `\t` actually doesn’t have a closing parenthesis at all. This means that if the parser ever encounters a `\t(` inside another `\t()`, it is the very last tag inside it, guaranteed, because it runs all the way to the closing parenthesis.

I could translate this observation directly into code, especially because I had to introduce a `nested` flag in the last commit anyway (see below), but I chose a more generic condition because it is clearer on its own (without knowing the story of how parentheses are parsed), it will not break if we ever decide to allow nested parentheses, and as an added bonus it can save an occasional recursive call even for unnested `\t` if the closing parenthesis is missing entirely.

While verifying that I didn’t break anything, I noticed that we disagreed with VSFilter on the meaning of `\t` with an incorrect number of arguments. Again due to the VSFilter 2.39–compatible tag parsing, the only case that actually matters is when there’s a `\t` with no parentheses inside a multiple-argument `\t()`. As soon as the inner `\t` is reached, the animation-controlling arguments (start, end, acceleration) of the outer `\t` are forgotten and reset to the defaults, effective for all subsequent tags inside the `\t()`. I haven’t seen real files using this, but I’ve confirmed in a practical test that our behaviour differs. The **last commit** in this pull request fixes this.

### DIFF
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -650,7 +650,17 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr)
                 k = pow(((double) (t - t1)) / delta_t, accel);
             }
             p = args[cnt].start;
-            p = parse_tags(render_priv, p, args[cnt].end, k);    // maybe k*pwr ? no, specs forbid nested \t's
+            if (args[cnt].end < end) {
+                p = parse_tags(render_priv, p, args[cnt].end, k);
+            } else {
+                assert(q == end);
+                // No other tags can possibly follow this \t tag,
+                // so we don't need to restore pwr after parsing \t.
+                // The recursive call is now essentially a tail call,
+                // so optimize it away.
+                pwr = k;
+                q = p;
+            }
         } else if (complex_tag("clip")) {
             if (nargs == 4) {
                 int x0, y0, x1, y1;
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -224,620 +224,620 @@ static int parse_vector_clip(ASS_Renderer *render_priv,
 }
 
 /**
- * \brief Parse style override tag.
+ * \brief Parse style override tags.
  * \param p string to parse
  * \param end end of string to parse, which must be '}', ')', or the first
  *            of a number of spaces immediately preceding '}' or ')'
  * \param pwr multiplier for some tag effects (comes from \t tags)
  */
-char *parse_tag(ASS_Renderer *render_priv, char *p, char *end, double pwr)
+char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr)
 {
-    while (*p != '\\' && p != end)
+    for (char *q; p < end; p = q) {
+        while (*p != '\\' && p != end)
+            ++p;
+        if (*p != '\\')
+            break;
         ++p;
-    if (*p != '\\')
-        return p;
-    ++p;
-    if (p != end)
-        skip_spaces(&p);
+        if (p != end)
+            skip_spaces(&p);
 
-    char *q = p;
-    while (*q != '(' && *q != '\\' && q != end)
-        ++q;
-    if (q == p)
-        return q;
+        q = p;
+        while (*q != '(' && *q != '\\' && q != end)
+            ++q;
+        if (q == p)
+            continue;
 
-    char *name_end = q;
+        char *name_end = q;
 
-    // Store one extra element to be able to detect excess arguments
-    struct arg args[MAX_VALID_NARGS + 1];
-    int nargs = 0;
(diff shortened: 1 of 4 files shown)


libass/ass_parse.c
```diff
@@ -232,6 +232,7 @@ static int parse_vector_clip(ASS_Renderer *render_priv,
  */
 char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr)
 {
+start:
```
wm4: Not loving this very much. Makes it a bit hard to reason about whether it really makes progress every time it gets here. (Seems `p` is updated as parsing progresses...) Also, goto is ugly.

Maybe this could be solved by another indirection by making a function call parse_tags itself in a loop?

libass/ass_parse.c
```diff
@@ -651,7 +652,17 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr)
                 k = pow(((double) (t - t1)) / delta_t, accel);
             }
             p = args[cnt].start;
-            p = parse_tags(render_priv, p, args[cnt].end, k);    // maybe k*pwr ? no, specs forbid nested \t's
+            if (args[cnt].end < end)
+                p = parse_tags(render_priv, p, args[cnt].end, k);
+            else {
```
wm4: (Minor nag: personally I'd prefer adding redundant `{` / `}` to the if body.)

wm4:
Generally fine, I guess.

libass/ass_parse.c
astiob: OK.

libass/ass_parse.c
astiob: I don’t think another function calling this in a loop would work, because the function’s arguments—`pwr` and later `nested`—need to be updated. Although I guess we could change them into pointers (in the inner function’s argument list only) so that the inner function call can update them.

One option to avoid the `goto` would be to use an actual tail call and rely on the compiler to optimize it, which I see you’ve mentioned on IRC. I think it should be safe but may look a bit weird when we go back and reread this code years later.

I thought of phrasing the loop like this:
```c
while (p < end) {
    ...
    p = q;
}
```
possibly with a `goto` _within_ the loop, but then I don’t see how to manage the `continue`s in tags such as `\fade`, which should execute the `p = q` assignment while the pseudo-tail-call in `\t` shouldn’t execute it. Either all those `continue`s need to be replaced with `goto`s—which feels wrong because the fewer `goto`s, the better—(and the `goto` in `\t` replaced with an actual `continue`), or the same effect can be achieved by wrapping the loop body in a `switch (0) { case 0: }` and using `break` instead of `goto`—which feels even more hackish—or a Boolean flag can be used to determine whether the `p = q` assignment should run, which I don’t like either.

Alternatively, perhaps I can just reassign `q = p` in the pseudo-tail-call path and let the loop-increment statement `p = q` happen as usual:
```c
for (char *q; p < end; p = q) {
    ...
        if (args[cnt].end < end) {
            ...
        } else {
            pwr = k;
            nested = true;
            q = p;
            continue;  // or not, because there’s no code after this anyway
        }
    ...
}
```

wm4: The function could pass in those arguments as pointers (or as a pointer to a struct).

Relying on tail call optimization if the goal is to avoid recursion is probably not a good idea, regrettably.

I think the goal should be to keep this somewhat readable - your last suggestion doesn't look too bad?

astiob:
Updated: added braces around the `if` body and replaced the `goto` by a `q = p` assignment.

wm4:
Dropping the goto is nice. I have no further comments.

astiob:
Updated again: turns out `end` was being modified when parsing `\fn`, so the loop ended too early if `\fn` was followed by anything else. The assertion in `ass_render.c` helped catch this. In the fixed commit, `end` is no longer modified within `parse_tags`.

astiob:
Final update (hopefully) before merging: added changelog entries, expanded the commit message of the recursion commit and mentioned this PR in the commit messages. If the changelog and commit messages look fine to you, I’ll push to `master`.

PR #298, opened 2018-02-24
Use FriBiDi 1.x API when available

khaledhosny:
This allows making use of the updated UBA in Unicode 6.3 and up.

### DIFF
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -39,6 +39,10 @@ enum {
 #define NUM_FEATURES 5
 #endif
 
+#if FRIBIDI_MAJOR_VERSION >= 1
+#define USE_FRIBIDI_EX_API
+#endif
+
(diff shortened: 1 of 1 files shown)


wm4:
On the first looks this seems reasonable, but I can't comment on the actual code.
What worries me a bit is the ifdeffery, but probably unavoidable since it'll take a long time until we can make newer fribidi a hard dependency.

khaledhosny:
Ping

astiob:
Hey, I’m terribly sorry for the lack of response. I was originally going to merge this immediately after merging #285, but it’s taken me crazy long to review that. Is this still relevant? If so, I’m definitely still interested in merging this.

astiob:
Finally merged this as df7c00c95ec6e526bf85e6bc8296eaca66f5db8d.

PR #311, opened 2018-07-19
Fix ass_render_frame's change detection when using BorderStyle=4

tmm1:
I discovered a bug where using BorderStyle=4 would cause `ass_render_frame` to always set `detect_change=2`, causing unnecessary re-rendering.

This PR tags any `ASS_Image` added as a background with `IMAGE_TYPE_BACKGROUND`, and uses that new type to ensure `ass_image_compare` works as expected.

### DIFF
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -53,7 +53,8 @@ typedef struct ass_image {
     enum {
         IMAGE_TYPE_CHARACTER,
         IMAGE_TYPE_OUTLINE,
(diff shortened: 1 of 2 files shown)


astiob:
It seems that we create a brand new background bitmap for each frame. @wm4, was this intended? The right fix would be to cache them (which might additionally help performance, but in most cases this is probably negligible) and keep `ass_image_compare` unchanged, rather than assume background images never change.

PR #312, opened 2018-07-24
Android font provider

tmm1:
Original proof-of-concept by @sfan5. I cleaned up the style, fixed some memory leaks and segvs, added support for `alias` definitions in fonts.xml, and tested on a variety of devices (API 22-26).

Works well enough for my needs. The hand-rolled xml parser is not ideal, but it's better than adding a heavy dependency like libxml2 (atleast imho).

The only hacky part of the patch is the `g_ftlibrary` global. I'd like to find a better way to make the freetype handle available to font providers which need it- suggestions welcome.


rcombs:
From my copy of the referenced XML file:
```
    WARNING: Parsing of this file by third-party apps is not supported. The
    file, and the font files it refers to, will be renamed and/or moved out
    from their respective location in the next Android release, and/or the
    format or syntax of the file may change significantly. If you parse this
    file for information about system fonts, do it at your own risk. Your
    application will almost certainly break with the next major Android
    release.
```

tmm1:
I see that warning in API26 and API27, but not earlier.

tmm1:
Looks like this warning was added in https://github.com/aosp-mirror/platform_frameworks_base/commit/19b05500b7ecc3998b77e1d03f36a73fe0531a9f#diff-59cec50835cfa2d987438f8b86daf314, along with advice to use a new java API  `android.text.FontManager#getSystemFonts()` to access the same information.

But then in https://github.com/aosp-mirror/platform_frameworks_base/commit/ff55115121a7a2753ba2265cb3201a3a14c0874d#diff-59cec50835cfa2d987438f8b86daf314 that API was removed, as was the suggestion to use it.

So this may or may not break in the future, at which point libass on Android would be just as broken as it is today.

rcombs:
I'd advise filing an issue against https://github.com/android-ndk/ndk asking what the intended behavior is here.

libass/ass_androidfonts.c
```diff
@@ -0,0 +1,659 @@
...
+
+/*****************************************/
+
+#define FONTS_XML "/system/etc/fonts.xml"
+#define FONTS_PATH "/system/fonts"
+#define MAX_FULLNAME 100
+
+#define ALOGE(fmt, ...) __android_log_print(ANDROID_LOG_ERROR,   "libass", fmt, ##__VA_ARGS__)
+#define ALOGV(fmt, ...) __android_log_print(ANDROID_LOG_VERBOSE, "libass", fmt, ##__VA_ARGS__)
+#define ALOGD(fmt, ...) __android_log_print(ANDROID_LOG_DEBUG,   "libass", fmt, ##__VA_ARGS__)
```
sfan5: I added these just for debugging, libass has its own proper logging system afaik

libass/ass_androidfonts.c
```diff
@@ -0,0 +1,659 @@
...
+                if (alias_read) {
+                    //ALOGD("adding alias from %s to %s\n", alias_name, alias_to);
+                    af->n_substitutions++;
+                    af->substitutions = realloc(af->substitutions, af->n_substitutions * sizeof(ASS_FontMapping));
+                    af->substitutions[af->n_substitutions - 1].from = alias_name;
+                    af->substitutions[af->n_substitutions - 1].to = alias_to;
+                    // TODO: use alias_weight
+                    alias_name = NULL;
+                    alias_to = NULL;
+                }
```
sfan5: this block would be better placed inside the if strcmp "alias" like so:
```c
if (!strcmp(t.tag.name, "alias")) {
    if (t.flags & XML_FLAG_OPENING_TAG) {
        alias_read = 1;
    } else {
        // do stuff
    }
}

File: libass/ass_fontselect.c [hunk 3]
```diff
@@ -1002,6 +1010,9 @@ ass_fontselect_init(ASS_Library *library,
     priv->path_default = path ? strdup(path) : NULL;
     priv->index_default = 0;
 
+#ifdef CONFIG_ANDROID
+    g_ftlibrary = ftlibrary;
```
sfan5: well yeah, global state
unfortunately I haven't found a nice way to avoid this

astiob:
https://issuetracker.google.com/issues/111801338 was eventually filed with Android. It was marked as “fixed” on 2 Nov 2018, but no explanation has ever been provided.

astiob:
Finally found something official: [“What’s new for text in Android Q”](https://android-developers.googleblog.com/2019/07/whats-new-for-text-in-android-q.html)

[NDK reference: Font](https://developer.android.com/ndk/reference/group/font.html#asystemfontiterator_close) shows three functions that are presumably new in Android Q:

```c
ASystemFontIterator *_Nullable ASystemFontIterator_open();
AFont *_Nullable ASystemFontIterator_next(ASystemFontIterator *_Nonnull iterator);
void ASystemFontIterator_close(ASystemFontIterator *_Nullable iterator);
```

The above blog post says:

> If you want to get all available system fonts, you can do this with a new font enumeration API. In Java, you can use [`SystemFonts.getAvailableFonts`](https://developer.android.com/reference/android/graphics/fonts/SystemFonts#getAvailableFonts()), or in the NDK, you can use [`ASystemFontIterator`](https://developer.android.com/ndk/reference/group/font.html#asystemfontiterator_next). The results of the font enumeration are changed only by a system update, so you should cache them.

(It is not clarified how to determine when the cache should be invalidated and rebuilt.)

So it seems that:

* in Android Q, we can use the `ASystemFontIterator` API, but it’s not clear whether we can call it every time on startup or need to build a custom cache;
* in earlier versions, we can use the code in this PR? (I’ll be honest: I haven’t even looked at it yet.)

tmm1:
Nice find! I wondered why they closed my issue and never told me how it was fixed.

Are there any devices with Android Q available yet? In order to support both new and old devices at the same time, I believe we would need to use `dlsym` to try to look up and use the new APIs conditionally.

astiob:
Ah, right. 😅 We should be able to compile once and run on both new and old Androids. I’m assuming Android probably provides some way to figure out its version at runtime, so perhaps we could just switch based on that, although feature-detection style is nice too.

I wish I could find the source code for the new API’s implementation… and see its commit history.

tmm1:
> I’m assuming Android probably provides some way to figure out its version at runtime

AFAIK only via java, so you have to use JNI even to get that.



> I wish I could find the source code for the new API’s implementation… and see its commit history.

Looks like it's just reading the same file we are: https://github.com/aosp-mirror/platform_frameworks_base/blob/master/native/android/system_fonts.cpp#L206

TheOneric:
> > I’m assuming Android probably provides some way to figure out its version at runtime
>
> AFAIK only via java, so you have to use JNI even to get that.

I don't know why they choose not to provide a more convenient way, but it is possible to get API at runtime natively using `popen` as detailed here:
 https://stackoverflow.com/a/33268925
I can confirm that using `popen` works on API level 23.

tmm1:
I've used __system_property_get on Android N and O. I think maybe the original removal was reverted.

You can use:

> #include <sys/system_properties.h>

rcombs:
I'd suggest checking for the presence of the individual symbols (via `dlsym(…, RTLD_DEFAULT)`) rather than trying to do a version check.

rcombs:
>in Android Q, we can use the ASystemFontIterator API, but it’s not clear whether we can call it every time on startup or need to build a custom cache;

I'd assume that by "cache" they mean "in memory, at startup".

As for the PR as it exists, parsing the XML manually on old Android versions fairly reasonable in concept, but I have a few issues with the implementation:
- Using the Android log facility instead of libass's (should be simple enough to deal with)
- Unnecessary global state (just create a local lib with `FT_Init_FreeType`)

Also, when running on Android Q, do we actually want to do this iteration at all? It looks like the `AFontMatcher_match` API should serve our use-case pretty well, in conjunction with `ass_get_font_info` (which should also replace the `get_extra_font_info` function from this PR).

lldsolitude:
Can this possibly be merged anytime soon? Non-native English speakers are really looking forward to this feature.

rcombs:
Looking back on this PR, I think this can be solved more simply and robustly with a few steps:
- Improve our fonts-dir handling to open the fonts on disk instead of loading them all into memory
- Hardcode a couple alternate default font names to search for in addition to Arial (Roboto, Noto Sans)
- Improve our fallback support to search all loaded fonts if we can't find a glyph for a character in the default family and the provider doesn't give any substitutions
- Have an Android font provider that just adds `/system/fonts` as a fonts dir

I think this gets us to effectively the same behavior as the current PR provides in substantially less code (and very little Android-specific code) in a way that improves robustness on all platforms rather than only Android. The only major downside is that it wouldn't handle aliases to fonts other than Roboto and Noto Sans, which I don't think is a huge deal in practice (if you really want a serif font you're probably attaching it anyway).

To improve support for newer versions and avoid potential issues if the fonts ever move out of `/system/fonts`, we can later improve the Android provider to use the first-class `AFontMatcher_match` function if available; that would also get us aliases for non-sans fonts.

TheOneric:
Closing as the PR appears abandoned and apparently the approach itself is problematic with better alternatives existing, see https://github.com/libass/libass/issues/608#issuecomment-1089436382.

PR #316, opened 2018-10-11
coretext: fix reading weights of some fonts

rcombs:
Some fonts have weights that can be expressed more precisely as doubles than as floats. In these cases, when writing to a float, CFNumberGetValue will set the value to the closest approximation, but return false (so we'd just clobber whatever it set with 0).
Easy fix: just store to a double instead.

### DIFF
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -116,17 +116,17 @@ static void get_name(CTFontDescriptorRef fontd, CFStringRef attr,
 }
 
 static void get_trait(CFDictionaryRef traits, CFStringRef attribute,
-                      float *trait)
+                      double *trait)
 {
     CFNumberRef cftrait = CFDictionaryGetValue(traits, attribute);
-    if (!CFNumberGetValue(cftrait, kCFNumberFloatType, trait))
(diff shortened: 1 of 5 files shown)


rcombs:
@astiob could you review the new changes as well? I can split this up into multiple PRs if there are any concerns.

astiob:
Is the 150 offset derived from VSFilter’s source code or experimentally?

Shouldn’t (or would it be a good idea to make) CoreText interpolate the weights like Fontconfig does?

rcombs:
>Is the 150 offset derived from VSFilter’s source code or experimentally?

Determined experimentally, using Open Sans (which has many weights). Note that it applies to weights <400 as well, so a font with weight 100 will be faux-bolded with `\b0` (verified this happens in VSFilter). This behavior appears to come from GDI, rather than from VSFilter itself.

>Shouldn’t (or would it be a good idea to make) CoreText interpolate the weights like Fontconfig does?

Probably.

astiob:
I’ll merge this as well to preserve the CoreText-specific changes in history in case they might ever be useful to look back at.

PR #321, opened 2018-11-24
Fix compilation with msvc

tgoyne:
msvc doesn't like a `const int` for the size of an array, so use an enum instead.


rcombs:
imo a macro would be more appropriate here; thoughts?

MrSmile:
That's already fixed in my PR #285.

PR #327, opened 2019-01-16
Use Freetype to read attrs on CoreText-discovered fonts

rcombs:
This makes results much more consistent with other platforms, particularly around cases where fonts have multiple conflicting names.

A successor to #316, with name handling improved in addition to weight.

### DIFF
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -103,117 +103,30 @@ static char *get_font_file(CTFontDescriptorRef fontd)
     return buffer;
 }
 
-static void get_name(CTFontDescriptorRef fontd, CFStringRef attr,
-                     char **array, int *idx)
+static char *get_name(CTFontDescriptorRef fontd, CFStringRef attr)
 {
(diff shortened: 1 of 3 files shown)


rcombs:
Looking at the Freetype-provided family name allows us to match e.g. "Open Sans" against all the fonts in that family, including different weights and italic variants. It also means that e.g. `\fnOpen Sans SemiBold\i1` gives the same results as `\fnOpen Sans SemiBold Italic`, where previously it gave faux-italics. My understanding is that this is consistent with VSFilter's behavior.

astiob:
Huh.

Shouldn’t “Open Sans SemiBold Italic” be a fullname of that font and get matched by `matches_full_or_postscript_name`?

astiob:
Oh, wait. You’re saying that does work but “Open Sans SemiBold” doesn’t. Oh.

OK, I’ll merge this for now. If this ever turns out to be a problem, we can revert or fix it later.

PR #329, opened 2019-02-09
rasterizer: add CONFIG_HIGH_DPI

CoffeeFlux:
The performance impact is minor, and it can cause boundaries to be off by a pixel or so. Originally discussed on IRC, and I've opened the pull request to get more formal feedback from everyone involved.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -49,7 +49,7 @@ AC_ARG_ENABLE([require-system-font-provider], AS_HELP_STRING([--disable-require-
     [allow compilation even if no system font provider was found @<:@default=enabled:>@]))
 AC_ARG_ENABLE([asm], AS_HELP_STRING([--disable-asm],
     [disable compiling with ASM @<:@default=check@:>@]))
-AC_ARG_ENABLE([render-lq], AS_HELP_STRING([--enable-render-lq],
+AC_ARG_ENABLE([high-dpi], AS_HELP_STRING([--enable-high-dpi],
     [adjust various spatial accuracy parameters for better performance but slightly worse quality @<:@default=disabled@:>@]))
 
 AS_IF([test x$enable_asm != xno], [
--- a/configure.ac
+++ b/configure.ac
@@ -49,8 +49,8 @@ AC_ARG_ENABLE([require-system-font-provider], AS_HELP_STRING([--disable-require-
     [allow compilation even if no system font provider was found @<:@default=enabled:>@]))
 AC_ARG_ENABLE([asm], AS_HELP_STRING([--disable-asm],
     [disable compiling with ASM @<:@default=check@:>@]))
-AC_ARG_ENABLE([large-tiles], AS_HELP_STRING([--enable-large-tiles],
-    [use larger tiles in the rasterizer (better performance, slightly worse quality) @<:@default=disabled@:>@]))
+AC_ARG_ENABLE([render-lq], AS_HELP_STRING([--enable-render-lq],
+    [adjust various spatial accuracy parameters for better performance but slightly worse quality @<:@default=disabled@:>@]))
 
 AS_IF([test x$enable_asm != xno], [
     AS_CASE([$host],
@@ -126,11 +126,11 @@ AM_COND_IF([ASM],
     [AC_DEFINE(CONFIG_ASM, 0, [ASM enabled])]
     )
 
-AM_CONDITIONAL([ENABLE_LARGE_TILES], [test x$enable_large_tiles = xyes])
+AM_CONDITIONAL([ENABLE_RENDER_LQ], [test x$enable_render_lq = xyes])
 
-AM_COND_IF([ENABLE_LARGE_TILES],
-    [AC_DEFINE(CONFIG_LARGE_TILES, 1, [use large tiles])]
-    [AC_DEFINE(CONFIG_LARGE_TILES, 0, [use small tiles])],
+AM_COND_IF([ENABLE_RENDER_LQ],
+    [AC_DEFINE(CONFIG_RENDER_LQ, 1, [enable optimizations])]
+    [AC_DEFINE(CONFIG_RENDER_LQ, 0, [disable optimizations])],
     )
 
 PKG_CHECK_MODULES([FREETYPE], freetype2 >= 9.10.3, [
--- a/libass/ass_func_template.h
+++ b/libass/ass_func_template.h
@@ -92,7 +92,7 @@ void DECORATE(blur8_vert)(int16_t *dst, const int16_t *src,
 const BitmapEngine DECORATE(bitmap_engine) = {
     .align_order = ALIGN,
 
-#if CONFIG_RENDER_LQ
+#if CONFIG_HIGH_DPI
     .tile_order = 5,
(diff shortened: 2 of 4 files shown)


MrSmile:
I should mention that there are several optimizations that can induce shift of boundaries and this line is only one of it. So this PR is incomplete as is and I think in the direction of renaming `CONFIG_LARGE_TILES` to something like `CONFIG_RENDER_LQ` and moving bunch of spatial accuracy parameters (especially from PR #285) under it.

CoffeeFlux:
That sounds fine to me; I'll update accordingly.

CoffeeFlux:
Looks like #285 is moving again, so once that's merged I'll rebase and add everything necessary to get this good to go.

CoffeeFlux:
MrSmile mentioned there are other optimizations that would potentially affect the bounding box, but I'm not sure what those are. This is the main one I've personally run into and care about having a compiler flag for so I can stop compiling with an extra patch, though I haven't tested anything really since #285 landed. Any suggestion on what else I need to do to get this landed?

MrSmile:
What I think should be done:
- Instead of `CONFIG_RENDER_LQ` something like `CONFIG_LARGE_SCREEN` or `CONFIG_HI_DPI` will be better.
- Spatial precision constants should be multiplied by the same factor (2 is good enough) under that mode, i. e.
```
#define RASTERIZER_PRECISION 32
#define POSITION_PRECISION 16.0
#define SUBPIXEL_ORDER 2
```
But such a low precision can be too extreme, maybe it's better to keep current values for `CONFIG_HI_DPI` and use improved values for default instead:
```
#define RASTERIZER_PRECISION 8
#define POSITION_PRECISION 4.0
#define SUBPIXEL_ORDER 4
```
Or maybe some mix of these two variants. `SUBPIXEL_ORDER` is dependent on `POSITION_PRECISION`, so only two other constants need to be tuned. Well, personally I think that tuning can be left for the future and first variant is good enough.

CoffeeFlux:
Thanks, I'll revisit this during the weekend.

astiob:
I’m getting dangerously close to wasteful bikeshedding, but:

Given that the original motivation was to provide accurate bitmap boundaries, isn’t naming it `...LARGE_SCREEN` or `...HIGH_DPI` kinda missing the point? @MrSmile has always asserted that the division approximation has negligible visual effect even in the default “higher quality” mode, so tying it to screen size/DPI of all things seems weird.

CoffeeFlux:
@astiob like you said, I only care about a way to get rid of that one optimization without patching libass. The name and other functionality don't really matter to me, so I'll do whatever is asked to get this merged.

MrSmile:
> Given that the original motivation was to provide accurate bitmap boundaries, isn’t naming it `...LARGE_SCREEN` or `...HIGH_DPI` kinda missing the point?

The original motivation is only about pixel-perfect axis-aligned boxes and it's almost not related to general use case of libass at all (also it's a classical example of mishandling of floating-points). So I've tried to extend the original problem in a way that can be related to some real-world use cases. And the resulting use case is adjustable spatial accuracy which can be useful due to the modern abundance of large resolution (4k, 8k) and high DPI (retina) displays. Hence the name.

I've done some benchmarking of different values of spatial accuracy constants, here are the results:
| LT|div| RP | PP| err   | 360    | 720    | 1080  |
|---|---|----|---|-------|--------|--------|-------|
| &minus; | &minus; | 16 | 8 | 1.813 | 25.854 | 35.021 | 44.695|
| &minus; | &minus; | 16 | 4 | 1.475 | 25.675 | 34.960 | 44.926|
| &minus; | &minus; | 8  | 8 | 1.850 | 27.633 | 39.893 | 51.677|
| &minus; | &minus; | 8  | 4 | 1.545 | 27.692 | 39.883 | 51.428|
| &minus; | + | 16 | 8 | 1.802 | 28.418 | 37.860 | 47.905|
| &minus; | + | 16 | 4 | 1.436 | 28.445 | 37.870 | 47.859|
| &minus; | + | 8  | 8 | 1.840 | 30.315 | 42.809 | 55.250|
| &minus; | + | 8  | 4 | 1.514 | 30.295 | 42.824 | 55.291|
| + | &minus; | 16 | 8 | 1.831 | 23.223 | 31.746 | 40.340|
| + | &minus; | 16 | 4 | 1.642 | 23.125 | 31.629 | 40.454|
| + | &minus; | 8  | 8 | 1.889 | 24.805 | 35.847 | 46.795|
| + | &minus; | 8  | 4 | 1.702 | 24.724 | 36.066 | 46.728|
| + | + | 16 | 8 | 1.804 | 25.663 | 34.274 | 43.148|
| + | + | 16 | 4 | 1.534 | 25.645 | 34.426 | 43.353|
| + | + | 8  | 8 | 1.831 | 27.399 | 39.038 | 49.994|
| + | + | 8  | 4 | 1.622 | 27.670 | 39.137 | 49.999|

Legend:
* **LT** = `CONFIG_LARGE_TILES`;
* **div** = presence of division (original motivation);
* **RP** = `RASTERIZER_PRECISION`;
* **PP** = `POSITION_PRECISION`;
* **err** = average error of some bunch of `compare` tests;
* **360**/**720**/**1080** = time of rendering of my rasterizer-heavy clip at different screen sizes.

My conclusions from that:
* `CONFIG_LARGE_TILES` is not that bad: insignificant drop of visual quality only at `POSITION_PRECISION = 4` with 9&ndash;10% faster rendering. Maybe it should be enabled by default? Also the current configure switch for it is broken.
* division is almost the opposite of the large tiles: even more insignificant quality increase at the cost of 7&ndash;12% rendering time.
* `RASTERIZER_PRECISION = 8` is useless. Quick experiments show that `RASTERIZER_PRECISION = 16` is nearly optimal, I'll repeat this benchmarking with larger values later.
* `POSITION_PRECISION = 4` don't have an impact on performance (as my test clip isn't slow animation). Potential performance impact is proportional to the value (i.e. it's possible to construct a test clip that will be almost two times slower), but such corner cases is quite rare. Given that it has the most visual impact, it's the best candidate for improvement.

So I don't think that division should be enabled by default. Instead, `CONFIG_LARGE_TILES` should be enabled (maybe only in presence of `CONFIG_ASM`, I should do some tests for that), `POSITION_PRECISION` set to 4 and maybe `RASTERIZER_PRECISION` to some value larger than 16. As for division I can create some configure switch (`EXTENDED_PRECISION`) with some extreme values optimal for precision. I'll do some additional testing and create PR myself later.

CoffeeFlux:
I'm inclined to agree the division shouldn't be enabled by default. I just want some way to get accurate bounds without having to continually update a patched libass fork, so an option existing but off by default was the original intention.

rcombs:
I sorta feel like we're at the point where we really should be offering a text-extents API. Like, pass in an ASS event, and we put it through all the parsing and part of the rendering, but then stop once we have extents and return that. Seems like a lot of people need that and currently people tend to approximate it by looking at the actual bitmap extents, which constrains us in awkward ways (and doesn't give any sub-pixel positioning information).

TheOneric:
Superseded by #449

PR #337, opened 2019-03-17
ass_parse: fix overflow in parse_tags around \t

rcombs:
Found by clusterfuzz: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=11632

### DIFF
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -620,12 +620,12 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
             long long t1, t2, t, delta_t;
             double k;
             if (cnt == 3) {
-                t1 = argtoll(args[0]);
-                t2 = argtoll(args[1]);
+                t1 = FFMIN(FFMAX(argtoll(args[0]), INT_MIN), INT_MAX);
+                t2 = FFMIN(FFMAX(argtoll(args[1]), INT_MIN), INT_MAX);
                 accel = argtod(args[2]);
             } else if (cnt == 2) {
-                t1 = argtoll(args[0]);
-                t2 = argtoll(args[1]);
+                t1 = FFMIN(FFMAX(argtoll(args[0]), INT_MIN), INT_MAX);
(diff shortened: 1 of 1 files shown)


MrSmile:
Personally I prefer more sensible limit values. Do negative time even make sense? Also I think that limit should be integrated into parsing function: currently we have `(long long) double_val` inside `argtoll()` and that can be UB too.

rcombs:
Negative values are uncommon, but result in well-"defined" numeric results. The clamp values I used here are the same as what VSFilter effectively has (since it stores these values in `int`s). Yeah, `argtoll` should be fixed as well.

astiob:
@rcombs I think this is also addressed by the older #261 (which you’ve just reviewed, thanks!). It does it in a more pedantic way which I think @MrSmile doesn’t like, but then if he doesn’t like your approach either, perhaps we can go with the more pedantic and less functionally impactful one? 😃 Of course, the source numbers are still long longs, which both can break and doesn’t match VSFilter, but from there on #261 actually computes the exact differences without any clamping.

rcombs:
I don't feel super strongly either way 🤷‍♀️, whatever matches better works for me.

astiob:
If you mean matching VSFilter, then for that we’d need to read values [as saturated int32 and as doubles truncated to int32](https://github.com/Cyberbeing/xy-VSFilter/blob/3.0.0.306/src/subtitles/RTS.cpp#L2767-L2774). I forgot those int32s were saturated; so your patch is indeed a better match.

But actually… matching VSFilter exactly is not hard at all, so I think I’ll create a patch to do just that. We even already have a [dedicated function to read saturated int32](https://github.com/libass/libass/blob/master@%7B2020-07-10%7D/libass/ass_utils.c#L180-L186).

As already noted, we currently read everything [as doubles rounded (with UB overflow) to long long or int](https://github.com/libass/libass/blob/master@%7B2020-07-10%7D/libass/ass_utils.c#L168-L169). I now realize that this makes my pedantic subtraction rather useless, as the long longs have gone through double already, so it’s no better than `a - (double) b`.

PR #342, opened 2019-05-20
Fix bunch of UB

MrSmile:
Fix potential problem found by Coverity Scan and couple of errors found by -fsanitize=undefined.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -188,7 +188,7 @@ static int test_and_set_read_order_bit(ASS_Track *track, int id)
     if (resize_read_order_bitmap(track, id) < 0)
         return -1;
     int index = id / 32;
-    int bit = 1 << (id % 32);
+    uint32_t bit = 1u << (id % 32);
     if (track->parser_priv->read_order_bitmap[index] & bit)
         return 1;
     track->parser_priv->read_order_bitmap[index] |= bit;
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2295,6 +2295,7 @@ static void render_and_combine_glyphs(ASS_Renderer *render_priv,
                 nb_bitmaps++;
             }
             last_info = info;
+            assert(current_info);
 
             ASS_Vector pos, pos_o;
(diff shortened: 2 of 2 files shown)


astiob:
Upon more careful inspection, it seems `!current_info` is impossible at that point. Whenever we reach the `last_info = info;` line, we’ve already initialized `current_info` to a valid pointer.

rcombs:
Maybe just `assert`, then? Unless there's some case we're missing (and I don't _think_ there is).

astiob:
This check was added in 30004491d2dc3ab84520333b1e13c82a99aa081f, which already acknowledged it was impossible but added it anyway to guard against future code changes that might make it a possibility. The direct trigger for that commit was a warning from Clang’s static analyser, which presumably failed to recognize that `is_new_bm_run` would always return `true` until the variable was initialized. I don’t know if Clang can be taught about this (an `assert` might work), but in any case, I can see some appeal in guarding against future changes, since I think this isn’t immediately obvious to a human either.

MrSmile:
I should add some comment about unreachable code to clarify that check then.

astiob:
I also like the `assert` suggestion. \*shrugs\*

rcombs:
The clang static analyzer respects assertions.

MrSmile:
Does it treat `assert(ptr)` and `ptr->something` different? It's the same class of problem for me.

rcombs:
If you `assert(ptr)` it assumes that `ptr` will always be non-null at that point.

MrSmile:
Even with something totally wrong like `assert(2 + 2 = 5)`? Okay, I replace check with assert then.

rcombs:
The way it actually works is that `assert(e)` is a macro defined like this:
```
#define assert(e) \
    (__builtin_expect(!(e), 0) ? __assert_rtn(__func__, __FILE__, __LINE__, #e) : (void)0)
```
and `__assert_rtn` is declared with `__attribute__((noreturn))`. So `assert(2 + 2 == 5)` is `assert(false)`, which is effectively equivalent to `abort();` from the static analyzer's perspective, so it'll assume that anything past it is unconditionally unreachable.

PR #344, opened 2019-05-26
Infrastructure for regression testing

MrSmile:
This PR introduces regression testing program but doesn't contain any tests.

Test program command line:
`compare [-i] <input-dir> [-o <output-dir>] [-s <scale:1-8>]`

`<input-dir>` here is test directory which consists of font files, subtitle files and images. For example
```
test/
  font1.ttf
  font2.otf
  sub1.ass
  sub1-0500.png
  sub1-1500.png
  sub1-2500.png
  sub2.ass
  sub2-153000.png
```
All required fonts should be present in test directory as libass configured to not use anything external (`ASS_FONTPROVIDER_NONE`). After loading all fonts in test directory program scans for subtitle files (`<sub_name>.ass`) and their corresponding image files in form of `<sub_name>-NNNN.png`. Subtitle file then would be rendered at `NNNN` milliseconds and compared with image file.

`<output-dir>` if present sets directory to store rendering results, and `-s` switch sets oversampling factor (default 1).


compare/compare.c
```diff
@@ -0,0 +1,639 @@
...
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <dirent.h>
+#include <string.h>
+
+
+#define FFMAX(a,b) ((a) > (b) ? (a) : (b))
+#define FFMIN(a,b) ((a) > (b) ? (b) : (a))
+
+static void blend_image(Image8 *frame, int32_t x0, int32_t y0,
```
rcombs: Not really a comment on the PR itself, but there are enough separate implementations of this particular blend routine floating around that I feel like it'd be worth having libass provide it itself (which could also feature asm optimizations that API consumers could benefit from).

compare/compare.c
```diff
@@ -0,0 +1,639 @@
...
+{
+    uint16_t res = 0;
+    for (int k = 0; k < 4; k++) {
+        uint16_t diff = abs_diff(a[k], b[k]);
+        res = FFMAX(res, diff);
+    }
+    return res;
+}
+
+static void calc_grad(const Image16 *target, uint16_t *grad)
```
rcombs: This needs some comments on what's being done here.

compare/compare.c
```diff
@@ -0,0 +1,639 @@
...
+        struct dirent *file = readdir(dir);
+        if (!file)
+            break;
+        const char *name = file->d_name;
+        if (name[0] == '.')
+            continue;
+        int len = strlen(name);
+        if (len <= 4)
+            continue;
+        switch (COMBINE(name[len - 4], name[len - 3],
```
rcombs: This is overkill. Just use `strrchr` and `strcmp`; it's much clearer.

compare/compare.c
wm4: I don't get what it does either. Some sort of fuzzy comparison?

wm4:
This is definitely something that is much needed. I think you should add an example test too.

It's a bit unfortunate that this seems to require putting test cases a png into the repo. Wouldn't that blow up the repo size after a while? Or maybe they're small enough and compress well enough that this won't be an issue? What about the required fonts?

I agree that this shouldn't use an OS font provider, and use hardcoded font files instead.

compare/compare.c
wm4: +1

compare/compare.c
MrSmile: It's to account for spatial error.
For example, 1/4 pixel offset can transform `0 0 0 255 255` into `0 0 64 255 255`, therefore error of 64 can be acceptable in one pixel position but not others.
`calc_grad()` estimates value change around each pixel to account for that.
It's not the best metric but I think it can be starting point.

compare/compare.c
MrSmile: I've seen many cases where chained string comparisons lead to significant performance degradation (and it's long-standing problem of libass parsing) so I consider it bad practice and avoid it when I can. Is that specific chunk of code hard to understand?

MrSmile:
Documentation isn't my strong point but I'll try to write something (does github show README.md is subdirs automatically?).

As for real tests, I avoided putting anything due to font license concerns. And size can be a problem too, especially for fonts with large unicode coverage. Maybe it's better to put tests in another repo altogether?

compare/compare.c
wm4: That may be true for subtitle parsing, but it's complete overkill in this case, and it's confusing. It'll also be obnoxious if someone wants to match a different number of characters, or decides to make the matching case-insensitive. Regarding bad practice, I think code should primarily be easy to read, and there need to be good reasons to do something else.

wm4:
>does github show README.md is subdirs automatically?

Yes.

>As for real tests, I avoided putting anything due to font license concerns. And size can be a problem too, especially for fonts with large unicode coverage. Maybe it's better to put tests in another repo altogether?

License problems shouldn't be an issue as long as the font is freely redistributable. The software license needs to apply to actual code only (it's not like we include the font as byte array into compiled libass or so).

A separate repo for this doesn't sound wrong. Or maybe a hybrid approach, where the main repository contains a list of filenames for test cases that must pass, while the actual test data is in a separate repo? (Might make it easier when bisecting? Not sure if it makes sense.)

compare/compare.c
rcombs: Yeah, the subtitle parsing code is a hot tight loop, and this isn't.

rcombs:
Whether the test images are in the main repo or a separate one, they should be included via git LFS, which avoids putting every version of them in the git history forever.

MrSmile:
Found some time to work with that PR again. I think most issues (except lack of test samples) are therefore resolved. Also any help with editing (or altogether rewriting) of `README.md` is welcome.

wm4:
Is there a complete example?

MrSmile:
> Is there a complete example?

No, there is no test examples in this commit. I have some tests that I used for development of this PR but it's not tailored to specifically test libass rendering (and also I used fonts with unclear origins). I can make simple test as example but we should decide a method for maintaining test files first.

wm4:
There should be at least a small one. For that one it probably doesn't matter if the files are in the libass repo itself, since it will be a small and one-time thing.

MrSmile:
I've added simple example (the same as described in `README.md`).

compare/compare.c
```diff
@@ -0,0 +1,627 @@
...
+
+static bool add_img_item(ItemList *list, const char *file, int len)
+{
+    // Parse image name:
+    // <subtitle_name>-<time_in_msec>.png
+
+    int pos = len, first = len;
+    while (true) {
+        if (!pos--)
+            return true;
```
TheOneric, line 425: Is `true` instead of `false` intentional in lines 425,429,433 ?

compare/compare.c
```diff
@@ -0,0 +1,627 @@
...
+        if (file[pos] == '-')
+            break;
+        if (file[pos] < '0' || file[pos] > '9')
+            return true;
```
TheOneric, line 429: s. comare.c:425

compare/compare.c
```diff
@@ -0,0 +1,627 @@
...
+        if (file[pos] != '0')
+            first = pos;
+    }
+    if (pos + 1 == len || first + 15 < len)
+        return true;
+
+    if (!add_item(list))
+        return false;
```
TheOneric, line 437: s. `compare.c:425`

compare/compare.c
```diff
@@ -0,0 +1,627 @@
...
+            return true;
+        if (file[pos] == '-')
+            break;
+        if (file[pos] < '0' || file[pos] > '9')
+            return true;
+        if (file[pos] != '0')
+            first = pos;
+    }
+    if (pos + 1 == len || first + 15 < len)
+        return true;
```
TheOneric, line 434: s. `compare.c:425`

compare/compare.c
MrSmile: Yes, `false` means unrecoverable error (allocation failure) that leads to immediate abort. Some stray `png` file in test directory is almost regular case so `true` is intended here.

TheOneric: Ok, that works, though a comment documenting the return values might be helpful.
While not strictly necessary, I think it might also be a beneficial to at least print a info about ignored images, to make sure no images are left out on accident.
_(Or alternatively, the test could be more strict and require additional images to be prefixed with a dot, as dotfiles are filtered out anyway in `main`.)_

MrSmile: I think that test directory can contain not only automatic test files but some human-readable description too. And that description can have images. Also I think that ability to disable tests by changing extension of `ass`-files only is useful too. I'm reporting all active test images anyway (with total count) so I think that's sufficient as error check at test creation.

PR #353, opened 2019-10-15
parse_tags: fix case where t==t1==t2

rcombs:
This previously gave the pre-transition value; VSFilter's behavior is to give the post-transition value.

hi @skiddiks


astiob:
This sounds familiar… \*scratches head\*

PR #359, opened 2019-11-06
coretext: fix error handling in get_font_file

### LINKED ISSUE #358: font selection for emoji and special symbols will crash on macOS
This problem may be caused by CoreText.framework which is libass's default font selection backends on macOS. 

For example, “❤” (U+2764) in chinese ass subtitle can cause crash.

**Log file**
Crash.log https://0x0.st/zgfy.log
mpv.log https://0x0.st/zgfx.log

**Sample files**
sample.ass https://0x0.st/zgfv.ass

**Reference**
https://github.com/mpv-player/mpv/issues/7123
https://github.com/iina/iina/pull/2638 
https://github.com/iina/iina/issues/1101

ePirat:
Fixes a crash in case a font does not has kCTFontURLAttribute,
which is the case for example on macOS 10.15.1 for the
".AppleSymbolsFB" font.

Fix #358

### DIFF
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -96,7 +96,13 @@ static bool check_glyph(void *priv, uint32_t code)
 static char *get_font_file(CTFontDescriptorRef fontd)
 {
     CFURLRef url = CTFontDescriptorCopyAttribute(fontd, kCTFontURLAttribute);
+    if (!url)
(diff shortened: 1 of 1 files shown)


wm4:
Seems sane and fixes an obvious bug -> merged.

PR #362, opened 2019-12-06
Add function to get the width of a font from the OS/2 table

downthomas:
I saw the FIXME in the source code on font width, and basically copied ass_face_get_weight while substituting usWidthClass for usWeightClass. So it will query the OS/2 table for the actual font width, and if it fails, it returns 100.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -87,7 +87,7 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
     if (!face->charmap)
         return symbol;
 
-    switch(face->charmap->encoding){
+    switch (face->charmap->encoding) {
     case FT_ENCODING_MS_SYMBOL:
         return 0xF000 | symbol;
     default:
@@ -317,6 +317,37 @@ int ass_face_get_weight(FT_Face face)
         return os2->usWeightClass;
     else
         return 300 * !!(face->style_flags & FT_STYLE_FLAG_BOLD) + 400;
+}
+
+/**
+ * \brief Get face width
+ **/
+int ass_face_get_width(FT_Face face)
+{
+#if FREETYPE_MAJOR > 2 || (FREETYPE_MAJOR == 2 && FREETYPE_MINOR >= 6)
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
+#else
+    // This old name is still included (as a macro), but deprecated as of 2.6, so avoid using it if we can
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, ft_sfnt_os2);
+#endif
+    if (os2 && os2->version != 0xffff && os2->usWidthClass) {
+		WidthClass wdth = os2->usWidthClass;
+		switch (wdth) {
+		case FWIDTH_ULTRA_CONDENSED:	return 50;
+		case FWIDTH_EXTRA_CONDENSED:	return 62;
+		case FWIDTH_CONDENSED:		return FONT_WIDTH_CONDENSED;
+		case FWIDTH_SEMI_CONDENSED:	return 87;
+		case FWIDTH_NORMAL:		return FONT_WIDTH_NORMAL;
+		case FWIDTH_SEMI_EXPANDED:	return 112;
+		case FWIDTH_EXPANDED:		return FONT_WIDTH_EXPANDED;
+		case FWIDTH_EXTRA_EXPANDED:	return 150;
+		case FWIDTH_ULTRA_EXPANDED:	return 200;
+		default:
+			return FONT_WIDTH_NORMAL;
+	}
+    }
+    else
+        return FONT_WIDTH_NORMAL;
 }
 
 /**
--- a/libass/ass_font.h
+++ b/libass/ass_font.h
@@ -57,11 +57,24 @@ struct ass_font {
     double size;
 };
 
+typedef enum WidthClass { 
+  FWIDTH_ULTRA_CONDENSED = 1, 
+  FWIDTH_EXTRA_CONDENSED = 2, 
+  FWIDTH_CONDENSED = 3, 
+  FWIDTH_SEMI_CONDENSED = 4, 
+  FWIDTH_NORMAL = 5, 
+  FWIDTH_SEMI_EXPANDED = 6, 
+  FWIDTH_EXPANDED = 7, 
+  FWIDTH_EXTRA_EXPANDED = 8, 
+  FWIDTH_ULTRA_EXPANDED = 9 
+} WidthClass;
+
 void charmap_magic(ASS_Library *library, FT_Face face);
 ASS_Font *ass_font_new(ASS_Renderer *render_priv, ASS_FontDesc *desc);
 void ass_face_set_size(FT_Face face, double size);
 void ass_font_set_size(ASS_Font *font, double size);
 int ass_face_get_weight(FT_Face face);
+int ass_face_get_width(FT_Face face);
 void ass_font_get_asc_desc(ASS_Font *font, int face_index,
                            int *asc, int *desc);
 int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -758,7 +758,7 @@ get_font_info(FT_Library lib, FT_Face face, ASS_FontProviderMetaData *info)
     int num_fullname = 0;
     int num_family   = 0;
     int num_names = FT_Get_Sfnt_Name_Count(face);
-    int slant, weight;
+    int slant, weight, width;
     char *fullnames[MAX_FULLNAME];
     char *families[MAX_FULLNAME];
 
@@ -809,11 +809,12 @@ get_font_info(FT_Library lib, FT_Face face, ASS_FontProviderMetaData *info)
     // calculate sensible slant and weight from style attributes
     slant  = 110 * !!(face->style_flags & FT_STYLE_FLAG_ITALIC);
     weight = ass_face_get_weight(face);
+    width = ass_face_get_width(face);
 
     // fill our struct
     info->slant  = slant;
     info->weight = weight;
-    info->width  = 100;     // FIXME, should probably query the OS/2 table
+    info->width  = width;
 
     info->postscript_name = (char *)FT_Get_Postscript_Name(face);
 


rcombs:
Hmmmm, this value is used only within font selection AFAIK; it's considered at the same level as weight and slant, and we always search for 100 (so this will only reduce the scores of fonts with other widths). The DirectWrite and fontconfig fontselect engines already read this, but it's unclear where in the font they get it from. Can we test to see how VSFilter behaves around this?

libass/ass_font.c
```diff
@@ -319,6 +319,23 @@ int ass_face_get_weight(FT_Face face)
...
+int ass_face_get_width(FT_Face face)
+{
+#if FREETYPE_MAJOR > 2 || (FREETYPE_MAJOR == 2 && FREETYPE_MINOR >= 6)
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
+#else
+    // This old name is still included (as a macro), but deprecated as of 2.6, so avoid using it if we can
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, ft_sfnt_os2);
+#endif
+    if (os2 && os2->version != 0xffff && os2->usWidthClass)
+        return os2->usWidthClass;
```
rcombs: This just copies the value in directly, but the value this is stored in is in fontconfig's units, defined in terms of the os2 value here: https://github.com/behdad/fontconfig/blob/5b41ded2b0ddb98a07ac86264b94403cb7a0fd82/src/fcfreetype.c#L1782
And also multiplied by a `wdth` value that comes from… somewhere? I dunno: https://github.com/behdad/fontconfig/blob/5b41ded2b0ddb98a07ac86264b94403cb7a0fd82/src/fcfreetype.c#L1362

downthomas: I should have read the [OS/2 table specs](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#uswidthclass) a bit more closely. The wdth value comes from the fvar table and STAT table it seems?

downthomas: I should have looked at the [OS/2 metrics table spec](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#uswidthclass) more closely. Looks like `wdth` comes from the fvar and STAT tables.

libass/ass_font.c
```diff
@@ -330,8 +330,29 @@ int ass_face_get_width(FT_Face face)
     // This old name is still included (as a macro), but deprecated as of 2.6, so avoid using it if we can
     TT_OS2 *os2 = FT_Get_Sfnt_Table(face, ft_sfnt_os2);
 #endif
-    if (os2 && os2->version != 0xffff && os2->usWidthClass)
-        return os2->usWidthClass;
+    if (os2 && os2->version != 0xffff && os2->usWidthClass){
```
rcombs: Nit: space between `)` and `{`

libass/ass_font.c
```diff
@@ -330,8 +330,29 @@ int ass_face_get_width(FT_Face face)
...
+		WidthClass wdth = os2->usWidthClass;
+		switch(wdth){
```
rcombs: Nit: spaces around `()`

downthomas: Fixed.

libass/ass_font.c
downthomas: Fixed.

rcombs:
>If we're trying to match VSFilter's behavior, the CoreText font matching was closer. But I'm not sure that should continue to be the goal.

This has been discussed on numerous occasions, and the conclusion has always been that backwards compatibility with existing scripts is critical to libass; software can be updated and bugs fixed, but generally subtitle files that have already been distributed will never be updated.

Re: "the CoreText font matching was closer", sorry, I'm not quite clear on what you meant here? Did the recent CoreText changes (e.g. be0d1613f79a95073d18d96a60e1394abf9316a2) result in libass failing to match VSFilter behavior where it previously had? If so, that wasn't the intention and I'd want to put the relevant changes behind a script-opt-in flag, and confirm the behavior of the other fontselect engines (I'd expected those changes to bring CoreText's behavior more in line with the others, and also to VSFilter, but it's possible I was wrong).

> Regardless of the fact that I'll have to do a font name change on hundreds of scripts,

If these scripts work in VSFilter but not in libass, we'll want to fix that here. Otherwise, we'd usually consider this a script bug (unless it's easily identifiable that a script was _deliberately_ relying on old libass behavior, but that's usually easier said than done).

To bring this back to the original point of this PR, how does VSFilter interact with font widths? For instance, to Avenir Roman and Avenir Book have different widths, and if so, which does VSFilter choose if you use `\b400`?

downthomas:
To answer the easy questions:
1. VSFilter does not interact with font width. The only thing VSFilter uses to choose a font is the font name. If VSFilter does not have a font name match, it uses the default font.
2. VSFilter only has (FW=font weight) FW_BOLD and FW_NORMAL. 
3. The only possibility here that libass would produce different results than VSFilter is if a script specifies a font that isn't matched and a substitute is searched for. Libass would score condensed or expanded fonts worse than before, so they would be less likely to be matched.

4. On Windows, when only "Avenir" is requested, both libass and VSFilter return "Avenir Roman," because of how DirectWrite returns the font request and the search order.
On macOS, the Coretext backend returns an list of matching fonts seemingly ordered by index and matches to the first font at index 0, "Avenir Book."
So in this case, libass on macOS never matched VSFilter's behavior.
There's something going on with the name matching on macOS that I don't quite understand just yet, and I've been working on a set of changes to the coretext backend to make it more reliably pick fonts, but I'm not done yet.

rcombs:
>VSFilter does not interact with font width. The only thing VSFilter uses to choose a font is the font name. If VSFilter does not have a font name match, it uses the default font.

Font name, weight, and italics, right? (in that order?)

>VSFilter only has (FW=font weight) FW_BOLD and FW_NORMAL.

I don't think this is correct; the VSFilter code seems to be similar to libass's: https://github.com/Cyberbeing/xy-VSFilter/blob/cf8f5b27de77fe649341bfab0fdfd498e1ad2fa6/src/subtitles/RTS.cpp#L2381

> The only possibility here that libass would produce different results than VSFilter is if a script specifies a font that isn't matched and a substitute is searched for. 

Aren't you seeing differing results in your test here, where libass on macOS picks Avenir Book, but VSFilter (and libass's DirectWrite fontselect) picks Avenir Roman? If so, we should fix that by figuring out how DirectWrite decides to prefer the Roman one (clearly not alphabetical…).

In any case, if VSFilter/GDI doesn't interact with width at all in font selection, we probably don't want to merge this PR, since it won't really improve behavior in practice; if anything we should drop the equivalent code in the other fontselects. I'll keep it open for continued discussion, though.

downthomas:
After some investigation, I've determined the current cause of libass selecting "Avenir Book" over "Avenir Roman" when asked for just "Avenir" to be commit 6c2120a7cb4a6fd648ef37ad8f0d961cbd60f500, "fontselect: expose the freetype-provided family as well." That commit should be reverted.

Before using freetype to get font info, CoreText used the system provided kCTFontFamilyNameAttribute (probably pulled from TT_NAME_ID_TYPOGRAPHIC_FAMILY) in families, which was just "Avenir." The freetype-provided family name is the same, "Avenir." But with 6c2120a7cb4a6fd648ef37ad8f0d961cbd60f500, now each font gets at least two family names: the freetype-provided family name, and whatever is provided by TT_NAME_ID_FONT_FAMILY.

Libass on Windows doesn't call get_font_info when adding fonts to the font library, so the freetype-provided family name is never added to the list of family names unless the font is embedded. Fontconfig doesn't call get_font_info when adding fonts to the font library, either.

So the family names on Windows end up being:
```
new font:
  families: 'Avenir Book'
  fullnames: 'Avenir Book'
new font:
  families: 'Avenir'
  fullnames: 'Avenir Roman'

```
But on macOS:
```
new font:
  families: 'Avenir' 'Avenir Book' 
  fullnames: 'Avenir Book' 
new font:
  families: 'Avenir' 'Avenir' 
  fullnames: 'Avenir Roman' 

```

Reverting 6c2120a7cb4a6fd648ef37ad8f0d961cbd60f500 causes this discrepancy in font matching to disappear.
macOS:
`[sub/ass] fontselect: (Avenir, 700, 0) -> /System/Library/Fonts/Avenir.ttc, -1, Avenir-Roman`
`[sub/ass] fontselect: (Avenir, 400, 0) -> /System/Library/Fonts/Avenir.ttc, -1, Avenir-Roman`

Windows:
`[sub/ass] fontselect: (Avenir, 700, 0) -> Avenir-Roman, 11, Avenir-Roman`
`[sub/ass] fontselect: (Avenir, 400, 0) -> Avenir-Roman, 11, Avenir-Roman`

astiob:
@rcombs Do you remember where this ended up at? This PR’s discussion thread dove deeper into font matching, and I gather some of the issues mentioned have been resolved by backing out “fontselect: expose the freetype-provided family as well”. Do we want to read the width? If not, why do we have mentions of it at all? Do I understand correctly that it will be used for finding similar fonts for font fallback, which is inherently VSFilter-incompatible (until we decide to bite the bullet and remove clever fallback altogether) so we have free reign to improve it?

rcombs:
Looking through some GDI information and finding a few useful details:
- They also use a "score" system to compare potential matches
- A non-italic font when asked for italic will get a score penalty of 1 (or 4 if unable to make faux-italic, which I don't think is ever true in vsfilter's use-case)
- An italic font when asked for non-italic gets a penalty of 4
- As we'd found via experimentation, faux-bold occurs when the requested weight is more than 150 (derived from (700-400)/2) greater than the font's weight
  - When faux-bold occurs, the rest of the algorithm treats the font as if it were 120 weight units heavier (derived from (700-400)*2/5)
  - After this, the font is penalized by about `3*x/10`, where `x` is the absolute value of the weight difference
    - …Except this is actually implemented as `(((((((x)<<3)+(x))<<3))+(x))>>8)`, which is 73*x/256, which is… not even a particularly good approximation? I have no idea what the heck happened here.
- If a width is requested, the font is penalized by 50*ABS(desired-got)… but VSFilter doesn't request a width so this doesn't happen

So yeah, we can drop width altogether, and should probably improve our scoring algorithm (e.g. we currently have italics on a numeric scale when it should just be a boolean).

TheOneric:
iiuc, the change from this PR would worsen VSFilter compatibility and we should instead implement what rcombs described above. And since this is already tracked in #437, there seems to be no reason to keep this open.

PR #364, opened 2020-01-02
Update VSFilter links

wangqr:
Change links of VSFilters in README.md to new address or maintained forks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,10 @@ Information about the ASS format:
 Other ASS/SSA implementations:
 ==============================
 - VSFilter:
-  - [xy-VSFilter/XySubFilter](https://code.google.com/p/xy-vsfilter/)
-  - VSFilter in [MPC-HC](http://mpc-hc.org/)
-  - [VSFilterMod](https://code.google.com/p/vsfiltermod/) (with custom format extensions)
+  - [xy-VSFilter/XySubFilter](https://github.com/Cyberbeing/xy-VSFilter/)
+    - [pfmod](https://github.com/pinterf/xy-VSFilter/)
+  - VSFilter in [MPC-HC](https://github.com/clsid2/mpc-hc/tree/develop/src/filters/transform/VSFilter/)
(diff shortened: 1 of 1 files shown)


astiob:
Why do people keep creating hard-to-find forks instead of submitting patches upstream and offering to take over as the official maintainer? 🙈

Why do people keep importing Subversion-controlled source code into Git without importing the history? 🙈

Anyway, yes, this is good. But since [sorayuki/VSFilterMod](https://github.com/sorayuki/VSFilterMod) doesn’t contain the history and doesn’t actually seem to pick up immediately after the last upstream commit—it seems to include commits up to and including 90 but nothing after that—it would be best to keep the Google Code link as well. If you could do that, I’d be happy to merge this.

astiob:
You may also wish to link to https://github.com/MewX/vsfiltermod which is a GitHub export of VSFilterMod’s Google Code repository. It contains the code history and the issues, but not the wiki.

Oh, you literally just pushed that other thing. I’m happy to merge it as-is, but if you want to add the MewX export link, I can wait.

wangqr:
> You may also wish to link to https://github.com/MewX/vsfiltermod which is a GitHub export of VSFilterMod’s Google Code repository. It contains the code history and the issues, but not the wiki.

Seems that MewX is just an export with no further maintenance. So I personally think it is more useful to link to the original repo, as long as Google is not deleting the google code archive.

astiob:
Merged. Thanks!

I definitely agree that a link to the original repo is good; if for no other reason, then for the fact it has the wiki. But do note that the Google Code Archive has no change history browser: you can see the list of commits but not their contents. You can still download [an SVN dump](https://storage.googleapis.com/google-code-archive-source/v2/code.google.com/vsfiltermod/repo.svndump.gz) if you [know](https://dominikdorn.com/2016/05/how-to-recover-a-google-code-svn-project-and-migrate-to-github/) [how](https://gist.github.com/yancyn/3f870ca6da4d4a5af5618fbd3ce4dd13) and roll out a local browsable copy, but the GitHub export allows you browse it on the Web without any additional effort.

PR #366, opened 2020-01-06
font: simplify metrics handling; fixes #361

### LINKED ISSUE #361: Incorrect metrics used on some fonts
Sample: https://puu.sh/EKGA3/7130de7b82.ass
```
[Script Info]
ScriptType: v4.00+
PlayResX: 1920
PlayResY: 1080
Timer: 0.0000
WrapStyle: 0
YCbCr Matrix: TV.709
ScaledBorderAndShadow: yes

[Aegisub Project Garbage]
Export Encoding: Unicode (UTF-8)
Video File: ?dummy:24.000000:2700:1920:1080:255:254:254:c

[V4+ Styles]
Style: Default,Profile Medium,70,&H00FFFFFF,&HC0000000,&H00000000,&HA8000000,0,0,0,0,95,100,0,0,1,3,1.5,2,150,150,60,1

[Events]
Dialogue: 10,0:00:00.00,0:00:06.35,Default,,0,0,0,,They showed clear animosity toward us, \Nunlike any other demonic beasts we've encountered before.
```

This renders as 3 lines in libass, at a larger font size than in VSFilter or _apparently_ older libass versions, though I've been unable to find a version on which the issue doesn't repro in order to bisect. I could be missing something obvious, though; it's late and I'm sleepy.

Looks to be a separate issue from #238?

rcombs:
We used to do a lot of work to get freetype to imitate VSFilter's font-metrics handling. Instead, we now just duplicate its behavior directly early on. This seems to produce the same behavior as we had with previous FreeType versions (matching VSFilter) in much less code.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -95,14 +95,30 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
     }
 }
 
-static void buggy_font_workaround(FT_Face face)
+static void set_font_metrics(FT_Face face)
 {
-    // Some fonts have zero Ascender/Descender fields in 'hhea' table.
-    // In this case, get the information from 'os2' table or, as
-    // a last resort, from face.bbox.
+    // Mimicking GDI's behavior for asc/desc/height.
+    // These fields are (apparently) sometimes used for signed values,
+    // despite being unsigned in the spec.
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, ft_sfnt_os2);
+    if (os2 && ((short)os2->usWinAscent + (short)os2->usWinDescent != 0)) {
+        face->ascender  =  (short)os2->usWinAscent;
+        face->descender = -(short)os2->usWinDescent;
+        face->height    = face->ascender - face->descender;
+    }
+
+    // If we didn't have usable Win values in the OS/2 table,
+    // then the values from FreeType will still be in these fields.
+    // It'll use either the OS/2 typo metrics or the hhea ones.
+    // If the font has typo metrics but FreeType didn't use them
+    // (either old FT or USE_TYPO_METRICS not set), we'll try those.
+    // In the case of a very broken font that has none of those options,
+    // we fall back on using face.bbox.
+    // Anything without valid OS/2 Win values isn't supported by VSFilter,
+    // so at this point compatibility's out the window and we're just
+    // trying to render _something_ readable.
     if (face->ascender + face->descender == 0 || face->height == 0) {
-        TT_OS2 *os2 = FT_Get_Sfnt_Table(face, ft_sfnt_os2);
-        if (os2) {
+        if (os2 && (os2->sTypoAscender - os2->sTypoDescender) != 0) {
             face->ascender = os2->sTypoAscender;
             face->descender = os2->sTypoDescender;
             face->height = face->ascender - face->descender;
@@ -213,7 +229,7 @@ static int add_face(ASS_FontSelector *fontsel, ASS_Font *font, uint32_t ch)
     }
 
     charmap_magic(font->library, face);
-    buggy_font_workaround(face);
+    set_font_metrics(face);
 
     font->faces[font->n_faces] = face;
     font->faces_uid[font->n_faces++] = uid;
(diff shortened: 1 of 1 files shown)


libass/ass_font.c
```diff
@@ -95,8 +95,16 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
     }
 }
 
-static void buggy_font_workaround(FT_Face face)
+static void set_font_metrics(FT_Face face)
 {
+    // Mimicking GDI's behavior for asc/desc/height
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, ft_sfnt_os2);
```
MrSmile: There is the same os2 table later requested in this function. At least second request should be deleted, but even branch conditions have some overlap, maybe better to merge that too.

libass/ass_font.c
```diff
@@ -95,8 +95,16 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
...
+    if (os2 && (os2->usWinAscent + os2->usWinDescent != 0)) {
+        face->ascender  =  os2->usWinAscent;
+        face->descender = -os2->usWinDescent;
+        face->height    = face->ascender - face->descender;
+    }
+
     // Some fonts have zero Ascender/Descender fields in 'hhea' table.
```
MrSmile: Looks like that comment will be outdated too.

rcombs:
CC: @downthomas
Both `LastResort` (the classic one) and `.LastResort` (the modern one with a simple ?-in-box glyph for all code points) work fine for me with this change; not sure what you were referring to [here](https://github.com/libass/libass/issues/361#issuecomment-571403802).

libass/ass_font.c
```diff
@@ -95,14 +95,28 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
...
 
-static void buggy_font_workaround(FT_Face face)
+static void set_font_metrics(FT_Face face)
 {
-    // Some fonts have zero Ascender/Descender fields in 'hhea' table.
-    // In this case, get the information from 'os2' table or, as
-    // a last resort, from face.bbox.
+    // Mimicking GDI's behavior for asc/desc/height
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, ft_sfnt_os2);
+    if (os2 && (os2->usWinAscent + os2->usWinDescent != 0)) {
```
astiob, line 102: I think this needs extra `(short)` as mentioned in a comment that is removed by this commit (if we are to believe that comment). The assignments a few lines below don’t need it because they assign to `short` (although this goes through implementation-defined behaviour, but so does `(short)`), but this line promotes the `unsigned short`s to `int` during the addition, so the sum will never be zero unless both values are zero.

libass/ass_font.c
```diff
@@ -95,14 +95,28 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
...
+    // then the values from FreeType will still be in these fields.
+    // It'll use either the OS/2 typo metrics or the hhea ones.
+    // If the font has typo metrics but FreeType didn't use them
+    // (either old FT or USE_TYPO_METRICS not set), we'll try those.
+    // In the case of a very broken font that has none of those options,
+    // we fall back on using face.bbox.
+    // Anything without valid OS/2 Win values isn't supported by VSFilter,
+    // so at this point compatibility's out the window and we're just
+    // trying to render _something_ readable.
     if (face->ascender + face->descender == 0 || face->height == 0) {
```
astiob, line 118: This line is not modified by your commit so it’s by no means your fault, but while we’re at it, I think this is wrong. Shouldn’t this _subtract_ `descender` from `ascender`? As far as I understand surrounding code, `face->ascender` is normally positive and `face->descender` is normally negative, so as it is, this triggers the conditional for fonts whose baseline is exactly in the middle of the glyph, which hardly seems intended.

If you agree, we can probably fix this in a separate commit.

libass/ass_font.c
rcombs: Oh, totally overlooked that comment.
I'm adding the casts on the assignments as well for the sake of clarity, along with a corresponding comment.

libass/ass_font.c
rcombs: Hmmm, looking at 7c901187, I think the intent here was just "ascender == 0 && descender == 0"? Probably makes more sense as `face->ascender - face->descender` though, yeah.

MrSmile: Maybe it's better to use intent directly as `face->ascender == 0 && face->descender == 0`?
Also in the similar places above and below it too.

rcombs: Out of scope for the PR, but I think the subtraction is a more sensible thing to do (i.e. "check for 0 space between asc and desc") than checking if both are specifically 0.

PR #370, opened 2020-02-04
ass_blur: check for memory allocation size overflows

wm4:
Check for overflows that could happen with alignment and the
multiplication. The INT_MAX / 4 is somewhat approximate and assumes that
degenerate alignment values won't happen.

This still assumes that a possibly overflowing end_w/end_h calculation
doesn't make the compiler's optimizer destroy the overflow checks.

---

Sample at: https://github.com/mpv-player/mpv/issues/7435
This fixes a crash I encountered with the sample (at slightly below 4K display resolution), but doesn't fix much of the sample itself.

### DIFF
--- a/libass/ass_blur.c
+++ b/libass/ass_blur.c
@@ -849,8 +849,14 @@ bool ass_gaussian_blur(const BitmapEngine *engine, Bitmap *bm, double r2)
     int end_w = ((w + offset) & ~((1 << blur.level) - 1)) - 4;
     int end_h = ((h + offset) & ~((1 << blur.level) - 1)) - 4;
 
+    if (end_w >= INT_MAX / 4)
+        return false;
+
     const int stripe_width = 1 << (engine->align_order - 1);
-    int size = end_h * ((end_w + stripe_width - 1) & ~(stripe_width - 1));
+    int aligned_end_w = (end_w + stripe_width - 1) & ~(stripe_width - 1);
+    if (end_h >= INT_MAX / 8 / aligned_end_w)
+        return false;
+    int size = end_h * aligned_end_w;
     int16_t *tmp = ass_aligned_alloc(2 * stripe_width, 4 * size, false);
     if (!tmp)
         return false;


astiob:
@MrSmile Could you briefly look this over just to make sure this isn’t by any chance cutting off a fully intended path?

libass/ass_blur.c
```diff
@@ -849,8 +849,14 @@ bool ass_gaussian_blur(const BitmapEngine *engine, Bitmap *bm, double r2)
     int end_w = ((w + offset) & ~((1 << blur.level) - 1)) - 4;
     int end_h = ((h + offset) & ~((1 << blur.level) - 1)) - 4;
 
+    if (end_w < 1 || end_h < 1 || end_w >= INT_MAX / 4)
```
MrSmile: `< 1` check doesn't have sense here. It's impossible to produce negative values (and zero) in regular case. The only problem here is overflow in `w + offset` and `h + offset`, but it's UB and should be checked before operation.

wm4: I did it because of the ``(...) - 4`` above. Are you sure the left side can't be lower than 5?

MrSmile: Yes. It can be seen that `offset >= (12 << blur.level)` so `end_w > 8`. It's exact calculation of final size of bitmap after bunch of different stages, but each stage have positive minimal result size.

wm4:
> Maybe it's time to introduce special functions for that (add_overflow() and mul_overflow() for example)? It can be implemented through built-ins in GCC/clang and custom implementation otherwise.

That will look ugly, but probably more readable than doing it "manually".

I think the main problem with this is creating macros/functions that are idiot-proof. They should not allow implicit conversion of the arguments (at least not truncation), or they'd actively make things worse. They should be portable, which is probably very hard. We could drop support for non-gcc/clang compilers to make use of certain extensions, but I guess at least MSVC people would complain.

So how to do it?

MrSmile:
> They should not allow implicit conversion of the arguments (at least not truncation), or they'd actively make things worse.

Pass arguments by pointers, for example? But I think we don't have that much of 64-bit integers to worry about truncation.

> They should be portable, which is probably very hard.

Most problems with portability come from arbitrary size integers. But even in that case it's possible to write overflow checks without any intrinsics. Another solution is to replace `int` with `int32_t` for bitmap functions (I think I already started that in some of previous PRs). In that case overflow check can be done trivially by using `int64_t` intermediates.

wm4:
Pushed it with the requested change (not checking w/h for < 1). Ensuring consistency of use of types, and overflow checking macros might be something for the future.

PR #371, opened 2020-02-08
render: match cases where VSFilter doesn't paint fill in the shadow

rcombs:
Some releases rely on this.
See corresponding VSFilter code: https://github.com/Cyberbeing/xy-VSFilter/blob/cf8f5b27de77fe649341bfab0fdfd498e1ad2fa6/src/subtitles/RTS.cpp#L1270

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1954,6 +1954,7 @@ static bool parse_events(ASS_Renderer *render_priv, ASS_Event *event)
         for (int i = 0; i < 4; i++) {
             uint32_t clr = render_priv->state.c[i];
             // VSFilter compatibility: apply fade only when it's positive
+            info->a_pre_fade[i] = _a(clr);
             if (render_priv->state.fade > 0)
                 change_alpha(&clr,
                              mult_alpha(_a(clr), render_priv->state.fade), 1.);
@@ -2285,10 +2286,21 @@ static void render_and_combine_glyphs(ASS_Renderer *render_priv,
                 flags |= FILTER_NONZERO_BORDER;
             if (info->shadow_x || info->shadow_y)
                 flags |= FILTER_NONZERO_SHADOW;
-            // VSFilter compatibility: invisible fill and no border?
-            // In this case no shadow is supposed to be rendered.
-            if (flags == FILTER_NONZERO_SHADOW && (info->c[0] & 0xFF) == 0xFF)
-                flags = 0;
+            if (flags & FILTER_NONZERO_SHADOW &&
+                (info->effect_type == EF_KARAOKE_KF ||
+                 info->effect_type == EF_KARAOKE_KO ||
+                 (info->a_pre_fade[0]) != 0xFF ||
+                 info->border_style == 3))
+                flags |= FILTER_FILL_IN_SHADOW;
+            if (!(flags & FILTER_NONZERO_BORDER) &&
+                !(flags & FILTER_FILL_IN_SHADOW))
+                flags &= ~FILTER_NONZERO_SHADOW;
+            if ((flags & FILTER_NONZERO_BORDER &&
+                 info->a_pre_fade[0] == 0 &&
+                 info->a_pre_fade[1] == 0 &&
+                 _a(info->c[2]) == 0) ||
+                info->border_style == 3)
+                flags |= FILTER_FILL_IN_BORDER;
 
             if (linebreak || is_new_bm_run(info, last_info)) {
                 linebreak = 0;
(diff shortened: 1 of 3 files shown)


libass/ass_render.c
```diff
@@ -2462,14 +2469,17 @@ size_t ass_composite_construct(void *key, void *value, void *priv)
         } else
             copy_bitmap(render_priv->engine, &v->bm_s, &v->bm);
 
+        if (!(flags & FILTER_FILL_IN_SHADOW) && !no_blur)
+            fix_outline(&v->bm, &v->bm_o);
```
astiob: ~~Should this say `bm_s` instead of `bm_o`?~~

Scratch that. Why is this block here at all? If `!no_blur`, nothing should be done.

Or, well, you could leave this here (with `bm_s`) and remove the “In this case no shadow is supposed to be rendered.” check on line 2246. That check happens to save us some work copying & subtracting bitmaps, so we can keep it, or—better yet—we can remove that check, remove this block, and add `if (flags & FILTER_FILL_IN_SHADOW)` to the `else` a couple lines above here. `v` is zero-initialized, so if the flags specify there’s no border width, no 0-width style 3 border, and no filled-in shadow, `v->bm_s.buffer` will remain `NULL` and there will be no shadow bitmap, just as desired.

MrSmile:
I've never quite understood logic behind that piece of code. Can anyone show how should look all possible permutations of relevant options in VSFilter? I've used [this](https://gist.github.com/MrSmile/c618b3dff4ff938a7e57e6b4ad36bfd1) file for tests, is that enough or there should be some additional cases?

astiob:
Your file looks fine. You can now add some karaoke to it, since this also depends on karaoke as we can see from this PR/the link above. But it should already show you the main problem—that we don’t (but VSFilter does) exclude fill from shadow when `\1a&HFF\bord8\shad8`.

samoht0:
Just a note on current PR code state. Causes regression with \be and \blur for me, easily notable with \be1 and \blur1 showing a much greater area blurred.

rcombs:
I had to look at d1feaf3684c14a24c707b6c50033d83014a8053b to try to understand what this code was doing in the first place. The new version of the commit makes this all a lot more verbose about what it's trying to do, rather than relying on funky bitwise properties (which this change makes no longer reliable). This seems to match VSFilter in all cases I've tested now:
```
[Script Info]
; Script generated by Aegisub 8959-rcombs-local-bae0e529a
; http://www.aegisub.org/
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
YCbCr Matrix: None
PlayResX: 1920
PlayResY: 1080

[Aegisub Project Garbage]
Last Style Storage: Default
Video File: ?dummy:24.000000:2700:1920:1080:255:254:254:c
Video AR Value: 1.777778
Video Zoom Percent: 1.250000
Active Line: 1

[V4+ Styles]
Style: Default,Gandhi Sans,51,&H00FFFFFF,&H000000FF,&H001FFF53,&H3AFF5317,-1,0,0,0,100,100,0,0,1,3.17333,20,2,200,200,43,1
Style: BS3,Gandhi Sans,51,&H00FFFFFF,&H000000FF,&H001FFF53,&H3AFF5317,-1,0,0,0,100,100,0,0,3,3.17333,20,8,200,200,43,1

[Events]
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\blur5}TEST LINE - BS1 - {\1aFF}1aFF{\1a} - {\1aFE}1aFE{\1a} - {\1a01}1a01{\1a} - {\1a80}1a80
Dialogue: 0,0:00:00.00,0:00:05.00,BS3,,0,0,0,,{\blur5}TEST LINE - BS3 - {\1aFF}1aFF{\1a} - {\1aFE}1aFE{\1a} - {\1a01}1a01{\1a} - {\1a80}1a80

```

rcombs:
Notably, this also fixes an issue where we would unnecessarily subtract the fill from the outline even when the fill is fully opaque. This resulted in aliasing on the edges, most visible when the two bitmaps have the same color, see e.g. here:
```
Dialogue: 0,0:00:05.00,0:00:07.00,Default,,0,0,0,,{\fs900\bord30\1c0\3c0\1a0\shad100}TEST
```

astiob (approved):
LGTM.

rcombs:
Pushed again to fix an issue where `\1aFE` was treated like `\1aFF` during a fade before it went full-opaque. Now matching the VSFilter, where `\1aFE` gets the non-`\1aFF` for the entire duration of lines with fades.

astiob:
Oh. Nice catch!

Well, LGTM, take two.

Issue #360, opened 2019-11-17
Subtitles without spaces do not break to a new line when overfull


svenper:
by way of mpv-player/mpv#7158 :

> ### Expected behavior
> 
> Subtitle lines without spaces (Chinese, Japanese, etc.) wrap automatically when too long.
> ### Actual behavior
> 
> Such lines extend outside the window, with characters cropped.

See e.g. at 43:00 in [sub.ass](https://github.com/mpv-player/mpv/files/3855514/sub.txt) with a large font size:

```text
Dialogue: 0,0:43:00.01,0:43:06.16,Default,,0,0,0,,任何一个有远大抱负的男人都不会因为区区一个女人
```

(Many lines in the subtitles do indeed have spaces between short phrases, which is an exeption to the rule in East Asian scripts.)

siikamiika:
Found this function while investigating this myself:
https://github.com/libass/libass/blob/6a670b485536e0398b3fc414fc4317a7b65b5622/libass/ass_render.c#L1588

I'd implement line breaking so that first priority would be for spaces as listed [here](http://jkorpela.fi/chars/spaces.html), and if the line would still be too long I would break before any CJK character that goes above `max_text_width`.

However, I see the problem that the function loops the text char by char instead of unicode characters. I'm not sure if it can be assumed to be UTF-8.

wm4:
It's definitely UTF-8, except when it's not. The latter could happen if an .ass file contains invalid UTF-8 sequences, but then libass doesn't need to render anything sane. But it shouldn't crash or so.
Due to the properties of UTF-8, looping by char is not a problem. If you need to skip codepoints that consist of multiple bytes, then I guess the current code structure is going to be very annoying.

wm4:
PS: if you haven't seen it yet, ``ass_utf8_get_char`` should be useful to compute the length of UTF-8 sequences (although there are faster methods if you don't need the codepoint itself).

Issue #348, opened 2019-06-27
Text measing API?


wm4:
Are there any thought about text measuring APIs? I'd like to have something that can retrieve the typographic measurements of some text. Any ideas whether this would be an acceptable thing, and how to expose it as API? What measurements should it even return? (Typography is hard.)

CoffeeFlux:
Proper text measuring APIs would definitely be useful. Currently SubInspector does [this](https://github.com/TypesettingTools/SubInspector/blob/master/src/SubInspector.c#L268) for bounds detection, but a more proper method would be welcome. Related: https://github.com/libass/libass/pull/329

wm4:
Yes, computing the bounding box is possible, but kind of sucks. It's slower than necessary, and not typographically correct.
I also assume it could skip full rendering of the measured text. It will often happen that the library user measures text, and then renders only a part of it, or so.

wm4:
I hoped for some discussion about possible APIs. I guess I'll just make a PR one day with an API that suits my needs.

MrSmile:
There are two different types of text measurement: logical (text advance + font height and depth) and graphical (bbox of actual vector outline). The first is useful for alignment, tables and such; the second for allocating bitmaps. The first is also much cheaper to compute (just dump Harfbuzz output) but second makes sense in case of transformed text.

So, what kind of measurement are you thinking of?

astiob:
Ideally we would be able to output both. Also, intercharacter spacing (both from the font and from `\fsp`) may be included or excluded on the last (right in left-to-right) edge.

wm4:
Thanks for giving a thought about it. I'm actually not sure myself what makes sense.

I want measurements for layouting and input feedback purposes. For example, I'd want to compute the "logical" bounding box of text to draw a frame around it. A bitmap bounding box could return a tiny bounding box for "." (instead of the logical line height), or an arbitrary box (higher than the text height) if the text exceeds the line bounds somehow (think of "zalgo"). And of course it doesn't tell you the baseline. At least at this point, using the ASS property that the font size equals the "line height" stops being sufficient for vertical alignment. Or maybe I'm overthinking, and returning baseline and line count is all what is missing?

Moreover, I don't just want bounding boxes of the entire text. I also want to:
- map coordinates back to a character/codepoint position (think of mouse selection of text)
- return grapheme cluster boundaries and order
- know the bounds of a certain character or text run (this could result in multiple rectangles if text is wrapped)
- have enough information to perform ellipsization
[...]

PR #372, opened 2020-02-15
Use the Unicode line breaking algorithm

siikamiika:
Demo in mpv:
wrap-cjk-demo

I'm opening this before adding any huge unicode range definitions to catch other issues early on. ~~I know that it isn't correct to break on the [Korean Unicode ranges](https://en.wikipedia.org/wiki/Korean_language_and_computers#Hangul_in_Unicode), for example.~~
Edit: it might be easier to treat all CJK the same way and [web browsers also do this](https://www.w3.org/TR/css-text-3/#line-breaking). Without any special handling for the Korean ranges the rules in [this post](https://nojeokhill.koreanconsulting.com/2013/05/korean-translation-tip-cardinal-rules-of-korean-language-layout.html) would be followed except when the words would otherwise overflow off screen because Korean text has spaces between words.

I'm planning to make the primary line break character [any space](http://jkorpela.fi/chars/spaces.html) (except the non-breaking ones maybe), secondary any punctuation and finally any Chinese or Japanese character.

Fixes https://github.com/libass/libass/issues/360

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -51,6 +51,8 @@ AC_ARG_ENABLE([asm], AS_HELP_STRING([--disable-asm],
     [disable compiling with ASM @<:@default=check@:>@]))
 AC_ARG_ENABLE([large-tiles], AS_HELP_STRING([--enable-large-tiles],
     [use larger tiles in the rasterizer (better performance, slightly worse quality) @<:@default=disabled@:>@]))
+AC_ARG_ENABLE([libunibreak], AS_HELP_STRING([--disable-libunibreak],
+    [disable libunibreak support @<:@default=check@:>@]))
 
 AS_IF([test x$enable_asm != xno], [
     AS_CASE([$host],
@@ -217,6 +219,14 @@ PKG_CHECK_MODULES([HARFBUZZ], harfbuzz >= 0.9.5, [
     ], [harfbuzz=false])
 fi
 
+if test x$enable_libunibreak != xno; then
+PKG_CHECK_MODULES([LIBUNIBREAK], libunibreak >= 1.1, [
+    CFLAGS="$CFLAGS $LIBUNIBREAK_CFLAGS"
+    AC_DEFINE(CONFIG_LIBUNIBREAK, 1, [found libunibreak via pkg-config])
+	libunibreak=true
+    ], [libunibreak=false])
+fi
+
 libpng=false
 if test x$enable_test = xyes; then
 PKG_CHECK_MODULES([LIBPNG], libpng >= 1.2.0, [
@@ -244,6 +254,9 @@ if test x$fontconfig = xtrue; then
 fi
 if test x$harfbuzz = xtrue; then
     pkg_requires="harfbuzz >= 0.9.5, ${pkg_requires}"
+fi
+if test x$libunibreak = xtrue; then
+    pkg_requires="libunibreak >= 1.1, ${pkg_requires}"
 fi
 
 if test x$enable_require_system_font_provider != xno &&
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -1044,62 +1044,19 @@ utf32_t text_info_get_next_char_utf32(
     return ch;
 }
 
-char *text_info_choose_ub_lang(TextInfo *text_info, char *track_lang)
+int is_valid_ub_lang(char *lang)
 {
-    // use track_lang if it's supported
-    if (track_lang) {
-        if      (strncmp(track_lang, "de", 2) == 0) return "de";
-        else if (strncmp(track_lang, "en", 2) == 0) return "en";
-        else if (strncmp(track_lang, "es", 2) == 0) return "es";
-        else if (strncmp(track_lang, "fr", 2) == 0) return "fr";
-        else if (strncmp(track_lang, "ja", 2) == 0) return "ja";
-        else if (strncmp(track_lang, "ko", 2) == 0) return "ko";
-        else if (strncmp(track_lang, "ru", 2) == 0) return "ru";
-        else if (strncmp(track_lang, "zh", 2) == 0) return "zh";
+    if (lang) {
+        if      (strncmp(lang, "de", 2) == 0) return 1;
+        else if (strncmp(lang, "en", 2) == 0) return 1;
+        else if (strncmp(lang, "es", 2) == 0) return 1;
+        else if (strncmp(lang, "fr", 2) == 0) return 1;
+        else if (strncmp(lang, "ja", 2) == 0) return 1;
+        else if (strncmp(lang, "ko", 2) == 0) return 1;
+        else if (strncmp(lang, "ru", 2) == 0) return 1;
+        else if (strncmp(lang, "zh", 2) == 0) return 1;
     }
-
-    char *lang = NULL;
-    GlyphInfo *cur;
-    unsigned symbol;
-
-    int kana_count = 0;
-    int cyrillic_count = 0;
-    int hangul_count = 0;
-    int cjk_count = 0;
-    for (int i = 0; i < text_info->length; ++i) {
-        cur = text_info->glyphs + i;
-        symbol = cur->symbol;
-        if (symbol < CYRILLIC_START) { // optimize
-            continue;
-        } else if (CJK_START <= symbol && CJK_END >= symbol) {
-            cjk_count += 1;
-        } else if (KANA_START <= symbol && KANA_END >= symbol) {
-            kana_count += 1;
-            break;
-        } else if (CYRILLIC_START <= symbol && CYRILLIC_END >= symbol) {
-            cyrillic_count += 1;
-            break;
-        } else if ((HANGUL_SYLLABLES_START  <= symbol &&  HANGUL_SYLLABLES_END  >= symbol) ||
-                   (HANGUL_JAMO_START       <= symbol &&  HANGUL_JAMO_END       >= symbol) ||
-                   (HANGUL_COMP_JAMO_START  <= symbol &&  HANGUL_COMP_JAMO_END  >= symbol) ||
-                   (HANGUL_EXTENDED_A_START <= symbol &&  HANGUL_EXTENDED_A_END >= symbol) ||
-                   (HANGUL_EXTENDED_B_START <= symbol &&  HANGUL_EXTENDED_B_END >= symbol)) {
-            hangul_count += 1;
-            break;
-        }
-    }
-
-    if (kana_count > 0) {
-        lang = "ja";
-    } else if (cyrillic_count > 0) {
-        lang = "ru";
-    } else if (hangul_count > 0) {
-        lang = "ko";
-    } else if (cjk_count > 0) {
-        lang = "zh";
-    }
-
-    return lang;
+    return 0;
 }
 
 // Return 1 if the event contains tags that will apply overrides the selective
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -1044,6 +1044,64 @@ utf32_t text_info_get_next_char_utf32(
     return ch;
 }
 
+char *text_info_choose_ub_lang(TextInfo *text_info, char *track_lang)
+{
+    // use track_lang if it's supported
+    if (track_lang) {
+        if      (strncmp(track_lang, "de", 2) == 0) return "de";
+        else if (strncmp(track_lang, "en", 2) == 0) return "en";
+        else if (strncmp(track_lang, "es", 2) == 0) return "es";
+        else if (strncmp(track_lang, "fr", 2) == 0) return "fr";
+        else if (strncmp(track_lang, "ja", 2) == 0) return "ja";
+        else if (strncmp(track_lang, "ko", 2) == 0) return "ko";
+        else if (strncmp(track_lang, "ru", 2) == 0) return "ru";
+        else if (strncmp(track_lang, "zh", 2) == 0) return "zh";
+    }
+
+    char *lang = NULL;
+    GlyphInfo *cur;
+    unsigned symbol;
+
+    int kana_count = 0;
+    int cyrillic_count = 0;
+    int hangul_count = 0;
+    int cjk_count = 0;
+    for (int i = 0; i < text_info->length; ++i) {
+        cur = text_info->glyphs + i;
+        symbol = cur->symbol;
+        if (symbol < CYRILLIC_START) { // optimize
+            continue;
+        } else if (CJK_START <= symbol && CJK_END >= symbol) {
+            cjk_count += 1;
+        } else if (KANA_START <= symbol && KANA_END >= symbol) {
+            kana_count += 1;
+            break;
+        } else if (CYRILLIC_START <= symbol && CYRILLIC_END >= symbol) {
+            cyrillic_count += 1;
+            break;
+        } else if ((HANGUL_SYLLABLES_START  <= symbol &&  HANGUL_SYLLABLES_END  >= symbol) ||
+                   (HANGUL_JAMO_START       <= symbol &&  HANGUL_JAMO_END       >= symbol) ||
+                   (HANGUL_COMP_JAMO_START  <= symbol &&  HANGUL_COMP_JAMO_END  >= symbol) ||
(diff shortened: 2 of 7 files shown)


rcombs:
Is this consistent with VSFilter behavior? If not, we should add a compatibility flag to trigger the new behavior (along with an API to toggle it, since players will probably want to enable it when converting from plaintext formats).

wm4:
CC @astiob

siikamiika:
@rcombs I'm afraid not:
mpc-hc64_YACEauNCFW

wm4:
We've always worked towards vsfilter compatibility, even in cases when it made necessary to disimprove libass behavior, that was objectively better, to that of vsfilter.

But I fully support making libass able to provide better behavior. Also, you could question whether compatibility behavior still makes sense, now that most ASS subtitles are probably rendered by libass (or not the original vsfilter at least), and even where vsfilter is used, there is a confusing number of differently behaving forks, as well as mpc-hc's internal renderer, which does everything differently anyway.

astiob:
I think one part of the problem is that libass tries to do two jobs at once: render ASS, and render everything else that media players/decoders convert to ASS on the fly but that actually isn’t.

In ASS, I can easily imagine someone fully relying on the non-breaking behaviour for some long line. Although, of course, ideally they should use `\q2` or whatever it is that explicitly disables automatic line wrapping, but you know how sloppy subtitle authors are. Even when automatic wrapping is enabled, VSFilter only breaks on U+0020 and not on any of the other spaces. Unfortunately, this does hurt CJK subtitles, but I’m fairly sure that (for once) all ASS renderers do this consistently and have always done so.

On the other hand, our line wrapping currently already differs from VSFilter’s in some cases. But we still decide based on spaces just like VSFilter does, so that’s much less of a difference than if we start to wrap lines that VSFilter doesn’t wrap at all. In fact, as I write this, I’m starting to recall we may have actually had files where our different wrapping messed up the rendering? Maybe it’s my imagination.

Meanwhile, in stuff like SRT, it does seem sensible to wrap CJK text etc. by default. Maybe if we ever get that sign-vs-subtitles detection/distinction (or decide that our current user-override-applicability detection is good enough), we could also apply it by default to ASS text but not ASS signs.

However, I’m not convinced that the simple algorithm of choosing any space, then any CJK, then any character is good:

1. For starters, yes, non-breaking spaces shouldn’t break.
2. CJK has its own line breaking rules that are more complicated than “break anywhere”. For example, you shouldn’t break between some punctuation and CJK.
3. Unicode has a [whole report](https://unicode.org/reports/tr14/) detailing how lines should be broken, and if we’re going to try to do it right, I’d prefer us to follow that instead of a random ad-hoc thing. But _maybe_ a random ad-hoc thing is better than nothing at all. Admittedly, TR14 doesn’t seem to make provision for breaking very long alphabetical words, so we’d need to add that as a last-resort lowest-priority breaking opportunity.
4. As it is, the code in this PR works on shaped grapheme clusters. This is probably generally good, as it avoids splitting combining marks off their base characters. But ideally, you might sometimes want to break a line in the middle of a cluster such as a ligature. But this is hopefully low-priority (although it would be required for TR14).
5. Breaking lines even between clusters can presumably break shaping, i. e. the nearby clusters may need to be reshaped. Current versions of HarfBuzz [provide _some_ help](https://github.com/harfbuzz/harfbuzz/issues/224) for doing this efficiently.

~~I wonder if there’s a ready implementation of TR14 that we could use.~~ Well, [here’s one](https://github.com/adah1972/libunibreak). I haven’t looked inside.

Here’s a few relevant emails from the HarfBuzz mailing list:
* [[1]](https://lists.freedesktop.org/archives/harfbuzz/2013-August/003532.html) (the HarfBuzz part is now implemented as linked above)
* [[2]](https://lists.freedesktop.org/archives/harfbuzz/2014-February/004136.html) (recommends using line breaking primitives implemented in ICU, as does the HarfBuzz website)
* [[3]](https://lists.freedesktop.org/archives/harfbuzz/2014-February/004141.html) (mentions the rare case of hyphenated Arabic which should effectively _not_ be reshaped along the break if I understand correctly)

siikamiika:
@astiob Thanks for the evaluation and sorry for the late reply.

I found another related issue https://github.com/libass/libass/issues/348 discussing an API that could potentially solve this issue and others related to overflowing text. I can see that this is not a trivial thing and my PR as it is now could definitely break some older subtitles that are designed to work with VSFilter or older libass. I'll try to add support for toggling this feature before moving forward with other things, but bear with me because I haven't done much C before.

As for the line breaking algorithm, I agree that my planned approach is far from perfect. It has been my annoyance for quite some time that I have to resize Japanese or Chinese subtitles in mpv when the lines are too long. I first thought of making a script that automatically adds line breaks to my subtitles, but then sat down and started digging into mpv source. After some time and help from @wm4 I had made this PR. It's just to solve my specific problem, and I only aimed to emulate what browsers or other reference implementations do "about right". Full standards compliance would be a nightmare to implement for a single project without reference implementations like https://github.com/adah1972/libunibreak.

I'll come back to this issue after I have studied the links you posted further and hopefully I can make it good enough to merge some day.

wm4:
#348 is just an idea to extend the libass API itself, so I can use it for other stuff.

I guess libunibreak demonstrates how hard the problem is to solve if you really want it to be generic. Personally I don't mind if we added a primitive version of line breaking, as implemented in this PR, as long as it doesn't fail fatally (like crashing).

But astiob has a good point that we can't just add it and potentially break existing scripts. Maybe it's time to add some sort of non-vsfilter compliant mode? Either per API, per new ASS header, or both.

rcombs:
My biggest problem with adding VSFilter-incompatibility headers is that whenever we do it, we're basically locked into the _new_ behavior forever (in addition to supporting the old), so it's essential to make sure that new behavior is exactly what we want it to be (in this case, presumably that means TR14) before giving script authors the ability to depend on it. An API flag makes a lot of sense, though, since applications could enable it whenever converting from SRT or other formats where exact layout consistency isn't expected, and we could safely provide a simple "better than it used to be" implementation like this as a stopgap.

siikamiika:
https://github.com/libass/libass/pull/372/commits/b8ad7806947142ec79b31bdacf2fc9c6b3862e8a should work as before except if you call `ass_set_wrap_cjk(renderer, 1);`

siikamiika:
I've modified the PR to use libunibreak. Seems to be an improvement from the previous solution because now it doesn't put the ideographic full stop to the next line. Let me know if there's something else to fix. I can squash the commits before merge but I'll keep them for now.

(seems like the build is failing, probably because of missing deps)

wrap-libunibreak-demo

libass/ass_render.c
```diff
@@ -1593,16 +1591,24 @@ wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
...
     double pen_shift_x;
     double pen_shift_y;
     int cur_line;
     TextInfo *text_info = &render_priv->text_info;
+    char brks[text_info->length];
+    char *brksPtr;
+
+    if (render_priv->settings.wrap_tr14) {
+        brksPtr = brks;
+        set_linebreaks(text_info, text_info->length, "ja", brksPtr,
```
astiob: Why the `brksPtr`? Why not just pass `brks`?

siikamiika: Fixed. I also noticed that `lang` is hard-coded to `"ja"`. I'll have to investigate if it's possible to use the subtitle language and fall back to something based on the code points in the string.

astiob: Good point. I think we already have a language field somewhere.

MrSmile:
Looks like [HarfBuzz already have support for line breaks](https://github.com/agl/harfbuzz/blob/master/src/harfbuzz-shaper.cpp#L37). I think it's better to use that instead of additional dependency.

rcombs:
Hmmm, as mentioned in that source file, harfbuzz has a few (somewhat arbitrary) differences from TR14; I'm not sure we want to codify those in the format? Are these even guaranteed not to change?

MrSmile:
Looks like that functionality got deleted from HarfBuzz anyway (my link is from some ancient version).

wm4:
I don't really know if we want to depend on another library...

siikamiika:
I wouldn't mind having 3 different versions of line breaking: 1 the original, 2 ifdeffed libunibreak to make it an optional dependency and 3 my original version with some custom improvements for CJK specifically. Maybe I could also find a more general solution than the current if/else.

astiob:
I wouldn’t mind another dependency since this doesn’t seem trivial, but we can make it optional. I don’t think there’s much value in a fallback non-TR14 CJK wrapping in case the dependency is omitted, and we should recommend it to be included in publicly distributed builds. It’s slightly concerning that libunibreak isn’t in Homebrew on macOS, but to my surprise it actually is in my Linux distribution’s (openSUSE) system repositories. It seems to be maintained and kept up-to-date with the latest Unicode, so it seems like a solid choice.

We may want to allow user style overrides to enable TR14 line wrapping.

We may want to give the API option a catch-all name that means “do your best to achieve proper text rendering rather than emulate VSFilter”, as a universal tool aimed specifically at pure-text subtitle formats that gives us carte blanche to add further improvements in the future. (Including stuff like not double-scaling opaque border boxes. Right off the bat, we can make it enable kerning by default.)

It’s fantastic that you could get libunibreak to work seemingly easily! However, I still have some worries about the correctness of the current approach. It seems it can potentially break within a cluster or at a wrong point if shaping recombines or reorders clusters… but I can’t come up with an example off the top of my head. (In fact, this applies to our existing line wrapper, too! But given that it only breaks on spaces, it’s presumably even less likely to encounter this situation in practice.)

The line rebalancing loop after `#define DIFF(x,y)` may need adjustment to use libunibreak’s computed breaks too.

astiob:
Interesting approach to guessing the language. I think our shaper code guesses language based on the first character’s script as reported by HarfBuzz, but I guess that’s done for each shaping run individually. Note that libunibreak [allows the language to be entirely unset](https://github.com/adah1972/libunibreak/issues/25#issuecomment-508993313). Maybe that’s a safer option when the track language isn’t set.

siikamiika:
Ok, I'll revert the part that guesses language based on the code points. I did it because I noticed that mpv didn't set `Language` in ASS script info when only the track has it.

astiob:
Yeah, it would be nice if stream language from the demuxer could be propagated to libass, both for converted subtitles and for ASS. Maybe we need a new API function to set a track’s language (as a fallback value if the Language header in the ASS is not present).

siikamiika:
> The line rebalancing loop after `#define DIFF(x,y)` may need adjustment to use libunibreak’s computed breaks too.

If I understand that correctly, it should be already done. The libunibreak block sets `break_at` to `i` and it causes `GlyphInfo.linebreak` to be set. I don't entirely understand what's done below `#define DIFF`, but the linebreaks are available there too.

astiob:
That loop rebalances lines. It tries to turn things like:
```
A quick brown fox jumped over the lazy
dog.
```
into things like:
```
A quick brown fox jumped
over the lazy dog.
```

As the function comment says, it does this by repeatedly moving words from one line to another. As you can see in the code, “words” in this case are delimited by spaces. It should probably use the break markers computed by libunibreak instead.

rcombs:
Does anyone know where libunibreak's language-specific data comes from? The global table seems to be from the standard https://www.unicode.org/Public/UNIDATA/LineBreak.txt file, but I don't see any sources for the lang-specific ones. See e.g. this change, which looks like it's simply down to the library author's opinions? https://github.com/adah1972/libunibreak/commit/79d9162c41f9177a02fa2175fdfe927891c21956#diff-9475b723a269299ae6de824e5c58613e
Unless we're sure of some fairly-stable fixed standard source for this data, we might want to just use the global tailoring in all cases.

siikamiika:
I think I've addressed everything exept:

- breaking when text overflows but there are no line break opportunities
- line break in the middle of a cluster (low priority?); haven't investigated this
- header in ASS file (not sure if needed)
- user styles for wrapping
- language specific tailorings
- refactoring

I'd like to limit the scope of the PR to adding the functionality behind an API flag if possible, and agreeing on how the line breaking should behave in ambiguous cases that TR14 doesn't specify.

I don't have a strong opinion about language specific tailorings in libunibreak. For CJK it could for example mean that Greek alphabet within ideographic text wouldn't be allowed to break. [This is actually specified in TR14](https://unicode.org/reports/tr14/#Properties) while the things in `linebreakdef.c` might not be, but all in all I think the differences will be minor and track language data is also scarce in practice.

I think it could be useful to have a priority for different types of characters such as whitespace, but after learning about line rebalancing it seems quite difficult to implement. This also concerns breaking long overflowing words without break opportunities. This is [out of scope for TR14](https://unicode.org/reports/tr14/#Introduction).

configure.ac
```diff
@@ -217,6 +219,14 @@ PKG_CHECK_MODULES([HARFBUZZ], harfbuzz >= 0.9.5, [
     ], [harfbuzz=false])
 fi
 
+if test x$enable_libunibreak != xno; then
+PKG_CHECK_MODULES([LIBUNIBREAK], libunibreak >= 1.1, [
+    CFLAGS="$CFLAGS $LIBUNIBREAK_CFLAGS"
+    AC_DEFINE(CONFIG_LIBUNIBREAK, 1, [found libunibreak via pkg-config])
```
xlz, line 225: Missing `LIBS="$LIBS $LIBUNIBREAK_LIBS"` here. Breaks ffmpeg runtime.

libass/ass_parse.h
```diff
@@ -19,6 +19,11 @@
 #ifndef LIBASS_PARSE_H
 #define LIBASS_PARSE_H
 
+#ifdef CONFIG_LIBUNIBREAK
+#include <linebreakdef.h>
+#include <linebreak.h>
```
xlz, line 24: This header order breaks build. GCC complains about undefined `utf32_t`.

adah1972:
I am the creator and maintainer of libunibreak. Yes, it is true that the language-specific tailoring is not from publicly available data, but my own understanding of the different languages’ punctuation rules (especially from [the Wikipedia page on quotation marks](https://en.wikipedia.org/wiki/Quotation_mark)). It is my *best-effort* changes to the language-neutral data.

I would recommend using language-specific data, when a language tag is available. There are two reasons:

- If my data were wrong, I would need feedback to fix them. This is the open-source spirit.
- Language-specific data allow much better breaking chances in certain cases.

Take the following sentence in Chinese:

> 他说：“别来。”

With language-neutral data, there are only two breaking chances in the middle:

> 他|
> 说：“别|
> 来。”

If we know it is Chinese, a very natural breaking position between `：` and `“` will then be available.

libass/ass_parse.c
```diff
@@ -1025,6 +1025,42 @@ unsigned get_next_char(ASS_Renderer *render_priv, char **str)
...
+        return EOS;
+
+    cur = text_info->glyphs + *ip;
+    ch = cur->symbol;
+    *ip += 1;
+
+    return ch;
+}
+
+int is_valid_ub_lang(char *lang)
```
TheOneric, line 1048: Is this actually neccessary ?
If libunibreak doesn't recognize the language string it seeems to default to NULL anyway (as you are manually doing right now).  
https://github.com/adah1972/libunibreak/blob/3bf54a439e0224e37d84ec92ec652f1609b10e49/src/linebreak.c#L377

Should libass change the language tag to accept BCP 47 language tags with variants, you might want to reject all language codes whose base code (without variants) is longer than 2 letters to avoid false negatives.  
See: https://github.com/libass/libass/issues/404#issuecomment-648306753

astiob:
Now that we have the `ASS_Feature` mechanism, it would be nice to see it used in this PR. See #409 for an example.

By default, we should keep VSFilter-compatible only-0020-SPACE-and-explicit-line-break line wrapping. When a new individual feature flag or the catch-all feature is enabled, we should switch to libunibreak.

Perhaps libunibreak can be an optional dependency [for now]. Just like in #409, I think the ideal arrangement is to make the feature flag enum constant available unconditionally and to signal the lack of libunibreak at runtime as an error in `ass_track_set_feature`.

TheOneric:
Superseded by #486

Issue #375, opened 2020-03-06
[RFC] New tags for "boxing" text


wm4:
I'm suggesting the following tag:
```
\box(x0, y0, x1, y1)
```
This shall be somewhat similar to `\pos(x0, y0)`, but the text is line-broken to not exceed the box width as set by `x1 - x0`. If adding another line would exceed the box height, the text is ellipsized (`Foobar` becomes `Fo...` to fit the width).

It might also be possible to correctly align text to the right for RTL text.

I want to use this for GUI purposes because mpv unfortunately uses libass as rendering API for GUIs. That would be nicer than clipping or guessing.

Thoughts? Not sure if I can/will implement this.

astiob:
It should be currently possible to achieve all of this except the ellipsization by creating a style with the appropriate margins.

I’m wary of ellipsization because it seems like a pretty language-specific thing. Of course, it’s also a thing that’s probably done a million times over in all kinds of software without any regard for language. Users may be used to it. They may not think well of it nevertheless. Heck, even in European languages I often don’t think well of it, although I’ve kinda gotten used to it over the years.

If you really want to avoid the text from overflowing the box, perhaps an alternative to ellipsization could be scrolling the text up and down repeatedly. In fact, doesn’t ASS already have a “banner” feature for this? But it can probably only scroll one way, not there and back again repeatedly.

wm4:
Margins have at least one problem: top and bottom margins are forced to the same value.

Instead of ellipsization you could do a "fade out" (static, not animated). Firefox and Chrome do this for tabs now, I think. Maybe it would be simpler to implement too.

Animating it with scrolling would be weird, I think.

astiob:
> Margins have at least one problem: top and bottom margins are forced to the same value.

Yes, but _if_ you forget about ellipsization/overflow, then IIRC this doesn’t actually matter. Depending on the alignment your text will be positioned at the top or bottom margin and flow freely towards the other end of the screen without regard for the opposite margin, right? Or if it’s explicitly positioned, the vertical margins won’t matter at all.

Also, we could implement support for [ASSv2 syntax, which allows separate top and bottom margins](https://github.com/Cyberbeing/xy-VSFilter/blob/3.1.0.752/src/subtitles/STS.cpp#L1462).

wm4:
Well, I want to cut off the text. One could use \clip, but that won't look very nice. The whole purpose is to constrain text to a specific rectangle, even if the text is not statically known (such as user input or so).

[...]

PR #376, opened 2020-03-07
[RFC] \box tag

wm4:
This essentially sets the margins. But it allows you to set different
margins for top and bottom (instead of only a "MarginV"), and takes a
rectangle instead of right/bottom margin values. Rather similar to for
example \clip.

Example: \box(500,50,700,100)

For now, either the top or bottom margin is ignored, depending on
alignment. A future idea might be to cut off the text in a non-ugly way,
such as ellipsization. (This would be controlled by other tags.)

This tag should be a nice addition in contexts where setting the the
event or style margins is too bothersome.

### DIFF
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -741,6 +741,16 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
                 render_priv->state.be = val;
             } else
                 render_priv->state.be = 0;
+        } else if (complex_tag("box")) {
+            if (nargs == 4) {
+                render_priv->state.margin_x0 = argtoi(args[0]);
+                render_priv->state.margin_y0 = argtoi(args[1]);
+                render_priv->state.margin_x1 =
+                    render_priv->track->PlayResX - argtoi(args[2]);
+                render_priv->state.margin_y1 =
+                    render_priv->track->PlayResY - argtoi(args[3]);
+                render_priv->state.margin_set = 1;
+            }
         } else if (tag("b")) {
             int val = argtoi(*args);
             if (!nargs || !(val == 0 || val == 1 || val >= 100))
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1057,6 +1057,11 @@ init_render_context(ASS_Renderer *render_priv, ASS_Event *event)
     render_priv->state.effect_type = EF_NONE;
     render_priv->state.effect_timing = 0;
     render_priv->state.effect_skip_timing = 0;
+    render_priv->state.margin_x0 = 0;
+    render_priv->state.margin_y0 = 0;
+    render_priv->state.margin_x1 = 0;
+    render_priv->state.margin_y1 = 0;
+    render_priv->state.margin_set = 0;
 
     apply_transition_effects(render_priv, event);
     render_priv->state.explicit = render_priv->state.evt_type != EVENT_NORMAL ||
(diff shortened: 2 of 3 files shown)


wm4:
CC @astiob @MrSmile @rcombs 

Not sure if this has any value. It would for me because 1. possible text cut off thing, and 2. mpv's OSD code doesn't allow setting the event margins (lol).

astiob:
To summarize our discussion on IRC:

* As it is, this currently does nothing (useful) that can’t already be done by setting the Margin fields on the Dialogue line.
* wm4 hopes to extend this to add automatic fade-out or ellipsization at the opposite-from-anchor vertical margin (which is currently ignored), as described in #375.
* wm4 also wants this because mpv’s internal OSD APIs currently have no mechanism to set the Margin fields, and this seems easier and potentially more useful than adding such mechanism.
* I’m wary of adding new override tags that can be used from actual ASS files/tracks to avoid breaking other renderers and further fragmenting the already fragile world of ASS rendering.
* I suggested, and wm4 agreed, that this new override tag could be available only in a special new rendering mode such as introduced in #382 or #372, which would be disabled for real ASS tracks but enabled for mpv’s OSD.

rcombs:
As mentioned over in #375, I'd prefer to call this `\margin` (since that's what it'd do). Seems sensible to gate it behind #382 as well.

astiob:
IIRC this was mentioned on IRC too but I forget what the conclusion was: if you call it `\margin`, that sounds like the fade-out/ellipsization effect should be a separate tag, no? Mind, maybe it should. Could be a Boolean tag like `\i`.

rcombs:
Yeah, I think that makes sense as a separate thing (in part because imo setting per-line margins is a useful feature independent of that).

wm4:
> IIRC this was mentioned on IRC too but I forget what the conclusion was: if you call it \margin, that sounds like the fade-out/ellipsization effect should be a separate tag, no? Mind, maybe it should. Could be a Boolean tag like \i.

Yes, making a separate flag was always the idea.

libass/ass_parse.c
```diff
@@ -741,6 +741,16 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
...
             } else
                 render_priv->state.be = 0;
+        } else if (complex_tag("box")) {
+            if (nargs == 4) {
+                render_priv->state.margin_x0 = argtoi(args[0]);
+                render_priv->state.margin_y0 = argtoi(args[1]);
+                render_priv->state.margin_x1 =
+                    render_priv->track->PlayResX - argtoi(args[2]);
+                render_priv->state.margin_y1 =
+                    render_priv->track->PlayResY - argtoi(args[3]);
```
astiob: Does this work correctly when libass is told to draw subtitles in black bars around the video?

astiob:
Another thing to keep in mind, although that’s not the focus of _this_ PR, is that our line wrapping currently (and after #372, too) keeps long words that don’t contain any break opportunities unbroken and lets them overflow the margin box. With a fade-out/ellipsization thing on, this will probably be undesirable.

wm4:
Good point about collisions. I think anything with `\margin` should be treated as explicitly positioned; however, then it'd be significantly different from just setting margins in the event header.

rcombs:
>since the event margins are right next to the event text in all source formats that we support

Wait, what? Aren't they on the style?

wm4:
Both have them.

PR #382, opened 2020-03-19
Ass compliance field

wm4:
This is a much discussed thing, that is supposed to enable us to disable
weird VSFilter compatibility hacks. Currently it obviously does nothing.

If we ever extend the libass format, the plan is that we may add a new
header to the ASS [Script Info] section, which causes the compliance
field set to a new (to be added) enum value.

In theory, we could reuse the track_type field for this purpose, but
that seems questionable. Better not touch anything related to SSA
compatibility.

### DIFF
--- a/libass/Makefile.am
+++ b/libass/Makefile.am
@@ -5,7 +5,7 @@ AM_CFLAGS = -std=gnu99 -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter \
 
 LIBASS_LT_CURRENT = 9
 LIBASS_LT_REVISION = 2
-LIBASS_LT_AGE = 0
+LIBASS_LT_AGE = 1
 
 nasm_verbose = $(nasm_verbose_$(V))
 nasm_verbose_ = $(nasm_verbose_$(AM_DEFAULT_VERBOSITY))
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -58,6 +58,8 @@ struct parser_priv {
     uint32_t *read_order_bitmap;
     int read_order_elems; // size in uint32_t units of read_order_bitmap
     int check_readorder;
+
+    int enable_extensions;
 };
 
 #define ASS_STYLES_ALLOC 20
@@ -1325,6 +1327,17 @@ ASS_Track *ass_new_track(ASS_Library *library)
     }
     track->parser_priv->check_readorder = 1;
     return track;
+}
+
+int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable)
+{
+    switch (feature) {
+    case ASS_FEATURE_INCOMPATIBLE_EXTENSIONS:
+        track->parser_priv->enable_extensions = !!enable;
+        return 0;
+    default:
+        return -1;
+    }
 }
 
 /**
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -24,7 +24,7 @@
 #include <stdarg.h>
 #include "ass_types.h"
 
-#define LIBASS_VERSION 0x01400000
+#define LIBASS_VERSION 0x01400001
 
 #ifdef __cplusplus
 extern "C" {
@@ -198,6 +198,19 @@ typedef enum {
     ASS_FONTPROVIDER_FONTCONFIG,
     ASS_FONTPROVIDER_DIRECTWRITE,
 } ASS_DefaultFontProvider;
+
+typedef enum {
+    /**
+     * Enable libass extensions that would display ASS subtitles incorrectly.
+     * These may be useful for applications, which use libass as renderer for
+     * subtitles converted from another format, or which use libass for other
+     * purposes that do not involve actual ASS subtitles authored for
+     * distribution.
+     */
+    ASS_FEATURE_INCOMPATIBLE_EXTENSIONS,
+
+    // New enum values can be added here in new ABI-compatible library releases.
+} ASS_Feature;
 
 /**
  * \brief Initialize the library.
@@ -511,6 +524,23 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
  * \return pointer to empty track
  */
 ASS_Track *ass_new_track(ASS_Library *);
+
+/**
+ * \brief Enable or disable certain features
+ * This manages flags that control the behavior of the renderer and how certain
+ * tags etc. within the track are interpreted. The defaults on a newly created
+ * ASS_Track are such that rendering is compatible with traditional renderers
+ * like VSFilter, and/or old versions of libass. Calling ass_process_data() or
+ * ass_process_codec_private() may change some of these flags according to file
+ * headers. (ass_process_chunk() will not change any of the flags.)
+ * Additions to ASS_Feature are backward compatible to old libass releases (ABI
+ * compatibility).
+ * \param track track
+ * \param feature the specific feature to enable or disable
+ * \param enable 0 for disable, any non-0 value for enable
+ * \return 0 if feature set, -1 if feature is unknown
+ */
+int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable);
 
 /**
  * \brief Deallocate track and all its child objects (styles and events).
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -56,6 +56,7 @@ typedef struct ass_image {
         IMAGE_TYPE_SHADOW
     } type;
 
+    // New fields can be added here in new ABI-compatible library releases.
 } ASS_Image;
 
 /*
@@ -170,6 +171,7 @@ typedef enum {
      * On dialogue events override: Justify
      */
     ASS_OVERRIDE_BIT_JUSTIFY = 1 << 10,
+    // New enum values can be added here in new ABI-compatible library releases.
 } ASS_OverrideBits;
 
 /**
--- a/libass/ass_types.h
+++ b/libass/ass_types.h
@@ -166,6 +166,7 @@ typedef enum ASS_YCbCrMatrix {
     YCBCR_SMPTE240M_PC,
     YCBCR_FCC_TV,
     YCBCR_FCC_PC
+    // New enum values can be added here in new ABI-compatible library releases.
 } ASS_YCbCrMatrix;
 
 /*
@@ -205,6 +206,8 @@ typedef struct ass_track {
 
     ASS_Library *library;
     ASS_ParserPriv *parser_priv;
+
+    // New fields can be added here in new ABI-compatible library releases.
 } ASS_Track;
 
 #endif /* LIBASS_TYPES_H */
--- a/libass/libass.sym
+++ b/libass/libass.sym
@@ -43,3 +43,4 @@ ass_set_pixel_aspect
 ass_set_selective_style_override_enabled
 ass_set_selective_style_override
 ass_set_check_readorder
+ass_track_set_feature


astiob:
The code itself looks fine. I have various concerns, but if you feel none of them are important and you want to merge as-is, feel free. Sorry for the text wall (as always).

How is this conceptually different from `ass_set_improve_rendering` introduced in #372? I suppose making this track-specific rather than one-per-renderer is good, but the concept is the same, right?

The name and especially the description make it sound like this is intended for experiments only and we promise absolutely nothing about the resulting behaviour. Is this useful? In my imagination, media players should set the New Setting for all tracks that are not originally ASS, namely for all tracks converted from other subtitle formats and for all player UI. But to do this, they should have reasonable belief that we will render their tracks in a reasonable way, rather than in an unspecified experimental way. The comment that says “you should never set this…”, while technically correct because it only talks about actual ASS, instils fear as well. I’m thinking perhaps the comment, and possibly the name of the value or even the whole setting, should provide more positive encouragement and suggestions when this setting _should_ be used, rather than only when it _shouldn’t_.

Moreover, I’m thinking that we could make some of our existing extensions available _only_ with the new setting, like BorderStyle 4 and base text direction autodetection on Encoding −1. We may want to allow them in user style overrides regardless of the track’s compliance setting but emulate VSFilter (i. e. fall back on defaults or whatever) when we encounter them while reading from file.

You asked me on IRC whether I preferred a field in `ASS_Track` or a setter function. I don’t have a strong opinion either way, especially after being reminded in this PR that adding a field doesn’t break ABI if we make it clear that `sizeof(ASS_Track)` isn’t fixed (nice job on that!). I’m _slightly_ inclined to say setter method because, of course, it’s still possible for someone to negligently or accidentally rely on `sizeof(ASS_Track)`, so a private field and a setter method would still be the safest choice, and I’m not sure how much of a good practice it is to have public structs at all; but we’re already stuck with public structs in any case and it seems fairly unlikely/silly that anyone should store `ASS_Track` when our API works with `ASS_Track *` only and provides our own allocation functions for it.

On the other hand, if/when we do make some current extensions available _only_ with the new setting, I’m thinking we may want to _intentionally_ break ABI _and_ API compatibility. This would force developers of all clients—mpv, VLC, Kodi, FFmpeg, Libav? etc.—to stop and think which mode they actually want in each instance. For real SSA/ASS, pick the ASS/VSFilter mode. For everything else, pick the new mode. I assume this only _really_ matters for mpv, which is probably the only client that intentionally uses our extensions. But it would be a bit silly if mpv suddenly stopped displaying the right kind of borders after a seemingly innocuous ABI-compatible libass update or even after a recompile (if we were to break ABI but keep API compatibility). Then again, does mpv actually use any extensions by itself? Or did we only introduce them for use in user overrides?

The new setting could be a Boolean like in #372, but I suppose an enum will give us more freedom if we get any more crazy ideas in the future, so that’s good. (Original _Sub Station Alpha_ compatibility?… don’t suppose anyone wants that? VSFilterMod compatibility might be interesting. Crazy ideas are always abundant.)

wm4:
> How is this conceptually different from ass_set_improve_rendering introduced in #372? I suppose making this track-specific rather than one-per-renderer is good, but the concept is the same, right?

We discussed it vaguely, I wasn't aware of any concrete attempts to add such a thing and don't find any mention of this specific function name.

> The name and especially the description make it sound like this is intended for experiments only and we promise absolutely nothing about the resulting behaviour. Is this useful? In my imagination, media players should set the New Setting for all tracks that are not originally ASS, namely for all tracks converted from other subtitle formats and for all player UI. But to do this, they should have reasonable belief that we will render their tracks in a reasonable way, rather than in an unspecified experimental way. 

Do you know any software which uses libass for converted subtitles? I only know of mpv and ffmpeg. (A player might in theory use libavcodec's subtitle->ASS converter, but this still requires effort and is far from natural.)

Anyway, the enum is extensible (forgot to add an ABI comment I guess), and if we ever decide that we should add a mode for "converted" subtitles, we can add that.

For now, the ASS_COMPLIANCE_NONE mode is intentionally worded in a "scary" way to prevent API users from setting it frivolously. (An API user might for example think "of course I want all these shiny libass extensions", when it reality it might break subtitles authored for vsfilter.)

The main reason to introduce this mode is to introduce features without exposing them to normal subtitle authors, which would lead to files using these features getting into the wild. (Although personally I don't see much of a problem with this, but we're having endless discussions for years about this stuff without ever making any progress, so that's my solution for this.)

> I’m thinking perhaps the comment, and possibly the name of the value or even the whole setting, should provide more positive encouragement and suggestions when this setting should be used, rather than only when it shouldn’t.

Again, it's not the "new" mode you thought it is, but it provides a place to add it.

Regarding current libass exclusive features, these are already "in the wild".

Feel free to suggest any solution that doesn't require me to add a modified copy of libass to mpv's source tree.

astiob:
Wow, why so combative?

You’re suggesting a new feature that you’re clearly going to use in mpv. mpv isn’t a magical oracle toy; it needs to assume that it will get something out of using the new feature just as any other client would. So does FFmpeg, which is clearly able to convert things to ASS as you mention yourself. I don’t see how our goals differ. The wording you have chosen suggests that you’re explicitly allowing libass to break mpv in the future as it sees fit.

> Again, it's not the "new" mode you thought it is

So far I fail to see how it isn’t.

> We discussed it vaguely, I wasn't aware of any concrete attempts to add such a thing and don't find any mention of this specific function name.

It’s in https://github.com/libass/libass/pull/372/commits/43c5814a6038a7cd46e823885700fc5930f15047, in response to https://github.com/libass/libass/pull/372#issuecomment-587956643.

rcombs:
Hmmm, maybe instead of "no compliance" we could call the fast-and-loose value "experimental"? It's a good place to put new features as they're being worked on, without providing the ability for subtitle authors to depend on it.

It might also make sense to make it a bitmask, with `EXPERIMENTAL` having the value `0xFFFFFFFF`. We could internally define bits for individual features, mpv would be able to use them immediately, and later on if we decided we wanted to declare a feature "beta" (i.e., it will continue to be available indefinitely, but might not render in exactly the same way) we could add a public `#define` for it, then if we wanted to declare it "stable", we could add a `Features: [X], [Y], [Z]` header.

We could also define an `ass_set_track_converted()` function that enables a set of feature flags that we decide make sense for content sourced from SRT and the like, such as `{}` escaping (which lavc already does even though nothing else supports it, iirc).

astiob:
If you’re suggesting that we may want to separate the mode for converted subtitles from the mode for mpv’s UI, you might be right, although I don’t immediately see any differences—but perhaps some differences might appear in the future.

In that case, you’re suggesting that the `ASS_COMPLIANCE_NONE` you’re adding is for mpv’s UI? I’m not at all opposed to leaving the enum open for extension (that’s why it’s an enum rather than a Boolean), but the classification seems odd to me. For one thing, converted subtitles wouldn’t comply to anything either. The best alternative I’ve come up so far is “purpose”, as in “configure rendering for the purpose of VSFilter compatibility”. Maybe we can keep “compliance” with a slightly overloaded meaning, but before we commit to that, can we spare a minute to come up with a remotely good suggestion for how a mode for converted subtitles might be called, if we’re thinking this is the place for it?

If you want to keep this mode effectively exclusive to mpv, your phrasing may be fine, with tacit understanding that libass won’t break this mode without coordination with mpv (at least unless we somehow ever find ourselves in the unlikely situation where mpv is dead and disused while libass is alive).

> such as {} escaping (which lavc already does even though nothing else supports it, iirc).

Right, that was another thing I wanted to disable in VSFilter compliance mode.

> It might also make sense to make it a bitmask

This may be a good idea.

On the other hand, are we sure we won’t run out of bits? 😅

> Feel free to suggest any solution that doesn't require me to add a modified copy of libass to mpv's source tree.

Sorry if I mistook desperation for combativeness. At some point, one might indeed wonder whether using (vanilla) libass for mpv’s UI is a long-term viable solution.

wm4:
> Wow, why so combative?

I think it's a reasonable idea. I wouldn't use the internal libass for subtitles, and instead slowly refactor it to some sort of text renderer. There'd be no concern for compatibility, clean design (at first), or subtitle focus at all, so it would still be faster than doing it internally in libass. Needed features could be implemented immediately.

> You’re suggesting a new feature that you’re clearly going to use in mpv. mpv isn’t a magical oracle toy; it needs to assume that it will get something out of using the new feature just as any other client would.

Not necessarily. For the main part, it'd mean there's no per-pixel compatibility. Consider the line break change: there are concerns that it might break existing subtitles, but in the "none" compliance mode, the exact way how line breaking is done would not matter much.

> So does FFmpeg, which is clearly able to convert things to ASS as you mention yourself. I don’t see how our goals differ. The wording you have chosen suggests that you’re explicitly allowing libass to break mpv in the future as it sees fit.

FFmpeg assumes standard ASS, with the already mentioned exception of being able to escape ``{``.

> It might also make sense to make it a bitmask, with EXPERIMENTAL having the value 0xFFFFFFFF. We could internally define bits for individual features, 

There's already ``ASS_OverrideBits``, which works quite similarly. Could also be used for the line breaks PR.

On that note, I agree that it should probably not be a bit mask, since there's the danger we might run out of them.

wm4:
PS: in any case, decide! Here's some brainstorming about possible implementations:
1. Use the approach as implemented in this PR
2. Just use ASS_OverrideBits for this
3. Add a new boolean setter function for each new case (like in the linebreak PR currently)
4. A setter called ``ass_set_feature(ASS_Renderer *, ASS_Feature, bool)`` (where ``ASS_Feature`` is an enum, and you add a new entry for each new case)
5. A new public field in ``ASS_Track`` for each new case

I don't mind which one it is, as long as we get results.

astiob:
> I think it's a reasonable idea. I wouldn't use the internal libass for subtitles, and instead slowly refactor it to some sort of text renderer. There'd be no concern for compatibility, clean design (at first), or subtitle focus at all, so it would still be faster than doing it internally in libass. Needed features could be implemented immediately.

So I did misjudge the emotion there. To be very honest, that does sound reasonable. But, of course, that would be a lot of extra effort for you, duplication of quite a bit of code even if far from all of libass, etc. So I appreciate that you don’t want to do this. If this is ever considered _seriously,_ we might try to factor out some kind of common rendering backend, but it seems daunting and even more so with MrSmile’s long-planned tiled rendering thing.

What worries me about the “experimental” phrasing, both as-is in this PR and in rcombs’ suggestion, is that it implies no suitability to purpose and stability at all. For example, let’s say we put `\box` behind the experimental mode flag now. mpv immediately starts using it. But it’s “experimental”, so we’re free to remove it at any time? or make it work significantly differently? I know mpv and libass share a bit of a special relationship, but even so I find it strange that _stable_ versions of mpv should want to use an _explicitly completely unstable_ libass mode. By all means, I want us to have leeway to improve things like line breaking, advanced text rendering, font handling. But at the same time, shouldn’t we at least promise that the result should be suitable for… something? that the changes we’re free to make will be aimed only at improving rendering in sensible ways and adding further extensions? No? Maybe it’s just me. If you’re happy with it as-is, so be it.

> Consider the line break change: there are concerns that it might break existing subtitles, but in the "none" compliance mode, the exact way how line breaking is done would not matter much.

And that’s exactly how I imagine converted subtitles working, so I’m still having trouble understanding why you said this is not the mode I want. Maybe this should be treated as a separate issue, in which case feel free to disregard this; but I find it’s related to the naming of the noncompliant mode.

For converted subtitles, I want a “make it like ASS but (with) sensible (text rendering)” mode. And mpv may be the only player that actually uses libass for converted subtitles at the moment—I’m not sure if it actually is, I don’t know—but it seems prudent to keep this option available for other players if they should want to do the same. If nothing else, a DirectShow filter wrapping libass as a drop-in replacement for VSFilter may well want to use this mode (although if it wants to actually emulate VSFilter, it will want the VSFilter mode anyway; but that’s a big if).

For mpv’s UI, you want a “make it like ASS but with extensions” mode. Right?

I assume mpv’s UI wouldn’t mind more sensible text rendering, either. Like, if you’re displaying the current file name and it happens to be Arabic, I think you might like it to be right-to-left automatically (although you can already achieve this using the libass extension “Encoding −1”, and you may well already be doing that). And if you should insert override tags inside it, you would probably like it to stay right-to-left (this is our current behaviour, but it keeps coming up in issues reported by Arabic users and we really want to emulate VSFilter’s bidi here, at which point this will be relegated to the noncompliant mode). And if it should be a very long name and you decide to let it wrap to the next line, you might like it to do so sensibly (per #372). And if we should ever implement GDI’s always-Arial-if-possible font fallback in VSFilter compliance mode, you might like to keep our current font fallback mechanism instead.

So it seems to me that both modes could be one and the same.

Or they could be kept separate, and e. g. the converted subtitle mode could have sensible text rendering and `\{` but no nonessential extensions like `\box` while mpv UI mode could be a superset and have that _plus_ additional extensions. I’m not sure what the immediate gain from separating the two modes is, but it may become useful in the future. For converted subtitle mode, we’d promise to read basic ASS and honour user overrides, and reserve the right to tweak the exact rendering to optimize the display of plain text. For mpv UI mode, we’d promise to read full ASS and additional extensions and to tweak the extensions and basic rendering to suit mpv’s UI needs. Is this similar to what you had in mind?

Also, if we go with the `ASS_Track` field, the ABI version in the Makefile should be updated to:
```
LIBASS_LT_CURRENT = 10
LIBASS_LT_REVISION = 0
LIBASS_LT_AGE = 1
```
as old clients work with the new libass (assuming they don’t use `sizeof(ASS_Track)`) but new mpv compiled against the new libass won’t work with old libass at runtime.

wm4:
> But, of course, that would be a lot of extra effort for you, duplication of quite a bit of code even if far from all of libass, etc. So I appreciate that you don’t want to do this. If this is ever considered seriously, we might try to factor out some kind of common rendering backend, but it seems daunting and even more so with MrSmile’s long-planned tiled rendering thing.

Indeed. The alternative would be using something like cairo + pango (whatever you pick, it's either pretty bloated, or too insufficient, especially because of text rendering and font handling). libass works surprisingly well for this purpose, all while it has a low footprint. Of course, it wouldn't be too fun to maintain such code, especially because of all the rasterizer innards.

> What worries me about the “experimental” phrasing, both as-is in this PR and in rcombs’ suggestion, is that it implies no suitability to purpose and stability at all. For example, let’s say we put \box behind the experimental mode flag now. mpv immediately starts using it. But it’s “experimental”, so we’re free to remove it at any time? or make it work significantly differently? I know mpv and libass share a bit of a special relationship, but even so I find it strange that stable versions of mpv should want to use an explicitly completely unstable libass mode.

That is true, and it should be clarified. How about this:
- Pixel perfect rendering is not guaranteed, instead there is a vague guarantee for best effort. (Unlike true ASS subtitles, where script authors may rely on corner cases to get exactly the result they intended to see.)
- Features can be changed or removed after a deprecation period. They are like API changes, though not as strong as ABI, which needs to be more stable. (Unlike true subtitles, for which you need to support a feature forever.)

Regarding the topic of converted subtitles: that is a bit messy. srt subtitles actually tend to contain ASS tags (because VSFilter). So while handling them is usually different from real ASS subtitles, there is an annoying overlap.

wm4:
I changed to to the set_features idea. No more compliance anything.

PR #384, opened 2020-03-26
Fix margin bugs and confusion

wm4:
This affects various cases. The first commit fixes https://github.com/mpv-player/mpv/issues/7557.

The second commit fixes problems that become visible especially when zooming and panning the video. This tends to trigger all sorts of corner cases. Unfortunately, the change is incompatible, which is why I had to gate it behind a flag (NB: would prefer using ASS_Feature for this). A mpv branch which uses this is at: https://github.com/mpv-player/mpv/compare/ass_shit

### DIFF
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -170,6 +170,12 @@ typedef enum {
      * On dialogue events override: Justify
      */
     ASS_OVERRIDE_BIT_JUSTIFY = 1 << 10,
+    /**
+     * Change behavior if ass_set_margins() with non-0 values is called: do not
+     * attempt to apply scaling to normal dialogue text based on the margins,
+     * and use the screen size instead.
+     */
+    ASS_OVERRIDE_BIT_SANE_MARGINS = 1 << 11,
 } ASS_OverrideBits;
 
 /**
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -188,13 +188,27 @@ static ASS_Image *my_draw_bitmap(unsigned char *bitmap, int bitmap_w,
 /**
  * \brief Mapping between script and screen coordinates
  */
+
+static double x2scr_sane(ASS_Renderer *render_priv, double x, double scale)
+{
+    if (!render_priv->state.explicit && render_priv->settings.use_margins) {
+        return x * render_priv->settings.frame_width /
+               scale / render_priv->track->PlayResX;
+    }
+    return x * render_priv->orig_width / scale / render_priv->track->PlayResX +
+        render_priv->settings.left_margin;
+}
 static double x2scr_pos(ASS_Renderer *render_priv, double x)
 {
+    if (render_priv->settings.selective_style_overrides & ASS_OVERRIDE_BIT_SANE_MARGINS)
+        return x2scr_sane(render_priv, x, render_priv->font_scale_x);
     return x * render_priv->orig_width / render_priv->font_scale_x / render_priv->track->PlayResX +
         render_priv->settings.left_margin;
 }
 static double x2scr(ASS_Renderer *render_priv, double x)
 {
+    if (render_priv->settings.selective_style_overrides & ASS_OVERRIDE_BIT_SANE_MARGINS)
+        return x2scr_sane(render_priv, x, render_priv->font_scale_x);
     if (render_priv->state.explicit)
         return x2scr_pos(render_priv, x);
     return x * render_priv->orig_width_nocrop / render_priv->font_scale_x /
@@ -203,11 +217,15 @@ static double x2scr(ASS_Renderer *render_priv, double x)
 }
 static double x2scr_pos_scaled(ASS_Renderer *render_priv, double x)
 {
+    if (render_priv->settings.selective_style_overrides & ASS_OVERRIDE_BIT_SANE_MARGINS)
+        return x2scr_sane(render_priv, x, 1);
     return x * render_priv->orig_width / render_priv->track->PlayResX +
         render_priv->settings.left_margin;
 }
 static double x2scr_scaled(ASS_Renderer *render_priv, double x)
 {
+    if (render_priv->settings.selective_style_overrides & ASS_OVERRIDE_BIT_SANE_MARGINS)
+        return x2scr_sane(render_priv, x, 1);
     if (render_priv->state.explicit)
         return x2scr_pos_scaled(render_priv, x);
     return x * render_priv->orig_width_nocrop /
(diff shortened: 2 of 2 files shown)


astiob:
I don’t see how the new behaviour is saner or more desirable. Maybe I’m misunderstanding something because I’m not an advanced user of margins myself. I tried to read some of the linked mpv issues but only got confused.

When (good) subtitle authors create their subtitles, they generally carefully choose such phrasing and event splitting that ensures there’s never more than one or two lines of subtitles on screen at a time. This is desirable from a readability perspective. When they’re doing this, they’re looking at the video without any margins. (Although currently the scaling is the same even with margins, in both libass and XySubFilter.)

If I understand correctly, what you’re proposing here makes font sizes and line sizes _bigger_ when the user switches to full screen or otherwise adds margins without (significantly) changing video size. The bigger the screen relative to the video, and therefore the bigger the margins, the bigger the text will become. This will be visually jarring when switching to full-screen, and this will make the carefully worded & split subtitles spill over to additional lines. They will also simply waste screen estate, which seems to be exactly the opposite of what margins are supposed to do. If the user is able to change the media player window’s height without changing the width (assuming widescreen video with only top & bottom margins), the text will change size as they do this. Perhaps someone does want this—it seems mpv has an option for this, which is fine—but this seems undesirable to me, and I expect the text size to stay unchanged and just move out of the video to make use of the space in the margins. I think margins should only affect the positioning of text, while the size should be determined by the video alone; and that’s exactly what the current code already achieves. (For what it’s worth, XySubFilter does the same.)

Maybe this is one of those things where the best defaults differ between ASS and SRT, although I don’t immediately see why I’d want text size to vary with margins with SRT either.

cc @rcombs

wm4:
Have you even tried this?

astiob:
**TL;DR: good for `--sub-scale-with-window=yes` but breaks `--sub-scale-with-window=no` which is important for styled subs. My main complaint is the size scaling pre zoom & pan. For styled subs, XySubFilter/madVR have nice but significantly different margin handling that we could emulate, too.**

I have now. It behaves just as I thought it would.

I’ve also tried various combinations of mpv’s pan, scan and `--sub-scale-with-window` with libass master/only the first commit from this PR, since you didn’t explain the problems you were trying to fix. I finally see what those problems are. They are real, but I don’t think the solution is universally correct.

**This fixes edge-case bugs but also changes normal-case behaviour.**

### This subsumes and improves `--sub-scale-with-window=yes`

As is, this seems to subsume mpv’s `--sub-scale-with-window=yes`, and **I’m happy to keep it as a libass option so mpv can use it when `--sub-scale-with-window=yes`** and remove whatever hacks it currently has for it. But it should have a descriptive name.

Above, I explained why I think `scale-with-window=yes` is wrong for styled subtitles (and mpv’s default `--sub-ass-scale-with-window=no` agrees). For unstyled subs, it’s a personal choice and some SRT subs are actually styled, so it can be useful to have the switch. Of course, mpv defaults to disabling margins for ASS entirely, but it has an option to enable them (`--sub-ass-force-margins`) and it seems useful to keep it.

### Current problems with `--sub-scale-with-window=no`

For anyone else who doesn’t know:

* Zoom ⇒ text size scales with the video.
* Zoom in a lot ⇒ scales up to a limit and then stays constant as part of libass’s effort to “fit the subtitles on screen”.

This seems acceptable, given that size scaling with video is requested.

* Pan vertically ⇒ middle-aligned subtitles move (pan) too, but no further than to the edge of the screen.
* Pan horizontally ⇒ all subtitles move (pan) too, but no further than to the edge of the screen.

Somewhat weird/inconsistent but makes some sense too.

Real trouble starts when panning around the video crops a part of it out. Presumably, this results in libass seeing negative margins, and the mechanism by which libass tries to “fit the subtitles on screen” kicks in. Evidently, this mechanism doesn’t work well when one margin is negative and the other positive:

* Pan vertically a lot ⇒ subtitles get smaller because they try to fit in the visible height of the video: the entire screen height should be used instead.
* Pan horizontally a lot ⇒ subtitle lines get squished (font size unchanged; extra line breaks) to fit in the visible part of the video: just no.

### Testing XySubFilter+madVR

I’ve tested XySubFilter too to see how it tackles these problems. It seems that XySubFilter only tells madVR which frame’s subtitles can be repositioned and which frame’s can’t, and the actual positioning is done by madVR.

Its approach is interesting: it moves subtitles out of the video and into the margins, but only as far as necessary to stop obstructing the video. It _doesn’t_ move bottom-titles to the very bottom of the screen as we do: it only moves them _just below_ the video. When zooming and panning, it zooms and pans subtitles together with the video without any limit (if you pan too far, the subtitles are cropped just as the video is). _Except_ when the bottom margin is negative due to either zooming-in or panning, it keeps bottom-titles at the bottom of the screen.

(It’s weirdly inconsistent and doesn’t move out top-titles at all and moves middle-titles to the bottom, but I think top-titles could be handled similarly to bottom-titles.)

\an2 line, zoomed out and panned, video entirely on screen: subtitle just below the video, zoomed and panned like the video is.

\an2 line, zoomed out and panned, video’s right and bottom parts off screen: subtitle just above the bottom of the screen (on top of the video), zoomed and panned only horizontally like the video is, so the right part of the line is cropped out.

I like this for ASS.

### Left/right margins

We (nor XySubFilter/madVR) don’t currently make use of the left & right margins to display text when e. g. displaying 4:3 video on a wide screen. I’ve tried hacking this in before and using it for some 4:3 videos, but I remember not liking the result so I never submitted it upstream. But this could be an answer to the problem of squishing on horizontal pan.

wm4:
Why the fuck do you care about whatever bullshit vsfilter does here? Margin handling is not part of the "standard" and thus don't have any relevance for us. My changes make handling of normal ASS dialogue similar to how SRT subtitles are treated (if margins and overrides are enabled), and otherwise tries not to corrupt subtitles. It also fixes the complete fucking nonsense that happens without these commit.

I have to think harder about not just copying the whole libass into mpv, so I don't have to spend weeks arguing about dumb shit.

astiob:
Literally the only reason I mentioned XySubFilter was because it handles this better. If you’re so allergic to that, just ignore its mention and that’s it.

You’re breaking `--sub-ass-scale-with-window=no`. I want `--sub-ass-scale-with-window=no`. That’s all.

wm4:
Your mention of vsfilter just makes this discussion more complicated and verbose and has no relevancy to this. These commits mainly fix bugs, you can introduce whatever behavior you want separately.

> You’re breaking --sub-ass-scale-with-window=no. I want --sub-ass-scale-with-window=no. That’s all.

These options never made sense in the first place and I have no idea what they do.

wm4:
PS: in addition, the buggy behavior is explicitly retained as default, so if you want that, you can keep it for the player you use.

astiob:
Replaced by #385.

The discussion above was an unfortunate chain of misunderstandings. The original post did not explain what problems the PR had the goal to fix, and the suggested code had its own issues that seemed like they might have been partly intentional. I assumed they were, but they weren’t. In the end, I believe #385 does everything that was originally desired and additionally avoids breaking ASS subtitles and existing mpv settings for SRT.

PR #385, opened 2020-03-28
Fix & rework margins (black bars)

astiob:
Follow-up to #384 as discussed on IRC.

Like #384, this makes all affected subtitles fixed within the screen/window frame and unaffected by any further pan & zoom in the media player, and disables all of this entirely when `!use_margins`.

The difference is that elements are scaled to the same sizes they would be if the screen/window was displaying a fit-to-screen letterboxed video at 100% zoom & no pan, sans the black bars from letterboxing. For example, opening a 1920×720 video on a 1920×1200 screen and then going full screen keeps element sizes unchanged (except very slightly to adjust for the disappearing menu bars etc.). _(In mpv, element sizes are enlarged because it explicitly does this with its `--sub-scale-with-video` option. Use `--sub-ass-force-margins` and play ASS to see it in action.)_

This is desirable especially for ASS or other styled subtitles where the font size is explicitly set in the subtitles and lines are manually broken to ensure a nice amount of text on screen. This also preserves the scaling mpv experiences currently, so user-configured font sizes stay the same and mpv’s sub scaling options keep working.

I also fixed a couple aspect ratio bugs that I found while figuring out the difference between `x2scr` and `x2scr_scaled`.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2795,13 +2795,13 @@ ass_start_frame(ASS_Renderer *render_priv, ASS_Track *track,
     // PAR correction
     double par = render_priv->settings.par;
     if (par == 0.) {
-        if (settings_priv->frame_width && settings_priv->frame_height &&
+        if (render_priv->orig_width && render_priv->orig_height &&
             settings_priv->storage_width && settings_priv->storage_height) {
-            double dar = ((double) settings_priv->frame_width) /
-                         settings_priv->frame_height;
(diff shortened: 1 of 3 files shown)


astiob:
@wm4 This still achieves what you wanted, right?

And it achieves the scaling that I wanted. We can both be happy.

And when no zoom & pan is active, it’s the same scaling as before, so no mpv changes are needed and users will be happy, too.

PR #390, opened 2020-04-14
ScaledBorderAndShadow

TheOneric:
This pr changes the `ScaledBorderAndShadow` default to be *VSFilter compatible and introduces a detection mechanism for subs converted to ASS by ffmpeg/libav, which did rely on libass' *VSFilter incompatible defaults.
Currently this only sets the `ScaledBorderAndShadow` header for the converted subs, however it might be worth to consider to also set other headers eg `Kerning: yes` for converted subs.


libass/ass.c
```diff
@@ -629,6 +670,58 @@ static void event_format_fallback(ASS_Track *track)
...
+        return;
+
+    if(track->track_type != TRACK_TYPE_ASS)
+        track->parser_priv->conv_score -= 5;
+
+    //2 not 1 because libass also adds a def style
+    if(track->n_styles != 2)
+        track->parser_priv->conv_score -= 3;
+    else if(!strncmp(track->styles[1].Name, "Default", 7))
+        track->parser_priv->conv_score += +3;
```
astiob, line 689: Nit: is this unary plus intentional?

libass/ass.c
```diff
@@ -629,6 +670,58 @@ static void event_format_fallback(ASS_Track *track)
...
+    // libav and past-ffmpeg used short Event formats
+    if(!strncmp(track->event_format, "Layer, Start, End, Style, Text", 30)
+       || !strncmp(track->event_format, "Layer, Start, End, Text", 23))
+        track->parser_priv->conv_score += 4;
+
+
+    //Set headers if track seems to be legacy converted
+    if(track->parser_priv->conv_score >= 12) {
+        track->ScaledBorderAndShadow = 1;
+        //RFC: Set additional headers eg Kerning ?
```
astiob, line 714: As mentioned on IRC in recent discussions, converted subs can unfortunately be quite dependent on VSFilter compatibility. So _probably_ not.

Although as it stands right now, FFmpeg strips out most ASS override tags (albeit it misses some), and without any override tags compatibility in things like kerning doesn’t matter much.

libass/ass.c
```diff
@@ -629,6 +670,58 @@ static void event_format_fallback(ASS_Track *track)
...
+    //2 not 1 because libass also adds a def style
+    if(track->n_styles != 2)
+        track->parser_priv->conv_score -= 3;
+    else if(!strncmp(track->styles[1].Name, "Default", 7))
+        track->parser_priv->conv_score += +3;
+
+    // Prior to signature ffmpeg either did not set PlayRes{X,Y},
+    // or explicitly set it to libass' default 384x288
+    if((track->PlayResX != 384 || track->PlayResY != 288)
+       && (track->PlayResX != 0 || track->PlayResY != 0 ))
```
astiob, line 694: Nit: the extra space between `0` and `))` could be removed.

TheOneric:
> Are the exact numbers arbitrary, or is there a particular logic behind them? If there is, it may be useful to explain it briefly.

The exact number are arbitrary, but the following was considered:

- |score(Signature)| ≫ threshold, so that all further score in/decrements do not change the outcome. The value is still small enough to never cause under/overflow (assuming a byte has at least 8bits)
- Apart from the signature comment, the style format line has the largest contribution, as this mix of ASS/SSA should probably be very unique. Shorter event format lines are unusual but not unique.
- Having both custom format lines puts the score very close to the threshold but it's not sufficient.
  - Unless there's a signature comment, both custom format lines are however _required_ to pass the threshold.
  - with the current scoring this means custom event lines and only one style named "Default" are required (or a signature comment and _any_ event format line)

I added further easy to check tests to reduce the possibility of false positives.
It would also be possible to check for the order of headers and the presence of CRLF newlines, or to decrement the score for each additional header _(might cause underflows)_. However this would add more complexity outside of `detect_legacy_conv_subs` and I doubt it would bring a reasonable benefit.

libass/ass.c
TheOneric: thx, fixed this and other styling issues

TheOneric:
- Rebased to master.
- ffmpeg now sets ScaledBorderAndShadow to yes by default.
- Swapped commit order to avoid having one odd commit in history behaving differently for old ffmpeg files.

TheOneric:
Overhauled to have a more general indication of which headers were set and the `generated by` signatures according to MrSmile's comments on IRC.

Code in `detect_legacy_..` could possibly simplified a bit, if "false positives" for all or at least some files with custom format lines would be acceptable. Iirc VSFilter can't display files with custom format files anyway, so all typesetting in files with custom format lines currently out there, would have been created with `SBAS=yes`. However this might also affect newly created files with custom format lines so that changing the format line unintuively also effectively changes the SBAS default value.
Not sure what's better.

libass/ass.c
```diff
@@ -683,6 +685,130 @@ static void event_format_fallback(ASS_Track *track)
...
+    // No definite clues: Continue to make an educated guess
+    // if this is an unmodified legacy ffmpeg file
+    if (track->track_type != TRACK_TYPE_ASS)
+        conv_score -= 5;
+
+    if ((track->PlayResX != 384 || track->PlayResY != 288)
+        && (
+            sinfo_set(track->parser_priv, SINFO_PLAYRESX)
+            || sinfo_set(track->parser_priv, SINFO_PLAYRESY)
+           )
```
MrSmile, line 783: Incorrect style here. Also can be simplified with the suitable enum:
`track->parser_priv->sinfo_flags & (SI_PLAYRESX | SI_PLAYRESY)`

libass/ass.c
```diff
@@ -683,6 +685,130 @@ static void event_format_fallback(ASS_Track *track)
...
+
+    //Final Score
+    ass_msg(track->library, MSGL_DBG2,
+            "Track legacy conv score: %d",
+            conv_score);
+
+    if (conv_score < 12)
+        goto not_legacy_ffmpeg;
+
+set_legacy_ffmpeg:
```
MrSmile, line 795: I think that `detect` function should not set anything but return a boolean flag or the `conv_score` directly. All updates should be done outside of it. Also it'd be much cleaner without a ton of gotos.

libass/ass.c
```diff
@@ -683,6 +685,130 @@ static void event_format_fallback(ASS_Track *track)
...
+     *    Signature comment
+     *  56bc0a6736cdc7edab837ff8f304661fd16de0e4  2015.02.08
+     *    Allow custom PlayRes(X|Y)
+     *  a8ba2a2c1294a330a0e79ae7f0d3a203a7599166  2020.04.17
+     *    Set 'ScaledBorderAndShadow: yes'
+     *
+     * libav outputs initial ffmpeg format. (no longer maintained)
+     */
+
+    signed char conv_score = 0;
```
MrSmile, line 725: `signed char` looks like overengineering and premature optimization. `int` is sufficient here.

libass/ass.c
```diff
@@ -591,27 +618,56 @@ static int process_styles_line(ASS_Track *track, char *str)
     return 0;
 }
 
+static inline void check_duplicate_info_line(
+                                    const ASS_Track *track,
+                                    const ScriptInfo si, const char* name
+) {
```
MrSmile, line 624: Incorrect style here.

TheOneric:
>     * Do not mix detection approaches: exact with boolean result and similarity with score result.
Yes, depending on the feedback to the approaches, I intended to change this.

libass/ass.c
```diff
@@ -683,6 +685,130 @@ static void event_format_fallback(ASS_Track *track)
...
+    tmp >>= sizeof(tmp)*CHAR_BIT - SINFO_GENBY;
+    // Note: If there's SINFO_SCRIPTTYPE in the future this needs to be updated
+    tmp &= ~(1 << SINFO_PLAYRESX | 1 << SINFO_PLAYRESY);
+    if (tmp != 0)
+        goto not_legacy_ffmpeg;
+
+    // Legacy ffmpeg only ever has one style
+    // 2 not 1 because libass also adds a def style
+    if (track->n_styles != 2)
+        conv_score -= 3;
```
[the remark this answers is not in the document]
TheOneric: Yes, depending on what approach will be chosen this can be simpified.

libass/ass.c
```diff
@@ -60,8 +72,14 @@ struct parser_priv {
     int check_readorder;
 
     int enable_extensions;
+
+    // track which [Script Info] headers have been set by the script
+    uint32_t sinfo_flags;
```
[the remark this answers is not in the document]
TheOneric: What name would you suggest ? `script_sinfo_flags`, `track_sinfo_flags`, … ?

libass/ass.c
```diff
@@ -84,8 +85,9 @@ struct parser_priv {
     uint32_t sinfo_flags;
 };
 
-#define set_sinfo(PP,SI) (PP->sinfo_flags |= 1 << SI)
-#define sinfo_set(PP,SI) ((PP->sinfo_flags & (1 << SI)) != 0)
+#define set_sinfo(PP,SI)   (PP->sinfo_flags |= 1 << SI)
```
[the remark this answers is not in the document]
TheOneric: Thanks. This is good idea and makes the macros unneccessary.

libass/ass.c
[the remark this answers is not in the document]
MrSmile: `script_info_flags` or simple `header_flags`.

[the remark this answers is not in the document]
TheOneric: I mainly added the comment to make clear that it SINFO_* only indicates if an Header has been set, **by the script file**/stream itself and not eg changed the regular default value for legacy cases.
This seems more sensible too me as otherwise the track->`header` value can just be used directly.
Expanding `sinfo` or swapping it for `header` doesn't make this clear imho. You can of course argue that these are better variable names in general.

libass/ass.c
```diff
@@ -651,6 +658,14 @@ static int process_info_line(ASS_Track *track, char *str)
         while (*p && ass_isspace(*p)) p++;
         free(track->Language);
         track->Language = strndup(p, 2);
+    } else if (!strncmp(str, "; Script generated by ", 22)) {
+        check_duplicate_info_line(track, SINFO_GENBY, "Generated by");
+        if (!strncmp(str + 22, "Aegisub", 7))
```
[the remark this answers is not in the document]
TheOneric: Done.

libass/ass.c
```diff
@@ -683,6 +685,130 @@ static void event_format_fallback(ASS_Track *track)
...
+     *  56bc0a6736cdc7edab837ff8f304661fd16de0e4  2015.02.08
+     *    Allow custom PlayRes(X|Y)
+     *  a8ba2a2c1294a330a0e79ae7f0d3a203a7599166  2020.04.17
+     *    Set 'ScaledBorderAndShadow: yes'
+     *
+     * libav outputs initial ffmpeg format. (no longer maintained)
+     */
+
+    signed char conv_score = 0;
+    uint32_t tmp;
```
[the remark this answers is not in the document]
TheOneric: SBAS=1 for all custom format lines and reducing `GENBY_*` allowed for much simpler `detect_legacy`. Scoring or bitflags are no longer neccessary.

libass/ass.c
TheOneric: Done.

libass/ass.c
TheOneric: Done.

libass/ass.c
TheOneric: Done *(probably?)*

TheOneric:
- Reduced GENBY_ to only FFMPEG
- All files with custom format lines now default to SBAS=1
→ As a result `detect_legacy_…` is now **much** simpler.

One thing though I'm not so sure about is the `Name`/`Actor` field in the default ASS Event format line.
libass had this as `Actor` all the way back to its initial commit:
https://github.com/libass/libass/blob/3ac944f8a6bea86e713e3bc225f6ceebdc5ae362/libass/ass.c#L470
**But** Aegisub, ffmpeg and the "spec" all use `Name`. I couldn't find any reference to `Actor` anywhere else. Is this just a long-lived typo, or are there other "standard variants" of the format line I'm not aware of that should be considered when checking for "non-standard" format lines ?

libass/ass.c
```diff
@@ -81,6 +81,23 @@ struct parser_priv {
     uint32_t sinfo_flags;
 };
 
+static const char *ASS_STYLE_FORMAT =
```
MrSmile, line 84: Not sure about all uppercase here. Personally I think that it should be all lowercase instead, but I'm not an expert on libass style conventions.

libass/ass.c
```diff
@@ -602,6 +630,8 @@ static int process_styles_line(ASS_Track *track, char *str)
         track->style_format = strdup(p);
         ass_msg(track->library, MSGL_DBG2, "Style format: %s",
                track->style_format);
+        if (strcmp(p, ASS_STYLE_FORMAT) && strcmp(p, SSA_STYLE_FORMAT))
```
MrSmile, line 633: Maybe it's better to have whitespace-agnostic comparison here.

libass/ass.c
```diff
@@ -55,7 +55,11 @@ typedef enum {
     SINFO_WRAPSTYLE    = 1 << 4,
     SINFO_SCALEDBORDER = 1 << 5,
     SINFO_COLOURMATRIX = 1 << 6,
-    SINFO_KERNING      = 1 << 7
+    SINFO_KERNING      = 1 << 7,
+    // SINFO_GENBY indicates if ANY GENBY has been set
+    SINFO_GENBY        = 1 << 8,
```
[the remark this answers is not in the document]
TheOneric: Only to warn about duplicate genby-signatures. But I guess that's not really necessary as the log messages in legacy detection should be clear enough. And all tested generators seem to override other genby-signatures anyway.
Will remove.

libass/ass.c
TheOneric: Upon closer inspection the parser doesn't care about lower-, uppercase. Case and whitespace agnostic comparison shouldn't have false positives and all false negatives would cause parser problems at another point anyway (eg `[…], Effec t, Te xt`).
It's probably better though to utilise `next_token` and `ass_strcasecmp` directly to compare the format lines.

TheOneric: Done.

libass/ass.c
```diff
@@ -591,23 +606,44 @@ static int process_styles_line(ASS_Track *track, char *str)
     return 0;
 }
 
+static inline void check_duplicate_info_line(const ASS_Track *const track,
+                                             const ScriptInfo si,
+                                             const char *const name)
+{
+    if ((track->parser_priv->sinfo_flags & si) != 0)
```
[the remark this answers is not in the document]
TheOneric: Done.

libass/ass.c
```diff
@@ -648,6 +652,10 @@ static int process_info_line(ASS_Track *track, char *str)
         while (*p && ass_isspace(*p)) p++;
         free(track->Language);
         track->Language = strndup(p, 2);
+    } else if (!strncmp(str, "; Script generated by ", 22)) {
+        check_duplicate_info_line(track, SINFO_GENBY, "Generated by");
```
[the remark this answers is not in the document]
TheOneric: Done.

libass/ass.c
TheOneric: libass does indeed seem to use lowercase for `static const`. Thanks!
Done.

TheOneric:
In addtition to changes suggested by MrSmile:
- Rebased to master
- Squashed "genby" and "ffmpeg" commit as genby is now used solely for legacy ffmpeg detection

libass/ass.c
```diff
@@ -590,15 +597,68 @@ static int process_style(ASS_Track *track, char *str)
 
 }
 
+// WILL change contents of fmt1 and fmt2
+static int format_line_compare(char *fmt1, char *fmt2)
```
MrSmile, line 603: Destructive comparison functions are evil. Maybe something like that would be better:
```c
// pattern is a list of words separated by ", "
bool format_line_compare(const char *fmt, const char *pattern)
{
    skip_spaces(&fmt);
    while (true) {
        if (*pattern == '\0') {
            skip_spaces(&fmt);
            return *fmt == '\0';
        }
        if (*pattern == ',') {
            skip_spaces(&fmt);
            if (*fmt++ != ',')
                return false;
            skip_spaces(&fmt);
            pattern += 2;  // skip space after comma
            continue;
        }
        if (lowertab[(uint8_t) *fmt++] != lowertab[(uint8_t) *pattern++])
            return false;
    }
}
```

TheOneric: I didn't like this either, but was reluctant to duplicate the token.
I now put most of the non-destructive token-logic into the `inline` function `advance_token_pos`.
This should not affect the performance of `next_token` negatively.
 I should point out though, that with this change `next_token` does no longer replace **all** `','` with `'\0'`, but only if a token is terminated by a `','`. As far as I can tell, replacing them all was superfluous anyway.

Unfortunately `advance_token_pos` can't just have `const char**` arguments, as `(r)skip_spaces` takes `char**`, even though they don't modify the `char`-data. The compiler doesn't like passing `char**` to `const char**`.
For this reason `strdup` is still used, the alternative would be casting between `const char**` and `char**` at the necessary places.

libass/ass.c
```diff
@@ -77,6 +77,22 @@ struct parser_priv {
     uint32_t sinfo_flags;
 };
 
+static const char *ass_style_format =
+        "Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, "
+        "OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, "
+        "ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, "
+        "Alignment, MarginL, MarginR, MarginV, Encoding";
+static const char *ass_event_format =
+        "Layer, Start, End, Style, Name, "
```
TheOneric, line 88: `Actor` *(used for the entire git history) changed to `Name`.
As I can't find any other reference to "Actor", this was probably a typo, but if there is some variant of "standard format lines" I'm not aware of, which use `Actor` instead of `Name`, please let me know.

libass/ass.c
```diff
@@ -301,25 +317,28 @@ static long long string2timecode(ASS_Library *library, char *p)
...
-    }
-    if (*p == '\0') {
-        *str = p;               // eos found, str will point to '\0' at exit
-    } else {
-        *p = '\0';
-        *str = p + 1;           // ',' found, str will point to the next char (beginning of the next token)
-    }
-    rskip_spaces(&p, start);    // end of current token: the first space character, or '\0'
+
+    advance_token_pos(str, &start, &p);
```
TheOneric, line 342: With this change `next_token` no longer replaces **all** `','` with `'\0'`, only if they terminate a token.
As far as I can tell, this was superfluous and removing it should not have any negative impact.

libass/ass.c
MrSmile: It's perfectly normal to cast to `(char *)` in such cases.

TheOneric: Done.

libass/ass.c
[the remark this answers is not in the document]
TheOneric: I renamed `sinfo_flags` to `header_flags`.

libass/ass.c
astiob: (_Technically,_ this isn’t even `static const`: it’s a nonconst pointer to const character.)

TheOneric: They probably should be const, thanks.
Pushed
```diff
-static const char *ass_style_format =
+static const char *const ass_style_format =
```
for all format line defaults

astiob:
Why do we even read `Timer:`? 🤯

Issue #383, opened 2020-03-21
add SubtitlesOctopus and vidcoder


realfinder:
both https://github.com/weizhenye/ASS and https://github.com/Dador/JavascriptSubtitlesOctopus to https://github.com/libass/libass#other-assssa-implementations

and https://vidcoder.net/ to https://github.com/libass/libass#related-links

astiob:
Do you still want VidCoder to be added? It seems to be an alternate GUI for HandBrake, which is present in the list.

PR #392, opened 2020-05-03
Purely non-functional nit-picks and issue 383

TheOneric:
This pr only contains functional changes not changing the functionality, most of them nits.
Aside from the pendatic style nits, this suppresses a gcc build warning in a portable way and addresses issue #383.

#### Style Nit-Picks
By fixing the ca 25 lines containing style issues, it becomes possible to automatically check for some common style issues via script, which I find convenient to lessen the chance of accidentally slipping style issues into a patch. The script should work on most Linux and BSD *(tested with FreeBSD's bsdgrep)* systems as well as OSX *(untested, but docs indicate so)*.
I can squash the Style-Nits if requested and will drop them(+script) entirely, if purely stylistics commits are not favoured. So feel free to dismiss them.

#### -Wimplicit-fallthrough
I'm pretty sure the 3 warnings are false positives.
Using the comment -Wimplicit-fallthrough warnings will be suppressed up to level 4. Attributes to suppress the warnings on level 5 are not portable.
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough

#### Regarding #383 
I did not add vidcoder as requested in the issue for now as it seems to only be an alternate Windows GUI for Handbrake, which is already present.


libass/ass_font.c
```diff
@@ -87,7 +87,7 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
     if (!face->charmap)
         return symbol;
 
-    switch(face->charmap->encoding){
+    switch(face->charmap->encoding) {
```
astiob, line 90: Should probably add a space after `switch`, too.

profile/profile.c
```diff
@@ -86,7 +86,7 @@ int main(int argc, char *argv[])
         exit(1);
     }
 
-    while(tm < time){
+    while(tm < time) {
```
astiob, line 89: Should probably add a space after `while`, too.

README.md
```diff
@@ -57,3 +57,5 @@ Other ASS/SSA implementations:
 - [Perian](https://github.com/MaddTheSane/perian)
 - [asa](http://git.spaceboyz.net/asa.git) (defunct)
 - [libjass](https://github.com/Arnavion/libjass)
+- [SubtitlesOctopus](https://github.com/Dador/JavascriptSubtitlesOctopus) (builds on wasm libass)
+- [Javascript ASS](https://github.com/weizhenye/ASS)
```
astiob, line 61: It seems the name is ASS.js, and it seems it doesn’t actually use libass.

checkStyle.sh
```diff
@@ -0,0 +1,70 @@
...
+# Apart from that, this should work on any POSIX-compliant system
+
+# This Script checks for some common styling issues. Ideally a commit should pass
+# ./checkStyle.sh, but passing ./checkStyle.sh does not guarantee the abscence
+# of styling issues
+
+
+cd "$(dirname $0)"
+
+#Setup vars and printErr function
```
astiob, line 14: Want to add a space after the `#`? :-)

astiob:
Hang on, I got that backwards. That’s a list of _alternative_ ASS implementations, so that’s the right place for something that doesn’t use libass. But then isn’t it the wrong place for things that _do_ use libass? Which I guess is only SubtitlesOctopus in that list. And I see you added a note that it does use libass. Hmm. Should it be moved up to the list of things that use libass? I’m not quite sure.

README.md
TheOneric: SubtitlesOctopus does pull in libass as a dependency.

TheOneric: I can also move it up the list, which is technically more correct. But fuctionally SubtitlesOctopus makes libass available for user on browsers, which makes it use-case more similar to ASS.js and libjass.

TheOneric:
> Maybe just check for any alphabetic character instead of specific keywords?

To check for **no** space alphanumeric works for brackets. In case of **too much** space:
For opening brackets alphanumeric currently works, but closing braket breaks for the `PARSE_END` macro. It might be more coherent to leave the check for too much space on opening brackets only to only check specific keywords too.
```
./libass/ass.c:260:#define PARSE_END   }
```

astiob:
> but closing braket breaks for the PARSE_END macro. It might be more coherent to leave the check for too much space on opening brackets only to only check specific keywords too.
>
> ```./libass/ass.c:260:#define PARSE_END   }```

Why check words before closing brace at all? With keywords, that’s not even valid syntax.

README.md
astiob: I’m _slightly_ inclined to think it should go in the first list because it’s more “software that uses libass” than “another ASS implementation”, but I do appreciate that it’s kinda middle ground. Would be good to hear other opinions, but ultimately this isn’t important.

TheOneric:
> Why check words before closing brace at all? With keywords, that’s not even valid syntax.

Ohh, I blundered. Even though I pointed out the matched macro I didn't notice there was a typo in line 64. This was supposed to check for **opening** braces. Now that this is fixed, alphanumeric does indeed work.
Thanks for pointing this out.

README.md
TheOneric: The above was only my reasoning for putting it to the ASS implementations, but I don't really have any strong opinion on this. I now moved SubtitleOctopus up to "software that uses libass".

rcombs:
I'm happy with the changes but wonder if we might do better to use clang-format, rather than a custom linter script. Something like this, for a start:
```
---
AlignConsecutiveAssignments: true
AlignConsecutiveMacros: true
AlignEscapedNewlines: Left
AllowShortBlocksOnASingleLine: true
AllowShortCaseLabelsOnASingleLine: true
AllowShortFunctionsOnASingleLine: true
AllowShortLoopsOnASingleLine: true
BreakBeforeBraces: Linux
ColumnLimit: 0
IncludeBlocks: Regroup
IncludeCategories:
  - Regex:    '^"config.h"$'
    Priority: 1
    SortPriority: 1
  - Regex:    '^"ass_compat.h"$'
    Priority: 1
    SortPriority: 2
  - Regex:    '^(<|".+/)'
    Priority: 3
IndentCaseLabels: false
IndentWidth: 4
Language: Cpp
MaxEmptyLinesToKeep: 1
PointerAlignment: Right
SortIncludes: true
SpaceBeforeAssignmentOperators: true
SpaceBeforeParens: ControlStatements
SpacesInCStyleCastParentheses: false
UseTab: Never

...
```

I'd say we should merge the rest of this and then continue discussion on linter stuff.

MrSmile:
I think that all style commits (whitespace-only) should be squashed into one.
And indeed an existing style checking utility is better than custom script.

libass/ass_utils.h
```diff
@@ -179,9 +179,9 @@ static inline uint32_t fnv_32a_buf(void *buf, size_t len, uint32_t hval)
     size_t n = (len + 3) / 4;
 
     switch (len % 4) {
-    case 0: do { hval ^= *bp++; hval *= FNV1_32A_PRIME;
-    case 3:      hval ^= *bp++; hval *= FNV1_32A_PRIME;
-    case 2:      hval ^= *bp++; hval *= FNV1_32A_PRIME;
+    case 0: do { hval ^= *bp++; hval *= FNV1_32A_PRIME; //-fallthrough
```
MrSmile, line 182: Is there any special meaning in `-` symbol? As far as I know, simple `// falthrough` is sufficient for gcc/clang.

astiob: See https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough (as linked in the OP). There’s a warning level option and this works at the highest level whereas simple `fallthrough` doesn’t.

astiob:
Eh, I like the separate commits because they clearly identify what kind of stuff each commit does. But squashing them is fine, too.

TheOneric:
Dropped style_check.sh

> I think that all style commits (whitespace-only) should be squashed into one.

> Eh, I like the separate commits because they clearly identify what kind of stuff each commit does. But squashing them is fine, too.

I pushed a squashed version for now, but if it's decided that separate commits are preferred, I've kept the separate commits locally and can un-squash them again.

TheOneric:
- Rebased to master
- Mark libjass and ffdshow as *defunct* the README
  - libjass is officially discontinued as written in its README
  - ffdshow's (ffdshow-tryout) last release was 2014 and the last activity in the source code in 2015

TheOneric:
Rebased to master and added an old commit improving the info about profiles arguments.
The added profile commit changes lines touched by the style commits of this pr.

@rcombs Are you happy with how I integrated your commit or would you prefer some changes ?

---

Otherwise: Is there anything left I need to chage so that this pr can be merged ?

rcombs:
Hah, I had indeed forgotten that commit. Sure, looks good to me. My only note on the entire PR is that it might be worth making Crunchyroll a sub-item under JavascriptSubtitlesOctopus, since they use that.

TheOneric:
> Hah, I had indeed forgotten that commit. Sure, looks good to me. My only note on the entire PR is that it might be worth making Crunchyroll a sub-item under JavascriptSubtitlesOctopus, since they use that.

Done.

PR #393, opened 2020-05-06
Static analysis warning fixes

rcombs:
Ran scan-build while poking around at clang-tidy and clang-format, and turned up some warnings. Some were false positives (but could easily be silenced in ways that improved code clarity); some were legitimate bugs. In the process, I ran into a couple additional bugs and went ahead and fixed them as well.

### DIFF
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -62,18 +62,19 @@ static void destroy_font(void *priv)
 
 static bool check_postscript(void *priv)
 {
+    bool ret = false;
     CTFontDescriptorRef fontd = priv;
     CFNumberRef cfformat =
         CTFontDescriptorCopyAttribute(fontd, kCTFontFormatAttribute);
     int format;
 
-    if (!CFNumberGetValue(cfformat, kCFNumberIntType, &format))
-        return false;
+    if (CFNumberGetValue(cfformat, kCFNumberIntType, &format)) {
+        ret = format == kCTFontFormatOpenTypePostScript ||
+              format == kCTFontFormatPostScript;
+    }
 
     SAFE_CFRelease(cfformat);
-
-    return format == kCTFontFormatOpenTypePostScript ||
-           format == kCTFontFormatPostScript;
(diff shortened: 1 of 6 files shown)


libass/ass_coretext.c
```diff
@@ -62,18 +62,21 @@ static void destroy_font(void *priv)
...
 {
+    bool ret = false;
     CTFontDescriptorRef fontd = priv;
     CFNumberRef cfformat =
         CTFontDescriptorCopyAttribute(fontd, kCTFontFormatAttribute);
     int format;
 
     if (!CFNumberGetValue(cfformat, kCFNumberIntType, &format))
-        return false;
+        goto fail;
```
MrSmile, line 72: Instead of this `goto` I think a simple assignment of the `ret` variable under (inverse) condition is better.

libass/ass_render.c
```diff
@@ -113,7 +111,13 @@ ASS_Renderer *ass_renderer_init(ASS_Library *library)
     priv->settings.shaper = ASS_SHAPING_SIMPLE;
 #endif
 
-  ass_init_exit:
+    if (!priv->cache.font_cache || !priv->cache.bitmap_cache || !priv->cache.composite_cache || !priv->cache.outline_cache ||
+        !priv->text_info.combined_bitmaps || priv->text_info.glyphs || !priv->text_info.lines || !priv->shaper) {
+fail:
```
MrSmile, line 116: Jump into a conditioned block looks ugly, especially as next condition is basically a continuation of this block. I think it's better to move `fail:` at the function end, together with deallocation, `priv = NULL` and message. And in this condition to do `goto fail` similar to failures above. Maybe even better to split that monstrous condition into parts and move them near corresponding allocations.

libass/ass_coretext.c
rcombs: Sure.

libass/ass_render.c
rcombs: Done.

PR #394, opened 2020-05-10
Blur algorithm overhaul

MrSmile:
My initial blur algorithm was overcomplicated due to (mostly arbitrary) decision to use filters not larger than 9-tap. That leads to use of an additional stage (prefilter) and irregular main filter pattern (1235, 1246). Over time I realized that it's not worth it and did some experiments. And now it's time to apply my new understanding to libass!

Also now I have AVX2-capable CPU and more practice with AVX2 assembly so there are some improvements in that area. Overall this new algorithm should have slightly better precision and performance than old.

### DIFF
--- a/libass/ass_blur.c
+++ b/libass/ass_blur.c
@@ -76,10 +76,10 @@ inline static void copy_line(int16_t *buf, const int16_t *ptr, uintptr_t offs, u
 void ass_stripe_unpack_c(int16_t *dst, const uint8_t *src, ptrdiff_t src_stride,
                          uintptr_t width, uintptr_t height)
 {
-    for (uintptr_t y = 0; y < height; ++y) {
+    for (uintptr_t y = 0; y < height; y++) {
         int16_t *ptr = dst;
         for (uintptr_t x = 0; x < width; x += STRIPE_WIDTH) {
-            for (int k = 0; k < STRIPE_WIDTH; ++k)
+            for (int k = 0; k < STRIPE_WIDTH; k++)
                 ptr[k] = (uint16_t) (((src[x + k] << 7) | (src[x + k] >> 1)) + 1) >> 1;
                 //ptr[k] = (0x4000 * src[x + k] + 127) / 255;
             ptr += STRIPE_WIDTH * height;
@@ -94,9 +94,9 @@ void ass_stripe_pack_c(uint8_t *dst, ptrdiff_t dst_stride, const int16_t *src,
 {
     for (uintptr_t x = 0; x < width; x += STRIPE_WIDTH) {
         uint8_t *ptr = dst;
-        for (uintptr_t y = 0; y < height; ++y) {
+        for (uintptr_t y = 0; y < height; y++) {
             const int16_t *dither = dither_line + (y & 1) * STRIPE_WIDTH;
-            for (int k = 0; k < STRIPE_WIDTH; ++k)
+            for (int k = 0; k < STRIPE_WIDTH; k++)
                 ptr[k] = (uint16_t) (src[k] - (src[k] >> 8) + dither[k]) >> 6;
                 //ptr[k] = (255 * src[k] + 0x1FFF) / 0x4000;
             ptr += dst_stride;
@@ -105,8 +105,8 @@ void ass_stripe_pack_c(uint8_t *dst, ptrdiff_t dst_stride, const int16_t *src,
         dst += STRIPE_WIDTH;
     }
     uintptr_t left = dst_stride - ((width + STRIPE_MASK) & ~STRIPE_MASK);
-    for (uintptr_t y = 0; y < height; ++y) {
-        for (uintptr_t x = 0; x < left; ++x)
+    for (uintptr_t y = 0; y < height; y++) {
+        for (uintptr_t x = 0; x < left; x++)
             dst[x] = 0;
         dst += dst_stride;
     }
@@ -142,15 +142,15 @@ void ass_shrink_horz_c(int16_t *dst, const int16_t *src,
     int16_t buf[3 * STRIPE_WIDTH];
     int16_t *ptr = buf + STRIPE_WIDTH;
     for (uintptr_t x = 0; x < dst_width; x += STRIPE_WIDTH) {
-        for (uintptr_t y = 0; y < src_height; ++y) {
+        for (uintptr_t y = 0; y < src_height; y++) {
             copy_line(ptr - 1 * STRIPE_WIDTH, src, offs - 1 * step, size);
             copy_line(ptr + 0 * STRIPE_WIDTH, src, offs + 0 * step, size);
             copy_line(ptr + 1 * STRIPE_WIDTH, src, offs + 1 * step, size);
-            for (int k = 0; k < STRIPE_WIDTH; ++k)
+            for (int k = 0; k < STRIPE_WIDTH; k++)
                 dst[k] = shrink_func(ptr[2 * k - 4], ptr[2 * k - 3],
                                      ptr[2 * k - 2], ptr[2 * k - 1],
                                      ptr[2 * k + 0], ptr[2 * k + 1]);
-            dst += STRIPE_WIDTH;
+            dst  += STRIPE_WIDTH;
             offs += STRIPE_WIDTH;
         }
         offs += step;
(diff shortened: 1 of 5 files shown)


libass/ass.c
```diff
@@ -19,20 +19,12 @@
...
-#include <errno.h>
-#include <sys/types.h>
-#include <sys/stat.h>
 #include <inttypes.h>
 
 #ifdef CONFIG_ICONV
 #include <iconv.h>
 #endif
 
-#include "ass.h"
```
rcombs, line 35: I'm very confused. This file is literally defining the functions declared in this header. How is it inappropriate for it to include it?

libass/ass.c
```diff
@@ -19,20 +19,12 @@
 #include "config.h"
 #include "ass_compat.h"
 
-#include <stdio.h>
-#include <stdlib.h>
```
rcombs, line 23: This file uses a lot of stuff in these system headers; just because it happens to include something else that happens to also include these doesn't mean we shouldn't include them here. I don't like this commit apart from the removal of includes that are completely unused in a particular file (like FT_OUTLINE_H in files that don't use anything from that header).

libass/ass_blur.c
```diff
@@ -26,16 +26,15 @@
...
  * Experiment shows that downscaling of factor 2 with kernel [1, 5, 10, 10, 5, 1] and
  * corresponding upscaling are enough for 8-bit precision.
  *
- * For central filter here is used generic 9-tap filter with one of 3 different patterns
- * combined with one of optional prefilters with fixed kernels. Kernel coefficients
- * of the main filter are obtained from solution of least squares problem
- * for Fourier transform of resulting kernel.
+ * For central filter here is used generic filter with 5 different kernel width (9 to 17-tap).
+ * Kernel coefficients of that filter are obtained from the solution of least squares problem
+ * for Fourier transform of the resulting kernel.
```
rcombs, line 37:
```suggestion
 * Here we use generic filters with 5 different kernel widths (9 to 17-tap).
 * Kernel coefficients of that filter are obtained from the solution of the least-squares problem
 * for the Fourier transform of the resulting kernel.
```

libass/ass_blur.c
```diff
@@ -26,16 +26,15 @@
  * Cascade Blur Algorithm
  *
  * The main idea is simple: to approximate gaussian blur with large radius
- * you can downscale, then apply filter with small pattern, then upscale back.
+ * you can downscale, then apply filter with relativly small pattern, then upscale back.
```
rcombs, line 29:
```suggestion
 * The main idea is simple: to approximate a gaussian blur with large radius,
 * you can scale down, apply a filter with a relatively small pattern, then scale back up.
```

libass/ass_blur.c
```diff
@@ -26,16 +26,15 @@
...
  *
  * To achieve desired precision down/upscaling should be done with sufficiently smooth kernel.
  * Experiment shows that downscaling of factor 2 with kernel [1, 5, 10, 10, 5, 1] and
  * corresponding upscaling are enough for 8-bit precision.
```
rcombs, line 33:
```suggestion
 * To achieve the desired precision, scaling should be done with sufficiently smooth kernel.
 * Experiments show that downscaling of factor 2 with kernel [1, 5, 10, 10, 5, 1] and
 * corresponding upscaling are enough for 8-bit precision.
```

libass/ass_blur.c
```diff
@@ -75,10 +72,10 @@ inline static void copy_line(int16_t *buf, const int16_t *ptr, uintptr_t offs, u
 void ass_stripe_unpack_c(int16_t *dst, const uint8_t *src, ptrdiff_t src_stride,
                          uintptr_t width, uintptr_t height)
 {
-    for (uintptr_t y = 0; y < height; ++y) {
+    for (uintptr_t y = 0; y < height; y++) {
```
rcombs, line 75: Would be nice to split these little stylistic changes into a separate commit.

libass/ass_blur.c
```diff
@@ -59,9 +58,7 @@ inline static const int16_t *get_line(const int16_t *ptr, uintptr_t offs, uintpt
 
 inline static void copy_line(int16_t *buf, const int16_t *ptr, uintptr_t offs, uintptr_t size)
 {
-    ptr = get_line(ptr, offs, size);
-    for (int k = 0; k < STRIPE_WIDTH; ++k)
-        buf[k] = ptr[k];
+    memcpy(buf, get_line(ptr, offs, size), STRIPE_WIDTH * sizeof(buf[0]));
```
rcombs, line 61: Similarly, this can be split out.

libass/ass_blur.c
```diff
@@ -250,404 +247,153 @@ void ass_expand_vert_c(int16_t *dst, const int16_t *src,
...
 {
-    uintptr_t dst_width = src_width + 4;
+    uintptr_t dst_width = src_width + 2 * n;
     uintptr_t size = ((src_width + STRIPE_MASK) & ~STRIPE_MASK) * src_height;
     uintptr_t step = STRIPE_WIDTH * src_height;
 
     uintptr_t offs = 0;
-    int16_t buf[2 * STRIPE_WIDTH];
-    int16_t *ptr = buf + STRIPE_WIDTH;
+    int16_t buf[3 * STRIPE_WIDTH], *ptr = buf + 2 * STRIPE_WIDTH;
```
rcombs, line 277: I think this'd be clearer if kept as 2 separate lines.

libass/ass_blur.c
```diff
@@ -693,142 +429,95 @@ static void coeff_filter(double *coeff, int n, const double kernel[4])
...
-    double vec[4];
-    for (int i = 0; i < 4; ++i)
-        vec[i] = mat_freq[0] - mat_freq[index[i]] - vec_freq[0] + vec_freq[index[i]];
-
-    for (int i = 0; i < 4; ++i) {
+    double mat_freq[17] = { kernel[0], kernel[1], kernel[2], kernel[3] };
+    coeff_filter(mat_freq, 7, kernel);
+
+    double vec_freq[12];
+    calc_gauss(vec_freq, n + 4, r2 * mul);
```
rcombs, line 480: Worth an `assert(n + 4 <= sizeof(vec_freq)/sizeof(vec_freq[0]))` here.

libass/ass_blur.c
```diff
@@ -861,38 +547,28 @@ bool ass_gaussian_blur(const BitmapEngine *engine, Bitmap *bm, double r2)
...
         index ^= 1;
     }
-    if (blur.prefilter) {
-        engine->pre_blur_horz[blur.prefilter - 1](buf[index ^ 1], buf[index], w, h);
-        w += 2 * blur.prefilter;
-        index ^= 1;
-    }
-    engine->main_blur_horz[blur.filter](buf[index ^ 1], buf[index], w, h, blur.coeff);
-    w += 2 * blur.filter + 8;
+    engine->blur_horz[blur.radius - 4](buf[index ^ 1], buf[index], w, h, blur.coeff);
```
rcombs, line 560: `assert(blur.radius >= 4);`

libass/ass.c
MrSmile: It's included through `ass_utils.h`. I don't think that header list has any impact on code clarity, so it's better for compiler to touch hard drive less.

libass/ass.c
MrSmile: `ass_utils.h` is essentially a sort of precompiled header with bunch of standard system includes that get included everywhere. So I think that it's not much sense to include them repeatedly. But I'm not holding that commit too dearly so if there is a consensus that excessive includes are necessary I'll drop it.

MrSmile:
> How's the perf look in benchmarking?

Slightly better on my large blur benchmark. But it's a complex test with bunch of blur radii, ideally blur should be benchmarked individually for each radius and for different CPU/OSes too.

> Can we add a couple higher-value \blur and \be tests to the compare infrastructure and use those to validate this?

I have a [special test](https://github.com/MrSmile/libass-tests/tree/master/blur-test) for that. Should trigger most of blur codepaths.

libass/ass_blur.c
MrSmile: I think that quite falls under "Simplify blur algorithm". If I split that into another commit then assembly modifications of the old functions (shrink/expand) should be separated too.

libass/ass_blur.c
MrSmile: `n + 4` is already copy-paste so using `sizeof(vec_freq)/sizeof(vec_freq[0])` instead of simply `12` don't make much sense. I think I'll go with `assert(n <= 8)` at the start of the function.

libass/ass_blur.c
MrSmile: Done

libass/ass_blur.c
MrSmile: Done

libass/ass_blur.c
MrSmile: Done

libass/ass_blur.c
MrSmile: Done

MrSmile:
I've rebased and moved the first unrelated commit to a separate PR #430.

rcombs (approved):

astiob:
I’m just gonna trust you on this one.

Issue #171, opened 2015-04-06
Various widths not scaled with PlayResX


aphirst:
```
<torque> uh that looks distinctly like border style 3 with non-100 fscx for anamorphic compensation
<aphirst> http://code.google.com/p/libass/wiki/IssuesAndDifferences
<aphirst> "BorderStyle 3 ("opaque box") does not behave like in VSFilter all the time. VSFilter's implementation is buggy and emulation of these bugs is hard. If possible, avoid using it."
```

**Sample file:** http://a.pomf.se/jgjpyg.mkv (approximately 7 seconds in)
**Rendering in MPC:** http://a.pomf.se/bmcwwd.jpg
**Rendering in MPV & libass:** http://a.pomf.se/hcljvr.jpg
As can clearly be seen, the opaque box between the video and a sign subtitle is the wrong size as compared to the "correct" rendering from MPC with xy-vsfilter.

**My system:**
- Arch Linux x86_64
- `libass 0.12.1`
- `mpv 0.8.3`
- `freetype2-infinality-ultimate 2.5.5`
- `fontconfig-infinality-ultimate 2.11.1`
- `harfbuzz 0.9.40`

astiob:
To clarify: the sample in the original post does not actually have BorderStyle 3 but rather an explicit vector drawing.

However (as illustrated in https://github.com/libass/libass/pull/395#issuecomment-629869816), it turns out there are a whole bunch of widths that we scale as `frame_height / PlayResY` but should actually be scaled as `frame_width / PlayResX`:
* [explicit vector drawing coordinates](https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4/src/subtitles/RTS.cpp#L893)
* [`\fsp`](https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4/src/subtitles/RTS.cpp#L2916)
* [`\xbord`](https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4/src/subtitles/RTS.cpp#L2918)
* [`\xshad`](https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4/src/subtitles/RTS.cpp#L2920)

possibly more; whenever anyone starts work on fixing this, audit by searching RTS.cpp for `m_scalex`.

[...]

PR #395, opened 2020-05-16
Border collision and \an456 stacking

### LINKED ISSUE #143: \an4/5/6 collision stacks the wrong way.
Pretty much what it says. I don't know why you'd want to have collision with \an5, but it seems like a simple enough thing it should probably be fixed.

[Script](https://gist.github.com/torque/9343c789813f90ce09be).

Photo evidence taken from xy-vsfilter shipped with Aegisub 3.2.1 and libass 26ccc19 on OS X.

libass:
libass

vsfilter:
vsfilter

### LINKED ISSUE #304: Collision detection doesn’t take borders into account
Originally reported in #154. I’ve also seen this myself.

The amount of space between two implicitly positioned lines should include the appropriate border widths of both lines.

TheOneric:
This pr changes implicit positioning to take Borders into account (like *VSFilter) and also changes the shift direction for `\an4,5,6` lines to match *VSFilter.
closes #304 
closes #143 

~~This pr is as of now not yet ready for merge.
However I'd would be niceto get some feedback if the chosen approach seems sensible.~~

--snip-- *(Moved the originally present/unidentified issues down as they have been adressed)*

As I couldn't locate the relevant part of VSFilter code, I relied on the info from #154 and test ASS files. VSFilter seems to only care about the maximal x-border, even if it's far from the edge of a line.

To test this pr I used these two ASS-Files
[Bord-Coll-ASS.zip](https://github.com/libass/libass/files/4638387/Bord-Coll-ASS.zip)
and
[More-ASS-files.zip](https://github.com/libass/libass/files/4641184/More-ASS-files.zip)

, rendered on an 640×480 video: 
[test-video.zip](https://github.com/libass/libass/files/4638382/test-video.zip)

**Still needs to be addressed:**

- [x] ~~1px VSF-libass difference in the `bordy-var` sample for small borders.~~
- [x] Evaluate possible side effects of changing `event_image` size
      or if it is better to only change the `Segment`
- [x] If no one objects, this patch will also change to `BorderStyle=4`, update the docs
    in the wiki https://github.com/libass/libass/wiki/Libass'-ASS-Extensions#borderstyle4 *(made preliminary changes to doc)*

---

**Problems with patch v1:** (have been adressed)
While testing I found some issues that don't seem to be caused by this pr; thoughts on what t do with those would also be appreciated.

1) ~~x-VSF moves colliding \an6 down; libass up (`bordx-coll2.ass`)~~ ~~#143~~ **[Patch included]**

*Only with Border-Style=3:  (related to: https://code.google.com/archive/p/libass/wikis/IssuesAndDifferences.wiki ? )*

2) ~~x-VSF moves \an4 1px down if there's any \bord30 compared to only \bord1; libass does not (`bordx-coll2.ass`).  Idk why VSFilter does this. *(EDIT: with different values a shift >1px can be achieved)*~~ **#396 [xy-VSFilter bug]**
3) ~~In xy-VSF there's sometimes a thin 1px line visible between stacked bordstyl3 (`bordy-coll.ass`),
   but vertical text distance is identical to libass. Might not be positioning issue, but with how border-style3 is drawn~~ *[confirmed to also happen without patch and **out of scope**]*

`2.` and `3.` do not occur if using Borderstyle=1.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1493,6 +1493,15 @@ size_t ass_bitmap_construct(void *key, void *value, void *priv)
     return sizeof(BitmapHashKey) + sizeof(Bitmap) + bitmap_size(bm);
 }
 
+static void measure_text_on_eol(ASS_Renderer *render_priv, double scale, int cur_line,
+                                int max_asc, int max_desc)
+{
+    render_priv->text_info.lines[cur_line].asc  = scale * max_asc;
+    render_priv->text_info.lines[cur_line].desc = scale * max_desc;
+    render_priv->text_info.height += scale * max_asc + scale * max_desc;
+}
+
+
 /**
  * This function goes through text_info and calculates text parameters.
  * The following text_info fields are filled:
@@ -1511,9 +1520,7 @@ static void measure_text(ASS_Renderer *render_priv)
     int max_asc = 0, max_desc = 0;
     for (int i = 0; i < text_info->length; i++) {
         if (text_info->glyphs[i].linebreak) {
-            text_info->lines[cur_line].asc  = scale * max_asc;
-            text_info->lines[cur_line].desc = scale * max_desc;
-            text_info->height += scale * max_asc + scale * max_desc;
+            measure_text_on_eol(render_priv, scale, cur_line, max_asc, max_desc);
             max_asc = max_desc = 0;
             scale = 0.5 / 64;
             cur_line++;
@@ -1525,9 +1532,7 @@ static void measure_text(ASS_Renderer *render_priv)
             scale = 1.0 / 64;
     }
     assert(cur_line == text_info->n_lines - 1);
-    text_info->lines[cur_line].asc  = scale * max_asc;
-    text_info->lines[cur_line].desc = scale * max_desc;
-    text_info->height += scale * max_asc + scale * max_desc;
+    measure_text_on_eol(render_priv, scale, cur_line, max_asc, max_desc);
     text_info->height += cur_line * render_priv->settings.line_spacing;
 }
 
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1494,11 +1494,16 @@ size_t ass_bitmap_construct(void *key, void *value, void *priv)
 }
 
 static void measure_text_on_eol(ASS_Renderer *render_priv, double scale, int cur_line,
-                                int max_asc, int max_desc)
+                                int max_asc, int max_desc, double max_border_y)
 {
     render_priv->text_info.lines[cur_line].asc  = scale * max_asc;
     render_priv->text_info.lines[cur_line].desc = scale * max_desc;
     render_priv->text_info.height += scale * max_asc + scale * max_desc;
+    // For *VSFilter compatibility do biased rounding on max_border*
+    // https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4@%7B2020-05-17%7D/src/subtitles/RTS.cpp#L1465
+    render_priv->text_info.border_bottom = (int) (render_priv->border_scale * max_border_y + 0.5);
+    if (cur_line == 0)
+        render_priv->text_info.border_top = render_priv->text_info.border_bottom;
 }
 
 
@@ -1506,7 +1511,9 @@ static void measure_text_on_eol(ASS_Renderer *render_priv, double scale, int cur
  * This function goes through text_info and calculates text parameters.
  * The following text_info fields are filled:
  *   height
- *   lines[].height
+ *   border_top
+ *   border_bottom
+ *   border_x
  *   lines[].asc
  *   lines[].desc
  */
@@ -1518,22 +1525,29 @@ static void measure_text(ASS_Renderer *render_priv)
     int cur_line = 0;
     double scale = 0.5 / 64;
     int max_asc = 0, max_desc = 0;
+    double max_border_y = 0, max_border_x = 0;
     for (int i = 0; i < text_info->length; i++) {
         if (text_info->glyphs[i].linebreak) {
-            measure_text_on_eol(render_priv, scale, cur_line, max_asc, max_desc);
+            measure_text_on_eol(render_priv, scale, cur_line, max_asc, max_desc, max_border_y);
             max_asc = max_desc = 0;
+            max_border_y = 0;
             scale = 0.5 / 64;
             cur_line++;
         }
         GlyphInfo *cur = text_info->glyphs + i;
         max_asc  = FFMAX(max_asc,  cur->asc);
         max_desc = FFMAX(max_desc, cur->desc);
+        max_border_y = FFMAX(max_border_y, cur->border_y);
+        max_border_x = FFMAX(max_border_x, cur->border_x);
         if (cur->symbol != '\n' && cur->symbol != 0)
             scale = 1.0 / 64;
     }
     assert(cur_line == text_info->n_lines - 1);
-    measure_text_on_eol(render_priv, scale, cur_line, max_asc, max_desc);
+    measure_text_on_eol(render_priv, scale, cur_line, max_asc, max_desc, max_border_y);
     text_info->height += cur_line * render_priv->settings.line_spacing;
+    // VSF takes max \bordx into account for collision, even if far from edge
+    text_info->border_x =
+        (int) (render_priv->border_scale * max_border_x + 0.5);
 }
 
 /**
@@ -2752,12 +2766,16 @@ ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
     render_and_combine_glyphs(render_priv, device_x, device_y);
 
     memset(event_images, 0, sizeof(*event_images));
-    event_images->top = device_y - text_info->lines[0].asc;
-    event_images->height = text_info->height;
+    // VSFilter does *not* shift lines with a border > margin to be within the
+    // frame, so negative values for top and left may occur
+    event_images->top = device_y - text_info->lines[0].asc - text_info->border_top;
+    event_images->height =
+        text_info->height + text_info->border_bottom + text_info->border_top;
     event_images->left =
-        (device_x + bbox.x_min) * render_priv->font_scale_x + 0.5;
+        (device_x + bbox.x_min) * render_priv->font_scale_x - text_info->border_x + 0.5;
     event_images->width =
-        (bbox.x_max - bbox.x_min) * render_priv->font_scale_x + 0.5;
+        (bbox.x_max - bbox.x_min) * render_priv->font_scale_x
+        + 2 * text_info->border_x + 0.5;
     event_images->detect_collisions = render_priv->state.detect_collisions;
     event_images->shift_direction = (valign == VALIGN_SUB) ? -1 : 1;
     event_images->event = event;
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2754,7 +2754,7 @@ ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
     event_images->width =
         (bbox.x_max - bbox.x_min) * render_priv->font_scale_x + 0.5;
     event_images->detect_collisions = render_priv->state.detect_collisions;
-    event_images->shift_direction = (valign == VALIGN_TOP) ? 1 : -1;
+    event_images->shift_direction = (valign == VALIGN_SUB) ? -1 : 1;
     event_images->event = event;
     event_images->imgs = render_text(render_priv);
 
--- a/libass/ass_render.h
+++ b/libass/ass_render.h
@@ -191,6 +191,9 @@ typedef struct {
     CombinedBitmapInfo *combined_bitmaps;
     unsigned n_bitmaps;
     double height;
+    int border_top;
+    int border_bottom;
+    int border_x;
     int max_glyphs;
     int max_lines;
     unsigned max_bitmaps;


TheOneric:
- a patch for the stacking direction is now also included
- The 1px line[¹] between stacked borderstyle-3 lines is probably out of scope for this pr.

---
¹: *(to clarify: there's still some border drawn, but it's translucent enough to be visible as a line on the blue background)*

astiob:
For reference, here’s the VSFilter code handling collisions:

* [MPC-HC](https://github.com/clsid2/mpc-hc/blob/develop@{2020-05-17}/src/Subtitles/RTS.cpp#L1520)
* [xy-VSFilter](https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4@{2020-05-17}/src/subtitles/RTS.cpp#L1605)

Neither appears to have had any changes (besides whitespace edits) since good old guliverkli (VSFilter 2.37 and possibly older), although changes in other code might have had a visible effect too.

TheOneric:
Currently, the only non-borderstyle=3 differnce to VSFilter I found, is that if small borders are involved VSFilter produces a 1px larger offset to some events:

See the distance between the 'N' in the first line of the middle-event and the 'e' in the second line of the event on top (also Xy-VSF border seems more *"wobbly"*, but that's not related to the patch):
xy-VSFilter:
grafik

libass-with-patch:
grafik

That's the file `bordy-var-coll.ass`, I'll add it and `frameedge_bs{1,3}.ass` to the OP.

TheOneric:
> For reference, here’s the VSFilter code handling collisions:
> 
>     * [MPC-HC](https://github.com/clsid2/mpc-hc/blob/develop@%7B2020-05-17%7D/src/Subtitles/RTS.cpp#L1520)
> 
>     * [xy-VSFilter](https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4@%7B2020-05-17%7D/src/subtitles/RTS.cpp#L1605)

Thanks! If I read it correctly, it pretty much confirms my findings:
 - The max top and max bottom y-border are handled separately
 - The max x border is taken into account regardless of placing

On top of that:
 - VSFilter tracks the max_borders as integers (biased rounding) instead of double.

I updated the patch accordiingly. This did however not resolve the 1px discrepancy.

astiob:
Side note: the “more wobbly” borders are caused by a discrepancy between the video aspect ratio and the script aspect ratio. I have added this to #171.

libass/ass_render.h
```diff
@@ -191,6 +191,9 @@ typedef struct {
     CombinedBitmapInfo *combined_bitmaps;
     unsigned n_bitmaps;
     double height;
+    int border_top;
+    int border_bot;
```
rcombs, line 195: I'd use "bottom" here; "bot" isn't exactly a common abbreviation for this.

libass/ass_render.c
```diff
@@ -1495,25 +1497,45 @@ static void measure_text(ASS_Renderer *render_priv)
...
+            // https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4@%7B2020-05-17%7D/src/subtitles/RTS.cpp#L1465
+            text_info->border_bot =
+                (int) (render_priv->border_scale * max_border_y + 0.5);
+            if (cur_line == 0)
+                text_info->border_top = text_info->border_bot;
         }
     }
     assert(cur_line == text_info->n_lines - 1);
     text_info->height += cur_line * render_priv->settings.line_spacing;
+    // VSFilter does not care if the max x-border is actuallly close to an edge
```
rcombs, line 1529: `actuallly`

libass/ass_render.c
```diff
@@ -1495,25 +1497,45 @@ static void measure_text(ASS_Renderer *render_priv)
...
+        max_border_y = FFMAX(max_border_y, cur->border_y);
+        max_border_x = FFMAX(max_border_x, cur->border_x);
         if (cur->symbol != '\n' && cur->symbol != 0)
             scale = 1.0 / 64;
         if (i == text_info->length - 1 || text_info->glyphs[i + 1].linebreak) {
             text_info->lines[cur_line].asc  = scale * max_asc;
             text_info->lines[cur_line].desc = scale * max_desc;
             text_info->height += scale * max_asc + scale * max_desc;
+            // For *VSFilter compatibility do biased rounding on max_border*
+            // https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4@%7B2020-05-17%7D/src/subtitles/RTS.cpp#L1465
```
rcombs, line 1520: What's going on with the conditional in this?

libass/ass_render.h
TheOneric: Done.

libass/ass_render.c
TheOneric: I've reworded the whole comment and shortened it to one line

libass/ass_render.c
TheOneric: xy-VSFilter does not have a local `max_bord_x`. In this patch the if part is taken care of with FFMAX and `max_bord_x`and so the rounding and casting only happens once without condition.
I'm pretty sure    `max_i  { round(b_i) }  = round( max_i {b_i} )`.

libass/ass_render.c
```diff
@@ -1509,11 +1506,13 @@ static void measure_text(ASS_Renderer *render_priv)
         max_desc = FFMAX(max_desc, cur->desc);
         if (cur->symbol != '\n' && cur->symbol != 0)
             scale = 1.0 / 64;
+        if (i == text_info->length - 1 || text_info->glyphs[i + 1].linebreak) {
```
MrSmile, line 1515: That's avoiding a code duplication at the cost of a logic complication. I think it's better to move common code into a separate function instead.

libass/ass_render.c
```diff
@@ -1495,25 +1497,43 @@ static void measure_text(ASS_Renderer *render_priv)
...
         if (cur->symbol != '\n' && cur->symbol != 0)
             scale = 1.0 / 64;
         if (i == text_info->length - 1 || text_info->glyphs[i + 1].linebreak) {
             text_info->lines[cur_line].asc  = scale * max_asc;
             text_info->lines[cur_line].desc = scale * max_desc;
             text_info->height += scale * max_asc + scale * max_desc;
+            // For *VSFilter compatibility do biased rounding on max_border*
+            // https://github.com/Cyberbeing/xy-VSFilter/blob/xy_sub_filter_rc4@%7B2020-05-17%7D/src/subtitles/RTS.cpp#L1465
+            text_info->border_bottom =
+                (int) (render_priv->border_scale * max_border_y + 0.5);
```
MrSmile, line 1522: C99 have `lrint()` for such rounding.

astiob: I think this is rather `lround` (when the argument is nonnegative, which I assume it is).

Also note that we have many more instances of this that could use the C99 functions but don’t. But sure, we can aim to use them in new code and maybe convert old code eventually.

MrSmile: `lrint` usually converts to more optimized assembly than `lround` and in practice there is no difference between results.

astiob: What do you mean “there is no difference in results”? They do different things. As the comment explicitly says here, we want `lround`.

MrSmile: They do the same in default rounding mode except some edge cases.

astiob: The “edge cases” you speak of are precisely the entire point. They round 1.5 an 2.5 differently.

TheOneric: `grep`ping for `'+ 0.5'`, most instances of rounding like this are already touched by this commit and could be changed to `lround`, assuming it is safe to assume `>=0` in the respective case.
In case of the `max_border_…`s `>=0` is guaranteed to be true.  
In case of `event_image->left` this is **not** guaranteed  
*(but I don't know if `+0.5` or `lround` would match VSFilter; if it's not `lround/lrint` this would create a cosmetic local inconsistency with `width` and `height` using `lround/lrint` and `left` using `+0.5`, but feel free to disregard this)*

`lround` is currently not being used in libass. 
`lrint` is used in `ass_blur.c` and `ass_outline.c`.

The behaviour of lrint depends on `fe{s,g}etround()` though. libass doesn't seem to set it to any particular value and just uses the machines default. I haven't seen a mention of a standard default mode, but it's possible I missed that.
Also when setting the mode; quoting C99-draft:
> [each of the rounding mode macros] is defined if and only if the implementation supports getting and setting the represented
rounding direction by means of the fegetround and fesetround functions.
[note: and `FE_TONEAREST` seems to do **un**biased rounding]

*(POSIX probably guarantees some to exist, haven't checked though)*

What assumptions does libass make about the target machine besides supporting C99 btw ?

MrSmile: Nobody changes rounding mode in 2020. It's always safe to assume nearest rounding mode.

TheOneric:
To give more people the chance to see this change and raise objections:  

As discussed on IRC, this patch, in its current state, would change the size of the `BorderStyle=4` box (a libass extension) from `\shad` to `\bord+\shad`.
There currently is no documentation on what `BS=4` is exactly supposed to do.  
As this extensions primary use case seems to be style overrides at playback time, or in some cases in static files to raise the readability of the text (while avoiding undesirable overlaps from BorderStyle=3) but **not** complex typesetting, this change actually seems kind of reasonable and like something that might actually be *desired* by its users as it avoids the border spilling over the BS=4 box.  
`\shad` would no longer control the box padding as measured **from the text**, but the padding from the **border of the text**.

A precedent for changing `BS=4`'s behaviour would be libass 0.13.7, where controlling the size of the box with `\shad` was introduced. Before the box tight fitted the text (without taking the border or shadow into account).

If there are objections to changing `BS=4`, it wouldn't be much trouble keeping the current behaviour. 

@wm4 : You authored the commit introducing `BS=4` and mpv seems to use it. What's your opinion on the proposed `BS=4` change ?

libass/ass_render.c
astiob: I agree with both of you here. 😅

On the one hand, it’s true that technically `lrint` actually isn’t … portable? fully defined? reliable? whatever you want to call it, if we don’t set or check the rounding mode. But on the other hand, here, indeed, I think we can say that in practice it’s always round-to-nearest-and-half-to-even.

At most, it might not be on some relatively exotic hardware platforms, but I can’t find anything with a quick Google search. One source also says, “The IEEE standard requires that the default rounding mode be round to nearest.” and I think we kinda already assume IEEE floating point (in `ass_strtod` in particular), although, of course, whatever software is calling libass could change the rounding mode before invoking our functions.

I think it’s safe to assume `lrint` and primitive floating-point arithmetic operations do half-to-even while `lround` does half-away-from-zero. But I do think the rounding mode is a big deal, and substituting `lrint` for `lround` is wrong.

As for more efficient machine code (even though saving a couple assembler instructions per character while knowingly sacrificing correctness doesn’t seem to make sense to me anyway), without `-ffast-math`, both `lround` and `lrint` [are compiled](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:___c,source:'%23include+%3Cstdio.h%3E%0A%23include+%3Cmath.h%3E%0A%0Atypedef+union+%7B%0A%09long+long+i%3B%0A%09double+f%3B%0A%7D+foo%3B%0A%0A__attribute__((noinline))%0Avoid+func(double*+f,+long+long*+i)+%7B%0A++++double+a+%3D+*f%3B%0A++++*i+%3D+4%3B%0A++++double+b+%3D+*f%3B%0A++++printf(%22%25f+%25f%5Cn%22,+a,+b)%3B%0A%7D%0A%0Avoid+func_caller()+%7B%0A++++foo+f%3B%0A++++func(%26f.f,+%26f.i)%3B%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cclang600,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',trim:'1'),lang:___c,libs:!(),options:'-O3',source:1),l:'5',n:'0',o:'x86-64+clang+6.0.0+(Editor+%231,+Compiler+%231)+C',t:'0')),header:(),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) into actual library function calls. Even with `-ffast-math`, [only GCC](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:___c,source:'%23include+%3Cstdio.h%3E%0A%23include+%3Cmath.h%3E%0A%0Atypedef+union+%7B%0A%09long+long+i%3B%0A%09double+f%3B%0A%7D+foo%3B%0A%0A__attribute__((noinline))%0Avoid+func(double*+f,+long+long*+i)+%7B%0A++++double+a+%3D+*f%3B%0A++++*i+%3D+4%3B%0A++++double+b+%3D+*f%3B%0A++++printf(%22%25f+%25f%5Cn%22,+a,+b)%3B%0A%7D%0A%0Avoid+func_caller()+%7B%0A++++foo+f%3B%0A++++func(%26f.f,+%26f.i)%3B%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cclang600,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',trim:'1'),lang:___c,libs:!(),options:'-O3',source:1),l:'5',n:'0',o:'x86-64+clang+6.0.0+(Editor+%231,+Compiler+%231)+C',t:'0')),header:(),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) (not Clang, not Intel, [not MSVC](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:___c,source:'%23include+%3Cstdio.h%3E%0A%23include+%3Cmath.h%3E%0A%0Atypedef+union+%7B%0A%09long+long+i%3B%0A%09double+f%3B%0A%7D+foo%3B%0A%0A__attribute__((noinline))%0Avoid+func(double*+f,+long+long*+i)+%7B%0A++++double+a+%3D+*f%3B%0A++++*i+%3D+4%3B%0A++++double+b+%3D+*f%3B%0A++++printf(%22%25f+%25f%5Cn%22,+a,+b)%3B%0A%7D%0A%0Avoid+func_caller()+%7B%0A++++foo+f%3B%0A++++func(%26f.f,+%26f.i)%3B%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cclang600,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',trim:'1'),lang:___c,libs:!(),options:'-O3',source:1),l:'5',n:'0',o:'x86-64+clang+6.0.0+(Editor+%231,+Compiler+%231)+C',t:'0')),header:(),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)) substitutes builtins and even then `lrint` is only one instruction shorter than `(int) (x + 0.5)`, although `lround` is indeed considerably longer.

So if speed is (somehow) important here, the current code, `(int) (x + 0.5)`, is actually best.

And if we need the `+ 0.5` in a case where numbers can be negative and in the next line we have numbers that can’t be negative, it would be more consistent/easier to read to use `(int) (x + 0.5)` in both cases. I remember explicitly using `+ 0.5` with negative numbers in… parsing, maybe? for VSFilter compatibility.

> What assumptions does libass make about the target machine besides supporting C99 btw ?

We definitely assume `int` holds 32 bits, and possibly that it’s two’s complement. We probably assume some amount of IEEE float compatibility (especially in `ass_strtod`). We assume obscure things about integer types like: the sign bit of a signed type isn’t a padding bit in the corresponding unsigned type. There was some talk of this in #261 when I tried to make code handle various odd cases allowed by the standard and was shot down.

astiob: Sorry, my Godbolt links were broken. Check the updated comment for the right ones.

libass/ass_render.c
TheOneric: Done.

libass/ass_render.c
astiob: And speaking of edge cases, `(int) (x + 0.5)` differs from `lround(x)` (_and_ `lrint(x)`)
* if `x` is the largest representable number smaller than 0.5, namely 0.49999999999999994 (the naïve code rounds to 1, while `lround` to 0),
* or if `x` is any odd 53-bit integer (the naïve code adds 1, while `lround` preserves the same value).

Admittedly, the 53-bit integer case shouldn’t be very relevant to us given that we cast the whole thing to a 32-bit `int`…

All in all, I’d prefer us to stick to exactly the same code that VSFilter uses where it matters, and to use `lrint` where we want actually good rounding.

libass/ass_render.c
```diff
@@ -1479,6 +1479,17 @@ size_t ass_bitmap_construct(void *key, void *value, void *priv)
     return sizeof(BitmapHashKey) + sizeof(Bitmap) + bitmap_size(bm);
 }
 
+static void measure_text_on_eol(
+        const ASS_Renderer *const render_priv, TextInfo *const text_info,
```
MrSmile, line 1483: `render_priv` is unused here, should be removed. Also I don't see any sense in that `const`s everywhere (all except the first). It prohibits variable changes inside a function but doesn't have any effect outside, only complicates a function prototype. And functions long enough to benefit from such a protection should be split to several simpler functions instead.

It's possible to further reduce prototype by eliminating `scale` at the cost of longer invocation, but I'm not sure which variant would be better:
```c
measure_text_on_eol(text_info, cur_line, scale * max_asc, scale * max_desc);
```

libass/ass_render.c
```diff
@@ -2730,12 +2750,17 @@ ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
...
-    event_images->height = text_info->height;
+    // VSFilter does *not* shift lines with a border > margin to be within the
+    // frame, so negative values for top and left may occur
+    event_images->top = device_y - text_info->lines[0].asc - text_info->border_top;
+    event_images->height =
+        text_info->height + text_info->border_bottom + text_info->border_top;
     event_images->left =
-        (device_x + bbox.x_min * render_priv->font_scale_x) + 0.5;
+        (device_x + bbox.x_min * render_priv->font_scale_x) + 0.5
+        - text_info->border_x;
```
MrSmile, line 2760: It's perfectly fine to write that in the same line, nobody works with 80-symbol terminals nowadays. I think that line widths up to ~120 are fine (you already have line with width nearly 120 in that PR anyway).

TheOneric:
The observed 1px difference *""on small border""* seems to have been a subtle difference in how some characters are rasterized. After testing multiple other examples and remeasuring the differences in the original sample using different characters as reference points, there doesn't seem to be a general shift in the lines.

libass/ass_render.c
TheOneric: `render_priv` is later needed for `render_priv->border_scale`.
But `text_info` can be accesed via render_priv, so only one of them is necessary, my bad.

Personally I'd prefer to keep `scale` as an argument, from a style perspective (though I don't mind changing that if requested).  
While pre-multiplying would be a bit more time-efficient (assuming the compiler doesn't optimize it anyway).

libass/ass_render.c
TheOneric: Done.
*(The previous ~115 charater longer line was a comment with an URL Splitting this would imo not have been a good idea)*

libass/ass_render.c
MrSmile: If it's needed only later then that argument should be added later too. It also can be premultiplied into `max_border_y` to eliminate that dependency altogether, but merging `text_info` and `render_priv` is good too.

TheOneric: iinm you'd prefer the following version ?

```C
static void measure_text_set_maxima(TextInfo *text_info, int cur_line,
                                double max_asc_scaled, double max_desc_scaled,
                                double max_bord_y_scaled)
{
    text_info->lines[cur_line].asc  = max_asc_scaled;
    text_info->lines[cur_line].desc = max_desc_scaled;
    text_info->height += max_asc_scaled + max_desc_scaled;
    // For *VSFilter compatibility do biased rounding on max_border*
    text_info->border_bottom = (int) (max_border_y_scaled + 0.5);
    if (cur_line == 0)
        text_info->border_top = text_info->border_bottom;
}
```

Personally, I find the current version cleaner, as the whole logic to scale, cast and update the text_info members correctly is contained in the function. Imho it's also more readable despite the longer function prototype.  *(especially if the rounding and cast of the vertical border was also moved outside the function)*

Ofc your opinion may differ and I'd change it if your not happy with the current version.

MrSmile: This version is cleaner conceptually (less context has leaked from outside), your version is simpler (less duplication). So it's about the same for me, I don't have clear preference here.

astiob:
For reference, I saw some curious incompatible behaviour with \an5 in my tests in the last few days. In at least one instance, VSFilter reordered several \an5 lines such that the first line was smack in the middle of the screen but subsequent lines were placed _both_ above and below it.

TheOneric:
@astiob Do you have more information about when this happens? Just [straightforward stacking](https://github.com/libass/libass/files/5453295/an5.ass.txt) many \an5 lines gives me a persistent downward shift in xy-VSF:
grafik

astiob:
Here’s the sample I saw it on, slightly simplified:

```
WrapStyle: 0
PlayResX: 1280
PlayResY: 720

Style: Default,Arial,96,&H000000FF,&H80FFFF00,&H00000000,&H0000FF00,0,0,0,0,100,100,0,0,1,0,0,2,0,10,20,1

Dialogue: 0,0:00:00.00,0:00:10.00,Default,,0,0,0,,{\an5}अा
Dialogue: 0,0:00:00.00,0:00:10.00,Default,,0,0,0,,{\an5}̀
Dialogue: 0,0:00:00.00,0:00:10.00,Default,,0,0,0,,{\an5}ربية
```

I think the Devanagari gets a substitute font but the accent and the Arabic both use Arial.

Here’s what I see in VSFilter (both MPC-HC and xy-VSFilter) on a 1280×720 pure-green video:

VSFilter screenshot

The Devanagari line seems to be in the middle of the screen, the accent is _above_ it and the Arabic is _below_ it.

JerryKit:
> ****
> 
> 正如在IRC上讨论的那样，此修补程序在其当前状态下会将`BorderStyle=4`框的大小（libass扩展名）从`\shad`更改为`\bord+\shad`。

Hello, can I set the width of the border separately when borderstyle=4? I tried shadow, but it seems to change along with the height

TheOneric:
On Sun, Mar 28, 2021 at 00:03:17 -0700, JerryKit wrote:
> Hello, can I set the width of the border separately when borderstyle=4? I tried shadow, but it seems to change along with the height

Not sure, what you want to do. If it's about changing width and height 
seperately use `\xbord, \ybord` and `\xshad, \yshad`.
If you really want to change `\shad` independent from `\bord` for 
BorderStyle=4,  open a new issue  for discussion, but note that libass' 
VSFilter incompatible extensions have less stable behaviour then standard 
ASS and are *not* intended to be used in complex typesetting.

JerryKit:
> 在2021年3月28日，星期日，00：03：17 -0700，JerryKit写道：您好，当borderstyle = 4时，我可以分别设置边框的宽度吗？我尝试过阴影，但是似乎随着高度的变化而变化
> 不确定，您想做什么。如果要更改宽度和高度，请分别使用`\ xbord，\ ybord`和`\ xshad，\ yshad`。如果您真的想独立于BorderType = 4来更改\\\\\\\\\\\\\\\\\\\\\\\\\\\\\ shad`的讨论，但请注意，libass的VSFilter不兼容扩展的行为不如标准ASS稳定，并且*并非*用于复杂的排版。

Thanks for your reply, I have solved the problem through ybord.

Issue #320, opened 2018-11-05
Replace call to deprecated function hb_font_funcs_set_glyph_func


rcombs:
This is apparently replaced by `hb_font_funcs_set_nominal_glyph_func` and `hb_font_funcs_set_variation_glyph_func`, but I'm not sufficiently familiar with the relevant code to know how they should be used differently.

PR #398, opened 2020-05-27
ass_shaper: fix harfbuzz deprecation warning

rcombs:
tbh this duplication feels a little excessive but it's what hb wants out of us so 🤷‍♀️


TheOneric:
Closes #320

Issue #42, opened 2014-02-09
Carnival Phantasm positioning issue


rcombs:
libass: 
![libass](http://puu.sh/6P3gp.png)
xy-vs:
![xy-vs](https://dl.dropboxusercontent.com/s/vner39a2py41teh/%5BUTW%5D_Carnival_Phantasm_-_08_%5BBD%5D%5Bh264-1080p_FLAC%5D%5B9AD4509C%5D_001_5719.png?dl=1&token_hash=AAEW5ty9Hxix7iW63hdEB3FViUHpsGRsX7icHpf9EO6xMw)
File is 9AD4509C

```
Dialogue: 0,0:03:58.53,0:04:02.02,Sign-SaberWork,,0,0,0,,{\fad(0,745)\frz2.5\fax0.05\fscx81\fscy80\1a&H00&\4c&HFFFEFF\3a&H00&\bord4\fsp15\b1\blur1\c&HEFB94C&\pos(325,767)\3c&HEFB94C&}Saber
Dialogue: 1,0:03:58.53,0:04:02.32,Sign-SaberWork2,,0,0,0,,{\fad(0,745)\frz2.5\fax0.05\fscx96\fscy92\1a&H00&\b1\blur0.5\bord6\pos(444,835)\c&HFFFFFF&\3c&HEFB94C&}Saber{\alpha&HFF&}\h\h\h\h\h\h\h\N{\alpha&H00&}at Work
Dialogue: 2,0:03:58.53,0:04:02.32,Sign-SaberWork2,,0,0,0,,{\fad(0,745)\frz2.5\fax0.025\fscx94\fscy91\1a&HFF&\b1\blur1\bord2\pos(444,835)\c&HFFFFFF&\3c&HFDE882&}Sa{\alpha&H2F&\1a&HFF&}be{\alpha&H50&\1a&HFF&}r{\alpha&HFF&}\h\h\h\h\h\h\h\N{\alpha&H00&\1a&HFF&}at {\alpha&H1F&\1a&HFF&}Wo{\alpha&H40&\1a&HFF&\bord0.5\3c&HFEF7C2&}rk
```

rcombs:
This looks to be fixed by @astiob's 2014 commit, or at least the sign looks pretty sensible now and I vaguely recall it being a fine mess:
![image](https://user-images.githubusercontent.com/819638/78308939-68b6ef80-750f-11ea-9f93-e83a1d17f241.png)

astiob:
[...]

Issue #2, opened 2013-11-09
\h rendered as "character not found" box in certain fonts without HarfBuzz


rcombs:
![Bug image](https://dl.dropboxusercontent.com/u/46348978/Plex/Photo%20Nov%2006%2C%2021%2053%2008.png)
Looks like a missing nbsp character to me, but it should probably be replaced with a regular space.
Latest git master. Harfbuzz is disabled. No shaping hints set.

PR #399, opened 2020-05-27
Make harfbuzz a hard dependency

### LINKED ISSUE #199: Require HarfBuzz?
Lots of things break without HarfBuzz: Arabic (right? I think FriBidi had some sort of fallback thing but it isn’t good?), vertical Japanese, syllabic Korean and many other Asian scripts. (How do Indic scripts fare?) The kerning setting in libass also does nothing without HarfBuzz despite being exposed, and some fonts might rely on OpenType features for correct display even of European scripts.

HarfBuzz has long been very easy to build on all major platforms including Windows, Linux and OS X. HarfBuzz _does not require_ GLib, ICU, Cairo or any other external library. It optionally depends on Graphite2, and while it certainly doesn’t hurt, it isn’t required either (and Graphite2 itself is easy to build as long as you have CMake).

The only problem might be the circular dependency between HarfBuzz and FreeType, but it’s still optional on FreeType’s side and it’s not clear how much trouble it actually causes. And in the worst case, it’s easy to work around by building FreeType without HarfBuzz, then building HarfBuzz with FreeType and then rebuilding FreeType with HarfBuzz if so desired. (Note: HarfBuzz’s dependency on FreeType is usually optional, but _libass requires_ HarfBuzz to depend on FreeType.) (Note²: this means making HarfBuzz a hard dependency of libass would effectively make FreeType a hard dependency of HarfBuzz.) (Edit: when upgrading, the circular dependency actually doesn’t break any build; it just confuses HarfBuzz’s build system when it’s checking FreeType’s pkg-config file. Remove HarfBuzz from freetype2.pc, build HarfBuzz, add it back. Voilà.)

How about we make HarfBuzz a hard requirement to build libass?

If not, or as a transitional measure, how about we make HarfBuzz a “strongly recommended” dependency, so that `./configure` refuses to proceed without it unless you explicitly run it with `--disable-harfbuzz`?

rcombs:
Closes #199.
The hb-ft commit means harfbuzz no longer needs to depend on freetype to work with libass.
Also adds some missing malloc error checking.

### DIFF
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -27,7 +27,9 @@
 #include <limits.h>
 #include <stdbool.h>
 
-#include <hb-ft.h>
+#include <ft2build.h>
+#include FT_FREETYPE_H
+#include FT_TRUETYPE_TABLES_H
 enum {
     VERT = 0,
     VKNA,
@@ -249,6 +251,31 @@ size_t ass_glyph_metrics_construct(void *key, void *value, void *priv)
     return 1;
 }
 
+static hb_blob_t*
+get_reference_table(hb_face_t *hbface, hb_tag_t tag, void *font_data)
+{
+  FT_Face face = font_data;
+  FT_ULong len = 0;
+
+  if (FT_Load_Sfnt_Table(face, tag, 0, NULL, &len) != FT_Err_Ok)
+    return NULL;
+
+  char *buf = malloc(len);
+  if (!buf)
+    return NULL;
+
+  if (FT_Load_Sfnt_Table(face, tag, 0, (FT_Byte*)buf, &len) != FT_Err_Ok) {
+    free(buf);
+    return NULL;
+  }
+
+  hb_blob_t *blob = hb_blob_create(buf, len, HB_MEMORY_MODE_WRITABLE, buf, free);
+  if (!blob)
+      free(buf);
+
+  return blob;
+}
+
 static hb_bool_t
 get_glyph_nominal(hb_font_t *font, void *font_data, hb_codepoint_t unicode,
                   hb_codepoint_t *glyph, void *user_data)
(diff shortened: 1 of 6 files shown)


libass/ass_shaper.c
```diff
@@ -613,7 +619,7 @@ shape_harfbuzz_process_run(GlyphInfo *glyphs, hb_buffer_t *buf, int offset)
  * \param glyphs glyph clusters
  * \param len number of clusters
  */
-static void shape_harfbuzz(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
+static int shape_harfbuzz(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
```
astiob, line 659: Wouldn’t it be more consistent to make this return `bool` (true on success, false on failure)? I assume you’re mirroring the return value of `ass_shaper_shape` here, so shall we change _that_ to `bool` too?

libass/ass_shaper.c
```diff
@@ -411,11 +438,21 @@ static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
     hb_fonts = font->shaper_priv->fonts;
     if (!hb_fonts[info->face_index]) {
         FT_Face face = font->faces[info->face_index];
-        hb_font_t *hb_font = hb_fonts[info->face_index] =
-            hb_ft_font_create(face, NULL);
+        hb_face_t *hb_face = hb_face_create_for_tables(get_reference_table, face, NULL);
```
khaledhosny, line 441: Note that `hb_face_create_for_tables()` never return NULL. In case of failure it will return a singleton empty `hb_face_t` (as returned by `hb_face_get_empty()`).

libass/ass_shaper.c
```diff
@@ -411,11 +438,21 @@ static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
...
+        if (!hb_face)
+            return NULL;
```
khaledhosny, line 443: Note that `hb_face_create_for_tables()` never return NULL. In case of failure it will return a singleton empty `hb_face_t` (as returned by `hb_face_get_empty()`).

libass/ass_shaper.c
```diff
@@ -411,11 +438,21 @@ static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
...
+        hb_face_set_index(hb_face, face->face_index);
+        hb_face_set_upem(hb_face, face->units_per_EM);
+
+        hb_font_t *hb_font = hb_fonts[info->face_index] = hb_font_create(hb_face);
+        hb_face_destroy(hb_face);
         if (!hb_font)
             return NULL;
```
khaledhosny, line 450: Same here, `hb_font_create()` never returns `NULL`.

astiob:
@rcombs I went ahead and edited your PR to return `bool` from `ass_shaper_shape` and `shape_harfbuzz`. Could you glance over it?

In particular, I [moved the returns inside the switch and added a `default` label](https://github.com/libass/libass/pull/399/files#diff-88b8b5ba7fa04b02b453c70be0fbcf73R948-R957) to proactively placate static analysers (for reference, we validate the shaper selection in `render_api.c`).

TheOneric:
This also closes #2 iinm.
Judging by the comments in 2 and 42, 33a3ab2 can probably be safely reverted with HarfBuzz as a hard dependency.

astiob:
Good catch, but we allow HarfBuzz to be explicitly disabled at runtime, so #2 still applies in those cases.

Meanwhile, reverting 33a3ab2c8777241b73ad85024158e19fb1f9c699 is required to fix #42 (and #2) anyway. I’d love to know what necessitated that commit in the first place, but alas.

astiob (approved):

astiob:
Is anyone testing this? 😅 IIRC we haven’t merged this yet because @rcombs wasn’t quite sure about the hb-ft changes. Shall we just merge and hope for the best or for bug reports to come in?

TheOneric:
First three commits seem fine to me and I'm not familiar enough with harfbuzz and freetype to say anything about the fourth commit.

> Is anyone testing this?

With patches applied and a freetype-less harfbuzz lib preloaded, it did pass a small set of `compare` tests. `lsof` and `ldd` seem to verify that a freetype-less harfbuzz was used. Regular video players I tried all fail due to pango trying to use `hb_ft_…` symbols.
That's only a quick and limited test though.

PR #400, opened 2020-05-30
Add checkasm and arm64/aarch64 assembly

rcombs:
This comes from x264 -> ffmpeg -> dav1d, tweaked a little at each step along the way (and now tweaked a bit here). It has a few very useful applications:
- Testing existing functions to ensure they behave as expected for randomized input. From this testing I discovered that our be_blur ASM needed 1px more padding than we were giving it.
- Doing test-driven development of new ASM routines. No more just running code on some random test files and comparing PNG output.
- Benchmarking function variants (both during and after development) to see how their performance compares. While doing this, I discovered that a few of my old AVX2 routines are actually slower than the SSE2 versions, so I've disabled them until we can see if faster AVX2 versions are possible.
- Testing multiple variants of functions without screwing around with ifdefs in the source.

Run `make check` for the new tests, and `make bench` for the benchmarks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ Makefile.in
 /test/test*
 /compare/compare*
 /profile/profile*
+/checkasm/checkasm*
 
 # pkgconfig
 /libass.pc
@@ -37,3 +38,4 @@ Makefile.in
 /ltmain.sh
 /missing
 /m4
+/test-driver
--- a/.gitignore
+++ b/.gitignore
@@ -19,9 +19,14 @@ Makefile.in
 
 # Test programs
 /test/test*
+!/test/test.c
 /compare/compare*
+!/test/compare/compare.c
 /profile/profile*
+!/profile/profile.c
 /checkasm/checkasm*
+!/checkasm/checkasm.c
+!/checkasm/checkasm.h
 
 # pkgconfig
 /libass.pc
--- a/Makefile.am
+++ b/Makefile.am
@@ -16,5 +16,8 @@ if ENABLE_PROFILE
     profile = profile
 endif
 
-SUBDIRS = libass $(test) $(compare) $(profile)
+if ASM
+    checkasm = checkasm
+endif
 
+SUBDIRS = libass $(test) $(compare) $(profile) $(checkasm)
--- a/checkasm/be_blur.c
+++ b/checkasm/be_blur.c
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2020 rcombs <rcombs@rcombs.me>
+ *
+ * This file is part of libass.
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include "checkasm.h"
+
+#include "ass_bitmap.h"
+
+#define HEIGHT 8
+#define STRIDE 64
+#define MIN_WIDTH 2
+
+static void check_be_blur(BeBlurFunc func)
+{
+    if (check_func(func, "be_blur")) {
+        ALIGN(uint8_t buf_ref[STRIDE * HEIGHT], 32);
+        ALIGN(uint8_t buf_new[STRIDE * HEIGHT], 32);
+        ALIGN(uint16_t tmp[STRIDE * 2], 32);
+        declare_func(void,
+                     uint8_t *buf, intptr_t stride,
+                     intptr_t width, intptr_t height, uint16_t *tmp);
+
+        for (int w = MIN_WIDTH; w <= STRIDE; w++) {
+            memset(buf_ref, 0, sizeof(buf_ref));
+            memset(buf_new, 0, sizeof(buf_new));
+            for (int y = 0; y < HEIGHT; y++) {
+                for (int x = 0; x < w; x++)
+                    buf_ref[y * STRIDE + x] = buf_new[y * STRIDE + x] = rnd();
+            }
+
+            memset(tmp, 0, sizeof(tmp));
+            call_ref(buf_ref, STRIDE,
+                     w, HEIGHT, tmp);
+            memset(tmp, 0, sizeof(tmp));
+            call_new(buf_new, STRIDE,
+                     w, HEIGHT, tmp);
+
+            for (int i = 0; i < HEIGHT; i++) {
+                if (memcmp(buf_ref + STRIDE * i,
+                           buf_new + STRIDE * i,
+                           w)) {
+                    fail();
+                    goto fail;
+                }
+            }
+        }
+
+        fail:
+        bench_new(buf_new, STRIDE, STRIDE, HEIGHT, tmp);
+    }
+
+    report("be_blur");
+}
+
+void checkasm_check_be_blur(const BitmapEngine *engine)
+{
+    check_be_blur(engine->be_blur);
+}
--- a/checkasm/checkasm.c
+++ b/checkasm/checkasm.c
@@ -1,0 +1,797 @@
+/*
+ * Copyright © 2018, VideoLAN and dav1d authors
+ * Copyright © 2018, Two Orioles, LLC
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#include "tests/checkasm/checkasm.h"
+
+#include <math.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <string.h>
+
+#include "src/cpu.h"
+
+#ifdef _WIN32
+#include <windows.h>
+#define COLOR_RED    FOREGROUND_RED
+#define COLOR_GREEN  FOREGROUND_GREEN
+#define COLOR_YELLOW (FOREGROUND_RED|FOREGROUND_GREEN)
+
+static unsigned get_seed(void) {
+    return GetTickCount();
+}
+#else
+#include <unistd.h>
+#include <signal.h>
+#include <time.h>
+#ifdef __APPLE__
+#include <mach/mach_time.h>
+#endif
+#define COLOR_RED    1
+#define COLOR_GREEN  2
+#define COLOR_YELLOW 3
+
+static unsigned get_seed(void) {
+#ifdef __APPLE__
+    return (unsigned) mach_absolute_time();
+#elif defined(HAVE_CLOCK_GETTIME)
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    return (unsigned) (1000000000ULL * ts.tv_sec + ts.tv_nsec);
+#endif
+}
+#endif
+
+/* List of tests to invoke */
+static const struct {
+    const char *name;
+    void (*func)(void);
+} tests[] = {
+    { "msac", checkasm_check_msac },
+#if CONFIG_8BPC
+    { "cdef_8bpc", checkasm_check_cdef_8bpc },
+    { "filmgrain_8bpc", checkasm_check_filmgrain_8bpc },
+    { "ipred_8bpc", checkasm_check_ipred_8bpc },
+    { "itx_8bpc", checkasm_check_itx_8bpc },
+    { "loopfilter_8bpc", checkasm_check_loopfilter_8bpc },
+    { "looprestoration_8bpc", checkasm_check_looprestoration_8bpc },
+    { "mc_8bpc", checkasm_check_mc_8bpc },
+#endif
+#if CONFIG_16BPC
+    { "cdef_16bpc", checkasm_check_cdef_16bpc },
+    { "filmgrain_16bpc", checkasm_check_filmgrain_16bpc },
+    { "ipred_16bpc", checkasm_check_ipred_16bpc },
+    { "itx_16bpc", checkasm_check_itx_16bpc },
+    { "loopfilter_16bpc", checkasm_check_loopfilter_16bpc },
+    { "looprestoration_16bpc", checkasm_check_looprestoration_16bpc },
+    { "mc_16bpc", checkasm_check_mc_16bpc },
+#endif
+    { 0 }
+};
+
+/* List of cpu flags to check */
+static const struct {
+    const char *name;
+    const char *suffix;
+    unsigned flag;
+} cpus[] = {
+#if ARCH_X86
+    { "SSE2",               "sse2",      DAV1D_X86_CPU_FLAG_SSE2 },
+    { "SSSE3",              "ssse3",     DAV1D_X86_CPU_FLAG_SSSE3 },
+    { "SSE4.1",             "sse4",      DAV1D_X86_CPU_FLAG_SSE41 },
+    { "AVX2",               "avx2",      DAV1D_X86_CPU_FLAG_AVX2 },
+    { "AVX-512 (Ice Lake)", "avx512icl", DAV1D_X86_CPU_FLAG_AVX512ICL },
+#elif ARCH_AARCH64 || ARCH_ARM
+    { "NEON",               "neon",      DAV1D_ARM_CPU_FLAG_NEON },
+#elif ARCH_PPC64LE
+    { "VSX",                "vsx",       DAV1D_PPC_CPU_FLAG_VSX },
+#endif
+    { 0 }
+};
+
+typedef struct CheckasmFuncVersion {
+    struct CheckasmFuncVersion *next;
+    void *func;
+    int ok;
+    unsigned cpu;
+    int iterations;
+    uint64_t cycles;
+} CheckasmFuncVersion;
+
+/* Binary search tree node */
+typedef struct CheckasmFunc {
+    struct CheckasmFunc *child[2];
+    CheckasmFuncVersion versions;
+    uint8_t color; /* 0 = red, 1 = black */
+    char name[];
+} CheckasmFunc;
+
+/* Internal state */
+static struct {
+    CheckasmFunc *funcs;
+    CheckasmFunc *current_func;
+    CheckasmFuncVersion *current_func_ver;
+    const char *current_test_name;
+    const char *bench_pattern;
+    size_t bench_pattern_len;
+    int num_checked;
+    int num_failed;
+    int nop_time;
+    unsigned cpu_flag;
+    const char *cpu_flag_name;
+    const char *test_name;
+    unsigned seed;
+    int bench_c;
+    int verbose;
+    int function_listing;
+#if ARCH_X86_64
+    void (*simd_warmup)(void);
+#endif
+} state;
+
+/* float compare support code */
+typedef union {
+    float f;
+    uint32_t i;
+} intfloat;
+
+static uint32_t xs_state[4];
+
+static void xor128_srand(unsigned seed) {
+    xs_state[0] = seed;
+    xs_state[1] = ( seed & 0xffff0000) | (~seed & 0x0000ffff);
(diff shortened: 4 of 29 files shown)


libass/ass_bitmap.c
```diff
@@ -370,13 +370,15 @@ void be_blur_post(uint8_t *buf, intptr_t w, intptr_t h, intptr_t stride)
...
-        return be;
+        return be + 1;
     if (be <= 7)
-        return 4;
-    if (be <= 123)
         return 5;
-    return FFMAX(128 - be, 0);
+    if (be <= 123)
+        return 6;
+    return FFMAX(128 - be, 0) + 2;
```
astiob, line 387: Hm. This is odd.

For one thing, you’re adding 2 in the last line but only 1 elsewhere. Also, the comment above this function should be updated to say something like, “One extra pixel is added to allow for vectorized code.”

Is the problem that the asm code is vectorized and writes to one byte past the last column/row, overrunning the buffer? But doesn’t this function affect the computed width and height of the bitmap, and therefore doesn’t this change merely cause the asm code to write _another_ byte past the last column/row, _still_ overrunning the buffer?

rcombs: It doesn't overrun, it just produces incorrect output when the input has nonzero data within 2px of the edge (don't ask me why, that comes straight from VSFilter).

checkasm/be_blur.c
```diff
@@ -0,0 +1,70 @@
...
+        declare_func(void,
+                     uint8_t *buf, intptr_t w, intptr_t h,
+                     intptr_t stride, uint16_t *tmp);
+
+        memset(buf_ref, 0, sizeof(buf_ref));
+        memset(buf_new, 0, sizeof(buf_new));
+
+        for (int y = 2; y < HEIGHT - 2; y++) {
+            for (int x = 2; x < WIDTH - 2; x++)
+                buf_ref[x + y * STRIDE] = buf_new[x + y * STRIDE] = rnd();
```
astiob, line 42: The pixel value range for \be is 0–64. I think you’re supplying 0–255 here. Could this be why you seem to see insufficient padding?

libass/ass_bitmap.c
astiob, line 387: In other news, just as my own comment above this function already says, at huge \be values, enough padding is needed for all intermediate values, so I think the last line with the `FFMAX` is wrong, and it should just return 5 (before your commit, or 6 after) in the default case. Of course, I wrote this code so it’s my own bug, if that’s right. I’m not sure what I was thinking, _especially_ because the comment clearly notes that intermediate values need large padding.

checkasm/be_blur.c
rcombs: See `ass_synth_blur`; the final iteration is on potentially-0-255 input.

astiob: Ah, right. I’d forgotten that. (Even though I probably wrote that…)

libass/ass_cpu.h
```diff
@@ -0,0 +1,37 @@
...
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifndef LIBASS_CPU_H
+#define LIBASS_CPU_H
+
+typedef enum ASS_CPUFlags {
+    ASS_CPU_FLAG_NONE = 0x00,
+#define ASS_CPU_FLAG_NONE          0x00
```
MrSmile, line 24: Why that cannot be an enum member instead of the define?

libass/ass_cpu.h
```diff
@@ -0,0 +1,37 @@
...
+#define ASS_CPU_FLAG_X86_SSE2      0x02
+#define ASS_CPU_FLAG_X86_AVX       0x04
+#define ASS_CPU_FLAG_X86_AVX2      0x08
+#define ASS_CPU_FLAG_X86_AVX512ICL 0x10
+    ASS_CPU_FLAG_MAX  = (ASS_CPU_FLAG_X86_AVX512ICL << 1) - 1,
+} ASS_CPUFlags;
+
+void ass_set_cpu_flags_mask(ASS_CPUFlags mask);
+
+ASS_CPUFlags ass_get_cpu_flags(void);
```
MrSmile, line 35: I'm personally prefer not to use an enum name for values outside of an enum scope, but feel free to ignore me here.

libass/ass_cpu.c
```diff
@@ -0,0 +1,72 @@
...
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include "config.h"
+
+#include "ass_cpu.h"
+#include "x86/cpuid.h"
+
+static ASS_CPUFlags mask = ~(ASS_CPU_FLAG_X86_AVX512ICL);
```
MrSmile, line 24: I think that variables in global scope should have longer names, at least `cpu_mask` or something.

libass/ass_bitmap.h
```diff
@@ -62,6 +63,8 @@ typedef void (*ParamFilterFunc)(int16_t *dst, const int16_t *src,
 typedef struct {
     int align_order;  // log2(alignment)
 
+    ASS_CPUFlags cpu_flags;
```
MrSmile, line 66: I think this should be the first struct member, in the same group as `align_order`, as alignment is directly derived from `cpu_flags`. Maybe it's even better to incorporate `align_order` into `cpu_flags`, i.e. `align_order = cpu_flags & ASS_CPU_ALIGN` or something.

checkasm/be_blur.c
```diff
@@ -0,0 +1,70 @@
...
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include "checkasm.h"
+
+#include "ass_bitmap.h"
+
+#define WIDTH  61
```
MrSmile, line 23: I think that width and height should be randomized too, as alignment problems are quite important.

checkasm/checkasm.c
```diff
@@ -494,18 +473,26 @@ static void check_cpu_flag(const char *const name, unsigned flag) {
     const unsigned old_cpu_flag = state.cpu_flag;
 
     flag |= old_cpu_flag;
-    dav1d_set_cpu_flags_mask(flag);
-    state.cpu_flag = dav1d_get_cpu_flags();
+    ass_set_cpu_flags_mask(flag);
+    state.cpu_flag = ass_get_cpu_flags();
 
     if (!flag || state.cpu_flag != old_cpu_flag) {
+        ASS_Renderer *renderer = ass_renderer_init(state.library);
```
MrSmile, line 480: Is there a need to initialize a renderer for assembly functions? Looks like overkill for me. Without the renderer even the global cpu masking can be localized into this function instead.

checkasm/checkasm.c
```diff
@@ -517,10 +504,22 @@ static void print_cpu_name(void) {
...
+static void ignore_msg(int level, const char *fmt, va_list va, void *data)
+{
+    return;
+}
+
 int main(int argc, char *argv[]) {
     (void)func_new, (void)func_ref;
     state.seed = get_seed();
     int ret = 0;
+    state.library = ass_library_init();
```
MrSmile, line 516: Same as `ass_renderer_init()`.

checkasm/checkasm.c
```diff
@@ -0,0 +1,794 @@
...
+#include <mach/mach_time.h>
+#endif
+#define COLOR_RED    1
+#define COLOR_GREEN  2
+#define COLOR_YELLOW 3
+
+static unsigned get_seed(void) {
+#ifdef __APPLE__
+    return (unsigned) mach_absolute_time();
+#elif defined(HAVE_CLOCK_GETTIME)
```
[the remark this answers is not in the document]
rcombs: Yup, fixed.

libass/ass_bitmap.c
rcombs: Removed this commit.

rcombs: Removed this commit.

checkasm/checkasm.c
rcombs: `ass_renderer_init` is responsible for selecting (or producing) a BitmapEngine struct.

libass/ass_cpu.h
rcombs: No particular reason, as it turns out; fixed.

libass/ass_cpu.c
rcombs: Sure.

libass/ass_bitmap.h
rcombs: Moved to first.

checkasm/be_blur.c
rcombs: Adjusted to test all possible mod-32 sizes.

rcombs:
Re-pushed (several times) with a number of improvements:
- Addressed (or responded to) all review comments
- `be_blur` ASM updated to pass checkasm without requiring any additional accommodations
- Now integrated with automake's test system
- Now passes `distcheck`
- Should work on all platforms (i.e. checkasm isn't built if we don't have the required ASM for it)

Re: improving the AVX2 perf, that's a perfectly reasonable goal, but I don't think it's in-scope for this PR, which is primarily about correctness and unit testing, and I think it's perfectly reasonable to disable those AVX2 routines for now, until we have faster code.

libass/ass_cpu.h
```diff
@@ -0,0 +1,37 @@
...
+
+#ifndef LIBASS_CPU_H
+#define LIBASS_CPU_H
+
+typedef enum ASS_CPUFlags {
+    ASS_CPU_FLAG_NONE = 0x00,
+#define ASS_CPU_FLAG_NONE          0x00
+#define ASS_CPU_FLAG_X86           0x01
+#define ASS_CPU_FLAG_X86_SSE2      0x02
+#define ASS_CPU_FLAG_X86_AVX       0x04
```
MrSmile, line 27: Do we need AVX? It needlessly complicates `ass_get_cpu_flags()`.

libass/ass_cpu.h
```diff
@@ -0,0 +1,37 @@
...
+#define ASS_CPU_FLAG_X86_AVX2      0x08
+#define ASS_CPU_FLAG_X86_AVX512ICL 0x10
```
MrSmile, line 29: I think it should be mentioned that it's placeholder flag for checkasm that never set.

libass/ass_bitmap.c
```diff
@@ -30,31 +30,38 @@
...
 
+#include "ass_cpu.h"
 #include "ass_utils.h"
 #include "ass_outline.h"
 #include "ass_bitmap.h"
 #include "ass_render.h"
 
 
 #define ALIGN           C_ALIGN_ORDER
-#define DECORATE(func)  ass_##func##_c
```
MrSmile, line 40: Do we really need this change? It adds a couple of lines but not really helps that much.

libass/ass_bitmap.c
```diff
@@ -30,31 +30,38 @@
...
+#define ENGINE_FLAGS    ASS_CPU_FLAG_NONE
```
MrSmile, line 41: I think that alignment should be incorporated into cpu flags. It's also less error-prone to use something like `alignment_order(flags)` instead of hardcoded numbers.

libass/x86/be_blur.asm
```diff
@@ -31,85 +31,101 @@ SECTION .text
...
-    movdqu xmm5, [r12 + r6 * 2] ; __m128i old_col_sum = *(col_pix_sum+x);
-    movdqa xmm3, xmm2 ; temp = new_pix;
-    paddw xmm3, xmm4 ; temp = _mm_add_epi16(temp, old_col_pix);
-    movdqu [r12 + r6 * 2], xmm3 ; *(col_sum_buf+x) = temp;
-    paddw xmm5, xmm3 ; old_col_sum = _mm_add_epi16(old_col_sum, temp);
-    psrlw xmm5, 4 ; old_col_sum = old_col_sum >> 4;
-    packuswb xmm5, xmm5 ; old_col_sum = _mm_packus_epi16(old_col_sum, old_col_sum);
-    movq qword [r13 + r6 - 1], xmm5 ; *(dst+x-1) = old_col_sum;
-    add r6, 8; x += 8;
+    %if mmsize == 32
```
MrSmile, line 74: I think both branches should be merged.

libass/ass_func_template.h
```diff
@@ -117,8 +117,8 @@ const BitmapEngine DECORATE(bitmap_engine) = {
     .mul_bitmaps = ass_mul_bitmaps_c,
 #endif
 
-#ifdef __x86_64__
-    .be_blur = DECORATE(be_blur),
+#if defined(__x86_64__) && (ENGINE_FLAGS & ASS_CPU_FLAG_X86_SSE2)
```
MrSmile, line 120: First condition here is unnecessary.

libass/ass_func_template.h
```diff
@@ -117,8 +117,8 @@ const BitmapEngine DECORATE(bitmap_engine) = {
...
+    .be_blur = DECORATE_SUFFIX(be_blur, sse2),
```
MrSmile, line 121: I think that simple `ass_be_blur_sse2` is better and `DECORATE_SUFFIX` is unnecessary. But if you'll choose to keep it then `ass_be_blur_c` below should be replaced accordingly.

checkasm/be_blur.c
```diff
@@ -0,0 +1,71 @@
...
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include "checkasm.h"
+
+#include "ass_bitmap.h"
+
+#define HEIGHT 8
+#define STRIDE 64
+#define MIN_WIDTH (STRIDE - 31)
```
MrSmile, line 25: Maybe it's better to check all the way down to 1?

checkasm/be_blur.c
```diff
@@ -0,0 +1,71 @@
...
+        declare_func(void,
+                     uint8_t *buf, intptr_t w, intptr_t h,
+                     intptr_t stride, uint16_t *tmp);
+
+        memset(buf_ref, 0, sizeof(buf_ref));
+        memset(buf_new, 0, sizeof(buf_new));
+
+        for (int w = MIN_WIDTH; w <= STRIDE; w++) {
+            for (int i = 0; i < sizeof(buf_ref); i++)
+                buf_ref[i] = buf_new[i] = rnd();
```
MrSmile, line 42: AFAIR we require zero-padding of bitmap stride. I think it's better to check valid use cases only.

MrSmile (commented):
Added some comments to the source code.

Also, if we're planning to improve `be_blur` AVX2 performance then I think that "ass_func_template: disable be_blur_avx2" should be removed from this PR.

libass/ass_cpu.h
MrSmile: No, duplication with the same name is worse (also it's never used as define shadows enum value anyway). Those flags are defined due to the use in `#if` later, so there is no alternative here.

checkasm/checkasm.c
MrSmile: I think it's better to extract `BitmapEngine` selection from `ass_renderer_init()` into another function. In that case it's possible to make `flag_mask` into parameter of that function and eliminate global variable and access function altogether.

libass/ass_render.c
```diff
@@ -70,15 +71,15 @@ ASS_Renderer *ass_renderer_init(ASS_Library *library)
     priv->ftlibrary = ft;
     // images_root and related stuff is zero-filled in calloc
 
+    priv->engine = &ass_bitmap_engine_c;
+
+    ASS_CPUFlags flags = ass_get_cpu_flags();
```
TheOneric, line 76: Causes `-Wunused-variable` when compiling without ASM.
I think an `#if CONFIG_ASM` around this and the architecture checks avoids this and imho works better for future architecture additions.

.gitignore
```diff
@@ -21,6 +21,7 @@ Makefile.in
 /test/test*
 /compare/compare*
 /profile/profile*
+/checkasm/checkasm*
```
TheOneric, line 24: Further down in `# Autoconf files` we also need to ignore `/test-driver`.

Makefile.am
```diff
@@ -16,5 +16,8 @@ if ENABLE_PROFILE
     profile = profile
 endif
 
-SUBDIRS = libass $(test) $(compare) $(profile)
+if ASM
+    checkasm = checkasm
+endif
 
+SUBDIRS = libass $(test) $(compare) $(profile) $(checkasm)
```
TheOneric, line 23: Not something that needs to addressed in this pr, but more as a future remark:

if we keep adding more recursive subdirs this will start to slow down dev-builds, since recursive targets are guaranteed to run in order and thus won't be parallelised. Also each subdir will be entered for all targets, regardless of whether or not it contributes to the current target.
We might want to consider switching to a non-recursive build *(via Automake's `include` we can stil use separate files for each unit, but paths will be different)*

libass/x86/be_blur.asm
rcombs: They're too different for this to be reasonable, due to all the extra cross-lane operations required for AVX2 (which are also why it's slower).

libass/ass_cpu.h
rcombs: The extra complexity comes from checking for (OS)XSAVE, which we _do_ need; AVX2 won't work unless the conditions required by the AVX flag are met.

checkasm/be_blur.c
rcombs: Down to 1 is excessive (be_blur isn't meaningful for anything <2px in both dimensions, and in practice never runs on <3), but I did find a couple actual bugs with this set to 2, so I've set it to that.

checkasm/be_blur.c
rcombs: Done.

rcombs:
Re-pushed with:
- An actual bugfix for be_blur on bitmaps <8px wide
- No more global state
- An infrastructural overhaul to the BitmapEngine system, moving away from static prefilled objects (which got overly complex when not all functions are available), and towards a system similar to that used in the lav* libraries
- All enum values for the CPU flags instead of macros (the macros were required for the old predefined BitmapEngine structs to work)

libass/ass_cpu.h
MrSmile: But that can be done inside check function locally, without outputting global flag.

libass/x86/be_blur.asm
MrSmile: I've done similar things in my code (for example [here](https://github.com/libass/libass/blob/master/libass/x86/blur.asm#L306)) and as I can see, at least half of instructions in both branches are the same. It's strange that it's slower, for me all that excess `vperm2i128` don't have much impact on performance. I'll look into `be_blur` assembly later then.

libass/ass_bitmap.c
```diff
@@ -98,12 +75,11 @@ void ass_synth_blur(const BitmapEngine *engine, Bitmap *bm,
 bool alloc_bitmap(const BitmapEngine *engine, Bitmap *bm,
                   int32_t w, int32_t h, bool zero)
 {
-    unsigned align = 1 << engine->align_order;
-    size_t s = ass_align(align, w);
+    size_t s = ass_align(ASS_ALIGNMENT, w);
```
MrSmile, line 78: Use of the global alignment for bitmaps independent of the current asm mode? Indeed that looks like the simplest solution (I prefer `ASS_BITMAP_ALIGNMENT` though). Are your sure we don't want to check different alignments with checkasm later?

libass/ass_bitmap_engine.c
```diff
@@ -0,0 +1,80 @@
...
+#if (defined(__i386__) || defined(__x86_64__))
+#include "x86/init.h"
+#endif
+#endif
+
+#define ENGINE_SUFFIX   c
+#include "ass_func_template.h"
+#undef ENGINE_SUFFIX
+
+void ass_bitmap_engine_init(BitmapEngine* engine, ASS_CPUFlags mask)
```
MrSmile, line 35: I'm not a fan of this triple copy-paste spanning two different files. Previous approach is much cleaner and less error-prone. Indeed, setting functions without complete assembly is slightly cumbersome, but I think we should aim for the full assembly instead.

libass/ass_bitmap_engine.c
```diff
@@ -0,0 +1,80 @@
...
+    engine->expand_horz = ass_expand_horz_c;
+    engine->expand_vert = ass_expand_vert_c;
+
+    engine->blur_horz[0] = ass_blur4_horz_c;
+    engine->blur_horz[1] = ass_blur5_horz_c;
+    engine->blur_horz[2] = ass_blur6_horz_c;
+    engine->blur_horz[3] = ass_blur7_horz_c;
+    engine->blur_horz[4] = ass_blur8_horz_c;
+
+    engine->blur_horz[0] = ass_blur4_vert_c;
```
MrSmile, line 68: That should be `blur_vert`. Also in other two places.

MrSmile (commented):
I dislike that dynamic `BitmapEngine` approach. It don't see any advantages (current or future) in this approach but problematic error-prone copy-paste is evident (there is already a typo there). Other than that LGTM.

MrSmile:
I've decided to tweak some of the assembly and used this branch for tests. Couple of notes from the end user perspective:
* Better to not link with the whole libass but only relevant files. It's a quite hard requirement to provide all libass dependencies only to check some of the 32-bit assembly.
* Benchmark numbers are quite unstable. Maybe it's better to increase the run count?

libass/ass_bitmap.c
MrSmile: After some thinking, I've concluded that the old approach with `BitmapEngine`-local alignments is better. Larger than necessary padding can have significant negative impact on performance in the case of many small bitmaps, which quite corresponds to subs rendering with its ton of font glyphs.

.gitignore
```diff
@@ -19,9 +19,14 @@ Makefile.in
 
 # Test programs
 /test/test*
+!/test/test.c
 /compare/compare*
+!/test/compare/compare.c
```
TheOneric, line 24: stray `/test` at the beginning

checkasm/checkasm.h
```diff
@@ -114,8 +114,8 @@ static void *func_ref, *func_new;
      ((func_type *)func_ref)(__VA_ARGS__));\
     checkasm_set_signal_handler_state(0)
 
-#if HAVE_ASM
-#if ARCH_X86
+#if CONFIG_ASM
+#if defined(__x86_64__)
```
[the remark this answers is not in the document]
MrSmile: Yes, I hit it too in my experiments: bench doen't work in 32-bit mode.

MrSmile:
I've extended tests to all assembly functions [here](https://github.com/MrSmile/libass/commits/checkasm). It's a research code that is not production ready, but it can be used as a reference for improving this PR. Testing of blur/rasterizer assembly is quite hard due to different data layouts depending on version, so I've ended generating bunch of different function versions by [include trickery](https://github.com/MrSmile/libass/blob/checkasm/checkasm/bitmap_engine.c).

MrSmile:
I've done a massive overhaul of my [extension of checkasm branch](https://github.com/MrSmile/libass/commits/checkasm). I've removed all assembly fixes from it (already merged) so only checkasm-related commits are left. There are several infrastructure commits in it, so maybe they should have a dedicated PR too? Anyway, I think it's almost production-ready.

.gitignore
TheOneric: The stray `/test` is still an issue.

checkasm/checkasm.h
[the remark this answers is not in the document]
TheOneric: Still an issue, this is also needed on 32-bit systems.

TheOneric (changes requested):
Commit f7ff255e21bca25587342185a252a559a57ed30e (configure: explicitly disable ASM when it isn't available) and ac1a6a2d7c0ff12e45a1ebcecde04be40576de21 (configure: remove redundant asm-disabled message ) should be dropped; their union is a noop and the intended message has already been added *(taken from this PR)* as part of #551 in f975de5d96aafd804c66ea79f0780ce2b30e1928.

Can't say anything about the actual x86 and aarch64 assembly.

TheOneric:
With #561 merged mroe commits can be dropped here now; altogether the following should be dropped:
 - build: move private_prefix define to Makefile.am (d7c428e68d69569eef7c0214f5a4d1f847afa6b0)
 - configure: explicitly disable ASM when it isn't available (f7ff255e21bca25587342185a252a559a57ed30e)
 - .gitignore: avoid ignoring source files (7ae2ae5e117e3b09e2a86f650d7855dca29713cb)
 - x86: remove empty file (6fd4fae490b6763d7ac759dacea576db534256ff)
 - configure: remove redundant asm-disabled message (ac1a6a2d7c0ff12e45a1ebcecde04be40576de21)

rcombs:
Closing out as obsolete.

PR #402, opened 2020-06-11
Fix memory leak of font family string when error during add_face

adipose:
When using this library I consistently detected a memory leak tied to the family string.  It was set to NULL here without being freed.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -257,7 +257,7 @@ size_t ass_font_construct(void *key, void *value, void *priv)
         free(font->desc.family);
         font->desc.family = NULL;
     }
-    
+
     return 1;
 }
 
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -253,8 +253,11 @@ size_t ass_font_construct(void *key, void *value, void *priv)
     font->size = 0.;
 
     int error = add_face(render_priv->fontselect, font, 0);
-    if (error == -1)
+    if (error == -1) {
+        free(font->desc.family);
         font->desc.family = NULL;
+    }
+    
     return 1;
 }
 


libass/ass_font.c
```diff
@@ -253,8 +253,11 @@ size_t ass_font_construct(void *key, void *value, void *priv)
     font->size = 0.;
 
     int error = add_face(render_priv->fontselect, font, 0);
-    if (error == -1)
+    if (error == -1) {
+        free(font->desc.family);
         font->desc.family = NULL;
+    }
+
```
TheOneric, line 260: Style-Nit: There should be no trailing whitespaces (so no spaces at all on empty lines).

*(GitHub's diff view doesn't diplay any spaces here, but git am warned about it and checking out your branch directly also confirms the presence of spaces. That's a bit odd.)*

adipose: Should be fixed.

TheOneric:
Leads to failure in case of #414 .

```
==15822==ERROR: AddressSanitizer: heap-use-after-free on address 0x602000001930 at pc 0x7fea4a9adb00 bp 0x7ffc116ea640 sp 0x7ffc116e9df0
READ of size 1 at 0x602000001930 thread T0
    #0 0x7fea4a9adaff  (/lib/x86_64-linux-gnu/libasan.so.5+0xb4aff)
    #1 0x5640f6704101 in font_compare …/libass/libass/ass_cache.c:55
    #2 0x5640f67084b2 in ass_cache_get …/libass/libass/ass_cache.c:369
    #3 0x5640f670e856 in ass_font_new …/libass/libass/ass_font.c:245
    #4 0x5640f6680a95 in update_font …/libass/libass/ass_parse.c:128
    #5 0x5640f668691f in parse_tags …/libass/libass/ass_parse.c:496
    #6 0x5640f6667d27 in parse_events …/libass/libass/ass_render.c:1905
    #7 0x5640f6667d27 in ass_render_event …/libass/libass/ass_render.c:2578
    #8 0x5640f6667d27 in ass_render_frame …/libass/libass/ass_render.c:3136
    #9 0x5640f660faea in process_image …/libass/compare/compare.c:326
    #10 0x5640f6608487 in main …/libass/compare/compare.c:611
    #11 0x7fea4964409a in __libc_start_main ../csu/libc-start.c:308
    #12 0x5640f6609629 in _start (…/libass/compare/compare+0x164629)

0x602000001930 is located 0 bytes inside of 12-byte region [0x602000001930,0x60200000193c)
freed by thread T0 here:
    #0 0x7fea4a9e1fb0 in __interceptor_free (/lib/x86_64-linux-gnu/libasan.so.5+0xe8fb0)
    #1 0x5640f671141d in ass_font_construct …/libass/libass/ass_font.c:273

previously allocated by thread T0 here:
    #0 0x7fea4a933810 in strdup (/lib/x86_64-linux-gnu/libasan.so.5+0x3a810)
    #1 0x5640f6680b25 in update_font …/libass/libass/ass_parse.c:109
    #2 0x5640f673d949  (…/libass/compare/compare+0x298949)
```

MrSmile:
There is no memory leak here: ownership of `font->desc.family` string belongs to `desc->family` which is the part of the cache key and managed by the cache logic.

astiob:
The leak should be fixed by #433, which has just landed in master. Please let us know if it isn’t.

PR #408, opened 2020-07-02
Fix Travis builds

TheOneric:
Travis builds are currently failing on MacOS due to `brew install` throwing an error on already installed but outdated packages (see eg [recent build failure](https://travis-ci.org/github/libass/libass/jobs/703007253) from #392 ). By delegating the brew and apt parts to addons, `brew install` or `brew upgrade` will be called as necessary.

While I was at it, I also added harfbuzz to macOS builds and bumped Ubuntu distro to bionic to make sure harfbuzz >= 1.2.3 will be present, which is needed for prs #398 and #399 . This is in a separate commit. Discard, copy over to relevant pr or apply as preferred.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,7 @@ branches:
     - coverity_scan
 
 before_install:
-  - (sudo apt-get update || brew update)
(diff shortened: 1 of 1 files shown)


astiob:
Awesome, thanks!

Issue #374, opened 2020-03-04
Bidi bracket matching is VSFilter-incompatibile


OmrSi:
I opened this issue in the MPV repo and was told to ask here if anyone knows why this happens.

Using MPV, The line should show up like this:
![image](https://user-images.githubusercontent.com/20923700/75904414-d8bc5200-5e4b-11ea-8bea-9703d2da6b19.png)

But the line shows up like this:
![image](https://user-images.githubusercontent.com/20923700/75904444-e4a81400-5e4b-11ea-8f42-453d5c84967b.png)

Here is a test file:
[Test.zip](https://github.com/mpv-player/mpv/files/4288644/Test.zip)

Can anyone help me identify the problem?
Thank you.

wm4:
Someone determined a likely cause: https://github.com/mpv-player/mpv/issues/7507#issuecomment-595380614

CC @astiob @khaledhosny

astiob:
The new ordering is the same that I get when I copy the text from your SRT file to other programs with base left-to-right direction:

![Firefox 73](https://user-images.githubusercontent.com/515193/76027784-ea7c2300-5f39-11ea-92e3-8b4a223198aa.png)

![macOS 10.13.6 Terminal 2.8.3](https://user-images.githubusercontent.com/515193/76026379-501ae000-5f37-11ea-89bf-4262be637cca.png)

With base right-to-left direction, the _text_ is ordered as in your “old” screenshot, but the order of punctuation marks is flipped as well:

![macOS 10.13.6 TextEdit 1.13 (autodetected base direction)](https://user-images.githubusercontent.com/515193/76026386-527d3a00-5f37-11ea-8fb5-2386784b7596.png)

![Firefox 73 (explicit right-to-left base direction)](https://user-images.githubusercontent.com/515193/76027762-df28f780-5f39-11ea-8046-7ac1d0193a5d.png)

**All signs point to the new behaviour being correct for the source text.**

(Incidentally, my Word for Mac 2019 (16.29.1) screws this up even after explicitly setting right-to-left direction, possibly due to a lack of some Middle East–specific modules:
![Microsoft Word for Mac 16.29.1 (2019)](https://user-images.githubusercontent.com/515193/76027472-5742ed80-5f39-11ea-9817-2a4c0fba0c73.png)
I tried to switch the text’s language to Arabic to see if it would help, but it just ignores me and keeps whatever European language it happened to detect based on things like my keyboard layout.)

astiob:
[...]

PR #409, opened 2020-07-05
Disable bidi bracket matching unless enabled via ASS_Feature

astiob:
As shown in #374, bracket matching (introduced in commit df7c00c95ec6e526bf85e6bc8296eaca66f5db8d of #298, but not yet part of any release) is incompatible with VSFilter (even on modern Windows).

As decided in #374, disable it by default. But as bracket matching is generally a good thing (and simply more standards-compliant), keep it available as an ASS_Feature that can be enabled for things like media player UIs.

The new feature can be toggled individually or enabled as part of the catch-all `ASS_FEATURE_INCOMPATIBLE_EXTENSIONS`. I’ve removed the placeholder `enable_extensions` flag and instead directly toggle the new feature’s underlying flag when the catch-all is toggled.

### Fallback

If libass is compiled against FriBidi older than 1.0, bracket matching is impossible. Signal this at runtime by failing to recognize the `ASS_FEATURE_BIDI_BRACKETS` feature. This way, clients who want to use bracket matching can set the feature without any compile-time checks for FriBidi and can be freely linked against libass that is itself compiled against any version of FriBidi; and yet they can detect at runtime whether the feature is actually enabled.

Bump `LIBASS_VERSION` so that the presence of `ASS_FEATURE_BIDI_BRACKETS` enum constant can be detected at compile-time. As per above, the enum constant is available at compile-time regardless of FriBidi version.

### Small conflicts

It’s nothing to worry about, but FYI, this conflicts with a commit from #399 that changes the signature of `ass_shaper_shape`.

### Naming of `ASS_ParserPriv`

The new field has nothing to do with the high-level ASS & Matroska packet parsing in ass.c that the rest of `ASS_ParserPriv` is used for. Or with any kind of ASS parsing for that matter. This disturbs me, but I’m (currently) making no attempt to solve this in this PR.

### DIFF
--- a/libass/ass_priv.h
+++ b/libass/ass_priv.h
@@ -19,7 +19,10 @@
 #ifndef LIBASS_PRIV_H
 #define LIBASS_PRIV_H
 
+#include <stdbool.h>
 #include <stdint.h>
+
+#include "ass_shaper.h"
 
 typedef enum {
     PST_UNKNOWN = 0,
@@ -55,10 +58,12 @@ struct parser_priv {
     int read_order_elems; // size in uint32_t units of read_order_bitmap
     int check_readorder;
 
-    int enable_extensions;
-
     // tracks [Script Info] headers set by the script
     uint32_t header_flags;
+
+#ifdef USE_FRIBIDI_EX_API
+    bool bidi_brackets;
+#endif
 };
 
 #endif /* LIBASS_PRIV_H */
--- a/libass/ass_priv.h
+++ b/libass/ass_priv.h
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2006 Evgeniy Stepanov <eugeni.stepanov@gmail.com>
+ *
+ * This file is part of libass.
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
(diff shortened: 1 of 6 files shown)


libass/ass_priv.h
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (C) 2006 Evgeniy Stepanov <eugeni.stepanov@gmail.com>
```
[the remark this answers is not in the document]
astiob: You’re not wrong, but I don’t know better either. I figured since this code was moved from ass.c, I’d copy the copyright notice from there, too. It’s even somewhat true, as half of `struct parser_priv` does in fact date back to eugeni’s 2006 commits (although I guess that’s a small fraction of the current code lines overall).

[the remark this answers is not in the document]
astiob: We should probably switch to some kind of generic “libass contributors” copyright notice in all files. Or, you know, just drop copyright notices altogether, since they don’t have any legal significance anyway.

astiob:
OK, I’ve added a copy of the `bidi_brackets` field to `ASS_Shaper`, and ass_render now sets that (through a setter function) instead of passing in the `parser_priv` directly.

astiob:
Oh, the Travis CI build failure helped find a bug. But we may want to make it use a newer FriBidi… 😅 (Or keep it as is precisely to test building with older FriBidi? But we’ll forget, and eventually it will get updated anyway.)

libass/ass_shaper.c
```diff
@@ -758,7 +758,7 @@ static void shape_fribidi(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
  * \param shaper shaper instance
  * \param kern toggle kerning
  */
-void ass_shaper_set_kerning(ASS_Shaper *shaper, int kern)
+void ass_shaper_set_kerning(ASS_Shaper *shaper, bool kern)
 {
 #ifdef CONFIG_HARFBUZZ
     shaper->features[KERN].value = !!kern;
```
MrSmile, line 762: I usually use `!!` for `int` to `bool` conversion, but here it's in reverse. Maybe it's better to write directly as `kern ? 1 : 0`?

libass/ass_shaper.c
```diff
@@ -860,6 +858,13 @@ void ass_shaper_set_level(ASS_Shaper *shaper, ASS_ShapingLevel level)
     shaper->shaping_level = level;
 }
 
+#ifdef USE_FRIBIDI_EX_API
+void ass_shaper_set_bidi_brackets(ASS_Shaper *shaper, bool match_brackets)
+{
+    shaper->bidi_brackets = !!match_brackets;
```
MrSmile, line 864: I think that `!!` is redundant here as both `bidi_brackets` and `match_brackets` are bools.

astiob: Removed the `!!`.

libass/ass_shaper.c
astiob: The `!!` isn’t needed at all, as `_Bool` is an integer that can only be 0 or 1. I’ve now removed it.

PR #421, opened 2020-08-15
ass.h: Mark deprecated declarations as deprecated

TheOneric:
Some of the API is declared to be deprecated in source comments, but this is easy to miss. Now GCC, Clang and MSVC should issue a warning when deprecated API is used.
 
---

Version checks for GCC are a bit messy. If supporting deprecation-warnings on versions older than GCC 6 is not important this can be simplified. *(For reference: 6.1 released 27.04.2016; Debian Jessie (oldoldstable) shipped 4.9.2, but LTS support ended a month ago)*  
Since I could not find out when clang gained support for the deprecation attribute (for enums), I'm assuming it will always be supported.  

Tested with clang7 and gcc8

**Not tested with MSVC**.
It'd be good, if someone with an MSVC compiler could verify that deprecation warnings are working. Otherwise I'd prefer to drop the untested MSVC part.
Here's a test file: [depr_test.c.txt](https://github.com/libass/libass/files/5077892/depr_test.c.txt)

### DIFF
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -30,6 +30,22 @@
 extern "C" {
 #endif
 
+#if (defined(__GNUC__) && ((__GNUC__ > 3) || (__GNUC__ == 3 && __GNUC_MINOR__ >= 1))) || defined(__clang__)
+    #define ASS_DEPRECATED(msg) __attribute__((deprecated(msg)))
+    #if __GNUC__ > 5 || defined(__clang__)
+        #define ASS_DEPRECATED_ENUM(msg) __attribute__((deprecated(msg)))
+    #else
+        #define ASS_DEPRECATED_ENUM(msg)
+    #endif
+#elif defined(_MSC_VER)
+    #define ASS_DEPRECATED(msg) __declspec(deprecated(msg))
+    #define ASS_DEPRECATED_ENUM(msg)
+#else
+    #define ASS_DEPRECATED(msg)
+    #define ASS_DEPRECATED_ENUM(msg)
+#endif
+
+
 /*
  * A linked list of images produced by an ass renderer.
  *
@@ -129,7 +145,7 @@ typedef enum {
     /**
      * Old alias for ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE. Deprecated. Do not use.
      */
-    ASS_OVERRIDE_BIT_FONT_SIZE = 1 << 1,
+    ASS_OVERRIDE_BIT_FONT_SIZE ASS_DEPRECATED_ENUM("replaced by ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE") = 1 << 1,
     /**
      * On dialogue events override: FontSize, Spacing, Blur, ScaleX, ScaleY
      */
@@ -402,7 +418,7 @@ void ass_set_pixel_aspect(ASS_Renderer *priv, double par);
  * \param dar display aspect ratio (DAR), prescaled for output PAR
  * \param sar storage aspect ratio (SAR)
  */
-void ass_set_aspect_ratio(ASS_Renderer *priv, double dar, double sar);
+ASS_DEPRECATED("use 'ass_set_pixel_aspect' instead") void ass_set_aspect_ratio(ASS_Renderer *priv, double dar, double sar);
 
 /**
  * \brief Set a fixed font scaling factor.
@@ -501,7 +517,7 @@ void ass_set_selective_style_override(ASS_Renderer *priv, ASS_Style *style);
  * \param priv renderer handle
  * \return success
  */
-int ass_fonts_update(ASS_Renderer *priv);
+ASS_DEPRECATED("it does nothing") int ass_fonts_update(ASS_Renderer *priv);
 
 /**
  * \brief Set hard cache limits.  Do not set, or set to zero, for reasonable


libass/ass.h
```diff
@@ -30,6 +30,21 @@
 extern "C" {
 #endif
 
+#if (defined(__GNUC__) && ((__GNUC__ > 3) || (__GNUC__ == 3 && __GNUC_MINOR__ >= 1))) || defined(__clang__)
+    #define ASS_DEPRECATE(msg)  __attribute__((deprecated(msg)))
```
rcombs, line 34: Nit: I'd prefer `ASS_DEPRECATED` here. Also, there's a doubled space here for no apparent reason.

libass/ass.h
```diff
@@ -30,6 +30,21 @@
...
+    #if __GNUC__ > 5 || defined(__clang__)
+        #define ASS_DEPRECATE_ENUM(msg) __attribute__((deprecated(msg)))
+    #else
+        #define ASS_DEPRECATE_ENUM(msg)
+    #endif
+#elif defined(_MSC_VER)
+    #define ASS_DEPRECATE(msg)  __declspec(deprecated)
```
rcombs, line 41: MSVC apparently supports passing a message here (`__declspec(deprecated(msg))`), so if we're going to keep this I think we should include that. Also, another double-space.

libass/ass.h
TheOneric: Done.

libass/ass.h
TheOneric: Thanks, I missed that.

libass/ass.h
```diff
@@ -501,7 +517,7 @@ void ass_set_selective_style_override(ASS_Renderer *priv, ASS_Style *style);
  * \param priv renderer handle
  * \return success
  */
-int ass_fonts_update(ASS_Renderer *priv);
+ASS_DEPRECATED("No longer necessary") int ass_fonts_update(ASS_Renderer *priv);
```
astiob, line 520: I don’t have a strong preference, but consider also the phrasing `"it does nothing"` (which I see in Apple’s Objective-C code).

astiob (commented):
Overall looks good.

If anyone could quickly test MSVC, that’d be great. Otherwise, perhaps I’ll try to do it myself at some point. I think it’s not a blocking issue though: the code looks simple enough to be correct.

---

FWIW, Clang [supports a second argument](https://clang.llvm.org/docs/AttributeReference.html#deprecated) to suggest a new name as a drop-in replacement for the deprecated one:

```c
__attribute__((deprecated("replaced by ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE", "ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE"))
```

```
a.c:5:16: warning: 'ASS_OVERRIDE_BIT_FONT_SIZE' is deprecated: replaced by ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE [-Wdeprecated-declarations]
int override = ASS_OVERRIDE_BIT_FONT_SIZE;
               ^~~~~~~~~~~~~~~~~~~~~~~~~~
               ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE
```

This can be detected using `#if __has_extension(attribute_deprecated_with_replacement)` (although the only way to discover this was to read Clang’s source code or [the original patch review](https://reviews.llvm.org/D17865)).

This seems cool, although it’s not of terribly much use to us as `ASS_OVERRIDE_BIT_FONT_SIZE` is the only instance that can use it. If we do want to use it, I believe we have to separate `ASS_DEPRECATED_ALIAS` from `ASS_DEPRECATED`.

---

Some more nonblocking bikeshedding:

I’d prefer the messages to start with a lowercase letter, as I find this is more grammatical and looks better with the way compilers show these messages:

Clang and GCC:

```
a.c:5:XX: warning: 'ASS_OVERRIDE_BIT_FONT_SIZE' is deprecated: replaced by ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE [-Wdeprecated-declarations]
a.c:5:XX: warning: 'ASS_OVERRIDE_BIT_FONT_SIZE' is deprecated: Replaced by ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE [-Wdeprecated-declarations]
```

libass/ass.h
TheOneric: Changed this message and all messages now start with a lowercase letter.

TheOneric:
> FWIW, Clang [supports a second argument](https://clang.llvm.org/docs/AttributeReference.html#deprecated) to suggest a new name as a drop-in replacement for the deprecated one:
> 
> […]
> 
> This can be detected using `#if __has_extension(attribute_deprecated_with_replacement)` […].
> 
> This seems cool, although it’s not of terribly much use to us as `ASS_OVERRIDE_BIT_FONT_SIZE` is the only instance that can use it. If we do want to use it, I believe we have to separate `ASS_DEPRECATED_ENUM_ALIAS` from `ASS_DEPRECATED_ENUM` (or keep _only_ the two-argument version).

This does seem nice, but I'm not sure if it justifies the additional complexity it adds to the macro definition. The deprecation message already clearly communicates that there is a simple drop in replacement available.
Having _only_ the two-argument version ~~might cause problems if something is deprecated without drop-in replacement~~ might work as at least with calng7 passing `""` as replacement results in no replacement being suggested.
It would look like this:

```C
#if defined(__clang__) && __has_extension(attribute_deprecated_with_replacement)
    #define ASS_DEPRECATED(msg,repl) __attribute__((deprecated(msg, repl)))
    #define ASS_DEPRECATED_ENUM(msg,repl) __attribute__((deprecated(msg, repl)))
#elif defined(__clang__)
    #define ASS_DEPRECATED(msg,repl) __attribute__((deprecated(msg)))
    #define ASS_DEPRECATED_ENUM(msg,repl) __attribute__((deprecated(msg)))
#elif (defined(__GNUC__) && ((__GNUC__ > 3) || (__GNUC__ == 3 && __GNUC_MINOR__ >= 1)))
    #define ASS_DEPRECATED(msg,repl) __attribute__((deprecated(msg)))
    #if __GNUC__ > 5
        #define ASS_DEPRECATED_ENUM(msg,repl) __attribute__((deprecated(msg)))
    #else
        #define ASS_DEPRECATED_ENUM(msg,repl)
    #endif
#elif defined(_MSC_VER)
    #define ASS_DEPRECATED(msg,repl) __declspec(deprecated(msg))
    #define ASS_DEPRECATED_ENUM(msg,repl)
#else
    #define ASS_DEPRECATED(msg,repl)
    #define ASS_DEPRECATED_ENUM(msg,repl)
#endif
```

TheOneric:
I used https://rextester.com/l/c_online_compiler_visual to test the MSVC part. It seems to work as expected (`C4996` being the code for deprecation warnings; msvc needs `/W3` argument to display the warnings):

```
source_file.c(991): warning C4996: 'ass_set_aspect_ratio': use 'ass_set_pixel_aspect' instead
source_file.c(968): note: see declaration of 'ass_set_aspect_ratio'
source_file.c(993): warning C4996: 'ass_fonts_update': it does nothing
source_file.c(974): note: see declaration of 'ass_fonts_update'
```

PR #424, opened 2020-08-30
A couple small optimizations

rcombs:
See commit messages for details; these reduced CPU usage on a real script I was sent by over 30%.

### DIFF
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -73,12 +73,14 @@ static inline void push_arg(struct arg *args, int *nargs, char *start, char *end
  */
 static inline int mystrcmp(char **p, const char *sample)
 {
-    int len = strlen(sample);
-    if (strncmp(*p, sample, len) == 0) {
-        (*p) += len;
+    char *p2;
+    for (p2 = *p; *sample != 0 && *p2 == *sample; p2++, sample++)
+        ;
+    if (*sample == 0) {
+        *p = p2;
         return 1;
-    } else
-        return 0;
+    }
+    return 0;
 }
 
 double ensure_font_size(ASS_Renderer *priv, double size)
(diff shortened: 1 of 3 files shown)


libass/ass_parse.c
```diff
@@ -73,12 +73,14 @@ static inline void push_arg(struct arg *args, int *nargs, char *start, char *end
  */
 static inline int mystrcmp(char **p, const char *sample)
 {
-    int len = strlen(sample);
-    if (strncmp(*p, sample, len) == 0) {
-        (*p) += len;
+    char *p2;
+    for (p2 = *p; *p2 == *sample; p2++, sample++)
+        ;
```
TheOneric, line 78: Won't this read past the legal memory region, if `{\tag` is at the very end of an event and safe memory region, and sample is also `tag`  ?
This seems to be fine in `parse_tags` since iinm there will always be atleast an `}` in `p`, but never in `sample`. This does however not apply to `event_has_hard_overrides`.

rcombs: I don't think that matters, since both strings will always be null-terminated?

TheOneric: > since both strings will always be null-terminated?

Yes, which means `*p2 == *sample` is true, so both get incremented and dereferenced and read past their legal range.
Eg if both strings are `{'p', '\0'}` as in the above sample file.

rcombs: Ah! Good catch, fixing.

astiob (approved):
LGTM now. Thanks for catching that!

mystr… wrapper inlining is something that is also solved by compiling with LTO, right? But it won’t hurt to make it happen even without LTO.

rcombs:
Yeah, in theory at least. Just seemed silly to require LTO for that on those tiny trivial wrappers.

PR #426, opened 2020-09-17
Assorted bug fixes

rcombs:
See individual commit messages; most notably the last commit reverts a mistake of mine from last year (which thankfully never made it into a release).

### DIFF
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -96,7 +96,8 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
     for (i = 0; i < fonts->nfont; i++) {
         FcPattern *pat = fonts->fonts[i];
         FcBool outline;
-        int index, weight;
+        int index;
+        double weight;
         char *path;
         char *fullnames[MAX_NAME];
         char *families[MAX_NAME];
@@ -109,23 +110,26 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
         // simple types
         result  = FcPatternGetInteger(pat, FC_SLANT, 0, &meta.slant);
         result |= FcPatternGetInteger(pat, FC_WIDTH, 0, &meta.width);
-        result |= FcPatternGetInteger(pat, FC_WEIGHT, 0, &weight);
+        result |= FcPatternGetDouble(pat, FC_WEIGHT, 0, &weight);
         result |= FcPatternGetInteger(pat, FC_INDEX, 0, &index);
         if (result != FcResultMatch)
             continue;
 
         // fontconfig uses its own weight scale, apparently derived
         // from typographical weight. we're using truetype weights, so
-        // convert appropriately
-#if FC_VERSION >= 21191
-        meta.weight = FcWeightToOpenType(weight);
+        // convert appropriately.
+#if FC_VERSION >= 21292
+        meta.weight = FcWeightToOpenTypeDouble(weight) + 0.5;
+#elif FC_VERSION >= 21191
+        // Versions prior to 2.12.92 only had integer precision.
+        meta.weight = FcWeightToOpenType(weight + 0.5) + 0.5;
 #else
         // On older fontconfig, FcWeightToOpenType is unavailable, and its inverse was
         // implemented more simply, using an if/else ladder instead of linear interpolation.
         // We implement an inverse of that ladder here.
         // We don't expect actual FC caches from these versions to have intermediate
         // values, so the average checks are only for completeness.
-#define AVG(x, y) ((x + y) / 2)
+#define AVG(x, y) (((double)x + y) / 2)
 #ifndef FC_WEIGHT_SEMILIGHT
 #define FC_WEIGHT_SEMILIGHT 55
 #endif
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -128,21 +128,42 @@ void update_font(ASS_Renderer *render_priv)
     render_priv->state.font = ass_font_new(render_priv, &desc);
 }
 
+static double calc_anim(double new, double old, double pwr)
+{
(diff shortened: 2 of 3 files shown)


libass/ass_fontconfig.c
```diff
@@ -109,23 +110,26 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
...
         // fontconfig uses its own weight scale, apparently derived
         // from typographical weight. we're using truetype weights, so
-        // convert appropriately
-#if FC_VERSION >= 21191
+        // convert appropriately.
+#if FC_VERSION >= 21292
+        meta.weight = FcWeightToOpenTypeDouble(weight);
+#elif FC_VERSION >= 21191
+        // Versions prior to 2.12.92 only had integer precision.
         meta.weight = FcWeightToOpenType(weight);
```
astiob, line 125: Should this round the now-fractional `weight` rather than truncate it? Or does old Fontconfig always produce an integer `weight` despite the `FcPatternGetDouble` call?

libass/ass_parse.c
```diff
@@ -128,21 +128,53 @@ void update_font(ASS_Renderer *render_priv)
...
+   return (1 - pwr) * old + new * pwr;
+}
+
+static uint32_t calc_anim_uint32(uint32_t new, uint32_t old, double pwr)
+{
+    double ret = calc_anim(new, old, pwr);
+
+    // Avoid UB on out-of-range values; match x86 behavior
+    if (isnan(ret) || ret < 0 || ret > UINT32_MAX)
+        return INT32_MIN;
```
astiob, line 142: Is this correct for the `uint32_t` version?

Why are you using an `uint32_t` variant for alpha to begin with? As far as I see, for alpha VSFilter casts the double to `BYTE`, which is `unsigned char`. Wouldn’t that produce something different? From a quick & dirty test on Intel macOS (not sure how representative that is), I seem to get 0 from all kinds of unrepresentable values when casting `double` to `unsigned char`. For reference, I see `(int)CalcAnimation` and `(BYTE)CalcAnimation` in VSFilter, and no other variants.

astiob (commented):
Any details on why you thought the FreeType-provided family was beneficial for compatibility originally but this turned out to be false? 😅 I seem to recall a user complain about it, but I can’t find that comment now (perhaps I’m thinking of something else).

libass/ass_fontconfig.c
rcombs: Hmm, I suppose it's possible for an older fontconfig to give a fractional weight from `FcPatternGetDouble` if you upgraded, cached, then downgraded? (I don't think the cache version changed here). Seems pretty corner-case-y but I guess it doesn't hurt. Prior to the introduction of `FcWeightToOpenTypeDouble` and `FcWeightFromOpenTypeDouble`, the cached values were produced by the integer `FcWeightFromOpenType` function, so you couldn't end up with fractional weights.

As for rounding the return value from `FcWeightToOpenTypeDouble`, I suppose that wouldn't hurt either?

libass/ass_parse.c
rcombs: > Is this correct for the `uint32_t` version?

Oh, hmmm, I was looking at the documentation for `cvttsd2si`, which returns `0x80000000` or `0x8000000000000000` on error… but it looks like compilers actually explicitly handle this when converting to an integer type and give 0 instead? Huh. It doesn't affect the color case regardless, since the high bit will get truncated away, but sure I'll switch this to return 0.

Looking closer, apparently negative (but ≥INT32_MIN) values actually convert to their signed integer representations and then cast to unsigned from there, so I'll correct that (don't think this can happen in the color case either but might as well be thorough).

Fun fact I just found while investigating deeper: signed int behavior seems to be the same regardless, but converting double to unsigned int can give different results when building for 32- vs 64-bit! I think we should probably go with the 32-bit behavior.

So, my implementation of `calc_anim_int32` is correct, but `calc_anim_uint32` should:
- Return 0 for values above `UINT32_T`
- Return 0 for NAN
- Return `(uint32_t)INT_MIN` for values below `INT_MIN`
- Return `(uint32_t)(int32_t)ret` for values in `[INT_MIN, 0)`
- Return `(uint32_t)ret` for values in `[0, UINT32_MAX]`

As far as I can tell, casting from double to char (either signedness) always goes via converting to a signed int, then just extracting the low byte. So yes, we should be using the signed version there.

…Which also means the unsigned version isn't actually needed at all, is it? So I can just drop all that complexity.

rcombs:
> Any details on why you thought the FreeType-provided family was beneficial for compatibility originally but this turned out to be false?

Looking back on the fonts I was experimenting with at the time, I'm pretty sure I'd confused something in Open Sans with something in Noto Sans.

> I seem to recall a user complain about it, but I can’t find that comment now (perhaps I’m thinking of something else).

It was me, I'd been poking at fonts and saw behavior that I thought mismatched vsfilter, but it was just because I'd mixed things up when testing. Whoops.

libass/ass_fontconfig.c
```diff
@@ -109,23 +110,26 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
...
+#elif FC_VERSION >= 21191
+        // Versions prior to 2.12.92 only had integer precision.
+        meta.weight = FcWeightToOpenType(weight + 0.5) + 0.5;
 #else
         // On older fontconfig, FcWeightToOpenType is unavailable, and its inverse was
         // implemented more simply, using an if/else ladder instead of linear interpolation.
         // We implement an inverse of that ladder here.
         // We don't expect actual FC caches from these versions to have intermediate
         // values, so the average checks are only for completeness.
-#define AVG(x, y) ((x + y) / 2)
```
astiob, line 128: I’d have written this as `(x + y) * 0.5`. But it makes no difference given that we only ever call this with small integer constant expressions.

astiob (approved):

PR #430, opened 2020-09-19
Remove redundant includes

MrSmile:
Libass has too many redundant include directives, so I attempted to fix that.


libass/ass.c
```diff
@@ -19,20 +19,12 @@
 #include "config.h"
 #include "ass_compat.h"
 
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-#include <assert.h>
```
TheOneric, line 25: `ass.c` does directly use `assert`, so it should imho also include `assert.h`. This is not the only occurence.
As I explained in https://github.com/libass/libass/pull/479#discussion_r600879938, I think relying on transitive includes is prone to break, but I agree that libass' includes iwould benefit from some cleanup, especially the header files.

astiob: My impression (from previous discussion) is that @rcombs and I both agree with TheOneric that ideally, each file should include precisely the headers that declare the names it uses and avoid relying on transitive includes, while MrSmile is unfortunately alone in a minority that prefers for a few common headers to include everything we’ll need more than once and for other files to just include those fat headers.

And to be fair, I can see the appeal of that, too; and that’s how some of our dependency libraries work: e. g. we include `<fribidi.h>` and `<hb.h>` to get all of FriBidi and HarfBuzz at once, and we are spared knowing that internally they have _lots_ of modules, each with its own headers.

Of course, the real answer is somewhere in the middle: each header intentionally publishes some names—and some of those names _may_ originate from transitive includes!—and then includes some other names that are only needed inside the header but aren’t intended to be republished _by_ this header to its own consumers. (I may be starting to understand why C++ wanted to add “modules”.) MrSmile prefers to have a shared header that re-exports all common includes, whereas the other camp prefers to see each file’s dependencies in itself.

MrSmile: > I think relying on transitive includes is prone to break

Such a break is visible immediately during compilation and quite easy to fix. On the other hand, the rule of "including all the necessary stuff directly" is really hard to enforce and prone to silent breaks. So I think that rule of "removing includes until compilation breaks" is superior.

> while MrSmile is unfortunately alone in a minority that prefers for a few common headers to include everything we’ll need more than once and for other files to just include those fat headers.

This PR isn't really about common headers, but more in the line of "if we already have fat headers and use it then don't repeat includes from such headers".

TheOneric: > On the other hand, the rule of "including all the necessary stuff directly" is really hard to enforce and prone to silent breaks.

It's unlikely for "silent breaks" to happen if the headers only use minimal includes. As I pointed out before, headers only using minimal includes also has positive effects on building. In the rare event, some transitive dependency sneaked in at it leads to a break it can still be fixed easily. If transitive dependencies are actively relied upon however, not only are we losing build-advantages, but it will be common for commits removing includes from one place, because it's no longer needed there, to also require changes in ""unrelated"" files that reliead on the transitive include, polluting diffs.

> "if we already have fat headers and use it then don't repeat includes from such headers"

I wouldn't say we're really using fat headers right now, they arent minimal, but a lot of stuff is also included in the source files. Currently headers and includes are neither fat nor minimal, just messy.

PR #432, opened 2020-09-27
Fix overflows in outline processing

### LINKED ISSUE #431: Bug in  ass_outline.c:1354: _Bool outline_stroke(ASS_Outline *, ASS_Outline *, const ASS_Outline *, int, int, int): Assertion `rad >= eps' failed.
fuzzer & poc
[libass.zip](https://github.com/libass/libass/files/5286727/libass.zip)

gdb:
```
fstark@fstark-virtual-machine:~/libass$ gdb ./libass_fuzzer
GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./libass_fuzzer...done.
(gdb) r out/fuzz5/crashes/id\:000000\,sig\:06\,src\:000046+000020\,time\:11326439\,op\:splice\,rep\:128 
Starting program: /home/fstark/libass/libass_fuzzer out/fuzz5/crashes/id\:000000\,sig\:06\,src\:000046+000020\,time\:11326439\,op\:splice\,rep\:128
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
======================= INFO =========================
This binary is built for AFL-fuzz.
To run the target function on individual input(s) execute this:
  /home/fstark/libass/libass_fuzzer < INPUT_FILE
or
  /home/fstark/libass/libass_fuzzer INPUT_FILE1 [INPUT_FILE2 ... ]
To fuzz with afl-fuzz execute this:
  afl-fuzz [afl-flags] /home/fstark/libass/libass_fuzzer [-N]
afl-fuzz will run N iterations before re-spawning the process (default: 1000)
======================================================
Reading 11249 bytes from out/fuzz5/crashes/id:000000,sig:06,src:000046+000020,time:11326439,op:splice,rep:128
libass_fuzzer: ass_outline.c:1354: _Bool outline_stroke(ASS_Outline *, ASS_Outline *, const ASS_Outline *, int, int, int): Assertion `rad >= eps' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff6efa428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
(gdb) t
[Current thread is 1 (Thread 0x7ffff7fdb780 (LWP 25974))]
(gdb) bt
#0  0x00007ffff6efa428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff6efc02a in __GI_abort () at abort.c:89
#2  0x00007ffff6ef2bd7 in __assert_fail_base (fmt=<optimized out>, 
    assertion=assertion@entry=0x66d940 <.str.3> "rad >= eps", 
    file=file@entry=0x66d820 <.str.1> "ass_outline.c", line=line@entry=1354, 
    function=function@entry=0x66d980 <__PRETTY_FUNCTION__.outline_stroke> "_Bool outline_stroke(ASS_Outline *, ASS_Outline *, const ASS_Outline *, int, int, int)") at assert.c:92
#3  0x00007ffff6ef2c82 in __GI___assert_fail (assertion=0x66d940 <.str.3> "rad >= eps", 
    file=0x66d820 <.str.1> "ass_outline.c", line=1354, 
    function=0x66d980 <__PRETTY_FUNCTION__.outline_stroke> "_Bool outline_stroke(ASS_Outline *, ASS_Outline *, const ASS_Outline *, int, int, int)") at assert.c:101
#4  0x000000000050911f in outline_stroke () at ass_outline.c:1354
#5  0x00000000004de968 in ass_outline_construct () at ass_render.c:1222
#6  0x000000000052327d in ass_cache_get () at ass_cache.c:404
#7  0x00000000004f2bb8 in get_bitmap_glyph () at ass_render.c:1456
#8  0x00000000004ed5db in render_and_combine_glyphs () at ass_render.c:2359
#9  0x00000000004e39e9 in ass_render_event () at ass_render.c:2787
#10 0x00000000004e17d7 in ass_render_frame () at ass_render.c:3153
#11 0x00000000004cc94d in LLVMFuzzerTestOneInput () at /src/libass_fuzzer.cc:45
#12 0x00000000004ccf5f in ExecuteFilesOnyByOne () at /src/libfuzzer/afl/afl_driver.cpp:217
#13 main () at /src/libfuzzer/afl/afl_driver.cpp:254
```

MrSmile:
I've added a couple of checks for overflows in outline pipeline. Should fix #431, but I'm not familiar with that fuzzing business so I can't ascertain it for now.

### DIFF
--- a/libass/ass_outline.c
+++ b/libass/ass_outline.c
@@ -48,6 +48,11 @@ static void outline_clear(ASS_Outline *outline)
     outline->n_segments = outline->max_segments = 0;
 }
 
+static bool valid_point(const FT_Vector *pt)
+{
+    return abs(pt->x) <= OUTLINE_MAX && abs(pt->y) <= OUTLINE_MAX;
+}
+
 bool outline_convert(ASS_Outline *outline, const FT_Outline *source)
 {
     if (!source || !source->n_points) {
@@ -77,12 +82,16 @@ bool outline_convert(ASS_Outline *outline, const FT_Outline *source)
             continue;
         }
 
+        if (!valid_point(source->points + j))
+            goto fail;
         switch (FT_CURVE_TAG(source->tags[j])) {
         case FT_CURVE_TAG_ON:
             st = S_ON;
             break;
 
         case FT_CURVE_TAG_CONIC:
+            if (!valid_point(source->points + last))
+                goto fail;
             pt.x =  source->points[last].x;
             pt.y = -source->points[last].y;
             switch (FT_CURVE_TAG(source->tags[last])) {
@@ -110,6 +119,8 @@ bool outline_convert(ASS_Outline *outline, const FT_Outline *source)
         outline->points[outline->n_points++] = pt;
 
         for (j++; j <= last; j++) {
+            if (!valid_point(source->points + j))
+                goto fail;
             switch (FT_CURVE_TAG(source->tags[j])) {
             case FT_CURVE_TAG_ON:
                 switch (st) {
@@ -209,6 +220,23 @@ bool outline_scale_pow2(ASS_Outline *outline, const ASS_Outline *source,
         return true;
     }
 
+    int32_t lim_x = OUTLINE_MAX;
+    if (scale_ord_x > 0)
+        lim_x = scale_ord_x < 32 ? lim_x >> scale_ord_x : 0;
+    else
+        scale_ord_x = FFMAX(scale_ord_x, -32);
+
+    int32_t lim_y = OUTLINE_MAX;
+    if (scale_ord_y > 0)
+        lim_y = scale_ord_y < 32 ? lim_y >> scale_ord_y : 0;
+    else
+        scale_ord_y = FFMAX(scale_ord_y, -32);
+
+    if (!lim_x || !lim_y) {
+        outline_clear(outline);
+        return false;
+    }
+
     if (!outline_alloc(outline, source->n_points, source->n_segments))
         return false;
 
@@ -216,6 +244,10 @@ bool outline_scale_pow2(ASS_Outline *outline, const ASS_Outline *source,
     int sy = scale_ord_y + 32;
     const ASS_Vector *pt = source->points;
     for (size_t i = 0; i < source->n_points; i++) {
+        if (abs(pt[i].x) > lim_x || abs(pt[i].y) > lim_y) {
+            outline_free(outline);
+            return false;
+        }
         // that's equivalent to pt[i].x << scale_ord_x,
         // but works even for negative coordinate and/or shift amount
         outline->points[i].x = pt[i].x * ((int64_t) 1 << sx) >> 32;
@@ -244,6 +276,10 @@ bool outline_transform_2d(ASS_Outline *outline, const ASS_Outline *source,
         for (int k = 0; k < 2; k++)
             v[k] = m[k][0] * pt[i].x + m[k][1] * pt[i].y + m[k][2];
 
+        if (!(fabs(v[0]) < OUTLINE_MAX && fabs(v[1]) < OUTLINE_MAX)) {
+            outline_free(outline);
+            return false;
+        }
         outline->points[i].x = lrint(v[0]);
         outline->points[i].y = lrint(v[1]);
     }
@@ -271,8 +307,15 @@ bool outline_transform_3d(ASS_Outline *outline, const ASS_Outline *source,
             v[k] = m[k][0] * pt[i].x + m[k][1] * pt[i].y + m[k][2];
 
         double w = 1 / FFMAX(v[2], 0.1);
-        outline->points[i].x = lrint(v[0] * w);
-        outline->points[i].y = lrint(v[1] * w);
+        v[0] *= w;
+        v[1] *= w;
+
+        if (!(fabs(v[0]) < OUTLINE_MAX && fabs(v[1]) < OUTLINE_MAX)) {
+            outline_free(outline);
+            return false;
+        }
+        outline->points[i].x = lrint(v[0]);
+        outline->points[i].y = lrint(v[1]);
     }
     memcpy(outline->segments, source->segments, source->n_segments);
     outline->n_points = source->n_points;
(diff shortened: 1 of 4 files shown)


TheOneric:
Can confirm that this fixes #431. 

Overflow checks seem fine to me. *(Assuming `abs(OUTLINE_MAX) < abs(OUTLINE_MIN)` will stay true in future version, which it probably will, I guess)*

MrSmile:
> Assuming `abs(OUTLINE_MAX) < abs(OUTLINE_MIN)` will stay true in future version, which it probably will, I guess

Maybe it's even better to get rid of `OUTLINE_MIN` altogether and use `-OUTLINE_MAX` instead.

libass/ass_outline.c
```diff
@@ -209,13 +220,34 @@ bool outline_scale_pow2(ASS_Outline *outline, const ASS_Outline *source,
...
+    if (scale_ord_x > 0)
+        lim_x = scale_ord_x < 32 ? lim_x >> scale_ord_x : 0;
+    else
+        scale_ord_x = FFMAX(scale_ord_x, -32);
+
+    int32_t lim_y = OUTLINE_MAX;
+    if (scale_ord_y > 0)
+        lim_y = scale_ord_y < 32 ? lim_y >> scale_ord_y : 0;
+    else
+        scale_ord_y = FFMAX(scale_ord_y, -32);
```
astiob, line 233: Doesn’t this (if it ever is a no-op) affect the numbers and cause the outline to be scaled out of proportion? Shouldn’t this clamping be done on `k->scale_ord_x` as soon as it is obtained in `get_bitmap_glyph`?

MrSmile: Negative scale is an equivalent to the arithmetic shift right and such a shift of `int32_t` value reduces to the sign bit for shift amounts from 31 and up. So `-32` here can even be replaced with `-31` without any influence on the result.

MrSmile:
New, more strict version. `max(abs(x), abs(y)) <= OUTLINE_MAX` is now enforced invariant of `ASS_Outline` instead of some ad hoc check. Also I've got rid of `OUTLINE_MIN`.

libass/ass_outline.c
```diff
@@ -244,6 +276,10 @@ bool outline_transform_2d(ASS_Outline *outline, const ASS_Outline *source,
         for (int k = 0; k < 2; k++)
             v[k] = m[k][0] * pt[i].x + m[k][1] * pt[i].y + m[k][2];
 
+        if (!(fabs(v[0]) <= OUTLINE_MAX && fabs(v[1]) <= OUTLINE_MAX)) {
```
TheOneric, line 279: Nit-pick: When comparing with floating types OUTLINE_MAX will be cast to floating type as well — but an accurate representation may not necessarily exist. Checking `<=` instead of `<` to avoid overflow can thus be potentially unsafe.
After a quick glance it seems like all float-comparisons here are done with doubles. On probably most systems `double` is 64 bit and able to accurately represent the current value of `OUTLINE_MAX`. 

Still, imho it would be better to stick to `<` when floating types are involved to account for odd systems and potential future changes of `OUTLINE_MAX` *(or is this specific value required for some VSFilter-compatibility?)*.

With floats this would already be problem on my machine:
```C
printf("%d\n", ((float) (OUTLINE_MAX+1)) <= OUTLINE_MAX);  // 1
printf("%d\n", ((double)(OUTLINE_MAX+1)) <= OUTLINE_MAX);  // 0
```

*same applies to ass_outline.c:313 and ass_render.c:1443*

MrSmile: Done.

> is this specific value required for some VSFilter-compatibility?

No, it's to avoid overflow in cubic spline splitting, `(p0 + 3 * p1 + 3 * p2 + p3) / 8`.

astiob:
@MrSmile 

```
ass_outline.c:53:12: warning: absolute value function 'abs' given an argument of type 'const FT_Pos' (aka 'const long') but has parameter of type 'int' which may cause truncation of value [-Wabsolute-value]
    return abs(pt->x) <= OUTLINE_MAX && abs(pt->y) <= OUTLINE_MAX;
           ^
ass_outline.c:53:12: note: use function 'labs' instead
    return abs(pt->x) <= OUTLINE_MAX && abs(pt->y) <= OUTLINE_MAX;
           ^~~
           labs
ass_outline.c:53:41: warning: absolute value function 'abs' given an argument of type 'const FT_Pos' (aka 'const long') but has parameter of type 'int' which may cause truncation of value [-Wabsolute-value]
    return abs(pt->x) <= OUTLINE_MAX && abs(pt->y) <= OUTLINE_MAX;
                                        ^
ass_outline.c:53:41: note: use function 'labs' instead
    return abs(pt->x) <= OUTLINE_MAX && abs(pt->y) <= OUTLINE_MAX;
                                        ^~~
                                        labs
ass_parse.c:52:25: warning: unused function 'argtoll' [-Wunused-function]
static inline long long argtoll(struct arg arg)
                        ^
```

MrSmile:
FT docs specify that `FT_Pos` is 16.16 or 26.6 fixed-point representation, so I'm not sure what's better, to cast into `int32_t` or use `labs` here.

astiob:
Just use `labs`. In a way, this will also serve as a sanity check to ensure the FT docs aren’t lying.

Issue #414, opened 2020-07-10
Memory Leak if ass_font_new fails


TheOneric:
Use sample from #412 and render at 2ms and do not provide eg `DejaVu Sans` in test directory.
If compiled with ASAN, memleaks will be reported:
```
=================================================================
==19318==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 272 byte(s) in 1 object(s) allocated from:
    #0 0x7f8ac5bd6330 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
    #1 0x56441edb4544 in ass_cache_get …/libass/libass/ass_cache.c:391

Direct leak of 12 byte(s) in 1 object(s) allocated from:
    #0 0x7f8ac5b27810 in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x3a810)
    #1 0x56441ed2c6b3 in update_font …/libass/libass/ass_parse.c:107
    #2 0x56441ede9909  (…/libass/compare/compare+0x298909)

Indirect leak of 15 byte(s) in 1 object(s) allocated from:
    #0 0x7f8ac5b27810 in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.5+0x3a810)
    #1 0x56441ed2c6b3 in update_font …/libass/libass/ass_parse.c:107
    #2 0x56441ede9909  (…/libass/compare/compare+0x298909)
```
~~Not sure if this is only an issue with compare or whole libass.~~ *(Should mostly affect `FONTPROVIDER_NONE`)

TheOneric:
After looking a bit more into this:
This will probably always happen when ass_font_new fails — which I guess does not happen with the system font provider, which will just return a fallback instead (at least with fontconfig). With `ASS_FONTPROVIDER_NONE` this seems to be the case for every font not explicitly added by the API user.

Applying PR #402 will result in segfaults (heap-use-after-free).

MrSmile:
The whole problem arises from using `font->desc.family` as a failure flag. `ASS_FontDesc` cache key has an ownership of the family string and most cache operations treat it as such, but not final `font_destruct()`. Destruction is done through internal copy of `ASS_FontDesc` inside of `ASS_Font` structure, but in case of failure that copy of `family` is `NULL`.

A simple solution here will be to add the line
```c
((ASS_Font *) value)->desc.family = ((ASS_FontDesc *) key)->family;
```
[...]

PR #433, opened 2020-10-10
font: use library as a failure flag

MrSmile:
I think this is the most straightforward fix of #414.
I've tried to tinker with `desc` field of `ASS_Font` structure but that leads to much more ugly code.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -245,7 +245,7 @@ ASS_Font *ass_font_new(ASS_Renderer *render_priv, ASS_FontDesc *desc)
     ASS_Font *font = ass_cache_get(render_priv->cache.font_cache, desc, render_priv);
     if (!font)
         return NULL;
-    if (font->desc.family)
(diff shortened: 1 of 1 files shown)


astiob (approved):

rcombs (approved):

astiob:
Just to be sure: has #414 been retested with this? I tried, but as this was my first time running ASAN libass, I spent a lot of time tinkering to get it working, only to end up with this:
```
AddressSanitizer: detect_leaks is not supported on this platform.
```
🤪

MrSmile:
I've tried with valgrind and leak with missing font and `ASS_FONTPROVIDER_NONE` goes away. Not sure about exact sample from #414, but should work too.

TheOneric:
I've retested the (probably) same sample. Without the patch there's a leak, with it applied there's none.

astiob:
Awesome, thanks! Merging this.

PR #434, opened 2020-10-16
Fix Banner/Scroll effects & simplify clip coordinate conversion

astiob:
Both in one PR because the “Support Banner/Scroll effects with \pos/\move” patch assumes the clip coordinate conversion patch has been applied first.

Labelling this “cosmetic” for the clip coordinate patch; “compatibility” for all the rest.

All banner/scroll changes have been tested and verified in practice against xy-VSFilter and MPC-HC.


rcombs (approved):
Can't say I'm especially familiar with these effects (nor the code implementing them) but this all looks fine to me.

astiob:
Reworded the “Fix Scroll effect coordinates” commit; no code changes.

PR #436, opened 2020-10-18
Split glyph runs early and fix some run splitting bugs

astiob:
The Holy Grail of glyph runs is (almost) here. This wasn’t even hard, so far.

My next PR will be karaoke that makes use of the early-detected runs.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1858,11 +1858,9 @@ fix_glyph_scaling(ASS_Renderer *priv, GlyphInfo *glyph)
   */
 static bool is_new_bm_run(GlyphInfo *info, GlyphInfo *last)
 {
-    // FIXME: Don't break on glyph substitutions
     return !last || info->effect || info->drawing_text || last->drawing_text ||
         strcmp(last->font->desc.family, info->font->desc.family) ||
         last->font->desc.vertical != info->font->desc.vertical ||
-        last->face_index != info->face_index ||
         last->font_size != info->font_size ||
         last->c[0] != info->c[0] ||
         last->c[1] != info->c[1] ||
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1582,6 +1582,8 @@ static void trim_whitespace(ASS_Renderer *render_priv)
         cur->is_trimmed_whitespace = true;
         cur = ti->glyphs + ++i;
     }
+    if (i < ti->length)
+        cur->starts_new_run = true;
 
     // Mark all extraneous whitespace inbetween
     for (i = 0; i < ti->length; ++i) {
@@ -1610,6 +1612,8 @@ static void trim_whitespace(ASS_Renderer *render_priv)
                 }
                 i = j - 1;
             }
+            if (cur < ti->glyphs + ti->length)
+                cur->starts_new_run = true;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


astiob:
Gah, I messed up breaking runs on karaoke while splitting up the PRs. I'll fix this.

astiob:
Karaoke mishap fixed.

Added a commit replacing `emblevel % 2` by `FRIBIDI_LEVEL_IS_RTL(emblevel)`: the less we have to know about it, the better.

Ready for review.

libass/ass_render.c
```diff
@@ -1850,40 +1851,42 @@ fix_glyph_scaling(ASS_Renderer *priv, GlyphInfo *glyph)
...
-        last->border_style != info->border_style ||
-        last->border_x != info->border_x ||
-        last->border_y != info->border_y ||
-        last->hspacing != info->hspacing ||
-        last->italic != info->italic ||
-        last->bold != info->bold ||
-        ((last->flags ^ info->flags) & ~DECO_ROTATE);
+    for (int i = 1; i < render_priv->text_info.length; i++) {
+        GlyphInfo *info = render_priv->text_info.glyphs + i;
+        GlyphInfo *last = render_priv->text_info.glyphs + (i - 1);
```
MrSmile, line 1859: I think that `last` is misleading here, shouldn't it be `prev` or something?

MrSmile (approved):
LGTM.

On a side note, `GlyphInfo` currently is a huge mess. At some point in time we should merge that bunch of boolean flags together into one (and also eliminate use of pointers as boolean flags).

libass/ass_render.c
astiob: Well, “last” can also mean “prev”, and here I tried to keep as much code unmodified as I could, barring the indentation change. We use `last...` for “most recent X” in a few places, although some of them skip some elements so their “most recent” is not always the immediately preceding element. But e. g. in `ass_shaper_find_runs` it actually is.

astiob:
Pushed a two-line change to drop the `effect` field from `GlyphInfo` as it’s no longer used.

astiob:
Pushed a few changes:

1. “don’t mingle run ID and bidi embedding level” was somehow missing the hunk that would actually remove the addition. Now it is in place.

2. `glyphs[0].starts_new_run` is now true. I’m conflicted on whether I should rename the field to `starts_run`.

3. Shaping can mark the `starts_new_run` glyph infos with `skip`, so to ensure the run breaks are not missed when combining bitmaps (and in future in other places), it now sets `starts_new_run` on the next nonskipped glyph info.

4. A couple extra fixes to ensure drawings are absolutely not shaped by HarfBuzz.

libass/ass_shaper.c
```diff
@@ -956,6 +956,16 @@ int ass_shaper_shape(ASS_Shaper *shaper, TextInfo *text_info)
         ass_shaper_skip_characters(text_info);
 #endif
 
+    // Make sure we don't lose any run breaks due to skipped glyphs
+    for (i = 0; i < text_info->length; i++) {
```
MrSmile, line 935: It should be possible to rearrange checks without nested loops. Something like
```C
bool new_run = true;
for (int i = 0; i < text_info->length; i++) {
    if (glyphs[i].starts_new_run)
        new_run = true;
    if (glyphs[i].skip)
        continue;
    glyphs[i].starts_new_run = new_run;
    new_run = false;
}
```

libass/ass_shaper.c
```diff
@@ -639,18 +639,20 @@ static void shape_harfbuzz(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
...
         hb_font_t *font = get_hb_font(shaper, glyphs + offset);
-        int level = glyphs[offset].shape_run_id;
-        int direction = FRIBIDI_LEVEL_IS_RTL(shaper->emblevels[offset]);
+        int run_id = glyphs[offset].shape_run_id;
+        int level = shaper->emblevels[offset];
 
         // advance in text until end of run
-        while (i < (len - 1) && level == glyphs[i+1].shape_run_id)
+        while (i < (len - 1) && run_id == glyphs[i+1].shape_run_id &&
+                level == shaper->emblevels[i+1])
```
MrSmile, line 652: Shouldn't it be `[i + 1]` per libass style?

libass/ass_shaper.c
astiob: Is that better though? It seems harder to understand.

libass/ass_shaper.c
astiob: Changed both instances to `[i + 1]`.

astiob:
Added a commit to fix UB pointer decrement, which was most likely innocent in practice.

libass/ass_shaper.c
MrSmile: For me it's instead simpler to understand than nested loops with a common index variable, but feel free to ignore me.
I think that triple copy-paste of condition `i < text_info->length` isn't that great.

astiob:
Sorry for wasting a bit of review effort, but… I’ve rolled back this:

> to ensure the run breaks are not missed when combining bitmaps (and in future in other places), [shaping] now sets `starts_new_run` on the next nonskipped glyph info [when it marks the `starts_new_run` glyph infos with `skip`]

I originally thought this would be easier to work with, but now I’ve realized that it makes karaoke unable to distinguish between run-leading-skipped-infos-within-a-single-run and multiple-proper-run-breaks-with-only-skipped-glyphs-in-between.

So now, instead, `starts_new_run` is set only on the first glyph of each run, even if that glyph is skipped. The bitmap combining loop remembers this flag across skipped glyphs, just like it used to remember the `linebreak` flag before this PR.

Further, to avoid adding new bad code only to fix it in the upcoming karaoke PR, I’ve redone the way line breaks affect `starts_new_run`. Previously in this PR, `wrap_lines_smart` set `starts_new_run` on the first nonskipped glyph of the new line after each line break. This was an arbitrary choice I made before considering how this interacts with karaoke, although a pretty good one. But for karaoke compatibility, `starts_new_run` must be set on the first _non-trimmed-whitespace_ glyph of the line (which may still be skipped due to shaping), _including the first line._ So I’ve moved the code from `wrap_lines_smart` to `trim_whitespace` and adjusted accordingly.

Only the “Split glyph runs early” commit has changed, but please re-review it.

Finally: doesn’t the last loop in `wrap_lines_smart` write past the end of a buffer if the last line is entirely whitespace? https://github.com/libass/libass/blob/0aec12b3576e93f8387992de73bca99ab6e50d2f/libass/ass_render.c#L1760-L1774

libass/ass_render.c
```diff
@@ -1850,40 +1854,46 @@ fix_glyph_scaling(ASS_Renderer *priv, GlyphInfo *glyph)
...
-        last->border_x != info->border_x ||
-        last->border_y != info->border_y ||
-        last->hspacing != info->hspacing ||
-        last->italic != info->italic ||
-        last->bold != info->bold ||
-        ((last->flags ^ info->flags) & ~DECO_ROTATE);
+    for (int i = 0; i < render_priv->text_info.length; i++) {
+        GlyphInfo *info = render_priv->text_info.glyphs + i;
+        if (!i) {
+            info->starts_new_run = true;
```
MrSmile, line 1863: Is there a need for that? First run is handled by `bool new_run = true;` later.

astiob: Not for bitmap combining, no. But it may make other future uses of this flag simpler: for one, it makes karaoke code a _tiny_ bit simpler. Also, for reference, karaoke will need an `if (!i)` section here in any case.

Also, the flag will anyway be set on the first glyph by `trim_whitespace` if the first line has no leading whitespace. So it wouldn’t be wise to have any code relying on the _absence_ of this flag for the first glyph, either.

But if you prefer to avoid setting this on the first glyph explicitly, I don’t mind removing this line and adding a `|| !i` in karaoke later. We’ll need to keep in mind that the flag _may or may not_ be set on the first glyph, then.

MrSmile: I think it's better to keep `i = 0` case in this function but outside of this loop.

astiob: I missed the fact that `text_info.length` is guaranteed to be nonzero. (This is checked just before the call to `split_style_runs`.)

I’ve now moved this assignment out of the loop.

astiob:
I’ve largely finished my karaoke changes (some of which build on top of this), and I think I’m happy with this PR as it is now.

MrSmile:
I don't see any more problems, so if there are no issues with test samples you can merge that.

Issue #357, opened 2019-11-05
Karaoke effects \k and \kf are lagging/desynced


AxelTerizaki:
I posted this originally on mpv-player/mpv#7121 but they told me I should post it here as well, since it seems to be an upstream issue with libass.

As the title says, \kf and \k effects are slightly desynced, making karaoke less enjoyable :)

I'll copy-paste the issue, since it contains example files, and captures. 

This happens with mpv 0.30, but not with mpv 0.29, which I suppose do not use the same libass (but I have no idea which version they do use, maybe someone from mpv can tell us here.)

### Reproduction steps

Load the sample files listed below and see for yourself how the karaoke subs are desynced.

### Expected behavior

Captured using mpv 0.29 : 

https://cdn.discordapp.com/attachments/324208228680466434/640862388169736219/2019-11-04_11-36-40.mp4

### Actual behavior

Captured using mpv 0.30

https://cdn.discordapp.com/attachments/324208228680466434/640858951549714432/2019-11-04_11-22-27.mp4

As you can see the subtitles effect \k / \kf aren't synced properly anymore, as if there was some kind of lag.

### Log file

Here is a mpv log (not sure if this is relevant here for libass, but I'm keeping it just in case) : 

https://pastebin.com/AmzdQrbr

### Sample files

- Video : https://kara.moe/downloads/medias/ENG%20-%20Dokidoki!%20PreCure%20-%20OP%20-%20Glitter%20Force%20Doki%20Doki%20Theme%20Song.mp4
- Subfile : https://kara.moe/downloads/lyrics/ENG%20-%20Dokidoki!%20PreCure%20-%20OP%20-%20Glitter%20Force%20Doki%20Doki%20Theme%20Song.ass

Thanks in advance for your help!

wm4:
Here's a better description of the bug: https://github.com/mpv-player/mpv/issues/7121#issuecomment-549511652

Akemi:
bisected this problem.
last working commit: https://github.com/libass/libass/commit/0f0068b0ae5b0bad8472d4a2af6d10e97f982bfc
first broken commit: https://github.com/libass/libass/commit/d3286d8a5e3a0186457b8596c2809866fa59d052

Akemi:
also #343 fixes this issue.

AxelTerizaki:
[...]

Issue #216, opened 2016-03-24
A rendering issue in newer version of libass


esterTion:
When I was working on a subtitle of an anime, I found some weird rendering issue. I was using ffmpeg with libass 0.12.3, and vertical subtitles with karaoke effect rendering was broken, the character didn't appear in order. 
I tried recompile ffmpeg with libass 0.13.2 but the issue is still there. So I downgrade to libass 0.11.2, problem solved.
Not sure which code cause the problem, I'm not a pro about this. 
I've put the resources and ass I used in a repository [here](https://github.com/lcz970/libass_issue)

aphirst:
I can confirm this issue too. How come you distributed the sample broken up but not as one cohesive unit? (just curious :P)

lcz970:
Not that good in English, so I'm not sure what you are asking. :(
If you are asking about cutting a slice from source, I was using ffmpreg. First copy streams and cut a big range to prevent missing of I-frame, then reencode the part you want.

TheOneric:
This seems to be fixed in recent git master (not fixed in release 0.14.0).

TheOneric:
Taking a closer look at the first karaoke line (Start at `0:00:00.03` and the rendering result at `0:00:02.252` (frame 54): 
```
{\fad(0,100)\blur2}{\K37}そ{\K81}っと{\K65}開{\K39}い{\K40}た{\K20}ド{\K22}ア{\K58}の{\K25}向{\K36}こ{\K43}う{\K36}に
```

In libass 0.11, the い was roughly half-filled. *(used provided hardsub)*  
In current libass master, the た following the い is roughly half-filled.  
In libass-master with patch from #343 , the い is fully filled and た not yet filled.

What **should** happen:  
The い should be fully filled at `(37+81+65+39)*10 + 3  = 0:00:02.223` and the た should be fully filled at `(37+81+65+39+40)*10 + 3  = 0:00:02.623`. Meaning at the frame in question い should already have finished filling and た should just have started.

Only master with #343 yields the correct result here.  

![comparison](https://user-images.githubusercontent.com/1668471/84580081-f77b9e80-add3-11ea-9386-843f0aa8746c.png)

PR #438, opened 2020-10-22
Karaoke fixes & improvements

astiob:
OK, sooo…

* This makes karaoke quite sane in the common case where there are no 3D transforms, fixing our regressions and old bugs. So it should now mostly Just Work in practice.
    * The behaviour of \kf may be surprising if the text has serifs extending far to the left or far to the right: the karaoke fill line instantly jumps to the first glyph’s origin point when karaoke starts, moves smoothly to the point where the next-beyond-last glyph’s origin would be, and instantly jumps to +infinity when karaoke ends.
* This makes karaoke more compatible with VSFilter, fixing our handling of edge cases, adding special cases that we didn’t handle at all, and reproducing VSFilter’s handling of karaoke on glyph run boundaries.

However, the various flavours of VSFilter handle \kf differently, and weirdly, when additional effects are involved:

* This makes our \kf _pretty close_ to guliverkli/Guliverkli2/xy-VSFilter/XySubFilter’s on text set in plain fonts with xbord ≥ ybord, without any blur, and with or without 3D transforms.
* With 3D transforms, MPC-HC’s \kf [looks different](https://chortos.selfip.net/~astiob/screens/karaoke-transformed/).
    * This appears saner at first.
    * But add border and blur, and MPC-HC’s \kf becomes absurd again.
* This makes our \kf _not_ match _any_ VSFilter when the font has serifs extending far to the left or far to the right, or when the vector drawing’s minimum x coordinate is nonzero, or when ybord > xbord, or with \blur, or with large \be in xy-VSFilter/XySubFilter (or be shifted by 1px with \be in Guliverkli2/MPC-HC).
    * Also, the more of these effects are involved, the more VSFilter’s \kf may be affected by rounding error from floor & ceil, causing it to shift left by up to 2 or 3 pixels. But note that XySubFilter does not attempt to scale this rounding error from video resolution. (It seems xy simply didn’t notice these problems with \kf, as he’s also changed its behaviour with \blur and \be.)
    * And Guliverkli2, xy-VSFilter/XySubFilter and MPC-HC all handle \kf with blur differently! (guliverkli doesn’t support blur. Except \be1, in which case there’s a 1px difference from the other VSFilters.)

This _can_ be emulated. We’d need to pick one VSFilter flavour to emulate though; we can’t do all at once. For example, [here](https://github.com/astiob/libass/commit/2226dd348f24f2300f23dbcadaa38101e20d35ca)’s an imperfect attempt at closely emulating Guliverkli2.

Two of these quirks are common to all VSFilters, and we could at least emulate those:

* Start the fill (in absence of blur, border etc.) at the glyph run’s leftmost control point after 3D transforms, floored, rather than the first glyph’s origin. Our quickest way to emulate this would be to use `bm->left`, which is roughly the leftmost point. (@MrSmile’s #343 does this, inadvertently.) It’s not the leftmost _control_ point though—there’s a difference for cubic curves—and it’s shifted by 1/64px before flooring, and its floor is computed in display resolution.
* Start the fill another (\ybord−\xbord) pixels to the left, if \ybord > \xbord. Plus rounding error, even if \xbord ≤ \ybord. Our quickest way to emulate this would be to ignore the rounding error and subtract (\ybord−\xbord).

If we really wanted, we could even emulate all of the shared rounding quirks.

Blur is where the various VSFilters really diverge. Blur is a late addition to ASS, so perhaps it’s not a surprise. Guliverkli2 and xy-VSFilter/XySubFilter shift the \kf fill to the left again, but by different amounts. MPC-HC attempts to do it better but only ends up being more bizarre, starting the fill more to the left but progressing it faster. None of them act sanely by any measure.

### DIFF
--- a/libass/ass_outline.c
+++ b/libass/ass_outline.c
@@ -323,6 +323,20 @@ bool outline_transform_3d(ASS_Outline *outline, const ASS_Outline *source,
     return true;
 }
 
+void outline_update_min_transformed_x(const ASS_Outline *outline,
+                                      const double m[3][3],
+                                      int32_t *min_x) {
+    const ASS_Vector *pt = outline->points;
+    for (size_t i = 0; i < outline->n_points; i++) {
+        double z = m[2][0] * pt[i].x + m[2][1] * pt[i].y + m[2][2];
+        double x = (m[0][0] * pt[i].x + m[0][1] * pt[i].y + m[0][2]) / FFMAX(z, 0.1);
+        if (isnan(x))
+            continue;
+        int32_t ix = lrint(FFMINMAX(x, -OUTLINE_MAX, OUTLINE_MAX));
+        *min_x = FFMIN(*min_x, ix);
+    }
+}
+
 
 void outline_free(ASS_Outline *outline)
 {
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -975,16 +975,22 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
     int tm_current = render_priv->time - render_priv->state.event->Start;
 
     int timing = 0;
+    Effect effect_type = EF_NONE;
     GlyphInfo *last_boundary = NULL;
     for (int i = 0; i <= render_priv->text_info.length; i++) {
         if (i < render_priv->text_info.length &&
-            render_priv->text_info.glyphs[i].effect_type == EF_NONE)
+            !render_priv->text_info.glyphs[i].starts_new_run)
             continue;
 
         GlyphInfo *start = last_boundary;
         GlyphInfo *end = render_priv->text_info.glyphs + i;
         last_boundary = end;
         if (!start)
+            continue;
+
+        if (start->effect_type != EF_NONE)
+            effect_type = start->effect_type;
+        if (effect_type == EF_NONE)
             continue;
 
         int tm_start = timing + start->effect_skip_timing;
@@ -999,9 +1005,9 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
         }
 
         int x;
-        if (start->effect_type == EF_KARAOKE || start->effect_type == EF_KARAOKE_KO) {
+        if (effect_type == EF_KARAOKE || effect_type == EF_KARAOKE_KO) {
             x = tm_current < tm_start ? x_start : x_end + 1;
-        } else if (start->effect_type == EF_KARAOKE_KF) {
+        } else if (effect_type == EF_KARAOKE_KF) {
             double dt = (double) (tm_current - tm_start) / (tm_end - tm_start);
             x = x_start + (x_end - x_start) * dt;
         } else {
@@ -1011,7 +1017,7 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
         }
 
         for (GlyphInfo *info = start; info < end; info++) {
-            info->effect_type = start->effect_type;
+            info->effect_type = effect_type;
             info->effect_timing = x - d6_to_int(info->pos.x);
         }
     }
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -495,7 +495,6 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
             double val;
             if (nargs) {
                 val = argtod(*args);
-                val *= M_PI / 180;
                 render_priv->state.frx =
                     val * pwr + render_priv->state.frx * (1 - pwr);
             } else
@@ -504,7 +503,6 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
             double val;
             if (nargs) {
                 val = argtod(*args);
-                val *= M_PI / 180;
                 render_priv->state.fry =
                     val * pwr + render_priv->state.fry * (1 - pwr);
             } else
@@ -513,12 +511,11 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
             double val;
             if (nargs) {
                 val = argtod(*args);
-                val *= M_PI / 180;
                 render_priv->state.frz =
                     val * pwr + render_priv->state.frz * (1 - pwr);
             } else
                 render_priv->state.frz =
-                    M_PI * render_priv->state.style->Angle / 180.;
+                    render_priv->state.style->Angle;
         } else if (tag("fn")) {
             char *family;
             char *start = args->start;
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -972,59 +972,47 @@ void apply_transition_effects(ASS_Renderer *render_priv, ASS_Event *event)
  */
 void process_karaoke_effects(ASS_Renderer *render_priv)
 {
-    GlyphInfo *cur, *cur2;
-    GlyphInfo *s1, *e1;      // start and end of the current word
-    GlyphInfo *s2;           // start of the next word
-    int i;
-    int timing;                 // current timing
-    int tm_start, tm_end;       // timings at start and end of the current word
-    int tm_current;
-    double dt;
-    int x;
-    int x_start, x_end;
+    int tm_current = render_priv->time - render_priv->state.event->Start;
 
-    tm_current = render_priv->time - render_priv->state.event->Start;
-    timing = 0;
-    s1 = s2 = 0;
-    for (i = 0; i <= render_priv->text_info.length; ++i) {
-        cur = render_priv->text_info.glyphs + i;
-        if ((i == render_priv->text_info.length)
-            || (cur->effect_type != EF_NONE)) {
-            s1 = s2;
-            s2 = cur;
-            if (s1) {
-                e1 = s2 - 1;
-                tm_start = timing + s1->effect_skip_timing;
-                tm_end = tm_start + s1->effect_timing;
-                timing = tm_end;
-                x_start = 1000000;
-                x_end = -1000000;
-                for (cur2 = s1; cur2 <= e1; ++cur2) {
-                    x_start = FFMIN(x_start, d6_to_int(cur2->bbox.x_min + cur2->pos.x));
-                    x_end = FFMAX(x_end, d6_to_int(cur2->bbox.x_max + cur2->pos.x));
-                }
+    int timing = 0;
+    GlyphInfo *last_boundary = NULL;
+    for (int i = 0; i <= render_priv->text_info.length; i++) {
+        if (i < render_priv->text_info.length &&
+            render_priv->text_info.glyphs[i].effect_type == EF_NONE)
+            continue;
 
-                dt = (tm_current - tm_start);
-                if ((s1->effect_type == EF_KARAOKE)
-                    || (s1->effect_type == EF_KARAOKE_KO)) {
-                    if (dt >= 0)
-                        x = x_end + 1;
-                    else
-                        x = x_start;
-                } else if (s1->effect_type == EF_KARAOKE_KF) {
-                    dt /= (tm_end - tm_start);
-                    x = x_start + (x_end - x_start) * dt;
-                } else {
-                    ass_msg(render_priv->library, MSGL_ERR,
-                            "Unknown effect type");
-                    continue;
-                }
+        GlyphInfo *start = last_boundary;
+        GlyphInfo *end = render_priv->text_info.glyphs + i;
+        last_boundary = end;
+        if (!start)
+            continue;
 
-                for (cur2 = s1; cur2 <= e1; ++cur2) {
-                    cur2->effect_type = s1->effect_type;
-                    cur2->effect_timing = x - d6_to_int(cur2->pos.x);
-                }
-            }
+        int tm_start = timing + start->effect_skip_timing;
+        int tm_end = tm_start + start->effect_timing;
+        timing = tm_end;
+
+        int x_start = 1000000;
+        int x_end = -1000000;
+        for (GlyphInfo *info = start; info < end; info++) {
+            x_start = FFMIN(x_start, d6_to_int(info->bbox.x_min + info->pos.x));
+            x_end = FFMAX(x_end, d6_to_int(info->bbox.x_max + info->pos.x));
+        }
+
+        int x;
+        if (start->effect_type == EF_KARAOKE || start->effect_type == EF_KARAOKE_KO) {
+            x = tm_current < tm_start ? x_start : x_end + 1;
+        } else if (start->effect_type == EF_KARAOKE_KF) {
+            double dt = (double) (tm_current - tm_start) / (tm_end - tm_start);
+            x = x_start + (x_end - x_start) * dt;
+        } else {
+            ass_msg(render_priv->library, MSGL_ERR,
+                    "Unknown effect type");
+            continue;
+        }
+
+        for (GlyphInfo *info = start; info < end; info++) {
+            info->effect_type = start->effect_type;
+            info->effect_timing = x - d6_to_int(info->pos.x);
         }
     }
 }
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -1006,12 +1006,12 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
 
         int x;
         if (tm_current < tm_start)
-            x = -1000000;
+            x = -100000000;
         else if (tm_current >= tm_end)
-            x = 1000000;
+            x = 100000000;
         else {
-            int x_start = d6_to_int(start->pos.x);
-            int x_end = d6_to_int(end[-1].pos.x + end[-1].advance.x);
+            int x_start = start->pos.x;
+            int x_end = end[-1].pos.x + end[-1].advance.x;
             double dt = (double) (tm_current - tm_start) / (tm_end - tm_start);
             double frz = fmod(start->frz, 360);
             if (frz > 90 && frz < 270) {
@@ -1023,12 +1023,12 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
                     info->c[1] = tmp;
                 }
             }
-            x = x_start + (x_end - x_start) * dt;
+            x = x_start + lrint((x_end - x_start) * dt);
         }
 
         for (GlyphInfo *info = start; info < end; info++) {
             info->effect_type = effect_type;
-            info->effect_timing = x - d6_to_int(info->pos.x);
+            info->effect_timing = x - info->pos.x;
         }
     }
 }
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -997,22 +997,21 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
         int tm_end = tm_start + start->effect_timing;
         timing = tm_end;
 
-        int x_start = 1000000;
-        int x_end = -1000000;
-        for (GlyphInfo *info = start; info < end; info++) {
-            x_start = FFMIN(x_start, d6_to_int(info->bbox.x_min + info->pos.x));
-            x_end = FFMAX(x_end, d6_to_int(info->bbox.x_max + info->pos.x));
-        }
-
         if (effect_type != EF_KARAOKE_KF)
             tm_end = tm_start;
 
         int x;
         if (tm_current < tm_start)
-            x = x_start;
+            x = -1000000;
         else if (tm_current >= tm_end)
-            x = x_end + 1;
+            x = 1000000;
         else {
+            int x_start = 1000000;
+            int x_end = -1000000;
+            for (GlyphInfo *info = start; info < end; info++) {
+                x_start = FFMIN(x_start, d6_to_int(info->bbox.x_min + info->pos.x));
+                x_end = FFMAX(x_end, d6_to_int(info->bbox.x_max + info->pos.x));
+            }
             double dt = (double) (tm_current - tm_start) / (tm_end - tm_start);
             x = x_start + (x_end - x_start) * dt;
         }
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -1006,12 +1006,8 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
         else if (tm_current >= tm_end)
             x = 1000000;
         else {
-            int x_start = 1000000;
-            int x_end = -1000000;
-            for (GlyphInfo *info = start; info < end; info++) {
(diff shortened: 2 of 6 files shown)


astiob:
Also: our karaoke code doesn’t handle glyphs marked with `skip` by the shaper.

…or trailing whitespace marked `skip` by the whitespace trimmer. I just realized. So that’s still broken.

And karaoke isn’t the only offender: neither do
* `retrieve_glyphs` (does unnecessary work and doesn’t properly add extra spacing after italic ligatures; not sure if there may be problems with \fsp)
* `preliminary_layout` (no harm as HarfBuzz-skipped glyphs have zero advance from calloc, unless there may be problems with \fsp)
* `wrap_lines_smart` (might make mistaken decisions?)

astiob:
Also, our FriBidi “shaper” marks some normally-invisible characters with `skip`, but HarfBuzz seems to do no such thing. Is this beneficial, harmless or harfmul?

libass/ass_utils.h
```diff
@@ -160,10 +160,10 @@ static inline int double_to_d22(double x)
     return (int) (x * 0x400000);
 }
 
-// Calculate cache key for a rotational angle in radians
+// Calculate cache key for a rotational angle in degrees
 static inline int rot_key(double a)
```
MrSmile, line 164: I missed that function in my quantization PR: its uses have got eliminated there. So it should be safe to remove it.

libass/ass_parse.c
```diff
@@ -1010,6 +1010,16 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
                         x_end = d6_to_int(e1->pos.x + e1->advance.x);
                         dt = (tm_current - tm_start);
                         dt /= (tm_end - tm_start);
+                        frz = fmod(s1->frz, 360);
+                        if (frz > 90 && frz < 270) {
+                            // Fill from right to left
+                            dt = 1 - dt;
+                            for (cur2 = s1; cur2 <= e1; ++cur2) {
+                                uint32_t tmp = cur2->c[0];
+                                cur2->c[0] = cur2->c[1];
```
MrSmile, line 1019: If we can redefine colors here then for uniformity we should handle `tm_current < tm_start` and `tm_current >= tm_end` through the same trick and eliminate usages of `info->effect_type` in `render_text()`.

MrSmile (commented):
It's not that hard to add leftmost control point calculation in `get_bitmap_glyph()`, bypassing quantization/rasterization pipeline and associated rounding errors. Is `FFMIN(x)` all that needed for karaoke?

Concerning this PR, I think that `process_karaoke_effects()` should be thoroughly refactored, as it's quite outdated.
Also, shouldn't "Drop the now-useless, previously-broken first_pos_x" follow after "process_karaoke_effects: compute \kf from glyph advances, not bounds"?

TheOneric:
Tested the samples from #216 and #357 . While 357 pretty much does what I'd expect it to, this pr completely breaks the fill animtion of 216 , with the `Fix \kf positioning: convert effect_timing to screen coordinate` commit. Not sure if this is an error in the pr or an intended consequence of emulating some VSFilter quirk.
Before that commit, there is an fill animation, but it starts out with half the glyph filled, which is not what I'd expect and doesn't match https://github.com/libass/libass/issues/216#issuecomment-643685186 . *(Again not sure if bug or VSFilter quirk)*

astiob:
Huh. That’s not what’s supposed to happen in #216. I’ll investigate.

astiob:
What’s happening with #216 is that it’s trying to start at the first glyph’s origin… where it was before transforms, but the text is rotated. So the \kf line ends up somewhere beyond the right edge of the glyphs.

So that’s one more reason to start the fill from the leftmost point like VSFilter, instead of the first glyph’s origin.

astiob:
* Removed `rot_key`.
* Edited `process_karaoke_effects` to use C99, `continue`, `switch`, more expressive variables names, and no `e1`.
* Fixed \kf with trimmed trailing whitespace.

> It's not that hard to add leftmost control point calculation in get_bitmap_glyph(), bypassing quantization/rasterization pipeline and associated rounding errors. Is FFMIN(x) all that needed for karaoke?

Yes, min(x) is all that’s needed there. Are you thinking of adding it as a new field in one of the structs? (`ASS_Outline` would be most logical, although `Bitmap` would be more economical.) Or are you thinking we can change `bm->left` to be this value?

> Also, shouldn't "Drop the now-useless, previously-broken first_pos_x" follow after "process_karaoke_effects: compute \kf from glyph advances, not bounds"?

No (not necessarily). \k and \ko are fixed, and `first_pos_x` made obsolete, by the immediately preceding commit, “place line at infinity except during \kf”.

libass/ass_parse.c
astiob: Ironically, I’m actually not very happy with my solution of fudging the colours here.

On the other hand, I don’t immediately see any harm in what you’re suggesting.

Furthermore, it might help us move closer to a solution to another long-standing problem: the edge between the glyph body and the ASS border sometimes looks bad because the body and border are separate bitmaps and one is drawn on top of the other. I have [an old set of patches](https://github.com/astiob/libass/commits/alpha) that fixes this, but it doesn’t work with karaoke. The reason is: it needs to know the body’s alpha value when it’s rasterizing the border. But karaoke can change the body’s alpha, and it does this _way_ after the border has been rasterized. So if we were to make karaoke change the colour here, that would help! We’d only need to figure out what to do with \kf.

astiob:
@MrSmile Does it look like we can get karaoke merged very quickly before releasing 0.15.0? Or should I just make the release now?

MrSmile (commented):
> Are you thinking of adding it as a new field in one of the structs?

No, I don't want to pollute our rendering pipeline with obscure ASS features. I'm thinking of adding to `get_bitmap_glyph()` after matrix calculation something like
```c
    const size_t n = info->outline->n_points;
    const ASS_Vector *pt = info->outline->points;
    for (size_t i = 0; i < n; i++) {
        double x = m[0][0] * pt[i].x + m[0][1] * pt[i].y + m[0][2];
        double z = m[2][0] * pt[i].x + m[2][1] * pt[i].y + m[2][2];
        leftmost_x = fmin(leftmost_x, x / FFMAX(z, 0.1));
    }
```

As for release, I think we should try to fix karaoke in it as it's a long-standing problem. With out release frequency ±week is nothing.

libass/ass_parse.c
MrSmile: In my tiled rendering experiment I've transformed all sorts of clipping into generic multiply by bitmap. I think something in that line should be done for karaoke too.

libass/ass_parse.c
```diff
@@ -964,59 +964,54 @@ void apply_transition_effects(ASS_Renderer *render_priv, ASS_Event *event)
...
+        switch (start->effect_type) {
+        case EF_KARAOKE:
+        case EF_KARAOKE_KO:
+            if (tm_current >= tm_start)
+                x = x_end + 1;
+            else
+                x = x_start;
+            break;
+        case EF_KARAOKE_KF:
+            double dt = (double) (tm_current - tm_start) / (tm_end - tm_start);
```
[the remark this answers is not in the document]
astiob: Right; C++ does forbid that. I think C doesn’t. But I can also avoid this if you’d like: would you prefer braces around the case, or the variables to be declared earlier?

astiob:
What I rather want is this in `ass_bitmap_construct`:

```c
for (i = 0; i < outline[0].n_points; i++)
    minx = FFMIN(minx, outline[0].points[i].x);
```

libass/ass_parse.c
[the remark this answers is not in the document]
astiob: Of course, this code is modified in a later commit that moves this declaration inside a {block} anyway.

libass/ass_parse.c
```diff
@@ -1006,8 +1006,14 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
                 x = x_start;
             break;
         case EF_KARAOKE_KF:
-            double dt = (double) (tm_current - tm_start) / (tm_end - tm_start);
-            x = x_start + (x_end - x_start) * dt;
+            if (tm_current < tm_start)
```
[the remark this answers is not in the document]
astiob: Done.

libass/ass_parse.c
```diff
@@ -964,59 +964,54 @@ void apply_transition_effects(ASS_Renderer *render_priv, ASS_Event *event)
...
-                timing = tm_end;
-                x_start = 1000000;
-                x_end = -1000000;
-                for (cur2 = s1; cur2 <= e1; ++cur2) {
-                    x_start = FFMIN(x_start, d6_to_int(cur2->bbox.x_min + cur2->pos.x));
-                    x_end = FFMAX(x_end, d6_to_int(cur2->bbox.x_max + cur2->pos.x));
-                }
+    int timing = 0;
+    GlyphInfo *last_boundary = NULL;
+    for (int i = 0; i <= render_priv->text_info.length; ++i) {
```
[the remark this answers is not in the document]
astiob: Fixed, thanks.

libass/ass_parse.c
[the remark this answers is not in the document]
MrSmile: If C guidelines don't have a problem with that then keep it as is, especially if it's not relevant later. (It's one of the reasons why I prefer to merge commits with overlapping changeset: less issues with intermediate representation).

libass/ass_parse.c
```diff
@@ -1000,8 +1000,8 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
         switch (effect_type) {
         case EF_KARAOKE:
         case EF_KARAOKE_KO:
-            x = tm_current < tm_start ? x_start : x_end + 1;
-            break;
+            tm_end = tm_start;
+            //-fallthrough
```
MrSmile, line 1001: I think that instead of fallthrough it's better to move \kf branch after the switch.

MrSmile (commented):
> What I rather want is this in `ass_bitmap_construct` after the transform calls:

But it's quantized matrix there instead of original, so additional rounding errors will be introduced. Also, additional backward dependency between shaping and rendering stages isn't good for the future (for example, GPU rendering). There is potential loss of performance, but that calculation can be done only for active karaoke (specifically \kf with nontrivial splitting). And to minimize harm form copy-paste that code could be moved into `ass_outline.c` near `outline_transform_3d()`.

libass/ass_parse.c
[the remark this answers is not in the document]
TheOneric: Both `gcc` and `clang` complain about this, so this specific code apparently isn't valid C either:
```error: a label can only be part of a statement and a declaration is not a statement```

Adding any valid statement eg a simple `;` after the case but before the declaration, makes this work in C *(but not C++)* — adding braces seems imo cleaner though.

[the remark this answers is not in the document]
Comment (astiob) [reply libass/ass_parse.c]: Ah, right. Now _that’s_ forbidden in C indeed. OK, I’ll have to change this.

But given that MrSmile is effectively asking me to eliminate the `switch` in a later commit anyway, I think I’ll just not introduce it in the first place and return to the if-else cascade that was here before my rewrite.

File: libass/ass_parse.c (hunk 6 above)
Comment (astiob) [reply libass/ass_parse.c]: I’ve now removed the `switch` entirely and replaced the `if (k/ko) A; else if (kf) B; else error;` cascade by `if (!kf) A; B`.

**Comment by astiob** [comment]
Right; for some reason I thought the code you suggested did a different thing, but it does what I want.

I’ve gone ahead and used your suggested code, with slight tweaks.

**This should now be ready for merge, I think.** One remaining thing that we can decide is whether we want to emulate VSFilter’s behaviour when ybord ≥ xbord, and if so, whether we want to emulate the rounding error as well or just subtract ybord−xbord.

File: libass/ass_outline.c [hunk 7]
```diff
@@ -323,6 +323,21 @@ bool outline_transform_3d(ASS_Outline *outline, const ASS_Outline *source,
     return true;
 }
 
+int32_t outline_find_min_transformed_x(const ASS_Outline *outline,
```
MrSmile, line 326: I think it's more clear to add `int32_t *leftmost_x` as a parameter and update current minimum instead of overwriting it.

libass/ass_outline.c
```diff
@@ -323,6 +323,21 @@ bool outline_transform_3d(ASS_Outline *outline, const ASS_Outline *source,
...
+                                       const double m[3][3]) {
+    int32_t min_x = INT32_MAX;
+    const ASS_Vector *pt = outline->points;
+    for (size_t i = 0; i < outline->n_points; i++) {
+        double z = m[2][0] * pt[i].x + m[2][1] * pt[i].y + m[2][2];
+        double x = (m[0][0] * pt[i].x + m[0][1] * pt[i].y + m[0][2]) / FFMAX(z, 0.1);
+        if (!(fabs(x) < OUTLINE_MAX))
```
MrSmile, line 333: Not sure about additional failure mode here. Maybe it's better to clamp or even return full `double` without any rounding?

libass/ass_render.c
```diff
@@ -1300,9 +1300,11 @@ static void calc_transform_matrix(ASS_Renderer *render_priv,
  */
 static void
 get_bitmap_glyph(ASS_Renderer *render_priv, GlyphInfo *info,
+                 int32_t *leftmost_x,
                  ASS_Vector *pos, ASS_Vector *pos_o,
                  ASS_DVector *offset, bool first, int flags)
 {
+    *leftmost_x = INT32_MAX;
```
MrSmile, line 1308: This function should update instead of overwriting minimum value too. In that case `&current_info->leftmost_x` can be used directly as a parameter.

libass/ass_parse.c
```diff
@@ -996,22 +996,17 @@ void process_karaoke_effects(ASS_Renderer *render_priv)
...
-            if (tm_current < tm_start)
-                x = x_start;
-            else if (tm_current >= tm_end)
-                x = x_end + 1;
-            else {
-                double dt = (double) (tm_current - tm_start) / (tm_end - tm_start);
-                x = x_start + (x_end - x_start) * dt;
-            }
-        } else {
-            ass_msg(render_priv->library, MSGL_ERR,
```
MrSmile: Is it impossible to reach here? Maybe it's better to add the corresponding assert somewhere?

MrSmile (commented):
Nitpick: `\k/\ko` in commit messages looks strange, maybe `\k and \ko` is better?

astiob:
It was a deliberate decision to make `leftmost_x` like that. Let me explain:

I need it as an integer, so returning it as a double only delays the overflow check. Meanwhile, if a transformed control point coordinate goes out of range, then `outline_transform_*` will drop the whole outline. If I encounter an out-of-range transformed x, it seems to make more sense to ignore the whole outline, which I know will be dropped anyway, than to return an arbitrary number, whether it be `-OUTLINE_MAX` or a random in-range control point coordinate.

Of course, this is only an attempt to fail less inexplicably in a situation when a failure is already occurring. So if you still feel that passing a pointer and clamping out-of-range values is better (perhaps because you feel the karaoke position doesn’t matter if we’re failing), I can still do that.

libass/ass_parse.c
astiob: The Effect enum has no other values. We skip `EF_NONE` above, and we handle all other values here. This `ass_msg` could have been an `assert` all along. I don’t see any value in keeping it.

astiob:
Oh, forgot to mention: in the last push I also changed these kinds of conditionals:

```c
if (foo
    || bar)
```

into these:

```c
if (foo ||
        bar)
```

because I think this better matches the bulk of our codebase. But I may be wrong.

MrSmile:
At least `get_bitmap_glyph()` should update `leftmost_x`. But I think that `outline_find_min_transformed_x()` should update too (with clamp). Consider some overly-transformed text with the first letter on `OUTLINE_MAX` boundary and the second letter in the center of the screen, perfectly visible. Correct rendering should be with the kara split far far away from the screen, but if the first letter gets removed, then the split unnaturally appears on the second visible letter. And that disappearance can be conditional, for example, depending on the window size.

> I also changed these kinds of conditionals:

I also prefer new variant. Dunno what libass codestyle says about this. But there is a third variant: write it all in one line. Current line length limit looks too low (less than 1/3 of my screen width). Also we already have lines with length more than 120.

astiob:
Fair enough. I’ve changed all of it to directly update the target `leftmost_x` field, clamped to ±`OUTLINE_MAX`.

I’m concerned by one thing though… the sample from #216 isn’t _completely_ right. When the file starts, the first glyph is already slightly filled in in libass, but it’s completely empty in VSFilter. I don’t know why yet.

> But there is a third variant: write it all in one line. Current line length limit looks too low (less than 1/3 of my screen width). Also we already have lines with length more than 120.

I’m not usually one to refuse longer lines, but ironically, I’m currently viewing libass code in two columns and I have to scroll because it doesn’t fit on screen 😃

TheOneric:
The current (and previous) version leads to memory leaks ([sample file](https://github.com/libass/libass/files/5435894/h.ass.txt)) not present in the version from 18h ago:
```
==6472==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 290816 byte(s) in 142 object(s) allocated from:
    #0 0x7ff3493fb330 in __interceptor_malloc (/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
    #1 0x5581d6e35bff in render_and_combine_glyphs ./libass/ass_render.c:2372

SUMMARY: AddressSanitizer: 290816 byte(s) leaked in 142 allocation(s).

```

Regarding https://github.com/libass/libass/issues/216#issuecomment-643685186 there's slight difference compared to #343 in that a sliver of red is already visible, but that's probably more correct than before, so this seems alright.

However, testing the following excerpt from #357 at **6798ms** `our` is already filled in this and the previous version.
According to the time values, `our` should only be active after **6800ms**, which is in line with what happens in #343 and previous versions of this pr iirc.
```
[Script Info]
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
PlayResX: 1920
PlayResY: 1080
YCbCr Matrix: TV.601

[Aegisub Project Garbage]
Video File: ?dummy:23.976000:6000:1920:1080:9:169:225:

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline,
StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encodin
g
Style: Default,Arimo,95,&H00008AFF,&H00FFFFFF,&H00000000,&H00666666,-1,0,0,0,100,100,0,0,1,6,0,8,40,40,60,1

[Events]
Dialogue: 0,0:00:04.17,0:00:09.82,Default,,0,0,0,fx,{\k90\fad(300,200)}{\k42}So {\k42}we {\k43}can {\k46}find {\k37}ou
r {\kf245}way
```

astiob:
Thanks for catching that! If it’s what I think, then it should now be fixed, and it should have been happening with #343, too. I don’t immediately see why that file is triggering this, though. 🤔 Did you run it with a randomly-failing malloc or something similar?

TheOneric:
> If it’s what I think, then it should now be fixed, and it should have been happening with #343, too

Upon retesting #343 does indeed also cause memleaks here. I probably wasn't yet testing with ASAN, or didn't yet add this file to my test when initially testing #343. *(As well as triggering some UB with an unexpected NULL here: https://github.com/libass/libass/blob/255f8d79d36f39415998a272c2245326d35c4ab0/libass/ass_render.c#L3091 )*

> Did you run it with a randomly-failing malloc or something similar?

Nope, just ASAN and UBSAN before running tests.
`make clean && make CC="gcc -fsanitize=address -fsanitize=undefined" -j8`

I don't see memory leaks anymore, but the rendering/timing discrepancy is still there.

astiob:
OK, all fixed.

> at 6798ms our is already filled in this […] version

Turns out it was filled even at 0000ms. Turns out I broke \k and \ko when I made the new leftmost_x calculation run only for \kf (to avoid unnecessary work otherwise) but continued to add leftmost_x to effect_type without checking the karaoke type.

astiob:
Oh, and as for #216, it seems that the way the MP4 file is cut makes mpv think the first frame’s timestamp is 0:00.085, causing it to already have 55 ms worth of karaoke fill in libass. My MPC-HC setup labels the first frame with 0:00.000, so VSFilter faithfully renders no fill. (They somehow sync up after the first frame.) So it seems everything is correct there, after all.

MrSmile (approved):

TheOneric:
> Turns out it was filled even at 0000ms.

Thought I checked this — but the shell history tells me I didn't preload the correct library when looking at the file with mpv, whoops.

So, now everything seems to work as I'd expect. There's still the slight difference from  #343 in #216 at 2252ms, but this version actually seems a bit more correct to me in regard to karaoke, given that the た-filling is supposed to have started 2ms prior. In #343 no fill was yet visible, here a few pixels are already red.
*(At a side note, I just noticed there's also a difference in border-blurring, but this is probably due to d60f2dce1dcafaed0e21406213e66072f3036752 ff I guess)*

343:
216-vertical-2252
this:
216-vertical-2252

libass/ass_parse.c
```diff
@@ -964,59 +964,48 @@ void apply_transition_effects(ASS_Renderer *render_priv, ASS_Event *event)
...
+        int x_start = 1000000;
+        int x_end = -1000000;
+        for (GlyphInfo *info = start; info < end; info++) {
+            x_start = FFMIN(x_start, d6_to_int(info->bbox.x_min + info->pos.x));
+            x_end = FFMAX(x_end, d6_to_int(info->bbox.x_max + info->pos.x));
+        }
+
+        int x;
+        if (start->effect_type == EF_KARAOKE ||
+                start->effect_type == EF_KARAOKE_KO) {
```
rcombs: Not sure what's going on with the indent here?

rcombs (commented):
I'm not really familiar enough with most of this code to provide any substantial review, so all I've got is a complaint about a late-added style detail.

libass/ass_parse.c
astiob: It’s a style we use elsewhere (but not everywhere): the second line is indented by two levels. The good thing about it is that for control structures like `if` in particular, it helps separate the condition from the body that follows. The bad thing is that it may look weird at first 😃

I don’t strongly feel either way; I just thought this was the more common style. But I may be wrong about that.

astiob: After a discussion on IRC, I’ve changed this to line-trailing `||` and a single indent.

PR #439, opened 2020-10-22
ass_render: Replace Segment with Rect

TheOneric:
For a change from the alloc-stuff here's a small, cosmetic refactor.

---

Contrary to what the comments suggest Segment does not use one
absolute reference point and relative offsets, but two absolute points,
making it practically identical to the previously defined Rect with
  a ≙ y0,  b ≙ y1,  ha ≙ x0,  hb ≙ x1

For simplicity, replace Segment with Rect.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2896,17 +2896,17 @@ static ASS_RenderPriv *get_render_priv(ASS_Renderer *render_priv,
     return event->render_priv;
 }
 
-static int overlap(Segment *s1, Segment *s2)
+static int overlap(Rect *s1, Rect *s2)
 {
-    if (s1->a >= s2->b || s2->a >= s1->b ||
-        s1->ha >= s2->hb || s2->ha >= s1->hb)
+    if (s1->y0 >= s2->y1 || s2->y0 >= s1->y1 ||
+        s1->x0 >= s2->x1 || s2->x0 >= s1->x1)
         return 0;
     return 1;
 }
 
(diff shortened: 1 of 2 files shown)


libass/ass_render.c
```diff
@@ -3008,16 +3008,16 @@ fix_collisions(ASS_Renderer *render_priv, EventImages *imgs, int cnt)
...
+                used[cnt_used].y1 = priv->top + priv->height;
+                used[cnt_used].x0 = priv->left;
+                used[cnt_used].x1 = priv->left + priv->width;
                 cnt_used++;
                 shift_event(render_priv, imgs + i, priv->top - imgs[i].top);
             }
         }
     }
-    qsort(used, cnt_used, sizeof(Segment), cmp_segment);
+    qsort(used, cnt_used, sizeof(*used), cmp_rect_y0);
```
astiob, line 3020: Minor nit: you replaced `sizeof(type)` by `sizeof(object)` here, but not in the very similar line in `fit_rect`. Seems strangely inconsistent.

astiob (approved):

libass/ass_render.c
TheOneric: Thanks, fixed it.

rcombs (approved):
Sure, and as a bonus the member names are clearer.

MrSmile (approved):
Not sure about special symbols in commit message, but LGTM.

TheOneric:
Rebased and slightly reworded commit message, otherwise identical.

PR #440, opened 2020-10-22
coretext: return full/PostScript name on fallback

astiob:
CoreText's "family name" can use TT_NAME_ID_TYPOGRAPHIC_FAMILY names, which we ignore for compatibility with VSFilter. As a result, fallback returns a family name that we then fail to find, and fallback fails.

Return a fuller name (fullname or postscriptname as appropriate) instead, which should be more reliable.

This fixes broken fallback on kanji from Arial/Helvetica to Hiragino Sans on my machine. The current code returns “Hiragino Sans”, which is a TT_NAME_ID_TYPOGRAPHIC_FAMILY name, we fail to select it, and I end up seeing tofu. The new code returns “HiraginoSans-W3”, which works.

### DIFF
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -60,19 +60,23 @@ static void destroy_font(void *priv)
     SAFE_CFRelease(fontd);
 }
 
-static bool check_postscript(void *priv)
+static bool is_postscript_font_format(CFNumberRef cfformat)
 {
     bool ret = false;
-    CTFontDescriptorRef fontd = priv;
-    CFNumberRef cfformat =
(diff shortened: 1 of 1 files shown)


rcombs (approved):
Doesn't need to block this but it occurs to me that the `get_fallback` routine is extremely lacking in error/null-ret handling.

astiob:
My analysis may have been incorrect. While “Hiragino Sans” is indeed the font’s “typographic family” Microsoft name, it is also its “(plain) family” Apple name, which Core Text, of course, prefers.

Issue #148, opened 2014-11-16
underline, strickout and Arabic Render problem since bitmap combining


realfinder:
vsfilter (and almost libass before bitmap)
![vsfilter](https://cloud.githubusercontent.com/assets/2512153/5061866/8a44668e-6dbb-11e4-95d9-5933f73b2905.png)

libass
![libass](https://cloud.githubusercontent.com/assets/2512153/5061864/869ea9ea-6dbb-11e4-8cd3-d8281fbe9d4e.png)

There are artifacts between letters

Script and font
http://www.mediafire.com/?39rbxfjgb8fmgmq

another one
http://www.mediafire.com/?dp70audha7rgpg7

more details and relation issue https://web.archive.org/web/20160120113701/https://code.google.com/p/libass/issues/detail?id=114#c6

note: this problem become worst after fix this https://github.com/libass/libass/issues/142

grigorig:
I think we should ideally rasterize runs of text as a whole, not glyphs.

astiob:
Yes! I’ve been of that opinion for ages, but it requires another fairly major change to the code, so there is no telling when we may get it done. Rasterizing whole runs will also help us considerably with VSFilter compatibility; even replicating VSFilter’s RTL behaviour will be trivial*.

\* Unless it actually resets to LTR every time, and it may well do; in which case a bit more work will be needed to get the expected behaviour for lines that use explicit RTL marks (overrides? I’m a bit hazy with these terms, sorry).

anoosi:
is there plan to fix this issue?

realfinder:
seems it not just the quality affected by this

[testing sub file.zip](https://github.com/libass/libass/files/5992583/testing.sub.file.zip) using avs assrender https://github.com/pinterf/assrender and avspmod
![image](https://user-images.githubusercontent.com/2512153/108142570-1656d100-70d7-11eb-93ff-fbcb7b1d9a37.png)

vsfilter (TextSub)
![image](https://user-images.githubusercontent.com/2512153/108142625-2d95be80-70d7-11eb-9db7-37055342902c.png)

both one by one
![test](https://user-images.githubusercontent.com/2512153/108144795-659f0080-70db-11eb-9912-42655c55345d.gif)

the bluring also is not same as vsfilter

also, as aside note, assrender don't load the font that inside the .ass file unlike vsfilter which did load it, it is problem in assrender or libass itself?

astiob:
[...]

Issue #71, opened 2014-04-01
another shaper issues in arabic


realfinder:
hi

using last mpv in windows (with HarfBuzz)

libass
![1 libass](https://cloud.githubusercontent.com/assets/2512153/2575335/f1be0b5c-b94e-11e3-8ae0-540b618aea4d.png)

libass with the Pointing
![1 libass a](https://cloud.githubusercontent.com/assets/2512153/2575386/2d3fb9ae-b950-11e3-99ec-6abb588bb13c.png)

vsfilter
![1 vsfilter](https://cloud.githubusercontent.com/assets/2512153/2575337/f1cb0208-b94e-11e3-87c9-6dc391b26953.png)

---

libass
![2 libass](https://cloud.githubusercontent.com/assets/2512153/2575336/f1c60ae6-b94e-11e3-813c-8ddb3de84d62.png)

vsfilter
![2 vsfilter](https://cloud.githubusercontent.com/assets/2512153/2575334/f1bc7d6e-b94e-11e3-8ae8-12ed93dc7dbb.png)

wrong word order because of underline code

---

video sample http://www.solidfiles.com/d/9287427ae4/[Devil-Fansub]_Hyouka_-_11.mkv

sub + font http://www.solidfiles.com/d/1c8cb1aace/sub_%2B_font.rar

thanks

grigorig:
Regarding the "wrong order", at the moment, the behaviour on override tags is somewhat intentional incompatibility with VSFilter. As far as I can tell VSFilter's behaviour (always reset to LTR after an override) is really unusual and inconsistent compared to BiDi interaction in other markup languages like HTML. Do we really want to change that?

I'll take a look at the other problems. We should especially filter out control characters much earlier. (Font fallback happens way before shaping)

realfinder:
hi again

since a lot of people used mobile now, and I saw there are some complaints about these compatibility cases (like in this "another shaper issues in arabic #71" and others compatibility issues)

any intention to make libass more compatible with vsfilter?

grigorig:
I'm not sure, VSFilter behavior is very inconsistent. We've already done various changes to make libass behave closer to VSFilter, but there's a limit. You proposed to reset the text direction at each override tag, which seems like a pretty significant change. Let me think about it.

astiob:
[...]

PR #441, opened 2020-10-25
Emulate VSFilter’s bidi and \fay by default; add an ASS_Feature for better bidi and shaping

astiob:
Guess what.

### DIFF
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -971,25 +971,31 @@ bool ass_shaper_shape(ASS_Shaper *shaper, TextInfo *text_info)
     if (!check_allocations(shaper, text_info->length))
         return false;
 
-    // Get bidi character types and embedding levels
+    for (i = 0; i < text_info->length; i++)
+        shaper->event_text[i] = glyphs[i].symbol;
+
+    fribidi_get_bidi_types(shaper->event_text,
+            text_info->length, shaper->ctypes);
+
+#ifdef USE_FRIBIDI_EX_API
+    if (shaper->bidi_brackets) {
+        fribidi_get_bracket_types(shaper->event_text,
+                text_info->length, shaper->ctypes, shaper->btypes);
+    }
+#endif
+
+    // Get bidi embedding levels
     last_break = 0;
     for (i = 0; i < text_info->length; i++) {
-        shaper->event_text[i] = glyphs[i].symbol;
         // embedding levels should be calculated paragraph by paragraph
         if (glyphs[i].symbol == '\n' || i == text_info->length - 1 ||
                 (!shaper->whole_text_layout &&
                     (glyphs[i + 1].starts_new_run || glyphs[i].hspacing))) {
             dir = shaper->base_direction;
-            fribidi_get_bidi_types(shaper->event_text + last_break,
-                    i - last_break + 1, shaper->ctypes + last_break);
 #ifdef USE_FRIBIDI_EX_API
             FriBidiBracketType *btypes = NULL;
-            if (shaper->bidi_brackets) {
+            if (shaper->bidi_brackets)
                 btypes = shaper->btypes + last_break;
-                fribidi_get_bracket_types(shaper->event_text + last_break,
-                        i - last_break + 1, shaper->ctypes + last_break,
-                        btypes);
-            }
             ret = fribidi_get_par_embedding_levels_ex(
                     shaper->ctypes + last_break, btypes,
                     i - last_break + 1, &dir, shaper->emblevels + last_break);
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -703,19 +703,25 @@ static bool shape_harfbuzz(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
                 level == shaper->emblevels[i + 1])
             i++;
 
-        int lead_context = 0, trail_context = 0;
-        if (offset > 0 && !glyphs[offset].starts_new_run &&
-                is_shaping_control(glyphs[offset - 1].symbol))
-            lead_context = 1;
-        if (i < (len - 1) && !glyphs[i + 1].starts_new_run &&
-                is_shaping_control(glyphs[i + 1].symbol))
-            trail_context = 1;
-
         hb_buffer_pre_allocate(buf, i - offset + 1);
-        hb_buffer_add_utf32(buf,
-                shaper->event_text + offset - lead_context,
-                i - offset + 1 + lead_context + trail_context,
-                lead_context, i - offset + 1);
+
+        int lead_context = 0, trail_context = 0;
+        if (shaper->whole_text_layout) {
+            hb_buffer_add_utf32(buf, shaper->event_text, len,
+                    offset, i - offset + 1);
+        } else {
+            if (offset > 0 && !glyphs[offset].starts_new_run &&
+                    is_shaping_control(glyphs[offset - 1].symbol))
+                lead_context = 1;
+            if (i < (len - 1) && !glyphs[i + 1].starts_new_run &&
+                    is_shaping_control(glyphs[i + 1].symbol))
+                trail_context = 1;
+
+            hb_buffer_add_utf32(buf,
+                    shaper->event_text + offset - lead_context,
+                    i - offset + 1 + lead_context + trail_context,
+                    lead_context, i - offset + 1);
+        }
 
         props.direction = FRIBIDI_LEVEL_IS_RTL(level) ?
             HB_DIRECTION_RTL : HB_DIRECTION_LTR;
@@ -726,7 +732,8 @@ static bool shape_harfbuzz(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
         set_run_features(shaper, glyphs + offset);
         hb_shape(font, buf, shaper->features, shaper->n_features);
 
-        shape_harfbuzz_process_run(glyphs, buf, offset - lead_context);
+        shape_harfbuzz_process_run(glyphs, buf,
+                shaper->whole_text_layout ? 0 : offset - lead_context);
         hb_buffer_reset(buf);
     }
 
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -894,6 +894,7 @@ void ass_shaper_find_runs(ASS_Shaper *shaper, ASS_Renderer *render_priv,
                         last->face_index != info->face_index) ||
                     last->script != info->script ||
                     info->starts_new_run ||
+                    (!shaper->whole_text_layout && info->hspacing) ||
                     last->flags != info->flags))
                 shape_run++;
             else if (info->skip)
@@ -976,7 +977,8 @@ bool ass_shaper_shape(ASS_Shaper *shaper, TextInfo *text_info)
         shaper->event_text[i] = glyphs[i].symbol;
         // embedding levels should be calculated paragraph by paragraph
         if (glyphs[i].symbol == '\n' || i == text_info->length - 1 ||
-                (!shaper->whole_text_layout && glyphs[i + 1].starts_new_run)) {
+                (!shaper->whole_text_layout &&
+                    (glyphs[i + 1].starts_new_run || glyphs[i].hspacing))) {
             dir = shaper->base_direction;
             fribidi_get_bidi_types(shaper->event_text + last_break,
                     i - last_break + 1, shaper->ctypes + last_break);
@@ -1074,7 +1076,8 @@ FriBidiStrIndex *ass_shaper_reorder(ASS_Shaper *shaper, TextInfo *text_info)
     GlyphInfo *glyphs = text_info->glyphs;
     for (i = 0; i < text_info->length; i++) {
         if (i == text_info->length - 1 || glyphs[i + 1].linebreak ||
-                (!shaper->whole_text_layout && glyphs[i + 1].starts_new_run)) {
+                (!shaper->whole_text_layout &&
+                    (glyphs[i + 1].starts_new_run || glyphs[i].hspacing))) {
             FriBidiParType dir = FRIBIDI_PAR_ON;
 
             ret = fribidi_reorder_line(0,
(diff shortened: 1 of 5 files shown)


astiob:
Also, I think it makes more sense to run bidi on the whole event, rather than on each explicitly-delimited line (as we do currently), no?

The question is, given:

> [English][Arabic]\N[Arabic][English]

is it more likely to mean

1. English with embedded Arabic and manually wrapped lines? or
2. two distinct texts, the first English-with-embedded-Arabic and the second Arabic-with-embedded-English?

Because our current code (with Encoding=-1, otherwise there’s nothing to discuss) does 2.

astiob:
In fact, [Unicode seems to mandate the latter behaviour](https://unicode.org/reports/tr9/#The_Paragraph_Level), saying the text must first be split on paragraph separators, among which it lists the line feed, and then each paragraph’s base direction must be separately autodetected. This is _not_ what FriBidi does though: it does split paragraphs on line feeds, but it runs direction detection only once for the whole input buffer. And I still think that makes more sense for our use case.

astiob:
Pushed several additional fixes for bidi and shaping (and fixed the Travis CI build failure). And renamed the feature to `ASS_FEATURE_WHOLE_TEXT`.

This improves _both_ our compatibility with VSFilter when the new feature is off _and_ our rendering when the new feature is on.

rcombs:
Maybe call it `ASS_FEATURE_WHOLE_TEXT_BIDI` for better clarity?

astiob:
Two reasons I didn’t:

* It affects shaping as well. Our internal code calls bidi and shaping together “shaping”, but that isn’t right either. I haven’t been able to come up with a simple term that covers both. Whole text typography? (Not really.) Whole text processing?

* Perhaps the same feature could be extended in the future for more run un-compatibility-fixes, e. g. we could try to combine bitmaps (especially BorderStyle 3) for whole lines instead of per run. Or would this be too unstable of an API?

realfinder:
> 
> 
> Two reasons I didn’t:
> 
>     * Perhaps the same feature could be extended in the future for more run un-compatibility-fixes, e. g. we could try to combine bitmaps (especially BorderStyle 3) for whole lines instead of per run. Or would this be too unstable of an API?

I think I agree, this look bad https://github.com/libass/libass/issues/148 also here https://web.archive.org/web/20160120113701/https://code.google.com/p/libass/issues/detail?id=114

it can be happen even in English
image
vs vsfilter
image



> 
> 
> 
> As for the default setting:
> 
> How about we make it default to whole-line bidi for `Encoding=-1` and VSFilter-bidi for all other `Encoding` values? -1 is our existing bidi extension that changes base paragraph direction.
> 
> Unfortunately, being a left-to-right user myself, I don’t know (and haven’t investigated yet) whether this compromise would help in practice. But it sounds like it might.

seems fine, I really hope to fix those soon so I can let vsfilter RIP and only use libass

realfinder:
using assrender https://github.com/pinterf/assrender , both avs+ and assrender should work in Linux nowadays (I did use windows) and since its easy to build and test using avspmod now (at least for me) I did some tests

using the sample from https://github.com/libass/libass/issues/71

`LoadcPlugin("D:\assrender-master\bin\Release_Win32\assrender runs_bidi_shaping.dll")
ColorBars
converttoyv12
assrender("D:\AvisynthRepository\script.ass")`
image
it seems work now

BUT https://github.com/libass/libass/issues/374 now look like this
image

astiob:
Thanks! I can confirm that the sample in #374 is broken and I’ve tracked down the bug. The specific sample in #374, with the explicit line break, will now be fixed. But this has highlighted a problem.

We run bidi, then wrap lines.

But each line is its own run and needs its _own_ bidi.

I don’t know how to fix this reliably.

astiob:
For example: take the sample from #374 and replace the line break by a space. Use a font large enough that the automatic line wrapping breaks the line in the same place as before. VSFilter will show the same result that is seen in #374, but libass—*both with and without this PR*—will show the output in the screenshot above.

astiob:
I’ve pushed a new version:

* The first commit now enables whole-text processing for events with Encoding −1 regardless of the state of the new `ASS_Feature`. This preserves our current whole-line bidi, and the later commits further improve bidi and shaping.

* I’ve reverted my decision from the first two comments on this PR and removed the “autodetect base direction for whole event, not per line” commit. Several reasons for this:
    * I had misunderstood FriBidi code: `fribidi_get_par_embedding_levels` really expects only a single paragraph as input and does _not_ internally split it, but rather stops looking for explicit bidi controls after it reaches the first paragraph terminator (including `'\n'`). So my commit broke explicit bidi controls beyond the first line break.
    * WebVTT demands independent direction autodetection for each line within a multiline event. (See <i>[A text alignment](https://www.w3.org/TR/webvtt1/#webvtt-cue-text-alignment),</i> in particular Example 18.)
    * I checked macOS’s stock text editor (which generally does typography well), and it autodetects direction for each LF-delimited line separately, too. (Although my macOS is far from latest.)
    * This change also “broke” #374, which is what thankfully made me notice it, but there’s a deeper reason why it really broke; see previous comments and below.

* I’ve accordingly reworked the “remember resolved base direction” commit to account for the fact that each paragraph has its own direction. (Note that direction autodetection can only happen when `shaper->whole_text`, as it now enforces whole-text processing.)

---

What remains is to somehow re-bidi runs that line wrapping splits into multiple parts.

An obvious, dumb approach is to simply rerun bidi and shaping on such runs after line wrapping. But it may be possible that this changes their widths and invalidates the line wrapping itself.

realfinder:
it's seems fine now, I will do more test later

TheOneric (commented):
A first look at everything but the \fay-changes; will lokk at them later, possibly tomorrow.
For the parts I did look at, there are only some style nits. In general:

- I agree on running bidi per paragraph.
- I'm not completly happy with the name `WHOLE_TEXT`, as it's a bit too unspecific and doesn't quite fit, as eg bidi happens per paragraph. But I'm also lackin
g any better ideas, so …  
- I'm not sure why you wnat to re-bidi after line wrapping
  As I understand it (after skimming through) the Unicode/WebVTT standards require BiDi per *paragraph* not per *line*. So only forced line breaks should matte
 here.
  If it's about compatability, then the obvious answer to how to do this is to just do the same as VSF sans UB, non-terminating loops etc.

Can you rebase this, throwing out the commits from #445, so we can run regressions test on it?

astiob:
Thanks for the (initial) review! I’ve rebased on master and thrown out the duplicate commits. In retrospect, I should’ve done this before deleting the `fay-fscxy` branch to reduce the potential for human error during rebasing :-)

But why do we want leading operators again?

We have 32 instances of leading operators, in 21 distinct consecutive-line blocks, as per this regex:

```regex
^\s*([|^%>]|/[^/*]|\+[^+]|-[^-]|&[& ]|< )
```

including 11 operators in 3 line blocks in `ass_shaper.c` before this PR, none of which are modified in this PR if I’m not mistaken.

We have 165 instances of trailing operators, in 88 distinct consecutive-line blocks, as per this analogous regex:

```regex
([|^%<]|[^/*]/|[^+]\+|[^-]-|[& ]&| >)\s*$
```

including 11 operators in 5 line blocks in `ass_shaper.c` before this PR, none of which are modified in this PR if I’m not mistaken; but neighbouring lines are modified.

astiob:
> I'm not sure why you wnat to re-bidi after line wrapping
> As I understand it (after skimming through) the Unicode/WebVTT standards require BiDi per _paragraph_ not per _line_. So only forced line breaks should matter here.
> If it's about compatability, then the obvious answer to how to do this is to just do the same as VSF sans UB, non-terminating loops etc.

Yes, it’s for VSFilter compatibility when `WHOLE_TEXT` is disabled. Doing this literally the same as VSFilter isn’t entirely practical, because (1) that’s slow with lots of unnecessary re-layouts; (2) that pretty much requires rewriting the whole bidi+shaping+wrapping pipeline. I’d like to find a more optimal way to do it. However, it’ll have to involve _some_ re-layouts and probably need the pipeline rewritten anyway.

Also, regardless of `WHOLE_TEXT`, limited re-layout will/should be needed for Unicode line breaking anyway (and technically theoretically even for our current space-only line breaking if a weird font has ligatures or kerning spanning across spaces), but it will specifically involve re-shaping but not re-bidi.

libass/ass_shaper.c
```diff
@@ -1004,17 +1025,23 @@ FriBidiStrIndex *ass_shaper_reorder(ASS_Shaper *shaper, TextInfo *text_info)
...
         shaper->cmap[i] = i;
 
-    // Create reorder map line-by-line
-    for (i = 0; i < text_info->n_lines; i++) {
-        LineInfo *line = text_info->lines + i;
-        FriBidiParType dir = FRIBIDI_PAR_ON;
+    // Create reorder map line-by-line or run-by-run
+    int last_break = 0;
+    GlyphInfo *glyphs = text_info->glyphs;
+    for (i = 0; i < text_info->length; i++) {
```
MrSmile, line 1060: While at it, you can move `i` and `ret` definitions from the start of the function to the first use.

TheOneric:
> Yes, it’s for VSFilter compatibility when WHOLE_TEXT is disabled. Doing this literally the same as VSFilter isn’t entirely practical, because […]

Thanks for clarifying!

> But why do we want leading operators again?

uhh.. we don't; I derped and got confused about libass style guidelines.

astiob:
> we don't; I derped and got confused about libass style guidelines.

Right; I’ve just (finally) grepped my IRC logs and seen that we concluded we prefer trailing operators in a discussion we had before releasing 0.15.0. But as I said another time…

> tbh I just grep the codebase every time I have similar questions and try to pick the most common version :)

still true! 😅

TheOneric (approved):
Seems good, just to confirm my understanding of VSF's `\fay`-logic:

`\fay`'s argument is relative to the unscaled glyph **width**, but shear is still affected by *vertical* scaling. In particular, if it weren't for VSF resetting runs, that means having some `\fscx0` glyphs between other text would create a sudden jump (up|down)wards of the text's baseline.
But since VSF resets shear on new VSF-runs, there are discontinuities anyway, and they always reset to the original baseline position. Since the whole `\fscx0` segment isn't visible in the end anyway, we can just skip them in `apply baseline shear. Right?

Then a question about “Reset baseline shear …”'s commit message:
Why do we want `\fscx0`-glyphs to advance the baseline height with `WHOLE_TEXT`? Intuitively I'd expect them to just vanish. Do you feel different, or do you think there's a use for introducing discontinuities into `WHOLE_TEXT`'s `\fay`?

astiob:
Yep.

VSFilter applies shear (`\fax` and `\fay`) before scaling (`\fscx` and `\fscy`) while transforming each run’s vector outlines for rasterization. Therefore, shear works in unscaled coordinates, and the result then gets affected by scaling.

In case of a glyph with nonzero `\fay`, the baseline rises (or descends) under that glyph; the wider the glyph, the bigger the total change. If that sheared glyph is then scaled by `\fscx0`, it collapses to a zero-width vertical line segment, which is invisible, but the baseline’s position for the next glyph has already risen due to the shear that was applied earlier. All in all, if runs weren’t reset, the next glyph would seem to experience a sudden jump in baseline position.

And that’s what “should” happen with `WHOLE_TEXT`, as it currently handles shear the same way but does not reset runs. This would be consistent with the visible effects of shear in all other cases.

But I don’t think this is _right._ The thing is, it’s not `\fscx0\fay` handling that isn’t right—it’s shear handling in general. Intuitively, I’d expect scaling to apply _first_ and whole lines, or even whole events, to be sheared afterwards. I think, in fact, that this (with lines) is how libass originally implemented shear before it got a much needed fix for compatibility; but I’m not confident that it actually worked as well as I make it sound.

TheOneric:
Were there any blockers/todos left for this? *(Apart from it merge-conflicting with the also pending #545)*

astiob:
Todos yes: rerun bidi after line wrapping somehow. But I don’t think it should be a blocker, as the current PR still provides a strict improvement in VSFilter compatibility and hopefully covers most of the real-world cases.

Plus bikesheds: about the feature name and about the feature mechanism itself.

TheOneric:
> Plus bikesheds: about the feature name and about the feature mechanism itself.

iirc none of us could come up with a better name than `WHOLE_TEXT` and imho redesigning the internal feature handling is orthogonal to adding `WHOLE_TEXT`. LGTM.

TheOneric (approved):

astiob:
Quick summary of our naming bikeshed brainstorm a while ago on IRC:

* What I was really going for here is “text _layout_ without the quirks caused by VSFilter’s run splitting”, or in other words, “text _layout_ that you can make sense of without knowing about VSFilter’s (or libass’s) run splitting”. We concluded that the keyword is “layout”.

    This is why it covers bidi, shaping, `\fsp` and even shear, but not karaoke filling or opaque borders.

    Now that I’m writing this, I guess counting shaping as layout is a bit odd. It does definitely affect layout though.

* MrSmile has proposed `SEAMLESS_TEXT_LAYOUT`. I’ve grown to like it, except that “seamless” smells funny. What do native speakers think?

TheOneric:
> - MrSmile has proposed SEAMLESS_TEXT_LAYOUT. I’ve grown to like it, except that “seamless” smells funny. What do native speakers think?

In absence of any native speaker feedback on “semaless”, does `WHOLE_TEXT_LAYOUT` smell less funny and facilitate a merge?

realfinder:
hi, sorry for my late (I've been busy for over a month) I have no problem with SEAMLESS_TEXT_LAYOUT, if it about "funny" then I think it already is in ".ass"

astiob:
Rebased on top of current `master` with the new `ASS_Feature` internals, and renamed to `WHOLE_TEXT_LAYOUT`. If anyone wants to give this a final look-over or test-run just in case I messed something up during the rebase, please do so. Barring objections, I think I’ll merge this in a couple of days.

TheOneric:
The range-diff of the rebase seems perfectly fine to me and ART also finishes as expected.

TheOneric:
@astiob: shouldn't this have bumped `LIBASS_VERSION` in `ass.h` by 1 to make it possible to `#if` guard usage of the feature enum? *(late, but we can still bump it now I guess)*

astiob:
Indeed, thank you. Pushed.

PR #445, opened 2020-10-29
Fix \fscx0, \fscy0 and a positioning bug with right-to-left \fay

astiob:
Avoid the divisions by zero for \fscx0 and \fscy0. Regardless of the divisions by zero, avoid drawing any border. Two birds with one stone. Extra speed is a bonus.

~~Reset baseline everywhere that it should reset, rather than on the arbitrary condition that \fay must change.~~ in #441 now

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2130,6 +2130,8 @@ static void apply_baseline_shear(ASS_Renderer *render_priv)
         if (text_info->glyphs[i].linebreak || last_fay != info->fay)
             shear = 0;
         last_fay = info->fay;
+        if (!info->scale_x || !info->scale_y)
+            info->skip = true;
         if (info->skip)
             continue;
         for (GlyphInfo *cur = info; cur; cur = cur->next)
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2055,9 +2055,6 @@ static void retrieve_glyphs(ASS_Renderer *render_priv)
 
         // add horizontal letter spacing
         info->cluster_advance.x += info->hspacing_scaled;
-
-        // add displacement for vertical shearing
-        info->cluster_advance.y += (info->fay / info->scale_x * info->scale_y) * info->cluster_advance.x;
     }
 }
 
@@ -2098,23 +2095,15 @@ static void reorder_text(ASS_Renderer *render_priv)
     // Reposition according to the map
     ASS_Vector pen = { 0, 0 };
     int lineno = 1;
-    double last_pen_x = 0;
-    double last_fay = 0;
     for (int i = 0; i < text_info->length; i++) {
         GlyphInfo *info = text_info->glyphs + cmap[i];
         if (text_info->glyphs[i].linebreak) {
-            pen.y -= (last_fay / info->scale_x * info->scale_y) * (pen.x - last_pen_x);
-            last_pen_x = pen.x = 0;
+            pen.x = 0;
             pen.y += double_to_d6(text_info->lines[lineno-1].desc);
             pen.y += double_to_d6(text_info->lines[lineno].asc);
             pen.y += double_to_d6(render_priv->settings.line_spacing);
             lineno++;
         }
-        else if (last_fay != info->fay) {
-            pen.y -= (last_fay / info->scale_x * info->scale_y) * (pen.x - last_pen_x);
-            last_pen_x = pen.x;
-        }
-        last_fay = info->fay;
         if (info->skip)
             continue;
         ASS_Vector cluster_pen = pen;
@@ -2127,6 +2116,25 @@ static void reorder_text(ASS_Renderer *render_priv)
             cluster_pen.y += info->advance.y;
             info = info->next;
         }
+    }
+}
+
+static void apply_baseline_shear(ASS_Renderer *render_priv)
+{
+    TextInfo *text_info = &render_priv->text_info;
+    FriBidiStrIndex *cmap = ass_shaper_get_reorder_map(render_priv->shaper);
+    int32_t shear = 0;
+    double last_fay = 0;
+    for (int i = 0; i < text_info->length; i++) {
+        GlyphInfo *info = text_info->glyphs + cmap[i];
+        if (text_info->glyphs[i].linebreak || last_fay != info->fay)
+            shear = 0;
+        last_fay = info->fay;
+        if (info->skip)
+            continue;
+        for (GlyphInfo *cur = info; cur; cur = cur->next)
+            cur->pos.y += shear;
+        shear += (info->fay / info->scale_x * info->scale_y) * info->cluster_advance.x;
     }
 }
 
@@ -2669,6 +2677,8 @@ ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
     // determing text bounding box
     ASS_DRect bbox;
     compute_string_bbox(text_info, &bbox);
+
+    apply_baseline_shear(render_priv);
 
     // determine device coordinates for text
     double device_x = 0;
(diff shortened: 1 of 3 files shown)


TheOneric:
This breaks positioning for the following sample
```
[Script Info]
ScaledBorderAndShadow: yes
YCbCr Matrix: TV.709
PlayResX: 1920
PlayResY: 1080

[V4+ Styles]
Style: Type: Titel,Ubuntu Titling,82,&H008B7CEB,&H00CDB85F,&H004041C5,&H00000000,-1,0,0,0,100,100,5,0,1,0,0,5,10,10,10,1

[Events]
Dialogue: 0,0:00:00.00,0:00:01.76,Type: Titel,,0,0,0,,{\t(1460,1750,\fsp-49)}{\pos(960,350)}{\fade(0,0,255,0,1460,1490,1660)}ZI{\fscy0}MM{\fscy100}ER{\fscy0}C{\1a&HFF&\fscy130\fscx120\fnDejaVu Sans}▴{\fnUbuntu Titling\1a&H00&\fscy100\fscx100}MP
Dialogue: 0,0:00:00.00,0:00:01.76,Type: Titel,,0,0,0,,{\t(1460,1750,\fsp-49)}{\pos(960,350)\k200}{\k10}{\fade(0,0,255,0,1460,1586,1665)}{\fscy0}ZI{\fscy100}MM{\fscy0}ER{\fscy100}C{\1a&HFF&\fscy130\fscx120\fnDejaVu Sans}▴{\fnUbuntu Titling\fscx100}{\fscy0}MP
Dialogue: 0,0:00:00.00,0:00:01.76,Type: Titel,,0,0,0,,{\t(1460,1751,\fsp-49)}{\pos(960,350)\k200}{\k10}{\fscy0}ZIMMERC{\fscy100}{\fscy130\fscx120\fnDejaVu Sans}▴{\fnUbuntu Titling\fscx100}{\fscy0}MP
```

libass master & VSFilter:
master_71595ab

with patch:
patch_v01

astiob:
Ah, I missed `align_lines`.

On the other hand, while investigating this, I discovered _another_ VSFilter incompatibility in event positioning: for right-to-left lines that are vertically centred or explicitly positioned, we align the appropriate point on the screen with the _left edge of the rightmost glyph_ as it’s the first glyph in logical order. This doesn’t matter when \fay0, but it makes a difference with nonzero \fay.

So now what I want to do is factor out the vertical shear computations for positions into a separate function that will be called after `align_lines`.

But while doing _that,_ I realized… this shear thing is pretty broken, and I don’t know what the right way to progress is at all.

## The problem

As you remember, VSFilter splits text into runs, runs bidi on each run separately and shears each run separately. So run boundary = bidi boundary = shear reset.

In libass, we’re applying shear in _visual order_ of glyphs, because, well, we want the visual baseline to suffer shear. But we’re running bidi on the _whole text._ But we’re trying to replicate VSFilter’s _run-based_ shear reset, which uses run boundaries in _logical order_ of glyphs. What we get is… a mess. The visual result is different in current master and with this PR in its current state, but neither makes immediate sense to me, as the visual order and the logical order mix up horribly.

```
Dialogue: 0,0:00:00.00,0:00:10.00,BorderStyle3,,0,0,0,,{\bord1\an5\fnArial\fay-0.5}ربية foo ربيةربية{\fay-0.2}ربية
```

master:
Screenshot with libass master shows, from left to right: “ربية foo” with \fay-0.5, baseline reset, “ربية” with \fay-0.3, baseline reset, “ربيةربية” with \fay-0.5.

This PR:
Screenshot with this PR in its current state shows, from left to right: “ربية foo” with \fay-0.5, no baseline reset, “ربية” with \fay-0.3, no baseline reset, “ربية” with \fay-0.5, baseline reset, “ربية” with \fay-0.5.

Of course, this is related to #441, and together with it, this PR should produce the same result as VSFilter.

But since we want to keep #441 optional (even if on by default), what is the right thing to do when it’s disabled and until #441 lands in the first place?

Ironically, even before this discovery, I was thinking of augmenting #441 to disable the shear baseline reset altogether when the new feature (`WHOLE_TEXT` or whatever) is enabled. That may be the best way forward. But then we kinda can’t merge this until that lands? Or should I just move the commit that makes \fay reset on each run from this PR into #441?

astiob:
Since this PR’s last update, I believe your sample should be fixed, as well as the (pre-existing) right-to-left issue that I discovered while investigating it.

I’ve moved the VSFilter-compatible \fay baseline reset to #441 as https://github.com/libass/libass/pull/441/commits/267b49371a7371a41ac6a036a3885c98366dc1a8. The present PR preserves our current \fay behaviour. There’s a logical flaw in it, in that \fscx0 glyphs don’t affect baseline regardless of their \fay, but, now that I think about it, even this probably actually matches our current behaviour with the NaNs if you have an even number of \fscx0 glyphs (otherwise you don’t see anything at all).

TheOneric:
Can confirm the samples are now fixed. Moving the baseline reset to #441 also seems good to me.

In the second commit you are noting, that this one both *“fixes a regression in the last commit”* and that this commit *“would also have prevented the bug fixed in the previous commit”*. Can this be resolved without introducing a regression in between those two commits? This may cause trouble later on if we attempt something related is bisected.

astiob:
Ah, I think I actually already changed it so there’s no regression (I think it was related to the baseline change) but left the phrase in the commit message by accident. I’ll double-check and fix the commit message.

astiob:
I’ve rebased on current master (no conflicts), removed the outdated mention of a regression that no longer exists, and polished the commit messages a little more.

TheOneric:
The second commit message mistakenly mentions `last_pen_y`, otherwise seems reasonable to me.
I'm not too familiar with some of the touched code, so while not a fully-qualified review, fwiw consider this patchset acked by me.

astiob:
> The second commit message mistakenly mentions `last_pen_y`

It looks like I mistakenly copied that warning in the commit message from an earlier version of the commit, after I had already changed the commit so that the warning was inapplicable/unnecessary. I’ve just removed the warning entirely.

astiob:
If anyone else wants to review this PR, please say so. Otherwise, I’m going to merge this soon™.

libass/ass_render.c
```diff
@@ -2130,6 +2119,28 @@ static void reorder_text(ASS_Renderer *render_priv)
...
+    int32_t shear = 0;
+    double last_fay = 0;
+    for (int i = 0; i < text_info->length; i++) {
+        GlyphInfo *info = text_info->glyphs + cmap[i];
+        if (text_info->glyphs[i].linebreak || last_fay != info->fay)
+            shear = 0;
+        last_fay = info->fay;
+        if (info->skip)
+            continue;
+        while (info) {
```
rcombs, line 2137: I'd use a second variable here, rather than reusing the first one and then later resetting it to its previous value. Could even be a 2-line for loop then.

rcombs (commented):
One minor nit; otherwise looks fine to me.

libass/ass_render.c
astiob: Fair point. Before I go and commit a change: I tried to keep this function similar to `reorder_text` from which it is being extracted. Do you think it’s not worth it? Or perhaps should I edit the same code in `reorder_text` while I’m at it?

astiob: @rcombs I’ve changed this to a two-line `for` loop with a second variable. I’ve also added a commit that changes `reorder_text` to avoid similarly resetting `info` but _without_ a second variable, but I’m not 100% sure I’m happy with how it turned out, so I’d appreciate if you could glance it over and say if it looks fine to you.

TheOneric: fyi: the same pattern of reusing + resetting the variable is also used eg in `preliminary_layout`, but I agree using a new var is clearer. If we want to change it, I'd say we can do that whenever we happen to modify an affected function anyway. *(which is the case for `reorder_text` in this patch series)*

PR #446, opened 2020-10-29
Treat zero-area events like VSFilter

TheOneric:
Instead of calculating the intersection rect and testing for an zero area *(as VSF does)*, treating zero-area event's as fixed and otherwise keeping the current intersection code should yield identical results (while probably being a tad faster than VSF).

There's one more `shift_event` call which I didn't modify here: https://github.com/libass/libass/blob/700e131c78e6c917369c02161c4c8e2408d963e5/libass/ass_render.c#L3031-L3037
I'm not sure what this one does, but it seems to work without modifying this call.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -3005,7 +3005,9 @@ fix_collisions(ASS_Renderer *render_priv, EventImages *imgs, int cnt)
     // fill used[] with fixed events
     for (i = 0; i < cnt; ++i) {
         ASS_RenderPriv *priv;
-        if (!imgs[i].detect_collisions)
+        // VSFilter considers events colliding if their intersections area is non-zero,
+        // zero-area events are therefore effectively fixed as well
+        if (!imgs[i].detect_collisions || !imgs[i].height  || !imgs[i].width)
             continue;
         priv = get_render_priv(render_priv, imgs[i].event);
         if (priv && priv->height > 0) { // it's a fixed event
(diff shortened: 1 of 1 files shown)


astiob:
I think the other `shift_event` you mentioned above handles events that stay on screen from previous frames, putting them in the same position where they got previously moved to avoid collisions with _other_ events (which may themselves have already disappeared from screen). And I think no changes are necessary there.

LGTM.

astiob (commented):
Hang on. Possibly not LGTM after all; sorry.

In fact, you’re not just treating them as “fixed events” in the sense that this function uses, i. e. events that are subject to collision avoidance and have already gone through it and found their final screen position.

Those “fixed events” are in the `used[0:cnt_used]` array, and all further events are checked for intersections against them. Which, of course, you want to avoid. So you don’t put the zero-area events in the `used` array… but on a subsequent frame, you do, because the first loop does that.

Try a test with an extra event that appears only on the second frame. I _think_ it will be moved to avoid colliding with the zero-area event with your current code.

It seems to me that you want to treat zero-area events as _explicitly positioned_ events instead, by changing this:

```c
if (!imgs[i].detect_collisions)
```

to this:

```c
if (!imgs[i].detect_collisions || !imgs[i].height || !imgs[i].width)
```

astiob (approved):

TheOneric:
Reworded commit messages to fix typos and improve the explanation for the second commit (although it didn't get shorter).
Code identical.

PR #448, opened 2020-10-30
ass_library: Constify ass_add_font

chouquette:
Both parameters are copied and are immutable. Make it clear through the
function prototype


astiob (approved):
Seems fair.

PR #450, opened 2020-11-06
Autotools Cosmetics: Part 1

TheOneric:
Except for the first commit, all changes are purely cosmetic.

Commit 1 adds NASM support for Haiku and makes configure print an error when trying to use NASM on a supported architecture but unrecognised platform. *(currently it just silently produces borked files)*

The main cosmetic improvement of this PR is reodering the configure logic to be imo easier to understand. While doing so I noticed some odd bits; the last 3 commits remove these seemingly unnecessary parts. *(Please tell me if they did serve some purposes afterall)*

Configuration and *(non-static)* compilation was tested on Debian Buster, Debian Sid, OpenIndiana Hipster and Haiku R1/beta2.

---

Furthermore I'm planning to replace the bare shell control structures with proper Autotools-M4 macros and change the `configure.ac` style to better reflect the control structures some macros expand to in a future pr., Imho this would further improve readability.
The reason this is not done here in Part 1, is that any requested change on the reordering would likely cause many merge conflicts with a later style-change commit, which seems like unnecessary work.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -282,8 +282,8 @@ AM_COND_IF([ASM],
     )
 
 AM_COND_IF([ENABLE_LARGE_TILES],
-    [AC_DEFINE(CONFIG_LARGE_TILES, 1, [use large tiles])]
-    [AC_DEFINE(CONFIG_LARGE_TILES, 0, [use small tiles])],
+    [AC_DEFINE(CONFIG_LARGE_TILES, 1, [use large tiles])],
+    [AC_DEFINE(CONFIG_LARGE_TILES, 0, [use small tiles])]
     )
 
 ## Setup output beautifier.
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_INIT(libass, 0.15.0)
-AM_INIT_AUTOMAKE
+AM_INIT_AUTOMAKE([foreign])
 AC_CONFIG_MACRO_DIR([m4])
 # Disable Fortran checks
 define([AC_LIBTOOL_LANG_F77_CONFIG], [:])
--- a/configure.ac
+++ b/configure.ac
@@ -46,6 +46,10 @@ AC_ARG_ENABLE([large-tiles], AS_HELP_STRING([--enable-large-tiles],
 OLDLIBS="$LIBS"
 LIBS=
 use_libiconv=false
+## Some iconv libraries like GNU's libiconv define iconv_open as a macro to
+## libiconv_open. As SEARCH_LIBS tests linking not compilation, check for
+## libiconv_open first. SEARCH_LIBS is smart enough to not add -liconv a second
+## time in case both versions are defined in the local libiconv.
 AC_SEARCH_LIBS([libiconv_open], [iconv], use_libiconv=true)
 AC_SEARCH_LIBS([iconv_open], [iconv], use_libiconv=true)
 AS_IF([test "x$use_libiconv" = xtrue], [
--- a/configure.ac
+++ b/configure.ac
@@ -21,18 +21,7 @@ AC_CHECK_HEADERS([stdint.h iconv.h])
 # Checks for library functions.
 AC_CHECK_FUNCS([strdup strndup])
 
-# Checks for libraries.
-# Add them to pkg-config for static linking.
-OLDLIBS="$LIBS"
-LIBS=
-use_libiconv=false
-AC_SEARCH_LIBS([libiconv_open], [iconv], use_libiconv=true)
-AC_SEARCH_LIBS([iconv_open], [iconv], use_libiconv=true)
-AC_CHECK_LIB([m], [fabs])
-pkg_libs="$LIBS"
-LIBS="$OLDLIBS $LIBS"
-
-# Check for libraries via pkg-config
+# Query configuration parameters and set their description
 AC_ARG_ENABLE([test], AS_HELP_STRING([--enable-test],
     [enable test program (requires libpng) @<:@default=no@:>@]))
 AC_ARG_ENABLE([compare], AS_HELP_STRING([--enable-compare],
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -264,12 +168,121 @@ Either DirectWrite (on Windows), CoreText (on OSX), or Fontconfig \
...
+AC_SUBST([AS], ["$AS"])
+
+
+# Relay config results to output files
+
+## Setup conditionals for use in Makefiles
+AM_CONDITIONAL([ASM], [test x$enable_asm != xno])
+AM_CONDITIONAL([INTEL], [test x$INTEL = xtrue])
+AM_CONDITIONAL([X86], [test x$X86 = xtrue])
+AM_CONDITIONAL([X64], [test x$X64 = xtrue])
```
astiob, line 260: Should this perhaps be closer to `AC_SUBST([PKG_LIBS_DEFAULT]` etc.?

TheOneric: Now it is, thanks.

MrSmile:
You should also port the first commit from https://github.com/libass/libass/pull/449, as it's a configure bug and conflicts with this PR.

TheOneric:
> You should also port the first commit from #449

I was thinking of including it in Part 2 together with the re-style making errors like this one harder to slip in, but sure I can already include the basic fix now.


After some more testing here is a slightly changed version of the patch series.
 - `AC_SEARCH_LIBS` is smart enough to not add something to `LIBS` if the current `LIBS`-libraries already contain a function.
    Therefore having two checks for different iconv functions in series should be safe afterall. Instead of making the second check 
    conditional, I'm now adding a comment with an explanation.
 - I dropped the change from `AC_CHECK_LIB` to `AC_SEARCH_LIBS` for the math library.
   While this does indeed avoid a superflous `-lm` on platforms like Haiku or [MacOS](https://travis-ci.org/github/libass/libass/jobs/741778098#L231) that have standard math functions integrated into their libc, unlike glibc-Linux and (iirc) OpenIndiana,
    FreeBSD has the math functions only **partially** integrated into its libc. Namely `fabs`, which we check for, is in libc, while
    other functions  we use are in `lm`, leading to linking errors. *(On first glance it seems like C90 math functions are in libc, while
   C99 math functions need `-lm`, but I'm inclined to just leave a potentially unneeded `-lm` instead of trying to deal with all the possible ways a system might split up its math functions)*

---

On a side note there was another instance of strange behaviour with FreeBSD, which also affects unpatched master. When I tried to build libass outside of the ports-infrastructure, during configuration it would claim to not find `libiconv_open` and not requiring any additional lib for `iconv_open`. During linking however, it failed with unresolved references to `libiconv_open`. Building libasss with the ports-infrastructure resolved this, so this might also be due to me misconfiguring something.
Furthermore, FreeBSD's libasss port applies [this patch to NASM-configuration](https://svnweb.freebsd.org/ports/head/multimedia/libass/files/patch-configure.ac?view=markup). Given that Debian doesn't patch its libass and also builds for i386, this patch probably doesn't apply universally though. *(and I don't know enough about it to to make an assessment of how to properly handle this)*

astiob:
We're checking for different, mutually exclusive functions in the two iconv variants, so you do need to make the second conditional on the first failing to avoid linking in both iconvs. Neither iconv allows the other check succeed without linking in an extra lib (unless iconv is part of libc etc.).

astiob:
On the other hand, is it even possible that the same -liconv will bring in two different libiconvs in the two checks, allowing both of them to succeed? Probably not. So then it's safe to keep them both unconditional.

astiob:
I suppose there's a danger of linking in an extra iconv when libc already contains one. But if we're worried about that, something more clever than a fixed sequence of conditional searches is needed.

TheOneric:
> I suppose there's a danger of linking in an extra iconv when libc already contains one. But if we're worried about that, something more clever than a fixed sequence of conditional searches is needed.

Assuming libc’s iconv implementation is called plain `iconv_`checking for this variant first and then only conditionally for `libiconv_` would probably work out to prefer libc’s implementation in most scenarios.

However, **if** both a libc and a library implementation are present and
 - both use the plain `iconv_` symbols, there would be a symbol collision everytime one tries to link `-liconv` begging the question what use the lib even serves. This seems like a broken environment to me.
 - the lib implementation uses `libiconv_` symbols. In this case there are also twodiffeenret header files (for recefrence [glibc](https://elixir.bootlin.com/glibc/latest/source/iconv/iconv.h) and [GNU libiconv](https://git.savannah.gnu.org/gitweb/?p=libiconv.git;a=blob;f=include/iconv.h.in;h=eced350627d1f5efd97bca497315f1a8f8b80657;hb=HEAD)):
    - If the lib header is the one that ends up being included, we can assume the user _wants_ to override the libc implementation and we _must_ link `-liconv` to not end up with unresolved symbols.
    - If the libc header is included then, even though we may link `-liconv`, the lib implementation is never used. The superfluos `-liconv` is maybe not ideal, but afaict also not harmful.

*(Ofc there's still the weird behaviour I encountered on FreeBSD, with supposedly no `libiconv_` present, no lib for `iconv_` needed, but failing with unresolved `libiconv_` symbols. The only explanation I can think of for that, is that the libiconv-header was in the default include path, but the libiconv-library not in the default lib-path, which imho is a somewhat broken state and not something we need to take into account in our configuration)*

TheOneric:
I'm currently not sure if it's better to use `AC_SEARCH_LIBS` with `lrint` for libm and hope that no system splits up functions defined in the same even weirder than FreeBSD, or stick with `AC_CHECK_LIBS` causing unnecessary `-lm` on some systems.  
Unless I overlooked something with the iconv check (@astiob), the rest should be fine.

Here's an overview of lib integration based on a libass-compile-test:
 ✓ : is part of libc *(or at least all parts of it used in libass)*
 ✘ : needs an additional lib flag
 ~ : some parts are in libc, some in a seperate lib

```
               | math | iconv
---------------+------+-------
 glibc¹        |  ✘   |  ✓
---------------+------+-------
 musl²         |  ✓   |  ✓
---------------+------+-------
 illumos³      |  ✘   |  ✓
---------------+------+-------
 Haiku⁴        |  ✓   |  ✘
---------------+------+-------
 MacOS⁵        |  ✓   |  ✘
---------------+------+-------
 FreeBSD⁶      |  ~⁷  |  ✓

⁷: eg fabs is in libc, but lrint needs libm

¹: Debian Buster and Sid
²: Adélie Linux 1.0-rc2
³: OpenIndiana Hipster
⁴: R1/beta2
⁵: 10.14.6
⁶: 12.2-p0
```

If there are no further remarks, I'm going to push Part 1 either with `AC_CHECK_LIB([m], [fabs])` or `AC_SEARCH_LIBS([lrint], [m])` tomorrow or so.

astiob:
FWIW, `lrint` does sound like a better candidate than `fabs` to check for, as it’s a newer function dating only (heh) to C99. Then again, for the `AC_CHECK_LIB` version, I guess the exact choice of function makes no difference, as we don’t handle failure at all.

astiob:
> Furthermore, FreeBSD's libasss port applies [this patch to NASM-configuration](https://svnweb.freebsd.org/ports/head/multimedia/libass/files/patch-configure.ac?view=markup). Given that Debian doesn't patch its libass and also builds for i386, this patch probably doesn't apply universally though.

@rcombs Are you able to comment on this?

TheOneric:
> for the AC_CHECK_LIB version, I guess the exact choice of function makes no difference, as we don’t handle failure at all.

Unless the compiler ignores missing libs, `AC_CHECK_LIB` will fail if there's no libm at all. This could either mean libm is integrated into libc, or the platform doesn't have math functions, so this version is not suitable to do error handling.
Adélie and apparently all other tested platforms (as otherwise they wouldn't get the `-lm` flag), do provide a static stub libm.a *(only 8 bytes in Adélie)*. I suspect this is done to be compatible with broken config-scripts assuming libm is always seperate from libc. *(Renaming `libm.a` to something else iin Adélie causes `AC_CHECK_LIB([m], [fabs], [], AC_MSG_ERROR([No math lib!]))` to error out)*

`AC_SEARCH_LIBS` on the other hand only fails if the function is neither in libc nor in libm and can therefore be used to detect systems missing `<math.h>`-functions or having them in another lib. Eg:
`AC_SEARCH_LIBS([lrint], [m], [], [AC_MSG_ERROR([Unable to locate math functions!])])`

TheOneric:
Also I again need to correct myself: Reading [this commit message](https://salsa.debian.org/multimedia-team/libass/-/commit/92a5c15e7d5feae3087b3f04346469af0ae18c9d) I confused architectures (x32 ≠ i386).
The [i386-buildlog](https://buildd.debian.org/status/fetch.php?pkg=libass&arch=i386&ver=1%3A0.15.0-1&stamp=1603816961&raw=0) shows that nasm is **not** present in Debian's i386-libass-builds. It was been explicitly disabled on i386, due to [not producing PIC-code](https://salsa.debian.org/multimedia-team/libass/-/commit/d9eae1f0aefacf9b8c838cb5108dcc100f336e0b).

Issue #451, opened 2020-11-14
ASM config may be borked on some 32bit x86 ABIs


TheOneric:
As we found out in #450 , FreeBSD is patching the NASM-configuration⁵ to work for their 32-bit ABI.
Regarding Linux-distros, Debian consiciously disabled NASM for i386-architectures¹ (but it **is** enabled for x32²³), while iiuc Gentoo⁴ has NASM enabled for i386 and amd64 but not x32.

Unfortunately, this is the end of useful information I have about this, as I don't know about other BSD's and OpenIndiana⁶ has NASM explicitly disabled *(presumably because when libass was added to OI's repo, the Solaris-config-bug⁷ wasn't yet fixed)* and Haiku currently also has NASM disabled *(#450 fixes at least x86_64-Haiku)*.

¹:https://salsa.debian.org/multimedia-team/libass/-/commit/d9eae1f0aefacf9b8c838cb5108dcc100f336e0b
²:https://buildd.debian.org/status/fetch.php?pkg=libass&arch=x32&ver=1%3A0.15.0-1&stamp=1603818152&raw=0
³:https://salsa.debian.org/multimedia-team/libass/-/commit/92a5c15e7d5feae3087b3f04346469af0ae18c9d
⁴:https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/libass/libass-0.14.0.ebuild
⁵:https://svnweb.freebsd.org/ports/head/multimedia/libass/files/patch-configure.ac?view=markup
⁶:https://github.com/OpenIndiana/oi-userland/blob/26aaf0b/components/library/libass/Makefile
⁷:https://github.com/libass/libass/commit/62dba758ea64d9819d0021f1d6c9cbb9369d63b1
⁸:https://github.com/haikuports/haikuports/blob/45de8e68fb19e3a767c5b24b3a8ec09b881ab6e4/media-libs/libass/libass-0.14.0.recipe#L74

rcombs:
Hmmm, looking at the relevant code in ffmpeg, we've got this:
```
    if enabled x86_32; then
        # Clang doesn't support maintaining alignment without assuming the
        # same alignment in every function. If 16 byte alignment would be
        # enabled, one would also have to either add attribute_align_arg on
        # every single entry point into the libraries or enable -mstackrealign
        # (doing stack realignment in every single function).
        case $target_os in
        mingw32|win32|*bsd*)
            disable aligned_stack
            ;;
        *)
[...]

PR #456, opened 2020-11-21
x86: update x86inc

CoffeeFlux:
This should fix the warning introduced with nasm 2.15

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -196,7 +196,7 @@ AS_IF([test "x$enable_asm" != xno], [
             X64=true
             BITS=64
             BITTYPE=x32
-            ASFLAGS="$ASFLAGS -DARCH_X86_64=1 -DPIC"
+            ASFLAGS="$ASFLAGS -DARCH_X86_64=1"
         ],
         [x86_64-*|amd64-*], [
             AS=nasm
@@ -204,7 +204,7 @@ AS_IF([test "x$enable_asm" != xno], [
             X64=true
             BITS=64
             BITTYPE=64
-            ASFLAGS="$ASFLAGS -DARCH_X86_64=1 -DPIC"
+            ASFLAGS="$ASFLAGS -DARCH_X86_64=1"
         ],
         [ # default
             INTEL=false
@@ -219,25 +219,18 @@ AS_IF([test "x$enable_asm" != xno], [
         ], [
             AS_CASE([$host],
                 [*darwin*], [
-                    ASFLAGS="$ASFLAGS -f macho$BITTYPE -DPREFIX -DHAVE_ALIGNED_STACK=1"
+                    ASFLAGS="$ASFLAGS -f macho$BITTYPE -DPREFIX -DSTACK_ALIGNMENT=16"
                 ],
                 [*linux*|*solaris*|*haiku*], [
-                    ASFLAGS="$ASFLAGS -f elf$BITTYPE -DHAVE_ALIGNED_STACK=1"
+                    ASFLAGS="$ASFLAGS -f elf$BITTYPE -DSTACK_ALIGNMENT=16"
                 ],
                 [*dragonfly*|*bsd*], [
                     ASFLAGS="$ASFLAGS -f elf$BITTYPE"
-                    AS_IF([test "x$BITS" = x64], [
-                        ASFLAGS="$ASFLAGS -DHAVE_ALIGNED_STACK=1"
-                    ], [
-                        ASFLAGS="$ASFLAGS -DHAVE_ALIGNED_STACK=0"
-                    ])
                 ],
                 [*cygwin*|*mingw*], [
                     ASFLAGS="$ASFLAGS -f win$BITTYPE"
-                    AS_IF([test "x$BITS" = x64], [
-                        ASFLAGS="$ASFLAGS -DHAVE_ALIGNED_STACK=1"
-                    ], [
-                        ASFLAGS="$ASFLAGS -DHAVE_ALIGNED_STACK=0 -DPREFIX"
+                    AS_IF([test "x$BITS" = x32], [
+                        ASFLAGS="$ASFLAGS -DPREFIX"
                     ])
                 ],
                 [ # default
@@ -251,7 +244,7 @@ AS_IF([test "x$enable_asm" != xno], [
                     ))
                 ]
             )
-            ASFLAGS="$ASFLAGS -DHAVE_CPUNOP=0 -Dprivate_prefix=ass"
+            ASFLAGS="$ASFLAGS -Dprivate_prefix=ass"
             AC_MSG_CHECKING([if $AS supports vpmovzxwd])
             echo "vpmovzxwd ymm0, xmm0" > conftest.asm
             AS_IF([$AS conftest.asm $ASFLAGS -o conftest.o >conftest.log 2>&1], [
--- a/libass/x86/rasterizer.asm
+++ b/libass/x86/rasterizer.asm
@@ -488,7 +488,7 @@ endstruc
 ;------------------------------------------------------------------------------
 
 %macro GET_RES_ADDR 1
-%if mmsize <= 16 && HAVE_ALIGNED_STACK
+%if mmsize <= STACK_ALIGNMENT
     mov %1, rstk
 %else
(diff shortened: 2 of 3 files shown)


TheOneric:
It seems like we'll also need to update our `ASFLAGS` along with this update; looking at the macros we set:
 - `HAVE_ALIGNED_STACK` was removed, instead `STACK_ALIGNMENT` can be used where necessary I guess
 - `PIC` is now handled differently *(can we just always set it now that it is force-disabled for win32?)*
- `HAVE_CPUNOP` was removed

CoffeeFlux:
`HAVE_ALIGNED_STACK` checks have been replaced with `STACK_ALIGNMENT == 16`.

`PIC` is no longer defined in the configure scripts, since it seems like the existing behavior will just be preserved with the new x86inc internal logic. If we want to always set it I can do that instead, but I don't have context on how this impacts 32-bit linux so it's not my call to make.

`HAVE_CUPNOP` is no longer defined.

TheOneric:
> PIC is no longer defined in the configure scripts, since it seems like the existing behavior will just be preserved with the new x86inc internal logic. If we want to always set it I can do that instead, but I don't have context on how this impacts 32-bit linux

Afaik it should be fine on all supported i386 Linux kernels and to my knowledge non-PIC code being included can break ASLR and potentially cause issues in shared libs. Not being PIC was also the reason ASM got [disabled in Debian's i386-builds of libass](https://salsa.debian.org/multimedia-team/libass/-/commit/d9eae1f0aefacf9b8c838cb5108dcc100f336e0b). 

However I'm not completely sure that always enabling PIC won't cause problems on some other non-Windows system and I realised it is possible for packagers to pass `ASFLAGS=-DPIC` to `configure` to enable it if it is preferred.
Given that this better matches the current behaviour, it's probably safer to leave the decision about PIC on i386 to x86inc's defaults and user `ASFLAGS` instead of always enabling it, as I previously hinted at. So I think this should be fine.

---

By changing `configure.ac` this pr now conflicts with #452. As #452 basically touches almost _everything_ in `configure.ac`, I'd prefer to merge #452 _before_ this one, as merging it the other way around would result in all of #452's commits needing to be adjusted, meaning new errors could slip in and I'd have to reverify it on all 6 systems.
Merging #452 first on the other hand only requires a few lines of this patch to be adjusted.

configure.ac
```diff
@@ -207,15 +207,13 @@ AS_IF([test x$enable_asm != xno], [
         ], [
             AS_CASE([$host],
                 [*darwin*], [
-                    ASFLAGS="$ASFLAGS -f macho$BITTYPE -DPREFIX -DHAVE_ALIGNED_STACK=1" ],
+                    ASFLAGS="$ASFLAGS -f macho$BITTYPE -DPREFIX" ],
                 [*linux*|*dragonfly*|*bsd*|*solaris*|*haiku*], [
-                    ASFLAGS="$ASFLAGS -f elf$BITTYPE -DHAVE_ALIGNED_STACK=1" ],
+                    ASFLAGS="$ASFLAGS -f elf$BITTYPE" ],
```
TheOneric, line 212: Won't this change stack alignment on 32-bit MacOS, Linux, Solaris and Haiku? Afaict x86inc defaults to what would previously have been `HAVE_ALIGNED_STACK=0` on i386-systems.
 *(Dragonfly and other BSDs shouldn't have 16Byte alignment on 32-bit anyway as demonstrated by #451 and fixed as part of #452)*

rcombs:
The reason lots of code doesn't support PIC on i386 (including ffmpeg, which debian seems to have no issue continuing to package with assembly enabled) is that there's no IP-relative addressing mode, so doing data loads ends up being really complex and expensive (it requires use of at least one extra register, and i386 is already pretty starved for those). I don't see much reason to try to port stuff to PIC-i386 at this point tbh; it's not like there are a whole lot of people clamoring for it when just about everybody not on windows has moved to x86_64.

CoffeeFlux:
I'm sympathetic to your reasons for merging #452 first, particularly given I've had the meson PR sitting on the vine for a couple years now, but that's a rather large PR making mostly cosmetic changes and this one fixes an immediate issue. But well, not my call either way.

Will check on the stack alignment.

libass/x86/rasterizer.asm
```diff
@@ -488,7 +488,7 @@ endstruc
 ;------------------------------------------------------------------------------
 
 %macro GET_RES_ADDR 1
-%if mmsize <= 16 && HAVE_ALIGNED_STACK
+%if mmsize <= 16 && STACK_ALIGNMENT == 16
```
MrSmile, line 491: To correctly support higher stack alignments that should be
```asm
%if mmsize <= STACK_ALIGNMENT
```

CoffeeFlux: Right, sorry. Fixed.

rcombs (approved):

TheOneric:
With #452 merged, and relying on `x86inc.asm` alignment defaults, I think the configure.ac-part should now look like this:
```diff
diff --git a/configure.ac b/configure.ac
index 2e5cae6..6cb4948 100644
--- a/configure.ac
+++ b/configure.ac
@@ -196,7 +196,7 @@ AS_IF([test "x$enable_asm" != xno], [
             X64=true
             BITS=64
             BITTYPE=x32
-            ASFLAGS="$ASFLAGS -DARCH_X86_64=1 -DPIC"
+            ASFLAGS="$ASFLAGS -DARCH_X86_64=1"
         ],
         [x86_64-*|amd64-*], [
             AS=nasm
@@ -204,7 +204,7 @@ AS_IF([test "x$enable_asm" != xno], [
             X64=true
             BITS=64
             BITTYPE=64
-            ASFLAGS="$ASFLAGS -DARCH_X86_64=1 -DPIC"
+            ASFLAGS="$ASFLAGS -DARCH_X86_64=1"
         ],
         [ # default
             INTEL=false
@@ -219,25 +219,18 @@ AS_IF([test "x$enable_asm" != xno], [
         ], [
             AS_CASE([$host],
                 [*darwin*], [
-                    ASFLAGS="$ASFLAGS -f macho$BITTYPE -DPREFIX -DHAVE_ALIGNED_STACK=1"
+                    ASFLAGS="$ASFLAGS -f macho$BITTYPE -DPREFIX -DSTACK_ALIGNMENT=16"
                 ],
                 [*linux*|*solaris*|*haiku*], [
-                    ASFLAGS="$ASFLAGS -f elf$BITTYPE -DHAVE_ALIGNED_STACK=1"
+                    ASFLAGS="$ASFLAGS -f elf$BITTYPE -DSTACK_ALIGNMENT=16"
                 ],
                 [*dragonfly*|*bsd*], [
                     ASFLAGS="$ASFLAGS -f elf$BITTYPE"
-                    AS_IF([test "x$BITS" = x64], [
-                        ASFLAGS="$ASFLAGS -DHAVE_ALIGNED_STACK=1"
-                    ], [
-                        ASFLAGS="$ASFLAGS -DHAVE_ALIGNED_STACK=0"
-                    ])
                 ],
                 [*cygwin*|*mingw*], [
                     ASFLAGS="$ASFLAGS -f win$BITTYPE"
-                    AS_IF([test "x$BITS" = x64], [
-                        ASFLAGS="$ASFLAGS -DHAVE_ALIGNED_STACK=1"
-                    ], [
-                        ASFLAGS="$ASFLAGS -DHAVE_ALIGNED_STACK=0 -DPREFIX"
+                    AS_IF([test "x$BITS" = x32], [
+                        ASFLAGS="$ASFLAGS -DPREFIX"
                     ])
                 ],
                 [ # default
@@ -251,7 +244,7 @@ AS_IF([test "x$enable_asm" != xno], [
                     ))
                 ]
             )
-            ASFLAGS="$ASFLAGS -DHAVE_CPUNOP=0 -Dprivate_prefix=ass"
+            ASFLAGS="$ASFLAGS -Dprivate_prefix=ass"
             AC_MSG_CHECKING([if $AS supports vpmovzxwd])
             echo "vpmovzxwd ymm0, xmm0" > conftest.asm
             AS_IF([$AS conftest.asm $ASFLAGS -o conftest.o >conftest.log 2>&1], [

```

TheOneric (approved):

PR #458, opened 2020-12-06
Use GitHubActions

TheOneric:
`travis-ci.org` will [close on 31.12.2020](https://docs.travis-ci.com/user/migrate/open-source-repository-migration#q-when-will-the-migration-from-travis-ciorg-to-travis-cicom-be-completed) and `travis-ci.com` [does no longer offer MacOS builds to free users](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing).
Since the project is currently hosted on GitHub using GHA as a travis replacement seemed like the least amount of work.

Compared to our travis builds GHA is currently missing IRC notifications and Coverity.
To my knowledge there's unfortunately no easy way to get IRC notifications with GHA.
I don't think it's possible for me to migrate Coverity to GHA as I don't know about the tokens. @astiob, iimn your familiar with the Coverity stuff, could you take a look at if Coverity can be integrated into the GHA builds? GHA's docs on secret values can be found here: https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets

An example of it in work can be found [here](https://github.com/TheOneric/libass/actions/runs/403298501). *(Note that GitHub only shows the full log if logged in, which is a bit annoying. I'm not sure if all looged in users can see the logs or only those with some form of repo access)*

### DIFF
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -20,6 +20,10 @@ jobs:
           - os: ubuntu-18.04
             cc: gcc
             do_distc: yes
+          # Run Coverity on a clang build; Coverity's gcc causes issues
+          - os: ubuntu-18.04
+            cc: clang
+            do_coverity: yes
 
     steps:
       - name: checkout code
@@ -56,3 +60,87 @@ jobs:
 
       - name: distcheck
         run: if [ "x${{ matrix.do_distc }}" = "xyes" ] ; then make -j 2 distcheck; fi
+
+      - name: Coverity scan
+        env:
+          COVERITY_SCAN_TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
+          PROJECT_NAME: libass/libass
+          NOTIFY_EMAIL: none@example.com
+          TOOL_URL: https://scan.coverity.com/download/
+          UPLOAD_URL: https://scan.coverity.com/builds?project=libass%2Flibass
+          SCAN_URL: https://scan.coverity.com
+          RES_DIR: cov-int
+        run: |
+          if [ "x${{ matrix.do_coverity }}" = "xyes" ] \
+             && [ "x${{ github.repository }}" = "xlibass/libass" ] \
+             && [ "x${{ github.event_name }}" != "xpull_request" ]
+          then
+            exit_code=0
+            echo "Running Coverity ..."
+            # Remove previous build output
+            make clean
+            # The upstream script is borked and always exits with 1 even on success
+            # To get meaningful success/error status we're using our own script
+            # but we still want to be informed about upstream script changes
+            if curl -s https://scan.coverity.com/scripts/travisci_build_coverity_scan.sh \
+                | shasum -a 256 \
+                | grep -Eq '^234d71b4a5257a79559e66dd3ba5765576d2af4845da83af4975b77b14ab536b '
+            then
+               : remote unchanged
+            else
+              echo "Coverity's travis script changed!"
+              exit_code=1
+            fi
+
+            # Check if we are within quoata
+            quota_res="$(curl -s --form project="$PROJECT_NAME" \
+                                 --form token="$COVERITY_SCAN_TOKEN" \
+                                 "$SCAN_URL"/api/upload_permitted)"
+            if [ "$?" -ne 0 ] || [ "x$quota_res" = "xAccess denied" ] ; then
+              echo "Coverity denied access or did not respond!"
+              exit 1
+            elif echo "$quota_res" | grep -Eq 'upload_permitted": *true' ; then
+              echo "Within Coverity quota."
+            else
+              echo "Exceeding Coverity quota! Try again later."
+              echo "$quota_res" | grep -Eo 'next_upload_permitted_at":[^,}]*'
+              exit 0
+            fi
+
+            # Download cov tool and make it available
+            wget -nv "$TOOL_URL""$(uname)" \
+                 --post-data "project=$PROJECT_NAME&token=$COVERITY_SCAN_TOKEN" \
+                 -O cov-analysis-tool.tar.gz
+            mkdir cov-analysis-tool
+            tar xzf cov-analysis-tool.tar.gz --strip 1 -C cov-analysis-tool
+            export PATH="$(pwd)/cov-analysis-tool/bin:$PATH"
+
+            # Coverity Build
+            echo "Starting Coverity build..."
+            #mkdir "$RES_DIR" # already done by cov-build
+            COVERITY_UNSUPPORTED=1 cov-build --dir "$RES_DIR" make -j 2
+            cov-import-scm --dir "$RES_DIR" --scm git --log "$RES_DIR/scm_log.txt" 2>&1
+
+            # Submit results to Coverity's server
+            tar czf libass.tar.gz "$RES_DIR"
+            upstat="$(curl --silent --write-out "\n%{http_code}\n" \
+                       --form project="PROJECT_NAME"               \
+                       --form token="$COVERITY_SCAN_TOKEN"         \
+                       --form email="$NOTIFY_EMAIL"                \
+                       --form file=@libass.tar.gz                  \
+                       --form version="${{ github.sha }}"          \
+                       --form description="GitHubActions CI build" \
+                        "$UPLOAD_URL")"
+            if [ "$?" -ne 0 ] ; then
+              echo "Upload failed (curl error)"
+              exit_code=1
+            elif echo "$upstat" | tail -n 1 | grep -Eq '^2[0-9]{2}$' ; then
+              echo "Upload successful."
+            else
+              echo "Upload failed (server error)"
+              exit_code=1
+            fi
+            echo "$upstat" | head
+
+            exit $exit_code
+          fi
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -1,0 +1,50 @@
+name: GitHub CI tests
+
+on:
+  push:
+    branches: [master, ci, coverity_scan]
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, macos-10.15]
+        cc: [gcc, clang]
+        exclude:
+          - os: macos-10.15
+            cc: gcc
+        include:
+          # Enable distcheck for one build
+          - os: ubuntu-18.04
+            cc: gcc
+            do_distc: yes
+
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v2
+
+      - name: install deps
+        run: |
+          if echo "${{ matrix.os }}" | grep -qE '^macos-' ; then
+            #brew update
+            # fontconfig, freetype, autoconf and libtool are preinstalled
+            # and `brew install` fails if a non-uptodate version is already installed
+            #brew upgrade fontconfig freetype autoconf libtool
+            brew install automake fribidi harfbuzz nasm
+          else
+            sudo apt-get update #&& sudo apt-get upgrade
+            sudo apt-get install -y \
+                 libfontconfig1-dev libfreetype6-dev libfribidi-dev \
+                 libharfbuzz-dev nasm ${{ matrix.cc }}
+          fi
+
+      - name: configure
+        run: ./autogen.sh && CC="${{ matrix.cc }}" ./configure
+
+      - name: compile
+        run: test "x${{ matrix.do_distc }}" = "xyes" || make -j 2
+
+      - name: distcheck
+        run: if [ "x${{ matrix.do_distc }}" = "xyes" ] ; then make -j 2 distcheck; fi
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -25,6 +25,14 @@ jobs:
       - name: checkout code
         uses: actions/checkout@v2
 
+      - name: retrieve caches
+        uses: actions/cache@v2
+        with:
+          path: |
+            $HOME/Library/Caches/Homebrew
+            /usr/local/Homebrew
+          key: ${{ runner.os }}-${{ matrix.os }}-build
+
       - name: install deps
         run: |
           if echo "${{ matrix.os }}" | grep -qE '^macos-' ; then


mia-0:
Re IRC notifications: GitHub Webhooks provide notifications for check statuses. I could add these to our IRC bot.

astiob:
Should we merge my Travis CI changes from the coverity_scan branch first? TBH I simply forgot about them.

Integrating Coverity Scan shouldn’t be hard, as the Coverity plugin for Travis CI just executes an external shell script, hosted not even by Travis but by Coverity.

By the way, I’m not too happy with that Coverity Scan shell script, as it makes it impossible to distinguish between build configuration/Coverity error and build failure during the build that is made for upload to Coverity Scan. Ideally, I’d like to treat the former as an error and the latter as a build failure, or treat the former as an error and ignore the latter; but we currently have to ignore both. But I’m also not eager to maintain our own modified copy of the shell script and potentially miss upstream changes (although I’m not sure they’ve ever changed that script until now).

By the way, did we figure out those Coverity Scan compilation errors in the end? The last thing I remember was myself musing that it seemed as if Coverity Scan was using an older GCC and lying about its version number (copying the system GCC’s version macros) when running the C preprocessor on our source code.

TheOneric:
> Re IRC notifications: GitHub Webhooks provide notifications for check statuses. I could add these to our IRC bot.

If it's not to bothersome, that would be nice.


> By the way, did we figure out those Coverity Scan compilation errors in the
> end? The last thing I remember was myself musing that it seemed as if
> Coverity Scan was using an older GCC and lying about its version number
> (copying the system GCC’s version macros) when running the C preprocessor on
> our source code.

We did a verbose test build, indicating such.
Coverity's clang compiler happens to already have all features we use *(eventhough it's likely also reporting a false version)*. So we can just run Coverity with clang instead of gcc.

https://travis-ci.org/github/libass/libass/builds/743655400

> Should we merge my Travis CI changes from the coverity_scan branch first?                                

iirc except for the Coverity change, those were all Travis specific. If we want to merge them anyway, Coverity should run on the clang build instead of gcc.

> Travis CI just executes an external shell script, hosted not even by Travis but by Coverity.

[The script from coverity_scan](https://scan.coverity.com/scripts/travisci_build_coverity_scan.sh) expects some Travis-specific env vars like `TRAVIS_BRANCH` and `TRAVIS_PULL_REQUEST`. Ideally Coverity would also offer a general script and/or GHA-specific script, but I guess we can also set all required vars manually.


> By the way, I’m not too happy with that Coverity Scan shell script, as it makes it impossible to distinguish between
 build configuration/Coverity error and build failure during the build that is made for upload to Coverity Scan.
Ideally, I’d like to treat the former as an error and the latter as a build failure, or treat the former as an error and
ignore the latter; but we currently have to ignore both. But I’m also not eager to maintain our own modified copy of the
 shell script and potentially miss upstream changes (although I’m not sure they’ve ever changed that script until now)

I'm not sure I can follow. Whats the benefit of ignoring Coverity build errors? If the quota was exceeded their script exits with `0`, otherwise I can only think of Coverity timing out during up/download again as something we might want to ignore.


I'd imagine Coverity integration might work something like the following with Coverity running *instead* of a regular compile, maybe also splitting regular compile and Coverity into separate steps to make the jobs easier to distinguish on failure. Can't test it though.
```diff
diff --git a/.github/workflows/ghci.yml b/.github/workflows/ghci.yml
index 12a7819..dd3a0f6 100644
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -20,6 +20,10 @@ jobs:
           - os: ubuntu-18.04
             cc: gcc
             do_distc: yes
+          # Do one Coverity build with clang; Coverity's gcc causes issues
+          - os: ubuntu-18.04
+            cc: clang
+            do_coverity: [no, yes]
 
     steps:
       - name: checkout code
@@ -52,7 +56,19 @@ jobs:
         run: ./autogen.sh && CC="${{ matrix.cc }}" ./configure
 
       - name: compile
-        run: make -j 2
+        env:
+          COVERITY_SCAN_TOKEN="${{ secrets.cov_token }}"
+        run: |
+          if [ "x${{ matrix.do_coverity }}" = "xyes" ] ; then
+            echo "This is a Coverity build!"
+            TRAVIS_BRANCH=… \
+            COVERITY_SCAN_BRANCH_PATTERN=… \
+              curl -s https://scan.coverity.com/scripts/travisci_build_coverity_scan.sh | bash
+          else
+            make -j 2
+          fi
+
+      - name: Coverity scan
 
       - name: distcheck
         run: if [ "x${{ matrix.do_distc }}" == "xyes" ] ; then make -j 2 distcheck; fi
```

jtanx:
Here's a [working coverity scan workflow](https://github.com/fontforge/fontforge/blob/master/.github/workflows/coverity.yml), although it only runs when manually triggered. Only thing you really need is the coverity scan token as a secret.

TheOneric:
So, Coverity's script apparently **always** exits with 1, even if the build was successful. Going by other complaints about this on the web this has been like this for years. (see: https://travis-ci.org/github/libass/libass/jobs/748250419#L439 )

So our options are, either emulate a Travis env and continue to ignore Coverity's exit code, or use a custom script for useful error codes. Here are commits for both:
Travis env: https://github.com/TheOneric/libass/commit/1bbbd207cfaff2bc05938b5929bad5e06b2d47b9
Custom script: https://github.com/libass/libass/pull/458/commits/fc342a09d36a9d802262ab283adafb4654e69aa0

If someone who knows the Coverity token adds it as an repo or organisation secret ([docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets)), I can test and finish the Coverity integration for whichever version is preferred.

astiob:
I think I’ve added the token. The name is `COVERITY_SCAN_TOKEN`, which I think is what your code expects.

astiob:
On that note, there’s a whole bunch of alternative static analysers GitHub offers to integrate with on https://github.com/libass/libass/security/code-scanning (I’m not sure who has access to this page). ~~Shall we enable all of them?~~

TheOneric:
> I’ve added the token to libass/libass

Thanks, both versions should now work.

Custom script:
Commit: https://github.com/libass/libass/pull/458/commits/4cdc10d62d6cee3582bb43e1be2ddf0246f78182
Build log : https://github.com/libass/libass/runs/1533823674?check_suite_focus=true
*(log is missing the `cov-import-scm`-call, which is now added)*

Travis script:
Commit: https://github.com/TheOneric/libass/commit/dd489377fbd3e268105c185e0b2196ea5e291dab
Build log: https://github.com/libass/libass/runs/1533914073?check_suite_focus=true

TheOneric:
It seems to me like the remaining SCM warning present in both our TRavis and this GHA builds, can be safely ignored. Here's the detailed log: https://github.com/libass/libass/runs/1551570451?check_suite_focus=true#step:8:159 .

Given that it seems to work well and report errors better than the Travis script, I'd be in favour of using the custom Coverity script.

Regarding other static analysers, if they are available to us, checking them out and adding one or a few does seems like a good idea for the future.

TheOneric:
Fixed some typos and changed Coverity script to also inform about next available build on exceeding quota:
```diff
@@ -103,7 +103,8 @@ jobs:
               echo "Within Coverity quota."
             else
               echo "Exceeding Coverity quota! Try again later."
-              exit 0;
+              echo "$quota_res" | grep -Eo 'next_upload_permitted_at":[^,}]*'
+              exit 0
             fi
```

With travis-ci-org closing in 3 days, unless changes are requested, I'll merge this pr tomorrow.

.github/workflows/ghci.yml
```diff
@@ -0,0 +1,146 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, macos-10.15]
+        cc: [gcc, clang]
+        exclude:
+          - os: macos-10.15
+            cc: gcc
+        include:
+          # Enable distcheck for one build
```
astiob, line 19: 👍

.github/workflows/ghci.yml
```diff
@@ -0,0 +1,146 @@
...
+            make clean
+            # The upstream script is borked and always exits with 1 even on success
+            # To get meaningful success/error status we're using our own script
+            # but we still want to be informed about upstream script changes
+            if curl -s https://scan.coverity.com/scripts/travisci_build_coverity_scan.sh \
+                | shasum -a 256 \
+                | grep -Eq '^234d71b4a5257a79559e66dd3ba5765576d2af4845da83af4975b77b14ab536b '
+            then
+               echo "" > /dev/null
+            else
```
astiob, line 90: Why not invert the condition with `if ! curl...` and drop the successful branch altogether?

.github/workflows/ghci.yml
```diff
@@ -0,0 +1,146 @@
...
+            sudo apt-get install -y \
+                 libfontconfig1-dev libfreetype6-dev libfribidi-dev \
+                 libharfbuzz-dev nasm ${{ matrix.cc }}
+          fi
+
+      - name: configure
+        run: ./autogen.sh && CC="${{ matrix.cc }}" ./configure
+
+      - name: compile
+        run: make -j 2
```
astiob, line 59: Does `make distcheck` require the regular compile step to be run beforehand? Could we save time (without sacrificing completeness or output clarity) by skipping this when distcheck is enabled?

Similarly, if we’re running Coverity Scan, it will recompile everything, so we could save time by skipping this step. However, we must be careful to fall back to plain compile if the Coverity Scan quota is exceeded.

.github/workflows/ghci.yml
```diff
@@ -0,0 +1,146 @@
...
+        uses: actions/cache@v2
+        with:
+          path: |
+            $HOME/Library/Caches/Homebrew
+            /usr/local/Homebrew
+          key: ${{ runner.os }}-${{ matrix.os }}-build
+
+      - name: install deps
+        run: |
+          if (echo "${{ matrix.os }}" | grep -qE '^macos-') ; then
```
astiob, line 42: Nit: the parentheses aren’t necessary here.

.github/workflows/ghci.yml
```diff
@@ -0,0 +1,146 @@
...
+            #brew update
+            # freetype, autoconf and libtool are preinstalled
+            # and `brew install` fails if a non-uptodate version is already installed
+            #brew upgrade freetype autoconf libtool
+            brew install automake fribidi harfbuzz nasm
+          else
+            #sudo apt-get update && sudo apt-get upgrade
```
astiob, line 49: Are these commented `update` and `upgrade` invocations kept in case we at some point require a dependency newer than GitHub’s default version?

.github/workflows/ghci.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v2
+
+      - name: install deps
+        run: |
+          if (echo "${{ matrix.os }}" | grep -qE '^macos-') ; then
+            #brew update
+            # freetype, autoconf and libtool are preinstalled
+            # and `brew install` fails if a non-uptodate version is already installed
```
astiob, line 45: Do you know if Fontconfig is also preinstalled by any chance? It is in Travis CI (but we can neither explicitly `brew install fontconfig` as it complains about the version being out of date nor `brew upgrade fontconfig` as it’s slow).

astiob (commented):
Nit: the last commit’s message has a typo where “Coverity” is “Coverit”. Also, it may be better to mention https://travis-ci.org/github/libass/libass/builds/740511824 instead of or in addition to 743655400, as it actually shows the Clang build working.

astiob:
For reference, travis-ci.org currently says:

> Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information.

which suggests it may not be shutting quite this week. Of course, this doesn’t mean we shouldn’t move away, but perhaps we aren’t in as much rush.

---

Presumably, we’ll want to drop the Travis CI configuration file as well.

Tying into your question on IRC:

> [do] you want to merge your Travis changes on coverity_scan before [merging this PR]?

I’m now thinking that if we want to remove the Travis CI config with this PR, we can ignore my coverity_scan branch changes, but if we want to keep it around for another while, we ought to merge them (but change the job number that runs Coverity Scan so that it corresponds to Clang instead of GCC).

.github/workflows/ghci.yml
TheOneric: While `test ! …` is part of [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html) and [considered portable](https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Limitations-of-Builtins.html#test), `if !` apparently [is not](https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Limitations-of-Builtins.html#if). 
On the other hand, in this case we know the code runs in `bash` or `dash`, both of which support `if !`.

.github/workflows/ghci.yml
TheOneric: > Re: distcheck

It does not require a regular compile beforehand; currently we _could_ skip make if distcheck is enabled. However when we add regression-tests in the future, those will require a regular compile beforehand.

> Re: Coverity

As we already noticed, Coverity uses custom compiler builds which might change the build success. Imho it's better to also run a build with "real" clang.

We could also run coverity-only and  distcheck-only builds in parallel to regular builds. However, with the start, dependency-install and configure overhead, I'm not sure if that would actually be faster.

.github/workflows/ghci.yml
TheOneric: Fontconfig is also [preinstalled](https://github.com/TheOneric/libass/runs/1623767314?check_suite_focus=true#step:4:16).

.github/workflows/ghci.yml
TheOneric: Yes. In case of `brew` it also lists the used preinstalled dependencies which will be handy should we actually need update & upgrade. The preinstalled-distinction is not necessary with `apt`, its comment is just there to match the `brew` clause.

.github/workflows/ghci.yml
TheOneric: thx, removed them.

TheOneric:
Fixed typos and added the reference to a working Coverity-clang build in last commit message.
Normal compile is now not run if distcheck is enabled; `fontconfig` is now also listed as preinstalled on macOS and some other minor changes.

---

> For reference, travis-ci.org currently says: […] which suggests it may not be shutting quite this week. 

Their [migration docs](https://docs.travis-ci.com/user/migrate/open-source-repository-migration#q-when-will-the-migration-from-travis-ciorg-to-travis-cicom-be-completed) still refer to the deadline as “no later than December 31st, 2020”. Perhaps they extended the deadline a bit.

---

>  if we want to keep [the Travis config] around for another while, we ought to merge [my coverity_scan changes]
>  (but change the job number that runs Coverity Scan so that it corresponds to Clang instead of GCC).

I already changed it to the clang build when I rebased `coverity_scan` 22 days ago:
> \<Oneric\> ^ rebased and moved Coverity to linux-clang instead of linux-gcc to workaround Coverity's gcc-version issue

If we get assembly on more architectures, adding Travis again _might_ be interesting as they [offer more cpu architectures](https://docs.travis-ci.com/user/multi-cpu-architectures/) beside amd64 (arm64, POWER and IBM Z). *(And afaik they are still available to OSS projects without payment)*.
On the other hand there is a [community action](https://github.com/marketplace/actions/run-on-architecture) we can use to relatively easily emulate those same architectures via qemu in GHA and I don't know if Travis' non-amd64 builds are native or also emulated.

.github/workflows/ghci.yml
TheOneric: Apparently GitHub does not keep their images up-to-date enough to allow us to skip `apt update`. (see: [this build failure](https://github.com/wangqr/Aegisub/runs/1665185971?check_suite_focus=true#step:3:447))

I guess this means we should also be calling`brew update`, but then *(MacOS builöds are significantly slower and)* [this](https://github.com/TheOneric/Aegisub/runs/1674614692?check_suite_focus=true#step:3:4318) brew-link-error appears again...
For now I've enabled `update` for `apt` only.

TheOneric: Other projects like [TS-Aegisub](https://github.com/TypesettingTools/Aegisub/blob/417ba4f/.github/workflows/ci.yml#L83-L88) and  [hugo](https://github.com/gohugoio/hugo/blob/master/.github/workflows/test.yml#L41-L43) omit `brew update` in their GHA-scripts as well; so I guess it's probably fine.

astiob (approved):

PR #466, opened 2020-12-13
Reduce strdup() usage

MrSmile:
I've removed most of strdups from rendering. Error handling should be simpler too.
There is a bunch of strdups in `ass.c` but that's impossible to remove without API change.

There is some concern that low-level parsing functions `mystrto*()` do not respect new string boundaries, but in our case we always have appropriate delimiter after string reference (`argto*()` rely on that too).

### DIFF
--- a/libass/ass_cache.c
+++ b/libass/ass_cache.c
@@ -40,12 +40,13 @@
 // font cache
 static bool font_key_move(void *dst, void *src)
 {
-    ASS_FontDesc *k = src;
-    if (dst)
-        memcpy(dst, src, sizeof(ASS_FontDesc));
-    else
-        free(k->family);
-    return true;
+    if (!dst)
+        return true;
+
+    ASS_FontDesc *d = dst, *s = dst;
+    memcpy(dst, src, sizeof(ASS_FontDesc));
+    d->family.str = ass_copy_string(s->family);
+    return d->family.str;
 }
 
 static void font_destruct(void *key, void *value)
@@ -207,8 +208,8 @@ static bool outline_key_move(void *dst, void *src)
     }
     memcpy(dst, src, sizeof(OutlineHashKey));
     if (s->type == OUTLINE_DRAWING) {
-        d->u.drawing.text = strdup(s->u.drawing.text);
-        return d->u.drawing.text;
+        d->u.drawing.text.str = ass_copy_string(s->u.drawing.text);
+        return d->u.drawing.text.str;
     }
     if (s->type == OUTLINE_BORDER)
         ass_cache_inc_ref(s->u.border.outline);
@@ -226,7 +227,7 @@ static void outline_destruct(void *key, void *value)
         ass_cache_dec_ref(k->u.glyph.font);
         break;
     case OUTLINE_DRAWING:
-        free(k->u.drawing.text);
+        free((char *) k->u.drawing.text.str);
         break;
     case OUTLINE_BORDER:
         ass_cache_dec_ref(k->u.border.outline);
--- a/libass/ass_cache.c
+++ b/libass/ass_cache.c
@@ -38,31 +38,6 @@
 #include "ass_cache_template.h"
 
 // font cache
-static uint32_t font_hash(void *buf, uint32_t hval)
-{
-    ASS_FontDesc *desc = buf;
-    hval = fnv_32a_str(desc->family, hval);
-    hval = fnv_32a_buf(&desc->bold, sizeof(desc->bold), hval);
-    hval = fnv_32a_buf(&desc->italic, sizeof(desc->italic), hval);
-    hval = fnv_32a_buf(&desc->vertical, sizeof(desc->vertical), hval);
-    return hval;
-}
-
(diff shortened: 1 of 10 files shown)


libass/ass_utils.h
```diff
@@ -52,6 +52,26 @@ int has_avx(void);
 int has_avx2(void);
 #endif
 
+typedef struct {
+    const char *str;
+    size_t len;
+} ASS_StringView;
```
TheOneric, line 58: Several places in the code rely on StringView's `str` to be zero terminated. From the struct alone it may seem like StringView is a fat char pointer — without the need for zero termination. I'd prefer to have a short comment here documenting what StringView is intended to be, iiuc something like this would work:
```C
/* References a substring of some valid C string (0-terminated) */
```

TheOneric (commented):
Apart from length info leaks, seems good to me.

As you already mentioned, StringView's `str` is passed along as a regular C-string to `mystrto*` and so does `argto` with its args. As `mystro*` is afaict only called via `argto`, which knows the length, and inside `ass_drawing.c: drawing_tokenize`, which also knows the length via StringView, those instances of length info being lost should be straightforward to fix by adding a length parameter to `mystro*`.

Additionally, at least in `ass_fontselect.c: ass_font_select` StringView's `str` is also passed along to several other functions without the length info.
For font strings coming from a Style this won't be a problem since they are zero-terminated right after the font's name.
For font strings originating from a `\fn` string it is not obvious to me why it works, though practical tests suggest that it does in fact work. Those fontnames are substringed from the whole Event text, which usually does not end after the fontname. Afaict it's not until the fontname goes through `font_key_move` *(does it always do this?)*, that a zero terminator is added. In `update_font` it's added to the desc, without any zero-termination.

I'd prefer to have at least a comment in `ass_font_select` explaining why it is safe to discard length info here.

MrSmile:
> should be straightforward to fix by adding a length parameter to `mystro*`.

It's not so simple as that requires rewrite of `ass_strtod()` and `strtoll()`. Due to that I decided to defer that change to the future PR.

> Additionally, at least in `ass_fontselect.c`: `ass_font_select` StringView's str is also passed along to several other functions without the length info.

It's intended (I should add corresponding comment there). `ass_font_select()` works with `ASS_Font` copy of cache item key `ASS_FontDesc`. And cache item key is always zero-terminated. Ideally, a cache should have two different key types for request and storage, but that is quite a complication too.

rcombs (approved):

MrSmile:
Updated and rebased version. I've reverted changes to `ass_drawing.*` as it's another case where `ASS_StringView` is always zero-terminated, old code works just fine.

TheOneric (approved):

libass/ass_cache.c
```diff
@@ -40,12 +40,13 @@
...
-    ASS_FontDesc *k = src;
-    if (dst)
-        memcpy(dst, src, sizeof(ASS_FontDesc));
-    else
-        free(k->family);
-    return true;
+    if (!dst)
+        return true;
+
+    ASS_FontDesc *d = dst, *s = dst;
```
astiob, line 46: I’m confused by these two variables.

libass/ass_cache.c
```diff
@@ -40,12 +40,13 @@
...
+    memcpy(dst, src, sizeof(ASS_FontDesc));
```
astiob, line 47: If `s` were to equal `src`, then this could be simplified to `*d = *s;`

libass/ass_cache_template.h
```diff
@@ -50,6 +50,13 @@
 
 
 
+START(font, ass_font_desc )
+    STRING(family)
```
astiob, line 54: `STRING` references `fnv_32a_str`, which you’ve removed.

astiob: No, wait, opposite commit order? GitHub had me.

libass/ass_cache_template.h
```diff
@@ -50,6 +50,13 @@
...
+    GENERIC(unsigned, bold)
+    GENERIC(unsigned, italic)
+    GENERIC(int, vertical)  // @font vertical layout
```
astiob, line 57: (could have made this a `bool` while you were at it)

libass/ass_cache.c
MrSmile: Yes, there is a typo here but it's magically not leading to any error.

libass/ass_cache.c
MrSmile: I were thinking about that too, but there is a bunch of `memcpy()` in the same context so I decided to use it. Maybe it's indeed better to replace all such `memcpy()` with assignment as that is less error-prone.

libass/ass_cache.c
astiob: Yep 😃 The code works equally well as it is now, and with `*s = src`, and with just a single variable. But either of the last two options would have been easier to read.

astiob:
Looks good otherwise, and the issues I was too late commenting on are only cosmetic.

Thanks!

PR #469, opened 2020-12-29
Faster format line processing

TheOneric:
This is an alternative to #35 .
Unlike #35 however this approach does not special case any particular Format line content and order and applies to all parsing entries, not just ass_process_chunk. This is done by tokenising the format line when it's first encountered. Notably this also removes strdup as a failure source for format lines specifically, ASS_track's format strings are still strduped, but they are not used anywhere internally and are only retained for ABI compatibility.

**Performance:**
I'm using callgrind's cost estimate for performance measuring.

In general the (inclusive) cost reduction of only `process_text` was usually around 20-25% and for `process_event_tail` around 25-30%.  
The _total_ cost reduction however is lower and varies much more depending on the file. It never causes performance regressions.

For basic dialogue the gains are utterly irrelevant, while if there are many events the impact is stronger — but if those events are costly enough the relative gains stay low. Here are some benchmarks: `event(s)PerFrame` files are synthetic and use very cheap events, while all others are taken from real-world files.

```
Old data removed to avoid confusion (Still available in edit history)
Below analysis still refers to old data samples though.
```

[Synthetic test files](https://github.com/libass/libass/files/6039033/469_syntest.zip)

`dialogue` and `1eventPerFrame` are examples of few (but cheap) events per Frame.

`fxkara_1` and `mpv-7712` have many events in total but the cost of rendering the events is more outweighs this.

The synthetic tests as well as the `zoinks-neptunia` sample, do have many cheap enough events for the gains to become noticeable. In `zoink-neptunia`'s case a frame-by-frame animated logo consisting of around 1000 ASS-drawings (each in its own Dialogue-event) is responsible for this.

`YuruCamp2_02_rock` also uses animated multiple ASS-Drawings, but unlike `zoinks-neptunia` it's not animated frame by frame but split into section which are each animated via tags *(epf switches between ~80 and ~700 sections)*.  Also many of them use much more drawing commands then the ones from `zoink-neptunia` making them more expensive individually. If we limit the sample to a ~700epf section, the relative overall gain rises to 1.2%.

~~Apart from performance, in it's current form, this also removes `strdup` as a failure source for format lines for valid ASS files.~~

**Note:** Sections about API-change and RFC parts removed (see history)

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -39,7 +39,7 @@
 #include "ass_shaper.h"
 #include "ass_string.h"
 
-#define ass_atof(STR) (ass_strtod((STR),NULL))
+#define ass_atof(STR) (ass_strtod((STR), NULL))
 
 static const ASS_FormatToken ass_style_format[] = {
     ASS_FMT_Name, ASS_FMT_FontName, ASS_FMT_FontSize, ASS_FMT_PrimaryColour,
@@ -262,7 +262,7 @@ static long long string2timecode(ASS_Library *library, char *p)
     return tm;
 }
 
-#define NEXT(str,token) \
+#define NEXT(str, token) \
     token = next_token(&str); \
     if (!token) break;
 
@@ -283,7 +283,7 @@ static long long string2timecode(ASS_Library *library, char *p)
         target->name = strdup(token); \
         break;
 
-#define TIMEVAL_ALTNAME(fieldname,formatname) \
+#define TIMEVAL_ALTNAME(fieldname, formatname) \
     case ASS_FMT_ ## formatname : \
         target->fieldname = string2timecode(track->library, token); \
         break;
@@ -293,14 +293,14 @@ static long long string2timecode(ASS_Library *library, char *p)
         target->name = lookup_style(track, token); \
         break;
 
-#define ANYVAL(name,func) \
+#define ANYVAL(name, func) \
     case ASS_FMT_ ## name : \
         target->name = func(token); \
         break;
 
-#define COLORVAL(name) ANYVAL(name,parse_color_header)
-#define INTVAL(name)   ANYVAL(name,atoi)
-#define FPVAL(name)    ANYVAL(name,ass_atof)
+#define COLORVAL(name) ANYVAL(name, parse_color_header)
+#define INTVAL(name)   ANYVAL(name, atoi)
+#define FPVAL(name)    ANYVAL(name, ass_atof)
 
 // skip spaces in str beforehand, or trim leading spaces afterwards
 static inline void advance_token_pos(const char **const str,
@@ -432,8 +432,8 @@ static int process_event_tail(ASS_Track *track, ASS_Event *event,
             INTVAL(MarginL)
             INTVAL(MarginR)
             INTVAL(MarginV)
-            TIMEVAL_ALTNAME(Start,Start)
-            TIMEVAL_ALTNAME(Duration,End)
+            TIMEVAL_ALTNAME(Start, Start)
+            TIMEVAL_ALTNAME(Duration, End)
             default:
                 //Nothing
                 break;
@@ -466,7 +466,7 @@ static int process_event_tail(ASS_Track *track, ASS_Event *event,
 #define PARSE_START_STR if (0) {
 #define PARSE_END_STR   }
 
-#define ANYVAL_STR(name,func) \
+#define ANYVAL_STR(name, func) \
     } else if (ass_strcasecmp(tname, #name) == 0) { \
         target->name = func(token);
 
@@ -478,9 +478,9 @@ static int process_event_tail(ASS_Track *track, ASS_Event *event,
             target->name = new_str; \
         }
 
-#define COLORVAL_STR(name) ANYVAL_STR(name,parse_color_header)
-#define INTVAL_STR(name) ANYVAL_STR(name,atoi)
-#define FPVAL_STR(name) ANYVAL_STR(name,ass_atof)
+#define COLORVAL_STR(name) ANYVAL_STR(name, parse_color_header)
+#define INTVAL_STR(name) ANYVAL_STR(name, atoi)
+#define FPVAL_STR(name) ANYVAL_STR(name, ass_atof)
 
 
 /**
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -267,55 +267,6 @@ static long long string2timecode(ASS_Library *library, char *p)
     if (!token) break;
 
 
-#define ALIAS(alias,name) \
-        if (ass_strcasecmp(tname, #alias) == 0) {tname = #name;}
-
-/* One section started with PARSE_START and PARSE_END parses a single token
- * (contained in the variable named token) for the header indicated by the
- * variable tname. It does so by chaining a number of else-if statements, each
- * of which checks if the tname variable indicates that this header should be
- * parsed. The first parameter of the macro gives the name of the header.
- *
- * The string that is passed is in str. str is advanced to the next token if
- * a header could be parsed. The parsed results are stored in the variable
- * target, which has the type ASS_Style* or ASS_Event*.
- */
-#define PARSE_START if (0) {
-#define PARSE_END   }
-
-#define ANYVAL(name,func) \
-    } else if (ass_strcasecmp(tname, #name) == 0) { \
-        target->name = func(token);
-
-#define STRVAL(name) \
-    } else if (ass_strcasecmp(tname, #name) == 0) { \
-        char *new_str = strdup(token); \
-        if (new_str) { \
-            free(target->name); \
-            target->name = new_str; \
-        }
-
-#define STARREDSTRVAL(name) \
-    } else if (ass_strcasecmp(tname, #name) == 0) { \
-        while (*token == '*') ++token; \
-        char *new_str = strdup(token); \
-        if (new_str) { \
-            free(target->name); \
-            target->name = new_str; \
-        }
-
-#define COLORVAL(name) ANYVAL(name,parse_color_header)
-#define INTVAL(name) ANYVAL(name,atoi)
-#define FPVAL(name) ANYVAL(name,ass_atof)
-#define TIMEVAL(name) \
-    } else if (ass_strcasecmp(tname, #name) == 0) { \
-        target->name = string2timecode(track->library, token);
-
-#define STYLEVAL(name) \
-    } else if (ass_strcasecmp(tname, #name) == 0) { \
-        target->name = lookup_style(track, token);
-
-
 #define PARSE_START_TK switch (fmt->tokens[tokenpos]) {
 #define PARSE_END_TK   }
 
@@ -501,6 +452,37 @@ static int process_event_tail(ASS_Track *track, ASS_Event *event,
     return event->Text ? 0 : -1;
 }
 
+
+/* One section started with PARSE_START and PARSE_END parses a single token
+ * (contained in the variable named token) for the header indicated by the
+ * variable tname. It does so by chaining a number of else-if statements, each
+ * of which checks if the tname variable indicates that this header should be
+ * parsed. The first parameter of the macro gives the name of the header.
+ *
+ * The string that is passed is in str. str is advanced to the next token if
+ * a header could be parsed. The parsed results are stored in the variable
+ * target, which has the type ASS_Style* (or ASS_Event*).
+ */
+#define PARSE_START if (0) {
+#define PARSE_END   }
+
+#define ANYVAL(name,func) \
+    } else if (ass_strcasecmp(tname, #name) == 0) { \
+        target->name = func(token);
+
+#define STRVAL(name) \
+    } else if (ass_strcasecmp(tname, #name) == 0) { \
+        char *new_str = strdup(token); \
+        if (new_str) { \
+            free(target->name); \
+            target->name = new_str; \
+        }
+
+#define COLORVAL(name) ANYVAL(name,parse_color_header)
+#define INTVAL(name) ANYVAL(name,atoi)
+#define FPVAL(name) ANYVAL(name,ass_atof)
+
+
 /**
  * \brief Parse command line style overrides (--ass-force-style option)
  * \param track track to apply overrides to
@@ -584,6 +566,13 @@ void ass_process_force_style(ASS_Track *track)
             *dt = '.';
     }
 }
+
+#undef PARSE_START
+#undef PARSE_END
+#undef STRVAL
+#undef COLORVAL
+#undef FPVAL
+#undef INTVAL
 
 /**
  * \brief Parse the Style line
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -694,20 +694,6 @@ static int process_style(ASS_Track *track, char *str)
     return 0;
 }
 
-static bool format_line_compare(const ASS_FormatToken *fmt1, size_t num1,
-                                const ASS_FormatToken *fmt2, size_t num2)
-{
-    if (num1 != num2)
-        return false;
-    for (size_t i = 0; i < num1; ++i) {
-        if (*fmt1 != *fmt2)
-            return false;
-        ++fmt1;
-        ++fmt2;
-    }
-    return true;
-}
-
 
 /**
  * \brief Set SBAS=1 if not set explicitly in case of custom format line
@@ -728,8 +714,8 @@ static void custom_format_line_compatibility(ASS_Track *const track,
                                              const ASS_FormatToken *std,
                                              size_t num_std)
 {
-    if (!(track->parser_priv->header_flags & SINFO_SCALEDBORDER)
-        && !format_line_compare(fmt, num_fmt, std, num_std)) {
+    if (!(track->parser_priv->header_flags & SINFO_SCALEDBORDER) &&
+            (num_std != num_fmt || memcmp(fmt, std, num_std * sizeof(*fmt)))) {
         ass_msg(track->library, MSGL_INFO,
                "Track has custom format line(s). "
                 "'ScaledBorderAndShadow' will default to 'yes'.");
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -267,40 +267,40 @@ static long long string2timecode(ASS_Library *library, char *p)
     if (!token) break;
 
 
-#define PARSE_START_TK switch (fmt->tokens[tokenpos]) {
-#define PARSE_END_TK   }
+#define PARSE_START switch (fmt->tokens[tokenpos]) {
+#define PARSE_END   }
 
-#define STRVAL_TK(name) \
+#define STRVAL(name) \
     case ASS_FMT_ ## name : \
         free(target->name); \
         target->name = strdup(token); \
         break;
 
-#define STARREDSTRVAL_TK(name) \
+#define STARREDSTRVAL(name) \
     case ASS_FMT_ ## name : \
         while (*token == '*') ++token; \
         free(target->name); \
         target->name = strdup(token); \
         break;
 
-#define TIMEVAL_ALTNAME_TK(fieldname,formatname) \
+#define TIMEVAL_ALTNAME(fieldname,formatname) \
     case ASS_FMT_ ## formatname : \
         target->fieldname = string2timecode(track->library, token); \
         break;
 
-#define STYLEVAL_TK(name) \
+#define STYLEVAL(name) \
     case ASS_FMT_ ## name : \
         target->name = lookup_style(track, token); \
         break;
 
-#define ANYVAL_TK(name,func) \
+#define ANYVAL(name,func) \
     case ASS_FMT_ ## name : \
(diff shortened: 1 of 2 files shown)


TheOneric:
Pushed new version with track-change check moved to `process_text` and `process_chunk` and a memleak on multiple format lines fixed. [DIff](https://github.com/libass/libass/compare/97f720d3b67ceeaff8d02eeeea5c2cfd6f17a7e6..33880bf7eb5fe02251e28b9205a0a4e1aaa95bd5)

The numbers from the previous version are still useful though, as they model the "worst-case" where every line is submitted separately, requiring a change-check each. The performance table from my previous comment has been updated to contain "best-case" numbers in addition to the previous  "worst-case" ones. For Aegisub, the "best-case" is representative of what happens.

TheOneric:
Added two cosmetic commits to reduce the old parsing macros to `ass_process_force_style` and limit them to only what's needed there, and also rename all parsing macros.

libass/ass.c
```diff
@@ -332,51 +467,40 @@ static int process_event_tail(ASS_Track *track, ASS_Event *event,
...
-        return -1;
-    char *q = format;           // format scanning pointer
-
-    for (i = 0; i < n_ignored; ++i) {
-        NEXT(q, tname);
-    }
-
-    while (1) {
-        NEXT(q, tname);
-        if (ass_strcasecmp(tname, "Text") == 0) {
```
MrSmile, line 351: Where does this check go?

libass/ass.c
```diff
@@ -473,36 +597,32 @@ static int process_style(ASS_Track *track, char *str)
...
-        if (track->track_type == TRACK_TYPE_SSA)
-            track->style_format = strdup(ssa_style_format);
-        else
-            track->style_format = strdup(ass_style_format);
-        if (!track->style_format)
-            return -1;
+        if (track->track_type == TRACK_TYPE_SSA) {
+            fmt->n_tokens = ssa_style_format_length;
+            memcpy(fmt->tokens, ssa_style_format,
+                    sizeof(*ssa_style_format) * ssa_style_format_length);
```
MrSmile, line 602: Isn't it always equals to `sizeof(ssa_style_format)`?

libass/ass.c
```diff
@@ -292,6 +312,50 @@ static long long string2timecode(ASS_Library *library, char *p)
     } else if (ass_strcasecmp(tname, #name) == 0) { \
         target->name = lookup_style(track, token);
 
+
+#define PARSE_START_TK switch (fmt->tokens[tokenpos]) {
+#define PARSE_END_TK   }
+
+#define ALIAS_FOR_NEXT_TK(name) \
```
MrSmile: Unused.

libass/ass.c
```diff
@@ -574,29 +707,17 @@ static int process_style(ASS_Track *track, char *str)
     if (strcmp(target->Name, "Default") == 0)
         track->default_style = sid;
     return 0;
-
 }
 
-static bool format_line_compare(const char *fmt1, const char *fmt2)
+static bool format_line_compare(const ASS_FormatToken *fmt1, const ASS_FormatToken *fmt2, size_t num)
```
MrSmile, line 701: Isn't it simply `memcmp()`?

libass/ass.c
```diff
@@ -612,11 +733,12 @@ static bool format_line_compare(const char *fmt1, const char *fmt2)
  * keep SBAS=1 default for custom format files.
  */
 static void custom_format_line_compatibility(ASS_Track *const track,
-                                             const char *const fmt,
-                                             const char *const std)
+                                             const ASS_FormatToken *fmt,
+                                             const ASS_FormatToken *std,
+                                             size_t num)
```
MrSmile, line 727: This function isn't equivalent to its previous version: should be separate `n_fmt` and `n_std` instead of one `num`.

libass/ass.c
```diff
@@ -321,6 +385,77 @@ static char *next_token(char **str)
     return start;
 }
 
+static int parse_format_line(ASS_FormatLine *fmt, const char *string)
```
MrSmile, line 339: This function can return `0` or `-1`. Better to replace with `bool`.

libass/ass.c
```diff
@@ -767,18 +900,25 @@ static bool detect_legacy_conv_subs(ASS_Track *track)
 
 static int process_events_line(ASS_Track *track, char *str)
 {
+    int ret = 0;
```
MrSmile, line 892: Why is that even needed? `ret` is zero everywhere except a couple of lines.

MrSmile (commented):
I'm not a fan of this sync approach. It's good to have an enum for fields instead of strings, but only if we can change API. Performance drop is due to the usual string switches, so if we solve the string switch problem then format processing would be fast enough even with the strings.

TheOneric:
> It's good to have an enum for fields instead of strings, but only if we can change API.

I'm open to restricting the API. The first version of this patch I posted 5 months ago did just that. It also added a resync function to the API, so that, **if** a API-user really wants to do this, it can be done, but requires a manually triggered resync afterwards. This also makes it relatively easy to support both newer and older libass-version by `#ifdef`-ing the resync call; but it's still an API change.
A prior example of adding API-restrictions is the `process_chunk`-change in de4ea4bbd1c5101bf0b572a4facd58875845f495.

> I'm not a fan of this sync approach.

Do you have a more performant sync mechanism in mind, that doesn't break API?
If this is a general concern about the performance impact of any sync-check:
As my performance measurements show, even in the worst case where this is executed every event, we still get nice performance boost in relative scenarios and even in the worst-case and worst-sample there's only a barely measurable performance regression.
This is because we don't need to reprocess the string, can drop a `strdup`, and since we compare the format lines as a whole, instead of each token separately we have a longer string and can better utilise optimised SIMD'ed versions of `strcmp`.

> Performance drop is due to the usual string switches, so if we solve the string switch problem then format processing would be fast enough even with the strings.

I highly doubt this. Not only are string-switches not a thing, the enum-switch can reach O(1) *(by using fixed binary size for each branch and a resulting fixed offset to jump to by something like `base_offset + fmt_enum * b_size`)* time if the compiler deems sacrificing some binary space for non-checked format tokens worth it. If not there's still a tree map for offsets giving us O(log_2(n)) at worst. *(before patch: `O(n)*O(short-strcmp)`)*
String switches are not only not a thing, if you compute hashes for them you first of all don't have a small range of values starting from zero, so the O(1) optimisation is already not feasible, and you still need to compare the strings since hashes can collide. So you cut down on string comparisons, but are still faring worse than the enum approach. *(assuming cheap hashes)*
Also you'd still need to reprocess the format string everytime adding more cost than just the string differentiation *(if you're caching the format positions you also need to resync might as well just use enums again)*.
I invite you to prove me wrong with reproducible performance tests. My tests above were conducted with default `CFLAGS` etc on top of 0.15.1 with callgrind.

astiob:
> by using fixed binary size for each branch and a resulting fixed offset to jump to by something like `base_offset + fmt_enum * b_size`

I believe it’s usually done with a table of pointers to code, each pointing to one case statement (whose code may vary in size). The table is indexed with the switch-object value, and an indirect jump is performed to the pointer. (Among other names, this is known as a jump table.)

TheOneric:
New version incorporating most of MrSmile's feedback, thanks for the review!
A new cosmetic commit was added to replace `format_line_compare` with a simple `memcmp`.

Diff of just the first commit each: https://github.com/libass/libass/compare/33880bf..3ce02565101523d7c1b214afe5c2c726bb96cb8b
Diff of all commits: https://github.com/libass/libass/compare/c5691d113e46a45cf1782d092e5ebed4a7204e11..f5a504c00ebdb92384975666405aed3d39d820a5

TheOneric:
> I believe it’s usually done with a table of pointers to code, each pointing to one case statement (whose code may vary in size). The table is indexed with the switch-object

Right, that's also possible and saves some binary size at the cost of an additional load. Still only feasible if all values are close to zero though.

libass/ass.c
```diff
@@ -292,6 +312,50 @@ static long long string2timecode(ASS_Library *library, char *p)
...
+        free(target->name); \
+        target->name = strdup(token); \
+        break;
+
+#define TIMEVAL_TK(name) \
+    case ASS_FMT_ ## name : \
+        target->name = string2timecode(track->library, token); \
+        break;
+
+#define TIMEVAL_ALTNAME_TK(fieldname,formatname) \
```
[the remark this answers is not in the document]
MrSmile: Yes, I mean comma without space.

libass/ass.c
```diff
@@ -321,6 +385,77 @@ static char *next_token(char **str)
...
+        }
+
+        #define FMTTOKEN(name) \
+            else if (strncasecmp(tk_start, #name, tk_end-tk_start) == 0) { \
+                fmt->tokens[fmt->n_tokens++] = ASS_FMT_ ## name ; \
+            }
+
+        if (0) {}
+        // Order matters!
+        // Tokens which are a prefix of other valid tokens must be tested before them.
```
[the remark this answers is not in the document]
MrSmile: Isn't it possible to get rid of that ordering by using exact string comparisons with length? It should also be faster.

libass/ass.c
MrSmile: > Seems like this got dropped at some point.

Yes, it should be moved to format line parsing but vanished instead.

libass/ass.c
MrSmile: This function has fully rewritten anyway, I don't see much sense in rewriting it to only delete later.

libass/ass.c
MrSmile: I mean old version checked exact equality, but current version checks that one token list is a prefix of another. I.e. there's missing `n_fmt == n_std` check.

libass/ass.c
MrSmile: I think that such "reserved for the future" features are a root of all evil. When that function really needed to differentiate errors then and only then change to `int` should be done.

libass/ass.c
MrSmile: It's a complication for no reason, direct violation of KISS principle. I've suggested to replace return value of `parse_format_line()` to `bool` exactly to keep it conceptually correct.

MrSmile:
> if you compute hashes for them you first of all don't have a small range of values starting from zero, so the O(1) optimisation is already not feasible

O(1) is always feasible for a static set of values. For example
```c
unsigned format_hash(const char *str, size_t len)
{
    unsigned c1 = str[0] | 0x20u;
    unsigned c2 = str[len < 7 ? len - 1 : 6] | 0x20u;
    return ((26 + 5 * (unsigned) len + 9 * c2) ^ c1) & 63;
    return ((5 + 2 * (unsigned) len + 22 * c2) ^ 15 * c1) % 47;
}
```
gives you a unique value for each format token in the range [0–63] (first return) or [0–46] (second return).

libass/ass.c
TheOneric: All other functions in this file use classic `int` for error return codes or zero on success, like `main` does. With `bool` you'd normally expect to return `true` on success and false otherwise. Using `bool`as an error indicator only here would only introduce inconsistency in failure checks. *(or create confusion by using `true` as an error indicator)*

libass/ass.c
TheOneric: Readded a warning to Event-Format parsing.

libass/ass.c
TheOneric: `ret` removed as it's currently only set in one place which (currently) is always a fatal error.

libass/ass.c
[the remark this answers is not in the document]
TheOneric: Added a `tk_end-tk_start == sizeof( #name ) - 1` check before the `strncasecmp` to make it order-insensitive.

libass/ass.c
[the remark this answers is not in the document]
TheOneric: As there are also spaces missing in macros not at all touched by this patch-series, I've opted to not add spaces in the initial commit, but add a later cosmetic commit to add spaces to *all* macro-arguments in this file so it stays locally consistent at all times.

MrSmile (commented):
I think that such miniscule performance difference isn't worth introduced code complexity (additional bookkeeping, three parsing mechanisms, for format line, old `*_STR` and new `*_TK`, instead of previous one). But at least we should wait for the merge of #506 as it could significantly lessen an already small performance difference.

Style nit: there are not enough spaces not only around `,`, but also `-`. Additional cosmetic commit could fix that too.

TheOneric:
After rebasing and including MrSmile's suggestions, here are new perf-measurements, callgrind *(without CPU-cache)* and one sample with time-based measurements. 
 For comparison a fixed-format implementation was also tested[1], but note that for simplicity the initial format line parsing was not removed, which shows for low event counts, but should be irrelevant in all other cases. The "worst case" of every event being fed as it's own is now also approximated with a patch[2].

First callgrind: The improvement is given as reduced cost, in percent of the unpatched cost, so: `100 - 100 * (patched cost)/(unpatched cost)`. With the simple callgrind configuration I used, the “cost” here is identical to the instructions executed.
As before, everything but the `eventsPerFrame`-set are real-world files, with everything but the `aegi_`-set being fully rendered after parsing. The `aegi_` set only renders a few seconds (but fully parses each file).
```
                        mkvchunk[%]    patch[%]    fixed-eventformat[%]

       aegi_dialogue		 2.21		 2.29		 2.19
         aegi_fxkara		 5.42		 5.49		 5.75
  aegi_jojo-e20_full		20.71		21.52		22.37
    aegi_neptunia-op		13.04		13.33		13.93
 aegi_snk-e01v2_full		18.93		19.23		20.23
  aegi_yurucamp-rock		 4.15		 4.26		 4.52      (<- probably also an artificat)

       jojo-e20_full		 0.46		 0.47		 0.52
      lightly-styled		 0.02		 0.02		 0.03
            mpv-7712		 0.02		 0.02		 0.02
   real-world-fxkara		 0.05		 0.05		 0.05
      snk-e01v2_full		 0.40		 0.41		 0.42
        snk-e01v2_op		 2.75		 2.75		 2.91
   YuruCamp2_02_rock		 0.83		 0.83		 0.87
     zoinks-neptunia		 2.03		 2.07		 2.19
zoinks-neptunia_logo		 2.29		 2.37		 2.46

      1eventPerFrame		 0.24		 0.21		 0.19      (<- artificats from imperfect simulation)
    32eventsPerFrame		 3.99		 4.79		 4.98
    64eventsPerFrame		 6.38		 6.47		 6.78
   128eventsPerFrame		 7.98		 8.10		 8.50
   512eventsPerFrame		 9.73		 9.92		10.41
 10240eventsPerFrame		10.42		10.59		11.15

 ```
The script used to run a set of callgrind test is attached to the previous perf table post.


---

Time-based profiling was done with an infalted version of the SnK sample *(`awk '/^Style: / || /^Dialogue: / {for(i=0; i < 32; ++i) print $0;} 1 {print $0}' snk_e01v2.ass > snk_inflated2_1411_1412_1.ass`)*, and tested both with multiple `profile` runs timed with `time`[3] and a patched profile.c which does the timing and looping[4]. Both approaches used 10 warmup runs followed by 50 measurement runs.

A note of caution: Those two approaches, also after readding the seperate results (and their uncertaintities) from the patched version, show a difference beyond the to be expected mere statistical uncertainty *(using linear propagation of uncertainty)*, indicating that at least one (but probably both) of those measurements have systematic and similar errors *(which isn't surprising and I don't know how to properly estimate and reduce this kind of benchmark error)*.

Taking the measurements at face value, the time reduction shown by time+profile was `10.3%`, for the patched-profile measurements including rendering `11.0%` and without rendering `12.0%`. The mere statistical uncertainty for those numbers using linear uncertainty propagation are `0.260`, `0.188` and `0.166` respectively.
More detailed run-data is attached[5].

So, I have no idea how accurate the results of the timed tests are, but this at least supports that there is some performance-improvement.

---
[1]: [Fixed Event Format patch](https://github.com/libass/libass/files/6528897/fixed-eventformat.diff.txt)
[2]: [Single Events Patch](https://github.com/libass/libass/files/6528961/worst-case-simul.diff.txt)
[3]: [perf.sh](https://github.com/libass/libass/files/6528991/perf.sh.txt)
[4]: [perf-measuring in profile.c patch](https://github.com/libass/libass/files/6528992/perf-profile.diff.txt)
[5]: [timed run results](https://github.com/libass/libass/files/6529036/469_timedruns_20210524.zip)

File: libass/ass.c [hunk 10]
```diff
@@ -75,6 +94,10 @@ void ass_free_track(ASS_Track *track)
         free(track->parser_priv->read_order_bitmap);
         free(track->parser_priv->fontname);
         free(track->parser_priv->fontdata);
+        free(track->parser_priv->style_format.tokens);
```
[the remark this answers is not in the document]
TheOneric: Custom format lines can be of arbitrary length (but not meaningful). The initial length is sufficient for all somewhat meaningful format lines.

libass/ass.c
```diff
@@ -511,48 +623,61 @@ static int process_style(ASS_Track *track, char *str)
...
+            INTVAL_TK(Alignment)
+            INTVAL_TK(MarginL)
+            INTVAL_TK(MarginR)
+            INTVAL_TK(MarginV)
+            INTVAL_TK(Encoding)
+            FPVAL_TK(ScaleX)
+            FPVAL_TK(ScaleY)
+            FPVAL_TK(Outline)
+            FPVAL_TK(Shadow)
+            default:
```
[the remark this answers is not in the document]
TheOneric: fwiw, I like the compat adjustments and sanitisation being bundled at the end more, but that's just a subjective preference

libass/ass.c
```diff
@@ -772,13 +899,24 @@ static int process_events_line(ASS_Track *track, char *str)
...
+        if (parse_format_line(&track->parser_priv->event_format, p))
             return -1;
-        ass_msg(track->library, MSGL_DBG2, "Event format: %s", track->event_format);
-        if (track->track_type == TRACK_TYPE_ASS)
-            custom_format_line_compatibility(track, p, ass_event_format);
-        else
-            custom_format_line_compatibility(track, p, ssa_event_format);
+        if (track->parser_priv->event_format.n_tokens > 0 &&
+                track->parser_priv->event_format.tokens[track->parser_priv->event_format.n_tokens-1] != ASS_FMT_Text) {
+            ass_msg(track->library, MSGL_WARN, "Event Format line does not end with Text! Pretend it does.");
```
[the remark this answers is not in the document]
TheOneric: Good point about Text being in the middle of the Format line.
Appending `Text` or replacing the last token fail in different ways (but the Format line is already broken): If the actual token count of events matches with the Format line but `Text` get's appended, we'll not render the event. If however, there's actually an additional Text token in the events, we don't parse a control token but include it in Text.

But in both of the above cases the Format line is already borked anyway. Do we want to preserve old behaviour for a broken Format line, or just log a warning about it?

libass/ass.c
```diff
@@ -574,49 +699,41 @@ static int process_style(ASS_Track *track, char *str)
...
-                                             const char *const fmt,
-                                             const char *const std)
+                                             const ASS_FormatToken *fmt,
+                                             size_t num_fmt,
+                                             const ASS_FormatToken *std,
+                                             size_t num_std)
 {
     if (!(track->parser_priv->header_flags & SINFO_SCALEDBORDER)
-        && !format_line_compare(fmt, std)) {
+        && !format_line_compare(fmt, std, FFMIN(num_fmt, num_std))) {
```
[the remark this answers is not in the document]
TheOneric: It did already use `memcmp(fmt, std, FFMIN(num_fmt, num_std) * sizeof(*fmt))` before, now changed to `(num_std != num_fmt || memcmp(fmt, std, num_std * sizeof(*fmt)))`

libass/ass.c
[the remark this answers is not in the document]
MrSmile: It's possible to use fixed-size buffers for that, in the form of pairs (n_skip, token_id). But it's a complication indeed.

libass/ass.c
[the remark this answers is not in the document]
MrSmile: At least `default:` should be (re)moved.

libass/ass.c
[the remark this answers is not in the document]
MrSmile: I personally prefer to fail hard, but it's another question of compatibility. At least, I think that behavior change shouldn't be in the same commit.

libass/ass.c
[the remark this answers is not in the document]
MrSmile: I'm saying that `num_fmt != num_std` and `FFMIN(num_fmt, num_std)` are not equivalent. Old behavior corresponds to former, so any `FFMIN()` here is plain wrong. It should be `(num_fmt != num_std || !format_line_compare(fmt, std, num_fmt))` if you want to keep commit count.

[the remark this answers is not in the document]
TheOneric: ohh… I only now got what you meant this whole time, sorry it took me so long. Yes, that would have caused false positives and is now fixed. Thanks for your insistence.

Issue #334, opened 2019-03-04
Slow fontselect when thousands of fonts are installed (ms-windows)


zc62:
Several related issues in mpv:
https://github.com/mpv-player/mpv/issues/2998
https://github.com/mpv-player/mpv/issues/6447
https://github.com/mpv-player/mpv/issues/6486

libass is built without fontconfig, so it is using directwrite (#79). Turns out it costs a few seconds even when the subtitle (or mpv's OSD) is requesting only a few kinds of fonts, if a huge amount of fonts are installed. XyVSFilter probably does not have this behavior (did not really test it).

Apache553:
The `scan_fonts` function of DirectWrite font provider will take much time when plenty of fonts installed. As a result of tracing API calls, the whole system font collection would be traversed about 3 times. Each traversal takes about 5 seconds with the API Monitor attached(slowing down the program being monitored) and 874 font families installed in my computer. I believe this function is the cause of the traversal. As far as I know, the xySubFilter uses CreateFont* to get font when the subtitle referred a font hasn't been loaded yet. Maybe we can just load those fonts by their names from system when they are required instead of scanning fonts. I have made some [modifications](https://github.com/Apache553/libass/commit/c141ff1e2bdeaf07d4f2224327c5642af7cee7c7) to the code to implement this and compiled and linked to mpv 0.33.0 release, traced again and found the mpv started faster than before and no longer did the traversal. The rendered subtitle looked correct, at least on my computer.

zc62:
Great! Could you make it a PR? @Apache553

ValZapod:
Regression? https://github.com/mpv-player/mpv/issues/8802#issuecomment-830799010

Apache553:
> Regression? [mpv-player/mpv#8802 (comment)](https://github.com/mpv-player/mpv/issues/8802#issuecomment-830799010)

I made a [pull request](https://github.com/libass/libass/pull/503) to fix this but it haven't been merged yet.

PR #477, opened 2021-02-22
directwrite: request font on demand

Apache553:
scan_fonts can be very slow when many fonts were installed in
the system. Use IDWriteGdiInterop to create a font by its name
when needed instead of scanning all installed fonts during the
initializing stage to get better performance. In case of a
non-existent font, the fallback mechanism is supposed to do its 
work.(I am not sure about the consequences of removing 'scan_fonts',
but I think it should be ok.)

### DIFF
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -50,6 +50,7 @@ typedef struct {
 typedef struct {
     HMODULE directwrite_lib;
     IDWriteFactory *factory;
+    IDWriteGdiInterop *gdi_interop;
 } ProviderPrivate;
 
 /**
@@ -389,6 +390,7 @@ static bool check_glyph(void *data, uint32_t code)
 static void destroy_provider(void *priv)
 {
     ProviderPrivate *provider_priv = (ProviderPrivate *)priv;
+    provider_priv->gdi_interop->lpVtbl->Release(provider_priv->gdi_interop);
     provider_priv->factory->lpVtbl->Release(provider_priv->factory);
     FreeLibrary(provider_priv->directwrite_lib);
     free(provider_priv);
@@ -668,50 +670,47 @@ cleanup:
 }
 
 /*
- * Scan every system font on the current machine and add it
- * to the libass lookup. Stores the FontPrivate as private data
- * for later memory reading
+ * When a new font name is requested, called to load that font from Windows
  */
-static void scan_fonts(IDWriteFactory *factory,
-                       ASS_FontProvider *provider)
+static void match_fonts(void *priv, ASS_Library *lib,
+                        ASS_FontProvider *provider, char *name)
 {
-    HRESULT hr = S_OK;
-    IDWriteFontCollection *fontCollection = NULL;
+    ProviderPrivate *provider_priv = (ProviderPrivate *)priv;
     IDWriteFont *font = NULL;
-    hr = IDWriteFactory_GetSystemFontCollection(factory, &fontCollection, FALSE);
+    IDWriteFontFamily *fontFamily = NULL;
+    LOGFONTW lf;
+    HRESULT hr = S_OK;
 
-    if (FAILED(hr) || !fontCollection)
+    memset(&lf, 0, sizeof(LOGFONTW));
+    lf.lfWeight = FW_DONTCARE;
+    lf.lfCharSet = DEFAULT_CHARSET;
+    lf.lfOutPrecision = OUT_TT_PRECIS;
+    lf.lfClipPrecision = CLIP_DEFAULT_PRECIS;
+    lf.lfQuality = ANTIALIASED_QUALITY;
+    lf.lfPitchAndFamily = DEFAULT_PITCH | FF_DONTCARE;
+
+    // lfFaceName can hold up to LF_FACESIZE wchars; truncate longer names
+    MultiByteToWideChar(CP_UTF8, 0, name, -1, lf.lfFaceName, LF_FACESIZE-1);
+
+    hr = IDWriteGdiInterop_CreateFontFromLOGFONT(provider_priv->gdi_interop, &lf, &font);
+    if (FAILED(hr) || !font)
         return;
 
-    UINT32 familyCount = IDWriteFontCollection_GetFontFamilyCount(fontCollection);
+    hr = IDWriteFont_GetFontFamily(font, &fontFamily);
+    if (FAILED(hr) || !fontFamily)
+        goto cleanup;
 
-    for (UINT32 i = 0; i < familyCount; ++i) {
-        IDWriteFontFamily *fontFamily = NULL;
+    add_font(font, fontFamily, provider);
 
-        hr = IDWriteFontCollection_GetFontFamily(fontCollection, i, &fontFamily);
-        if (FAILED(hr))
-            continue;
+    IDWriteFontFamily_Release(fontFamily);
 
-        UINT32 fontCount = IDWriteFontFamily_GetFontCount(fontFamily);
-        for (UINT32 j = 0; j < fontCount; ++j) {
-            hr = IDWriteFontFamily_GetFont(fontFamily, j, &font);
-            if (FAILED(hr))
-                continue;
+    return;
 
-            // Simulations for bold or oblique are sometimes synthesized by
-            // DirectWrite. We are only interested in physical fonts.
-            if (IDWriteFont_GetSimulations(font) != 0) {
-                IDWriteFont_Release(font);
-                continue;
-            }
-
-            add_font(font, fontFamily, provider);
-        }
-
+cleanup:
+    if (font)
+        IDWriteFont_Release(font);
+    if (fontFamily)
         IDWriteFontFamily_Release(fontFamily);
-    }
-
-    IDWriteFontCollection_Release(fontCollection);
 }
 
 static void get_substitutions(void *priv, const char *name,
@@ -731,6 +730,7 @@ static ASS_FontProviderFuncs directwrite_callbacks = {
     .check_glyph        = check_glyph,
     .destroy_font       = destroy_font,
     .destroy_provider   = destroy_provider,
+    .match_fonts        = match_fonts,
     .get_substitutions  = get_substitutions,
     .get_fallback       = get_fallback,
     .get_font_index     = get_font_index,
@@ -755,6 +755,7 @@ ASS_FontProvider *ass_directwrite_add_provider(ASS_Library *lib,
 {
     HRESULT hr = S_OK;
     IDWriteFactory *dwFactory = NULL;
+    IDWriteGdiInterop *dwGdiInterop = NULL;
     ASS_FontProvider *provider = NULL;
     DWriteCreateFactoryFn DWriteCreateFactoryPtr = NULL;
     ProviderPrivate *priv = NULL;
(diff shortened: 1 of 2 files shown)


Apache553:
#334

Masaiki:
This patch really helps on Windows, which saves memory and time to load fonts.
Especially for those who has a lot of fonts (subbers always).
Hope to merge.

astiob:
I’ve taken the liberty to update and force-push the patch to:
* rebase it on master (removing the provider-private hack in the process thanks to #491),
* remove `scan_fonts` entirely,
* use `LF_FACESIZE` instead of `32`,
* adjust style slightly to better match the existing content of the files.

Could anyone confirm that I haven’t broken anything, that this still builds and works on Windows?

astiob:
I also think we should get rid of `NAME_MAX_LENGTH`, but it existed before this patch, so I left it alone for now.

Apache553:
Thanks for your work! I have built libass from the source code from both master and this pr with default configuration, dynamically linked to mpv 0.33.0 mingw64 respectively. Both builds rendered subtitles in the same fonts. So the fonts are loaded as expected. However, I found some strange color blocks on subtitle lines or mpv ui.

def01
def02

Then I tried different compilation configurations. I noticed that configuring libass with `--disable-asm` specified can solve this problem.

good01

MrSmile:
> I noticed that configuring libass with `--disable-asm` specified can solve this problem.

Can you try without `--disable-asm` and with [this patch](https://github.com/MrSmile/libass/commit/db5813281393e39ea996118fd02ddb9442ad3889)?

Apache553:
Built with patched asm, now there are no longer color blocks. Looks good!

TheOneric:
Don't know how relevant this or if it's a false positive, and it also already happened before this PR, so feel free to ignore:
There's a compiler warning about an invalid cast at line 766 (pre-PR) and 773 (post-PR) of ass_directwrite.c respectively [(MSys2+MinGW64)](https://github.com/TheOneric/libass-regression-tests-tmp/runs/2476491018?check_suite_focus=true#step:4:193):
```
ass_directwrite.c: In function 'ass_directwrite_add_provider':
ass_directwrite.c:766:30: warning: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'HRESULT (*)(DWRITE_FACTORY_TYPE,  const IID * const,  IUnknown **)' {aka 'long int (*)(DWRITE_FACTORY_TYPE,  const GUID * const,  IUnknown **)'} [-Wcast-function-type]
  766 |     DWriteCreateFactoryPtr = (DWriteCreateFactoryFn)GetProcAddress(directwrite_lib,
      |                              ^
```

astiob:
Looks like just a generic function pointer cast warning, unless I’m missing something about function attributes etc. Seems intentional. (But thanks for checking!)

astiob (approved):
Re-pushed again to remove the `wide_name` intermediate buffer (and move an unchanged function so that it doesn’t show up in the diff). I thought `MultiByteToWideChar` simply failed if the buffer was too short without touching the buffer, because the documentation doesn’t make it clear what happens to the buffer but it returns an error and a 0 as the “written wchar count” return value, and this seemed in line with Microsoft’s “security first” approach to string functions also seen in `strcpy_s` and `strncpy_s`. But now I’ve run a simple test, and it shows that it does fill the whole buffer with the correct truncated output (ending in a lone surrogate if need be). This seems undocumented, but knowing Microsoft’s approach to backwards compatibility, it seems safe to rely on this.

I’ve also pushed a commit to master to quelch the function pointer cast warning (which should also appear in MSVC) by casting through `void*`. In perfect portable C, that would be even worse than the direct function pointer cast, but this is Windows-specific code.

Shall we merge this in 0.15.1? I assume people have tested this for a while. We’ve even found and fixed a directly-unrelated assembler bug :smiley: I like this and rcombs seemed to like this, so if there are no more objections, I think this can go in 0.15.1; unless we are finally sick of the endless Coverity & OSS-Fuzz waiting periods and want to just push the realease out the door.

astiob:
Also, many thanks to @Apache553 for writing the patch in the first place!

TheOneric:
> Shall we merge this in 0.15.1?

No objection from me. But as I'm not familiar with DIrectWrite, I can't give a actual review of this code.

> unless we are finally sick of the endless Coverity & OSS-Fuzz waiting periods and want to just push the realease out the door.

I'm in favour of keeping the Coverity&OSS-Fuzz freeze period *(which is short anyway)*.

astiob:
I’ve realized that this file is not compiled by either Coverity or OSS-Fuzz anyway. So if this PR is the only extra thing we merge, we need not obey the freeze period as it is useless.

libass/ass_directwrite.c
```diff
@@ -668,50 +670,49 @@ static void add_font(IDWriteFont *font, IDWriteFontFamily *fontFamily,
...
+    ProviderPrivate *provider_priv = (ProviderPrivate *)priv;
     IDWriteFont *font = NULL;
-    hr = IDWriteFactory_GetSystemFontCollection(factory, &fontCollection, FALSE);
+    IDWriteFontFamily *fontFamily = NULL;
+    LOGFONTW lf;
+    HRESULT hr = S_OK;
 
-    if (FAILED(hr) || !fontCollection)
+    memset(&lf, 0, sizeof(LOGFONTW));
+    lf.lfEscapement = GM_COMPATIBLE;
```
astiob, line 685: @Apache553 Can you explain this value? Looking at the [`LOGFONTW` docs](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-logfontw), it doesn’t seem that this constant goes in this field.

astiob: For comparison, VSFilter simply doesn’t set this field (leaves it zero-filled).

libass/ass_directwrite.c
```diff
@@ -668,50 +670,49 @@ static void add_font(IDWriteFont *font, IDWriteFontFamily *fontFamily,
...
+    lf.lfWeight = FW_DONTCARE;
+    lf.lfCharSet = DEFAULT_CHARSET;
+    lf.lfOutPrecision = OUT_TT_PRECIS;
+    lf.lfClipPrecision = CLIP_DEFAULT_PRECIS;
+    lf.lfQuality = DEFAULT_QUALITY;
```
astiob, line 690: I suspect this probably doesn’t matter for our use case, but VSFilter sets this field to `ANTIALIASED_QUALITY`.

libass/ass_directwrite.c
```diff
@@ -668,50 +670,49 @@ static void add_font(IDWriteFont *font, IDWriteFontFamily *fontFamily,
...
+    lf.lfPitchAndFamily = DEFAULT_PITCH;
+
+    // lfFaceName can hold up to LF_FACESIZE wchars; truncate longer names
+    MultiByteToWideChar(CP_UTF8, 0, name, -1, lf.lfFaceName, LF_FACESIZE-1);
+    lf.lfFaceName[LF_FACESIZE-1] = 0;
```
astiob, line 695: And I’ve just realized this line is redundant because the struct is prefilled with zeros.

libass/ass_directwrite.c
Apache553: This is snippet is copied from another project... After I read the MSDN again just now, I have to admit that this assignment is wrong. It seems that it didn't affect getting the font, since our purpose is to get physical font data used by freetype.

libass/ass_directwrite.c
Apache553: IIRC this field affects the result of GDI rendered text. So it doesn't matter in our use case.

libass/ass_directwrite.c
Apache553: This line is to make sure this field is null-terminated. `wcsncpy` will not append a null character if source string is longer than dest buffer. Since we get rid of that intermediate buffer which is not prefilled with zeros, we can just remove this line.

astiob: Ah, right, I’d forgotten this was a remnant of the intermediate buffer which wasn’t zero-filled.

libass/ass_directwrite.c
astiob: Yeah, that’s what I think, too. But since I was editing this file anyway (to remove the `lfEscapement` just in case, and to remove the NUL termination just because), I’ve changed this to match VSFilter. At worst this should make no difference; at best this should help.

astiob:
Thanks a lot!

astiob:
This change broke font families with multiple styles.

`match_fonts` receives a font family name and is supposed to return all fonts that match that name. In this PR, it returns a single font, specifically the Regular Roman variant (or the next closest thing if one doesn’t exist).

Fonts still work, so this is better than the macOS regression in 0.15.0 where font fallback resulted in completely illegible text. But instead of using the right font face from each family, libass now uses the default variant and synthesizes bold (a single weight), slanted and slanted bold from it.

Discussion in #502 suggests using the GDI function [`EnumFontFamiliesEx`](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-enumfontfamiliesexw), as is done [in Aegisub’s GDI font provider](https://github.com/TypesettingTools/Aegisub/blob/b00f5dd58eb36cc87109ca3529ad52a789a8db03/src/libass_gdi_fontselect.cpp#L88). I don’t see a direct [`IDWriteGdiInterop`](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/nn-dwrite-idwritegdiinterop) equivalent for it, so I assume the only way is to do the font lookup in actual GDI, like Aegisub does it (and like #502 is trying to do but taken a step further):
1. call `EnumFontFamiliesEx` to get a list and
2. * materialize each of the fonts it returns via `CreateFontIndirectW` + `SelectObject` + `CreateFontFaceFromHdc` as in #502; or
    * `CreateFontIndirectW` + keep processing the font via GDI directly as in Aegisub; or
    * possibly `CreateFontFromLOGFONT` if it works in those corner cases after `EnumFontFamiliesEx`.

Apache553:
OK this is my fault. I didn't know that `match_fonts` is supposed to add all fonts with that family name. We can simply do this:

1. Calling `CreateFontFromLOGFONT` as usual to get a font as `pFont`.
2. Call `pFont->GetFontFamily(&pFontFamily)` to get the whole font family as `pFontFamily` .
3. Call `pFontFamily->GetFontCount()` to get the count of fonts in that font family.
4. Call `pFontFamily->GetFont()` to get each font and add them. Every variant(different styles and weights) font of that font family will be added.

If I understand correctly, this should meet the requirements of `match_fonts`. I can make a PR to implement this if needed.

astiob:
To be fair, it’s not just your fault. We should’ve caught this in review (especially since just recently I fixed an issue in Core Text where confusion between single fonts and whole families was also involved). Finally, as much as I trusted your limited tests, I shouldn’t have rushed this out the door as part of a release before at least a few days of wider testing in `master`.

> […] that `match_fonts` is supposed to add all fonts with that family name.

For better understanding:

In general, it can get two kinds of names, because fonts can be referred by two kinds of names in ASS: the family name or the full-or-PostScript name (depending on whether the font has TrueType or PostScript outlines). When the full name is used, it should be safe to return just one font because there should be only one, and libass will select the first full-name match anyway. But when the family name is used, then yeah, we really want the whole family. `match_fonts` takes no other parameters that could specify which specific font face we need—so this “should” be clear from the API alone, but as so many other things, it is only clear in retrospect; and I totally forgot about this during code review, too.

PR #479, opened 2021-02-28
Improve ISO-C compatability

TheOneric:
Does not actually change our target to ISO C99 + POSIX `str(n)dup`.  
~~I'm not sure about putting `ASS_PI` into `ass_compat.h`, but couldn't come up with a better place for it.~~ `ASS_PI` has been moved to `ass_utils.h`

### DIFF
--- a/libass/ass_compat.h
+++ b/libass/ass_compat.h
@@ -26,4 +26,15 @@
 #define inline __inline
 #endif
 
+#ifndef HAVE_STRDUP
+char *ass_strdup_fallback(const char *s); // definition in ass_utils.c
+#define strdup ass_strdup_fallback
+#endif
+
+#ifndef HAVE_STRNDUP
+#include <stddef.h>
+char *ass_strndup_fallback(const char *s, size_t n); // definition in ass_utils.c
+#define strndup ass_strndup_fallback
+#endif
+
 #endif                          /* LIBASS_COMPAT_H */
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -510,15 +510,18 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
     // z = m_zx * x + m_zy * y + m_zz
     //  = m_zx * (x + sign(m_zx) * dx) + m_zy * (y + sign(m_zy) * dy) + z0.
 
-    // D(f)--absolute value of error in quantity f
-    // as function of error in matrix coefficients, i. e. D(m_ij) for i, j from {x, y, z}.
-    // Error in constant is zero, i. e. D(dx) = D(dy) = D(z0) = 0.
-    // In the following calculation errors are considered small
-    // and second- and higher-order terms are dropped.
-    // That approximation is valid as long as glyph dimensions are larger than couple of pixels.
-    // Therefore standard relations for derivatives can be used for D(?):
-    // D(A * B) <= D(A) * max|B| + max|A| * D(B),
-    // D(1 / C) <= D(C) * max|1 / C^2|.
+    // Let D(f) denote the absolute error of a quantity f.
+    // Our goal is to determine tolerable error for matrix coefficients,
+    // so that the total error of the output x_out, y_out is still acceptable.
+    // As glyph dimensions are usually larger than a couple of pixels, errors
+    // will be relatively small and we can use first order approximation.
+
+    // z0 is effectively a scale factor and can thus be treated as a constant.
+    // Error of constants is obviously zero, so:  D(dx) = D(dy) = D(z0) = 0.
+    // For arbitrary quantities A, B, C with C not zero, the following holds true:
+    //   D(A * B) <= D(A) * max|B| + max|A| * D(B),
+    //   D(1 / C) <= D(C) * max|1 / C^2|.
+    // Write ~ for 'same magnitude' and ~= for 'approximately'.
 
     // D(x_out) = D((m_xx * x + m_xy * y) / z)
     //  <= D(m_xx * x + m_xy * y) * max|1 / z| + max|m_xx * x + m_xy * y| * D(1 / z)
@@ -534,14 +537,14 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
     // D(y_out) <= (D(m_yx) * dx + D(m_yy) * dy) / z0
     //       + 2 * (D(m_zx) * dx + D(m_zy) * dy) * y_lim / z0^2.
 
-    // To estimate acceptable error in matrix coefficient
-    // set error in all other coefficients to zero and solve system
-    // D(x_out) <= ACCURACY & D(y_out) <= ACCURACY for desired D(m_ij).
-    // ACCURACY here is some part of total error, i. e. ACCURACY ~ POSITION_PRECISION.
-    // Note that POSITION_PRECISION isn't total error, it's convenient constant.
-    // True error can be up to several POSITION_PRECISION.
+    // To estimate acceptable error in a matrix coefficient, pick ACCURACY for this substep,
+    // set error in all other coefficients to zero and solve the system
+    // D(x_out) <= ACCURACY, D(y_out) <= ACCURACY for desired D(m_ij).
+    // Note that ACCURACY isn't equal to total error.
+    // Total error is larger than each ACCURACY, but still of the same magnitude.
+    // Via our choice of ACCURACY, we get a total error of up to several POSITION_PRECISION.
 
-    // Quantization steps (ACCURACY ~ POSITION_PRECISION):
+    // Quantization steps (pick: ACCURARY = POSITION_PRECISION):
     // D(m_xx), D(m_yx) ~ q_x = POSITION_PRECISION * z0 / dx,
     // D(m_xy), D(m_yy) ~ q_y = POSITION_PRECISION * z0 / dy,
     // qm_xx = round(m_xx / q_x), qm_xy = round(m_xy / q_y),
@@ -565,7 +568,7 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
     // max(x_lim, y_lim) / z0 ~= w
     //  = max(|qm_xx| + |qm_xy|, |qm_yx| + |qm_yy|) * POSITION_PRECISION.
 
-    // Quantization steps (ACCURACY ~ 2 * POSITION_PRECISION):
+    // Quantization steps (pick: ACCURACY = 2 * POSITION_PRECISION):
     // D(m_zx) ~ POSITION_PRECISION * z0^2 / max(x_lim, y_lim) / dx ~= q_zx = q_x / w,
     // D(m_zy) ~ POSITION_PRECISION * z0^2 / max(x_lim, y_lim) / dy ~= q_zy = q_y / w,
     // qm_zx = round(m_zx / q_zx), qm_zy = round(m_zy / q_zy).
@@ -1412,7 +1415,7 @@ get_bitmap_glyph(ASS_Renderer *render_priv, GlyphInfo *info,
         //  <= (|m_yx| / z0 + |m_zx| * y_lim / z0^2) * D(x)
         //   + (|m_yy| / z0 + |m_zy| * y_lim / z0^2) * D(y).
 
-        // Quantization steps (ACCURACY ~ POSITION_PRECISION):
+        // Quantization steps (pick: ACCURACY = POSITION_PRECISION):
         // STROKER_PRECISION / 2^scale_ord_x ~ D(x) ~ POSITION_PRECISION /
         //   (max(|m_xx|, |m_yx|) / z0 + |m_zx| * max(x_lim, y_lim) / z0^2),
         // STROKER_PRECISION / 2^scale_ord_y ~ D(y) ~ POSITION_PRECISION /
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1246,9 +1246,9 @@ size_t ass_outline_construct(void *key, void *value, void *priv)
 static void calc_transform_matrix(ASS_Renderer *render_priv,
                                   GlyphInfo *info, double m[3][3])
 {
-    double frx = M_PI / 180 * info->frx;
-    double fry = M_PI / 180 * info->fry;
-    double frz = M_PI / 180 * info->frz;
+    double frx = ASS_PI / 180 * info->frx;
+    double fry = ASS_PI / 180 * info->fry;
+    double frz = ASS_PI / 180 * info->frz;
 
     double sx = -sin(frx), cx = cos(frx);
     double sy =  sin(fry), cy = cos(fry);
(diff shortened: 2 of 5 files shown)


libass/ass_render.c
```diff
@@ -29,6 +29,7 @@
 #include "ass_parse.h"
 #include "ass_priv.h"
 #include "ass_shaper.h"
+#include "ass_utils.h"
```
MrSmile, line 32: I'm not a fan of excessive includes. `ass_render.c` already heavily depends on `ass_utils.h`, so adding one constant changes nothing.

libass/ass_render.c
```diff
@@ -508,7 +508,7 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
     //  = m_zx * (x + sign(m_zx) * dx) + m_zy * (y + sign(m_zy) * dy) + z0.
 
     // D(f)--absolute value of error in quantity f
-    // as function of error in matrix coefficients, i. e. D(m_??).
+    // as function of error in matrix coefficients, i. e. D(m_kl).
```
MrSmile, line 511: I'd prefer symbols instead of letters here (`m_**`, `m_##`, `m_@@` or `m_$$`). It's not good to introduce new identifiers without definition, especially if that definition isn't quite trivial: `k, l = {x, y, z}`.

MrSmile (commented):
Added a couple of comments.

astiob (approved):
LGTM.

(Also, I think you’ll enjoy https://sourceforge.net/p/guliverkli2/code/HEAD/tree/src/subtitles/RTS.cpp#l148)

libass/ass_compat.h
```diff
@@ -26,4 +26,15 @@
 #define inline __inline
 #endif
 
+#ifndef HAVE_STRDUP
+char *ass_strdup_fallback(const char *s); //definition in ass_utils.c
```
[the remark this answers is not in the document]
TheOneric: Added spaces, thx.
If with “inplace inline” you meant `static inline` function definitions inside `ass_compat.h`, I decided against that, because it results in identical two functions per source file being placed in the resulting lib *(if fallbacks are required)*. Also the "proper" `str(n)dup` functions aren't forced inline either; leaving this to the compiler and possibly LTO seems better overall imo.

libass/ass_render.c
TheOneric: I now see `ass_render.h` already includes `ass_utils.h`, which is definitely **not** where I'd expect this to be, as nothing in the header itself requires `ass_utils.h`.
Imo, a header file should only include the bare minimum it requires for the header to be usable, so if it eg declares a function returning `size_t`, it should include `<stddef.h>`, but *not* things only the associated source file needs. Each source file should include everything it uses directly, except for types etc inherited from its own header file.
This has the advantage of making implicit dependencies sneaking in less likely. If a header file removes an include, because its own source file no longer needs it, with implicit dependencies now suddenly *other* source files might break, because they relied on that implicit/transitive include. Also putting the includes for the *source* file in the *header* file, means a lot of unneeded garbage gets included in other source files including the header, which also results in unnecessary recompiles when *any* header gets edited.
Currently it seems libass' includes are just placed wherever, or is there some system to it?

… But this pr is not about cleaning up includes, so I just removed `ass_utils.h` from `ass_render.c` for now.

libass/ass_render.c
TheOneric: I find it harder to understand those two identical non-alphabetic symbols¹, instead of two different labels. Personally I'd assume any unbound label to be a placeholder for any valid value, especially for a common enumerators like`i, j, k, l`, but this might also be a non-universal convention like the `**, @@, …`.
Perhaps it's best to just add `for all k, l from {x, y, z}` to the first/all use(s)?

¹: In fact while context made the intent clear in the original `??` case, I thought the `??` would be used as shell globs, where `?` matches any single character.

astiob: Just to add another perspective:

> Personally I'd assume any unbound label to be a placeholder for any valid value, especially for a common enumerators like`i, j, k, l`

Personally, I’m not used to seeing enumerators go as far as ℓ (and k is borderline), so I’d be confused by a new, unexplained mention of something involving it. Even the new version with the explainer comment feels kinda weird to me.

I was going to say that I find it an improvement that this now explicitly describes exactly which index combinations are intended… but is that actually right here? meaningful? After another few glances at the surrounding text, this one sentence seems to mean to say:

> Let’s write D(thing) for the absolute error in the thing’s value that we’ll get after all transformations are done and reverted, relative to the value the thing would have in an ideal rendering. We will now find reduced expressions for D(various things) that involve only various combinations of D(any element of matrix m) and no D(other things).

which

1. isn’t immediately clear from the sentence at all, whether before or after this PR;
2. doesn’t necessary involve “all” combinations of x, y, z.

libass/ass_render.c
MrSmile: > I now see `ass_render.h` already includes `ass_utils.h`

It's also indirectly included dozen times through `ass_outline.h`.

> Currently it seems libass' includes are just placed wherever, or is there some system to it?

I've tried to minimize include count in [this PR](https://github.com/libass/libass/pull/430) but hadn't seen much enthusiasm about it.

libass/ass_render.c
MrSmile: > Personally, I’m not used to seeing enumerators go as far as ℓ

In real math texts there is always shortage of letters, so I've seen _ijklmn_ and even _pqrts_ as indices.
But in this particular case `m_ij` is indeed better than `m_kl` if we go with letters.

TheOneric: Now I've reworded the whole introductory paragraph, is this easier to understand and still correct?
@MrSmile: Why is `z0` a constant even though it depends on matrix coefficients? Also I don't understand what `ACCURACY` in the second touched paragraph is or why it is bit-anded with `D(y_out)`.

MrSmile: I think that rewording shouldn't be done in trigraph commit.

> `i, j` are already used as variables by the function.

It's unrelated context. Also mathematicians often reuse the same letters as indices (for example, index _i_ and imaginary unit _i_ in the same expression simultaneously).

> Why is `z0` a constant even though it depends on matrix coefficients?

It depends on initial non-quantized matrix coefficients. Also it's an arbitrary scale factor that cancels out later. So there are two equivalent approaches: divide all coefficients by `z0` first and consider `z0 = 1` later, or treat `z0` as constant.

> Also I don't understand what `ACCURACY` in the second touched paragraph is or why it is bit-anded with `D(y_out)`.

`ACCURACY` is target error, and all that text is in pseudo-math notation, where `&` is simply "and" and `~` is "have the same order as".

TheOneric: > I think that rewording shouldn't be done in trigraph commit.

Will split.

> > `i, j` are already used as variables by the function.
> 
> It's unrelated context. Also mathematicians often reuse the same letters as indices (for example, index _i_ and imaginary unit _i_ in the same expression simultaneously).

I've certainly seen that double use of `i` in the same expression before, but I always held the opinion, and still do, that this is poor style and should be avoided. In this case reusing `i, j` wouldn't be as bad, but we're not short on letters either, so using `k, l` is better imo and I don't see any advantage in using `i, j` instead.

> `ACCURACY` is target error, and all that text is in pseudo-math notation, where `&` is simply "and" and `~` is "have the same order as".

I'd rather use a comma instead of `&` and `≈` *(or just write out `same magnitude` if you want to stick to ASCII)* instead of `~`, which I *primarily* know as “proportional to”.

TheOneric: Later on you're also using `~=`, I assume this is supposed to be an ASCII-fallback for `≈`?

MrSmile: Actually, `i, j` have used in the code in exactly the same context: as vector/matrix indices. So reusing them is even better.

> which I _primarily_ know as “proportional to”.

It's not wrong too. Basically by `x ~ y` I mean `x ≈ C * y` with some constant not far from 1.

> Later on you're also using `~=`, I assume this is supposed to be an ASCII-fallback for `≈`?

Yea. Maybe we should decide that source encoding is strictly UTF-8 and permit special symbols? Math descriptions would be much better with them.

TheOneric: > Actually, i, j have used in the code in exactly the same context

It's used to enumerate various things, not only the matrix.

@MrSmile , @astiob : How's the new version?

libass/ass_render.c
```diff
@@ -530,14 +535,14 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
...
     //       + 2 * (D(m_zx) * dx + D(m_zy) * dy) * y_lim / z0^2.
 
-    // To estimate acceptable error in matrix coefficient
+    // To estimate acceptable error in matrix coefficient,
     // set error in all other coefficients to zero and solve system
-    // D(x_out) <= ACCURACY & D(y_out) <= ACCURACY for desired D(m_??).
-    // ACCURACY here is some part of total error, i. e. ACCURACY ~ POSITION_PRECISION.
-    // Note that POSITION_PRECISION isn't total error, it's convenient constant.
+    // D(x_out) <= ACCURACY, D(y_out) <= ACCURACY for desired D(m_kl).
+    // Note ACCURACY is of the same magnitude as POSITION_PRECISION (write as: ~)
```
astiob, line 541: English nit: “Note [A is B]” feels weird to me: I’d expect either “Note that A is B” or “Note: A is B” with a colon. Especially with “and that” following it, I think that leaves “Note that” as the only option.

Then again, I’m not actually a native speaker myself (so I may be wrong), and this paragraph has other (pre-existing) slight language problems (such as missing articles), and yet the meaning remains clear, so this isn’t super important.

astiob (approved):
I may have more questions about the mathematical _content_ of those paragraphs, but I accepted them when we first merged this (and I know I went through all the maths) and that’s not relevant to this PR. So I’ll refrain from any specific comments.

That leaves me with a few English-language nits, which I sincerely hope you don’t mind 😅 and other than that, it still looks good to me.

libass/ass_render.c
TheOneric: Not sure if it's necessary, but “Note *that* …” should be correct *(but in that case the second *that* seems superfluous)*; changed that and also added articles.

libass/ass_render.c
```diff
@@ -506,15 +506,20 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
...
+    // Our goal is to determine tolerable error for matrix coefficients,
+    // so that the error of the output x_out, y_out is still acceptable.
+    // As glyph dimensions are usually larger than a couple of pixels, errors
+    // will be relatively small and we can use first order approximation.
+    //
+    // z0 is effectively a scale factor and can thus be treated as a constant.
+    // The error of constants is obviously zero, so:
+    //   D(dx) = D(dy) = D(z0) = 0.
+    // We can also utilise the fact, that for arbitrary quantities A, B, C
+    // with C not zero the following holds true:
```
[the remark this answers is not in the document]
TheOneric: I often get confused with English punctuation; corrections are always welcome *(and are now applied, thanks)*.

libass/ass_render.c
```diff
@@ -553,17 +558,17 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
         }
 
     // x_lim = |m_xx| * dx + |m_xy| * dy
-    //  ~= |qm_xx| * q_x * dx + |qm_xy| * q_y * dy
+    //  ≈ |qm_xx| * q_x * dx + |qm_xy| * q_y * dy
```
MrSmile, line 561: If we permit unicode in source code then `<=` and `^2` should be replaced with `≤` and `²`, also better to use other symbols (`×`, `±`, `∝`, `≡`, `⋅`, `Δ`, `mᵢⱼ`, etc.) too.

libass/ass_render.c
```diff
@@ -553,17 +558,17 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
...
+    //  ≈ |qm_yx| * q_x * dx + |qm_yy| * q_y * dy
     //  = (|qm_yx| + |qm_yy|) * POSITION_PRECISION * z0,
-    // max(x_lim, y_lim) / z0 ~= w
+    // max(x_lim, y_lim) / z0 ≈ w
     //  = max(|qm_xx| + |qm_xy|, |qm_yx| + |qm_yy|) * POSITION_PRECISION.
 
-    // Quantization steps (ACCURACY ~ 2 * POSITION_PRECISION):
-    // D(m_zx) ~ POSITION_PRECISION * z0^2 / max(x_lim, y_lim) / dx ~= q_zx = q_x / w,
-    // D(m_zy) ~ POSITION_PRECISION * z0^2 / max(x_lim, y_lim) / dy ~= q_zy = q_y / w,
+    // Quantization steps (reminder: ACCURACY ~ 2 * POSITION_PRECISION):
```
MrSmile, line 569: Similarly, following expressions derived by substituting `ACCURACY` with `2 * POSITION_PRECISION`, it's not a reminder.

MrSmile (commented):
Not sure what consensus about unicode in source is. Also `≈` is quite similar to `=`, contrary to `~=`, not sure if that is a problem.

Also I really think that better to use `m_ij` due to
- there is at least one person confused by `kl`;
- math notation uses `ij` with priority and collisions outside of index context don't matter;
- variables `i` and `j` have extensively used in relevant functions as index variables and there are no other conflicting uses.

libass/ass_utils.h
```diff
@@ -46,6 +46,8 @@
 #define FFMIN(a,b) ((a) > (b) ? (b) : (a))
 #define FFMINMAX(c,a,b) FFMIN(FFMAX(c, a), b)
 
+#define ASS_PI 3.14159265358979323846 /* pi */
```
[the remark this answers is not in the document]
TheOneric: The comment was copied from glibc's and musl's definition; now removed.

TheOneric:
After going through this a bit more on IRC, I've now re-pushed a version of the rewording which should address everything from above, only has +3 lines and is imho easier to understand than the original version.

libass/ass_render.c
```diff
@@ -530,14 +533,14 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
...
-    // True error can be up to several POSITION_PRECISION.
+    // To estimate acceptable error in a matrix coefficient, pick ACCURACY for this substep,
+    // set error in all other coefficients to zero and solve the system
+    // D(x_out) <= ACCURACY, D(y_out) <= ACCURACY for desired D(m_ij).
+    // Note that ACCURACY isn't equal to total error.
+    // Total error is larger than each ACCURACY, but still of the same magnitude.
+    // Via our choice of ACCURACY, we get a total error of up to several POSITION_PRECISION.
 
-    // Quantization steps (ACCURACY ~ POSITION_PRECISION):
+    // Quantization steps (pick ACCURARY = POSITION_PRECISION):
```
MrSmile, line 543: Later you've used `pick:`, should be uniform. Also, maybe `use` or `set` here will be better?

MrSmile (commented):
Mostly LGTM, marked a couple of typos.

libass/ass_render.c
astiob: “Pick” conveys that the parameter is arbitrarily chosen to have this value. “Use” and “set” are more ambiguous. Frankly, I (still) don’t quite understand how `ACCURACY` is chosen or what it really is, and why that choice is OK, so “pick” would make me feel a little less lost.

TheOneric:
Will merge this in two days if there are no objections.

PR #480, opened 2021-03-01
bitmap: subtract fill from stroke without dividing

rcombs:
This is what vsfilter does, and looks better on most inputs.
I'm guessing we used to do it differently because it looked better before 12cf524b831289cf67ba3432264fa01a69c3bbb5 and 706f23d84180d36538c467e20b461677781cab16.

Test script; improvement most noticeable on the triangle drawing in the top-left:
[border_aliasing_with_1aFF.txt](https://github.com/libass/libass/files/6063280/border_aliasing_with_1aFF.txt)
VSFilter:
image
Before:
image
After:
image

This is a case multiple people have pointed out to me as problematic in libass. This _does_ seem to slightly worsen another related case, but that I'm not super worried about because it's comparatively uncommon and the difference isn't huge:
[border_aliasing_with_1aFF-2.txt](https://github.com/libass/libass/files/6063279/border_aliasing_with_1aFF-2.txt)
VSFilter:
image
Before:
image
After:
image

In the previous version, the aliasing here was extremely subtle (but present, and thus not matching vsfilter); with this patch it's a bit more noticeable. Somehow, vsfilter manages not to visibly alias on this edge at all; might have something to do with the way it does the actual drawing? Maybe a rounding thing, somehow? I'm not sure; definitely worth looking into.

### DIFF
--- a/libass/ass_bitmap.c
+++ b/libass/ass_bitmap.c
@@ -211,7 +211,7 @@ void fix_outline(Bitmap *bm_g, Bitmap *bm_o)
 
     for (int32_t y = 0; y < b - t; y++) {
         for (int32_t x = 0; x < r - l; x++)
-            o[x] = (o[x] > g[x]) ? o[x] - (g[x] / 2) : 0;
+            o[x] = (o[x] > g[x]) ? o[x] - g[x] : 0;
         g += bm_g->stride;
         o += bm_o->stride;
     }


petzku:
> This does seem to slightly worsen another related case, but that I'm not super worried about because it's comparatively uncommon and the difference isn't huge:

While I'm not terribly concerned about the latter case, I should note that one of the cases where it would come up (possibly a lot) is with fades.

> Somehow, vsfilter manages not to visibly alias on this edge at all; might have something to do with the way it does the actual drawing? Maybe a rounding thing, somehow? I'm not sure; definitely worth looking into.

This is possibly even more noticeable at very high alpha values, where vsfilter produces an entirely solid shape:
[border aliasing with 1aFF.ass.txt](https://github.com/libass/libass/files/6063622/border.aliasing.with.1aFF.ass.txt)

recent libass master (I believe 7a853255bb5083c9c6ab31a29f215875dc8d8251): 
libass

vsfilter (Cyberbeing/xy-VSFilter@fc01a8d)
vsfilter

For completeness, vsfiltermod (sorayuki/VSFilterMod@b0c62f9, R5.2.3)
image

VSFilterMod output is slightly different from xy-VSFilter, and seems to have very minor aliasing. The aliasing is much more apparent at `\1a1\3a1` (compare to images in original issue): 
image

astiob:
FYI, this code keeps getting switched back and forth when someone comes upon a different sample where one or the other looks better: this PR exactly reverts 0699a8640056535901a9aabec58e909ce7ae54ef and 37f063bfb45fe33239fe8569be4acf2f80bc7174, and see also d3cd5b0148d64ae0f6e288e69b09acfe639fdcce and 8407e392667a2f37eaca4fabfcf6bd7ca42aaed5.

I have an unfortunately stale, ancient branch that is meant to fix this for good: `alpha` in my personal fork on GitHub. The reason this branch never got merged was because it looked worse in some cases with karaoke, because it does not know which fill colour (primary or secondary) is used where, because this is decided in our karaoke machinery that sets to work only after the main rendering pipeline is done.

If we could push this decision into the main pipeline proper, before fills are subtracted from strokes, then my branch could be applied and we could finally put this issue to rest. I think this might be an opportunity also to make the karaoke code clearer and easier to maintain, as our current karaoke splitting machinery uses peculiar code and is a rather obscure & isolated corner of our codebase where we rarely look, so it might help if we could re-express the logic in terms of the more common mechanisms on the main pipeline.

astiob:
For reference, my sample ***on white background*** from https://code.google.com/archive/p/xy-vsfilter/issues/145:

```
[Script Info]
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
PlayResX: 106
PlayResY: 100
ScaledBorderAndShadow: yes
YCbCr Matrix: TV.709

[V4+ Styles]
Style: Default,Arial,60,&H00000000,&H01000000,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,5,0,2,10,10,20,1

[Events]
Dialogue: 0,0:00:00.00,0:00:10.00,Default,,0,0,0,,test
```

or equivalently just: `{\fs60\bord5\c&H000000&\2a&H01&\3c&H000000&}test`

xy-VSFilter and libass with this PR (bad):

mpv-shot0002

libass master (also bad but less obvious):

mpv-shot0003

libass with #483 (good):

mpv-shot0001

rcombs:
This is superseded by #521.

rcombs:
Wait, no it's not, I'm just mixed up.

rcombs:
…it is, however, superseded by #483.

PR #481, opened 2021-03-03
ass_render: Break at last position if no space was seen

cdown:
It's pretty common (in Chinese/Japanese/Korean especially) to not have any spaces to split on. Currently, these cases simply overflow the render area.

If this happens, try a last ditch attempt to wrap at the last character seen. This dramatically improves the usability of long subtitles with no clear break point.

Compared to https://github.com/libass/libass/pull/372, this pull request is much simpler (but complementary): we just split in any case where we will overflow the buffer as a last resort. Cases like https://github.com/mpv-player/mpv/issues/8115 demonstrate demand for this.

Example previous issues which show demand for this behaviour:

- https://github.com/mpv-player/mpv/issues/8115
- https://github.com/mpv-player/mpv/issues/2203
- https://github.com/libass/libass/issues/348
- https://github.com/libass/libass/issues/360

This PR also contains a change to how we do feature enumeration, in order to better support ASS_FEATURE_INCOMPATIBLE_EXTENSIONS as the feature list grows.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -1524,7 +1524,7 @@ static const uint64_t supported_features =
 #ifdef USE_FRIBIDI_EX_API
     1LL << ASS_FEATURE_BIDI_BRACKETS |
 #endif
-    0;
+    1LL << ASS_FEATURE_FORCE_WRAP_ON_OVERFLOW;
 
 int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable)
 {
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -1520,19 +1520,43 @@ fail:
     return NULL;
 }
 
+static const uint64_t supported_features =
+#ifdef USE_FRIBIDI_EX_API
+    1LL << ASS_FEATURE_BIDI_BRACKETS |
+#endif
+    0;
+
 int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable)
 {
-    switch (feature) {
-    case ASS_FEATURE_INCOMPATIBLE_EXTENSIONS:
-        //-fallthrough
-#ifdef USE_FRIBIDI_EX_API
-    case ASS_FEATURE_BIDI_BRACKETS:
-        track->parser_priv->bidi_brackets = !!enable;
-#endif
-        return 0;
-    default:
+    uint64_t requested;
+
+    if (feature >= 64)
         return -1;
-    }
+
+    requested = 1LL << feature;
+
+    if (feature == ASS_FEATURE_INCOMPATIBLE_EXTENSIONS)
+        requested = supported_features;
+    else if (!(requested & supported_features))
+        return -1;
+
+    if (enable)
+        track->parser_priv->features |= requested;
+    else
+        track->parser_priv->features &= ~requested;
+
+    return 0;
+}
+
+int ass_track_get_feature_state(ASS_Track *track, ASS_Feature feature)
+{
+    if (feature >= 64)
+        return -1;
+
+    if (feature == ASS_FEATURE_INCOMPATIBLE_EXTENSIONS)
+        return track->parser_priv->features == supported_features;
+
+    return !!(track->parser_priv->features & (1LL << feature));
 }
 
 /**
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -572,6 +572,15 @@ ASS_Track *ass_new_track(ASS_Library *);
 int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable);
 
 /**
+ * \brief Check if a feature is enabled
+ * \param track track
+ * \param feature the specific feature to check the state of
+ * \return 0 if disabled, a positive integer with feature-dependent meaning if
+ * enabled, -1 if feature is unknown
+ */
+int ass_track_get_feature_state(ASS_Track *track, ASS_Feature feature);
+
+/**
  * \brief Deallocate track and all its child objects (styles and events).
  * \param track track to deallocate or NULL
  */
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -24,7 +24,7 @@
 #include <stdarg.h>
 #include "ass_types.h"
 
-#define LIBASS_VERSION 0x01500000
+#define LIBASS_VERSION 0x01500001
 
 #ifdef __cplusplus
 extern "C" {
@@ -237,6 +237,14 @@ typedef enum {
      * may return -1) if libass was compiled against old FriBidi.
      */
     ASS_FEATURE_BIDI_BRACKETS,
+
+    /**
+     * If no explicit ("\n") or implicit (" ") wrap point is found, wrap at the
+     * last seen character when doing smart wrap.
+     *
+     * Has no effect if wrap_mode is 2.
+     */
+    ASS_FEATURE_FORCE_WRAP_ON_OVERFLOW,
 
     // New enum values can be added here in new ABI-compatible library releases.
 } ASS_Feature;
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2858,7 +2858,7 @@ ass_start_frame(ASS_Renderer *render_priv, ASS_Track *track,
     ass_shaper_set_level(render_priv->shaper, render_priv->settings.shaper);
 #ifdef USE_FRIBIDI_EX_API
     ass_shaper_set_bidi_brackets(render_priv->shaper,
-            track->parser_priv->bidi_brackets);
+            ass_track_get_feature_state(track, ASS_FEATURE_BIDI_BRACKETS));
 #endif
 
     // PAR correction
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1632,7 +1632,8 @@ static void trim_whitespace(ASS_Renderer *render_priv)
  * FIXME: implement style 0 and 3 correctly
  */
 static void
-wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
+wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width,
+                 bool always_wrap)
 {
     int i;
     GlyphInfo *cur, *s1, *e1, *s2, *s3;
@@ -1665,7 +1666,7 @@ wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
         } else if (len >= max_text_width
                    && (render_priv->state.wrap_style != 2)) {
             break_type = 1;
-            break_at = last_space;
+            break_at = (last_space == -1 && always_wrap) ? i - 1 : last_space;
             if (break_at >= 0)
                 ass_msg(render_priv->library, MSGL_DBG2, "line break at %d",
                         break_at);
@@ -2600,7 +2601,7 @@ static void add_background(ASS_Renderer *render_priv, EventImages *event_images)
  * Process event, appending resulting ASS_Image's to images_root.
  */
 static bool
-ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
+ass_render_event(ASS_Track *track, ASS_Renderer *render_priv, ASS_Event *event,
                  EventImages *event_images)
 {
     if (event->Style >= render_priv->track->n_styles) {
@@ -2657,7 +2658,8 @@ ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
         x2scr_left(render_priv, MarginL);
 
     // wrap lines
-    wrap_lines_smart(render_priv, max_text_width);
+    wrap_lines_smart(render_priv, max_text_width,
+                     ass_track_get_feature_state(track, ASS_FEATURE_FORCE_WRAP_ON_OVERFLOW));
 
     // depends on glyph x coordinates being monotonous within runs, so it should be done before reorder
     process_karaoke_effects(render_priv);
@@ -3157,7 +3159,7 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
                     realloc(priv->eimg,
                             priv->eimg_size * sizeof(EventImages));
             }
-            if (ass_render_event(priv, event, priv->eimg + cnt))
+            if (ass_render_event(track, priv, event, priv->eimg + cnt))
                 cnt++;
         }
     }
--- a/libass/ass_priv.h
+++ b/libass/ass_priv.h
@@ -61,9 +61,8 @@ struct parser_priv {
     // tracks [Script Info] headers set by the script
     uint32_t header_flags;
 
-#ifdef USE_FRIBIDI_EX_API
-    bool bidi_brackets;
-#endif
+    // Enabled ASS_Feature flags
+    uint32_t features;
 };
 
 #endif /* LIBASS_PRIV_H */
--- a/libass/libass.sym
+++ b/libass/libass.sym
@@ -44,3 +44,4 @@ ass_set_selective_style_override_enabled
 ass_set_selective_style_override
 ass_set_check_readorder
 ass_track_set_feature
+ass_track_get_feature_state


cdown:
Before:



After:

astiob:
To clarify a few things:

* This is _complementary_ to #372 and not mutually exclusive. What #372 actually does is apply Unicode line breaking, which would e. g. break the line after the last slash (/) in https://github.com/mpv-player/mpv/issues/8115. And conversely, what it _doesn’t_ do is break overfull lines that Unicode doesn’t allow breaking (e. g. a sequence of very many Latin letters), which is a void that this PR fills.
* Just as #372, I’m inclined to say this should be hidden behind an `ASS_Feature` flag and disabled for ASS subtitles (to be explicitly enabled by the media player for OSD and subtitles converted from other formats) to preserve compatibility.

cdown:
Thanks for the fast feedback!

> This is complementary to #372 and not mutually exclusive.

Sorry, what I originally wrote wasn't clear: yes, it's complementary. :-)

> I’m inclined to say this should be hidden behind an ASS_Feature flag and disabled for ASS subtitles (to be explicitly enabled by the media player for OSD and subtitles converted from other formats) to preserve compatibility.

Sure thing, can do.

cdown:
Ok, put it behind an `ASS_Feature` and tested in mpv.

Not sure if `ass_track` itself is the right place to store the state or not, let me know if you'd like it somewhere else. :-)

libass/ass.c
```diff
@@ -1528,11 +1528,16 @@ int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable)
 #ifdef USE_FRIBIDI_EX_API
     case ASS_FEATURE_BIDI_BRACKETS:
         track->parser_priv->bidi_brackets = !!enable;
+        break;
```
cdown, line 1531: This may or may not be the right thing -- I'm not entirely sure what the expected behaviour is here currently: is `ASS_FEATURE_INCOMPATIBLE_EXTENSIONS` expected to enable all incompatible features using fallthrough?

cdown: (and if so, one supposes the strategy needs some adjusting, since one presumably has to also be able to individually select features, which fallthrough would preclude)

astiob: It is just expected to enable all incompatible features. The fallthrough presumably has to be replaced by a list (and feel free to smack me for micro-optimizing using a fallthrough in the first place). Compare with: https://github.com/libass/libass/pull/441/files.

cdown: Sounds good, would you like that in a separate commit or in the same one?

cdown:
For now, pushed as two separate commits: the first one changes the way we implement ass_track_set_feature, and the second is now rebased to use that method.

libass/ass.c
```diff
@@ -1522,17 +1522,28 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
+        [ASS_FEATURE_INCOMPATIBLE_EXTENSIONS] = NULL,
 #ifdef USE_FRIBIDI_EX_API
-    case ASS_FEATURE_BIDI_BRACKETS:
-        track->parser_priv->bidi_brackets = !!enable;
+        [ASS_FEATURE_BIDI_BRACKETS] = &track->parser_priv->bidi_brackets,
 #endif
+        [ASS_FEATURE_ALWAYS_WRAP] = &track->always_wrap,
+    };
+
+    if (feature == ASS_FEATURE_INCOMPATIBLE_EXTENSIONS) {
```
cdown, line 1533: Oh wait, this won't work if `!USE_FRIBIDI_EX_API`, because it still takes up a space in the enum but isn't used. Hmm, let me think about a better way.

libass/ass.c
```diff
@@ -1522,17 +1522,29 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
+                i < sizeof(features) / sizeof(features[0]); i++) {
+            if (features[i])
+                *features[i] = !!enable;
+        }
         return 0;
-    default:
-        return -1;
     }
+
+    if (feature < 0 || feature > (sizeof(features) / sizeof(features[0])) || !features[feature])
```
astiob, line 1543: The middle check should say `>=`.

And you may or may not want to say `< ASS_FEATURE_INCOMPATIBLE_EXTENSIONS` instead of `< 0` to more obviously match the loop above.

libass/ass.c
```diff
@@ -1522,17 +1522,29 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
-    switch (feature) {
-    case ASS_FEATURE_INCOMPATIBLE_EXTENSIONS:
-        //-fallthrough
+    bool *features[] = {
 #ifdef USE_FRIBIDI_EX_API
-    case ASS_FEATURE_BIDI_BRACKETS:
-        track->parser_priv->bidi_brackets = !!enable;
+        [ASS_FEATURE_BIDI_BRACKETS] = &track->parser_priv->bidi_brackets,
+#else
+        [ASS_FEATURE_BIDI_BRACKETS] = NULL,
```
astiob, line 1529: Hmm, can’t this `#else` clause be omitted entirely? Not in the first commit because the array needs at least one element? But it would seem sensible to omit such clauses entirely in the future, as soon as at least one always-present element is added (which happens in the second commit).

libass/ass.c
```diff
@@ -1528,6 +1528,7 @@ int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable)
 #else
         [ASS_FEATURE_BIDI_BRACKETS] = NULL,
 #endif
+        [ASS_FEATURE_ALWAYS_WRAP] = &track->always_wrap,
```
astiob, line 1531: I think we’re currently preferring to put new fields in `track->parser_priv`, like the `bidi_brackets` above, to minimize ABI breakage risk now and in the future (if we ever decide to change the Feature), since `ASS_Track` is unfortunately public, even if we’ve added a comment that new fields may be appended.

astiob (changes requested):
Interesting approach to the implementation of `ass_track_set_feature`.

Please give `LIBASS_VERSION` a slight bump to signal the presence of a new `ASS_Feature` enum constant.

Further, I’d like to request that “which is CJK-specific” be removed from the commit message, because it isn’t. The Unicode line breaking algorithm applies even to English, allowing line breaks after punctuation (which we otherwise don’t do).

astiob:
We’re also discussing the `ASS_Feature` stuff on IRC at the moment (feel free to join in at #libass on Freenode), so perhaps don’t rush to amend this quite yet as we may just end up deciding we want an entire different implementation (a bit-mask has been mentioned).

libass/ass.c
cdown: Currently no, because ASS_FEATURE_BIDI_BRACKETS takes a space in the enum and will be iterated.

cdown:
> We’re also discussing the ASS_Feature stuff on IRC at the moment (feel free to join in at #libass on Freenode), so perhaps don’t rush to amend this quite yet as we may just end up deciding we want an entire different implementation (a bit-mask has been mentioned).

Sure thing -- was there any conclusion? More than happy to go with a bitmask or whatever solution you folks prefer, just let me know.

libass/ass.c
cdown: Sure thing. I wasn't quite sure, since it seemed like parser_priv seemed off (since this is rendering rather than parser), but it's certainly possible to do so. :-)

astiob: I do agree about the parsing vs rendering name! I think we might want to rename it at some point, or split, or something. But for now that’s where we’re putting these things.

libass/ass.c
astiob: Sure, but won’t the remainder of the array (any elements that aren’t explicitly designated) be zero-initialized?

cdown:
Okay, updated the feature code as per discussion on IRC. Also bumped LIBASS_VERSION and actioned all feedback above, I think :-)

libass/ass.h
```diff
@@ -223,7 +223,7 @@ typedef enum {
      * purposes that do not involve actual ASS subtitles authored for
      * distribution.
      */
-    ASS_FEATURE_INCOMPATIBLE_EXTENSIONS,
+    ASS_FEATURE_INCOMPATIBLE_EXTENSIONS = 1,
```
TheOneric, line 226: You must not change public enum values, this breaks ABI.
So `0` and `1 = 1 << 0`, which fortunately works well with a bit mask.

libass/ass.h
```diff
@@ -238,6 +238,14 @@ typedef enum {
      */
     ASS_FEATURE_BIDI_BRACKETS = 1 << 1,
 
+    /**
+     * If no explicit ("\n") or implicit (" ") wrap point is found, wrap at the
+     * last seen character when doing smart wrap.
+     *
+     * Has no effect if wrap_mode is 2.
+     */
+    ASS_FEATURE_ALWAYS_WRAP = 1 << 2,
```
TheOneric, line 247: I think a differnt name would be better. Eg  `WRAP_BREAKWORDS` or a bit longer but also clearer `WRAP_FORCE_ON_OVERFLOW`.

libass/ass.h
astiob: Yes, and to expand a bit, _this_ (public) enum is not expected to be bit-flags. It’s just an enum. The bit stuff should remain an encapsulated implementation detail.

astiob: So we can either use `1 << feature` in the bit-mask ([example](https://gist.github.com/MrSmile/ef9c0118163e46a0ba8e7a5529d6ead1/bcfca1a4e724acb4aa1911ff8406bcedc0b59630), but it needs an overflow check for the shift) or create a separate (internal) enum with bit-flags ([example](https://gist.github.com/MrSmile/ef9c0118163e46a0ba8e7a5529d6ead1/0ab3bb1ff8af4abdaef52d79f87c9989f4d35411)).

cdown:
Ready for another round of review. :-)

libass/ass.h
```diff
@@ -571,6 +579,15 @@ ASS_Track *ass_new_track(ASS_Library *);
  */
 int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable);
 
+/**
+ * \brief Check if a feature is enabled
+ * \param track track
+ * \param feature the specific feature to check the state of
+ * \return 0 if disabled, a positive integer with feature-dependent meaning if
+ * enabled, -1 if feature is unknown
```
cdown, line 587: "[P]ositive integer" left intentionally vague since I remember you folks mentioned you may want to have non-boolean states in future.

libass/ass.c
```diff
@@ -1520,19 +1520,43 @@ ASS_Track *ass_new_track(ASS_Library *library)
     return NULL;
 }
 
-int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable)
-{
-    switch (feature) {
-    case ASS_FEATURE_INCOMPATIBLE_EXTENSIONS:
-        //-fallthrough
+static const uint64_t supported_features =
```
MrSmile, line 1523: Type is different from `parser_priv->features`. Better to make it the same.

libass/ass.c
```diff
@@ -1520,19 +1520,43 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
 #ifdef USE_FRIBIDI_EX_API
-    case ASS_FEATURE_BIDI_BRACKETS:
-        track->parser_priv->bidi_brackets = !!enable;
+    1LL << ASS_FEATURE_BIDI_BRACKETS |
```
MrSmile, line 1525: Not an issue, but what libass style guide says about literal suffixes? Currently we have lowercase unsigned suffix (10u) and uppercase long long suffix (60LL). I think we should make that uniform. Personally I prefer uppercase hex digits and lowercase suffixes to not mix them.

libass/ass.c
```diff
@@ -1520,19 +1520,43 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
 #endif
-        return 0;
-    default:
+    0;
+
+int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable)
+{
+    uint64_t requested;
```
MrSmile, line 1531: We're using C99, so it's better to combine variable definition with its first use. Also type inconsistency.

libass/ass.c
```diff
@@ -1520,19 +1520,43 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
+
+    if (enable)
+        track->parser_priv->features |= requested;
+    else
+        track->parser_priv->features &= ~requested;
+
+    return 0;
+}
+
+int ass_track_get_feature_state(ASS_Track *track, ASS_Feature feature)
```
MrSmile, line 1551: I think it's better to name it `ass_track_get_feature` without `state` for symmetry.

libass/ass.c
```diff
@@ -1520,19 +1520,43 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
+{
+    if (feature >= 64)
+        return -1;
+
+    if (feature == ASS_FEATURE_INCOMPATIBLE_EXTENSIONS)
+        return track->parser_priv->features == supported_features;
+
+    return !!(track->parser_priv->features & (1LL << feature));
```
MrSmile, line 1559: Missing `requested & supported_features` check. Should return &minus;1 in case of unsupported feature.

libass/ass.h
MrSmile: Also should be mentioned behavior in case of macro features like `ASS_FEATURE_INCOMPATIBLE_EXTENSIONS`.

libass/ass_render.c
```diff
@@ -2858,7 +2858,7 @@ ass_start_frame(ASS_Renderer *render_priv, ASS_Track *track,
     ass_shaper_set_level(render_priv->shaper, render_priv->settings.shaper);
 #ifdef USE_FRIBIDI_EX_API
     ass_shaper_set_bidi_brackets(render_priv->shaper,
-            track->parser_priv->bidi_brackets);
+            ass_track_get_feature_state(track, ASS_FEATURE_BIDI_BRACKETS));
```
MrSmile, line 2863: Better to add ` > 0` here, as &minus;1 is also a possible return value of this function. It doesn't matter in this particular place as it's already guarded by `#ifdef`, but it's better to have a correct sample for the future copy-paste.

libass/ass.h
```diff
@@ -24,7 +24,7 @@
 #include <stdarg.h>
 #include "ass_types.h"
 
-#define LIBASS_VERSION 0x01500000
+#define LIBASS_VERSION 0x01500001
```
MrSmile, line 27: Not sure which commit should have this bump. Previous commit also changed API.

libass/ass.h
```diff
@@ -238,6 +238,14 @@ typedef enum {
      */
     ASS_FEATURE_BIDI_BRACKETS,
 
+    /**
+     * If no explicit ("\n") or implicit (" ") wrap point is found, wrap at the
+     * last seen character when doing smart wrap.
+     *
+     * Has no effect if wrap_mode is 2.
+     */
+    ASS_FEATURE_FORCE_WRAP_ON_OVERFLOW,
```
MrSmile, line 247: I think `FORCE` is excess here, especially if we're planning to have a bunch of `WRAP` features.

libass/ass_render.c
```diff
@@ -1632,7 +1632,8 @@ static void trim_whitespace(ASS_Renderer *render_priv)
  * FIXME: implement style 0 and 3 correctly
  */
 static void
-wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width)
+wrap_lines_smart(ASS_Renderer *render_priv, double max_text_width,
+                 bool always_wrap)
```
MrSmile, line 1636: Better to synchronize this name with the feature name (`wrap_on_overflow` or something).

libass/ass_render.c
```diff
@@ -2600,7 +2601,7 @@ static void add_background(ASS_Renderer *render_priv, EventImages *event_images)
  * Process event, appending resulting ASS_Image's to images_root.
  */
 static bool
-ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
+ass_render_event(ASS_Track *track, ASS_Renderer *render_priv, ASS_Event *event,
```
MrSmile, line 2604: I think it's better to swap `track` and `render_priv` parameters to make it similar to `ass_render_frame()`.

libass/ass_render.c
```diff
@@ -2657,7 +2658,8 @@ ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
         x2scr_left(render_priv, MarginL);
 
     // wrap lines
-    wrap_lines_smart(render_priv, max_text_width);
+    wrap_lines_smart(render_priv, max_text_width,
+                     ass_track_get_feature_state(track, ASS_FEATURE_FORCE_WRAP_ON_OVERFLOW));
```
MrSmile, line 2662: Also ` > 0`.

MrSmile (changes requested):
Good overall, with a few nitpicks.

astiob (commented):
Oh boy, this really is a “Remake ASS_Feature” PR now. You just keep digging it deeper 😃 But here’s hoping that we’ll get an all-the-better product out of it!

I’m not _sure_ I’m a fan of the public getter function. On the one hand, it seems like a fairly obvious addition, but on the other, it kinda locks us into storing each feature’s state as a distinctly tangible object just so we can return it from the getter function. Of course, this is fine for our current `BIDI_BRACKETS` and `WRAP...` features, which are indeed independent Boolean toggles. But we already have `INCOMPATIBLE_EXTENSIONS` which doesn’t have its own state and instead affects other feature toggles, and we might theoretically get something else similar in the future? something that you can reset to a known macro-state (e. g. full-on or full-off) but which decomposes into a variety of micro-configuration (which might even be settable in other ways) that you can’t really read back and recompose into a well-defined macro-state?

cdown:
Thanks @MrSmile and @astiob for the feedback!

> we might theoretically get something else similar in the future? something that you can reset to a known macro-state (e. g. full-on or full-off) but which decomposes into a variety of micro-configuration (which might even be settable in other ways) that you can’t really read back and recompose into a well-defined macro-state?

Hmm, is there even a well defined meaning of "on" or "off" for such features other than indicating that the eventual state that would be caused by enabling it would or would not change the current state?

Another option is to have ASS_Feature be boolean toggles, and have a separate ASS_FeatureSet which somehow has different semantics, but I think that really just punts a lot of the confusion over to the user.

cdown:
Ah, or if we're worried about non-boolean, how about `ass_track_feature_state(ASS_Feature feature, int state)`? That way, for now we only have boolean state, but can extend internally if more come up later.

cdown:
Hey folks, any thoughts on the above suggestions? :-)

astiob:
@cdown In your current proposed API, what do you think `ass_track_get_feature_state(ASS_FEATURE_INCOMPATIBLE_EXTENSIONS)` should return? Especially if I do this:

```c
ass_track_set_feature(track, ASS_FEATURE_INCOMPATIBLE_EXTENSIONS, true);
ass_track_set_feature(track, ASS_FEATURE_WRAP_ON_OVERFLOW, false);
ass_track_get_feature_state(track, ASS_FEATURE_INCOMPATIBLE_EXTENSIONS)
```

---

@MrSmile (cc @rcombs @TheOneric) It seems we may need to decide on some limits on what we really want the feature API to do.

Is a feature always either just a Boolean toggle or a meta-set of of other features?

What is our _goal_ with the feature API as opposed to introducing separate setter functions for new features?

Right now, all our features are Booleans (or feature-sets), and the API allows a nifty kind of ABI compatibility whereby an application compiled against a newer libass with a newer feature can still run with an older libass at runtime—the feature will, of course, be disabled, but the application can detect this if it cares. And all the features currently under consideration (this, `WHOLE_TEXT`, Unicode line breaking) are also just Boolean toggles.

But what if we ever want a new feature that is more than just a Boolean toggle? Say we decide to add a VSFilter emulation mode _enum_ that would allow you to choose among xy-VSFilter, MPC-HC and old-libass (and perhaps many more detailed options). Do we want that to be an ASS_Feature with a non-Boolean value, or a whole new API?

MrSmile:
I'm neutral about that `ass_track_get_feature_state()` business. There are merits in having this function but potential ABI problems are there too. For boolean feature flags it can return something in the line of
```c
enum { NotSupported = -1, Off, Partial, On }
```
but in case of the future non-boolean features and especially combinations of such non-boolean features this function doesn't make sense.

libass/ass.h
TheOneric: I agree that `WRAP` should come first, but to bikeshed: without `FORCE` anywhere the name no longer is a good description of the feature, as except for `\q2` we always try to wrap on overflow; this feature just forces wrap even if no legal breakpoints were found.
 But to keep the name short, maybe `ASS_FEATURE_WRAP_BREAK_WORDS` is better after all?

libass/ass_render.c
TheOneric: The `smart` comes from ASS, where wrap modes `0` and `3` are referred to as "smart-wrapping". I think it's better to keep this name. Also when Unicode line-breaking gets merged, `on_overflow` would no longer be the only Feature to influence this function's behaviour.

libass/ass.h
TheOneric: If `ass_get_feature_state` remains, both commits should have a bump imho.

TheOneric (commented):
Something to consider regarding the "catch-all" `INCOMPATIBLE_EXTENSIONS` and the getter-function: the proposed “Strict Mode” from #78, right now seems to be primarily about disabling font fallback. If that's implemented as an `ASS_FEATURE`, do we really want font fallback to be disabled by `INCOMPATIBLE_EXTENSIONS`? Right now this "catch-all" seems to be advertised for non-ASS renders to get a more sane™ rendering, disabling font-fallback may however not be desireable for those renderers.
Also as “Strict Mode” is targeted at ASS sub authors, it may conceivably be extended to also disable libass' ASS extensions *(like `Border-Style: 4`, `Encoding: -1`, etc)*, which then raises the question what `get_state` should return if strict and some other feature is enabled?

On the other hand having several mode-APIs *(feaures, strictness-level, different bug-emul-target, ...)* could also be confusing, so I'm not sure what's better.

If we decide to only use `ASS_FEATURE` for simple booleans, and `INCOMPAT…`as the only multi-switch, then `get_state` is probably fine. If we're not yet sure how we'll use `ASS_FEATURES` it may be better to not (yet) add a getter function to avoid future ABI/API problems. Since it initialises with all features disabled API-users could always track the current feature states themselves if needed. *(But perhaps a "reset" function would be useful?)*

---

*Ugh, I forgot GitHub will display my replies to other review comments twice and once without any context, when also adding a "general" review comment*

libass/ass_render.c
astiob: I think the earlier comment was about the new parameter’s name, rather than the function’s name.

astiob:
> If that's implemented as an `ASS_FEATURE`, do we really want font fallback to be disabled by `INCOMPATIBLE_EXTENSIONS`?

Doing it as an `ASS_Feature` does seem sensible, and no, I imagine we wouldn’t want it activated by `INCOMPATIBLE_EXTENSIONS`.

> it may conceivably be extended to also disable libass' ASS extensions _(like `Border-Style: 4`, `Encoding: -1`, etc)_, which then raises the question what `get_state` should return if strict and some other feature is enabled?

Indeed, this sounds good too. I imagine that enabling the strict-mode feature would _disable_ all the extension features (even if they were previously enabled one by one or via the catch-all), so `get_state` could, in fact, just report that they’re all disabled.

> Since it initialises with all features disabled API-users could always track the current feature states themselves if needed.

Personally, I imagine some future features might be enabled by default! (Not to mention non-Boolean features.)

libass/ass_render.c
TheOneric: > I think the earlier comment was about the new parameter’s name

Ahh, that makes sense `^^. Then just ignore my previous comment.

MrSmile:
> Doing it as an `ASS_Feature` does seem sensible, and no, I imagine we wouldn’t want it activated by `INCOMPATIBLE_EXTENSIONS`.

I even think that it shouldn't be a separate feature, but additional state of `INCOMPATIBLE_EXTENSIONS` feature as they are mutually exclusive, i. e.
* 0 — default value;
* 1 — enable all extensions;
* &minus;1 — strict mode.

> Personally, I imagine some future features might be enabled by default!

I think it's better to make all features disabled by default (zero state) and invert feature meaning in such cases. But well, I imagine we someday would change default value of an already introduced feature...

Anyway, looks like consensus is not to introduce `ass_track_get_feature_state()` API function, at least for this PR.

libass/ass_render.c
TheOneric: Adding `ASS_Track` to the arguments is superfluous. `ASS_Renderer *render_priv` already contains a reference to the associated track.

cdown:
Hey folks, I've been running into this more and more recently and I'm hoping to sort out what forward progress would look like with you folks. Looking at this and #486 from @TheOneric, I'm trying to read through the scrollback and work out what the remaining blockers are (but there's a lot of stuff there, so I may have missed it!).

Aside from the Unicode-specific stuff already discussed on #486, is there anything blocking this in the interim? Is the feature API mature enough now to support this?

cdown:
Hey folks, just checking in here. Any thoughts on a way forwards based on the above comment? :-)

TheOneric:
Sorry for not replyig to your earlier comment.

> Is the feature API mature enough now to support this?

Boolean features work well.

> Aside from the Unicode-specific stuff already discussed on #486, is there anything blocking this

#486 also refactors the linebreaking code for better readability, which made some preëxisting issues apparent. #486 also fixes those issues (one still pending). I'd prefer if any changes to linebreaking happen only after the refactoring and fixes are in. While #486 is close to done, unfortunately it is not a priority for me right now and I don't have the capacity to work on it in addition to the higher prio issues.  
Other than that, the feature name here needs more work, there are a couple of unaddressed review comments (not counting the track feature stuff) and of course it needs to be rebased.

> in the interim to make the situation better for the trivial case?

I may be reading to much into this, but given your motivating example was CJK text, which would be addressed in a better way by Unicode line breaking: Does the “interim” in the quote mean, force-wrapping would only be an interim-solution for you until Unicode linebreaking is available?
Given that any new feature needs to be supported for a long time, I'm not fond of adding a new feature if the only verified usecase will be obsoleteted rather quickly by another nearly-finished feature.
Regarding the other issues you referenced in OP and commit message:
 - The mpv-turned-libass issue #360 is also about CJK originating from conversion from a non-ASS format.
 - Regarding mpv’s OSD, iirc fading the overflowing text out or ellipsing it was preferred back then over force-wrapping *(presumably because long paths could otherwise cover up everything)* which is the source of the `\box` and text-measuring suggestions *(the latter being also useful for other purposes)*.
 - I'm not sure why you're referencing the text-measuring API proposal as a motivator for forced word breaking.

If you have an usecase for forced word breaking not addressed better by Unicode and impractical with text-measuring, I’d like to hear about it and it should be reflected in the patch descriptions.

PR #485, opened 2021-03-09
Asm overhaul

MrSmile:
This is the result of my tinkering with checkasm branch (my continuation of it is [here](https://github.com/MrSmile/libass/commits/checkasm)). There was some problematic assembly, which already has its checkasm tests, so I've tried to improve it. This PR is getting rid of 64-bit only, avx2 slower than sse2 assembly, while improving performance in general.

### DIFF
--- a/libass/Makefile.am
+++ b/libass/Makefile.am
@@ -14,9 +14,8 @@ nasm_verbose_0 = @echo "  NASM    " $@;
 .asm.lo:
 	$(nasm_verbose)$(LIBTOOL) $(AM_V_lt) --tag=CC --mode=compile $(AS) $(ASFLAGS) -I$(srcdir)/ -o $@ $< -prefer-non-pic
 
-SRC_INTEL = x86/rasterizer.asm x86/blend_bitmaps.asm x86/blur.asm x86/cpuid.asm \
+SRC_INTEL = x86/rasterizer.asm x86/blend_bitmaps.asm x86/be_blur.asm x86/blur.asm x86/cpuid.asm \
             x86/cpuid.h
-SRC_INTEL64 = x86/be_blur.asm
 
 SRC_FONTCONFIG = ass_fontconfig.c ass_fontconfig.h
 SRC_DIRECTWRITE = ass_directwrite.c ass_directwrite.h dwrite_c.h
@@ -51,9 +50,6 @@ endif
 if ASM
 if INTEL
 libass_la_SOURCES += $(SRC_INTEL)
-if X64
-libass_la_SOURCES += $(SRC_INTEL64)
-endif
 endif
 endif
 
--- a/libass/ass_bitmap.c
+++ b/libass/ass_bitmap.c
@@ -86,12 +86,10 @@ void ass_synth_blur(const BitmapEngine *engine, Bitmap *bm,
     if (--be) {
         be_blur_pre(buf, w, h, stride);
         do {
-            memset(tmp, 0, stride * 2);
             engine->be_blur(buf, w, h, stride, tmp);
         } while (--be);
         be_blur_post(buf, w, h, stride);
     }
-    memset(tmp, 0, stride * 2);
     engine->be_blur(buf, w, h, stride, tmp);
     ass_aligned_free(tmp);
 }
@@ -266,7 +264,6 @@ void ass_be_blur_c(uint8_t *buf, intptr_t w, intptr_t h,
     uint16_t *col_sum_buf = tmp + w;
     unsigned x, y, old_pix, old_sum, temp1, temp2;
     uint8_t *src, *dst;
-    memset(tmp, 0, sizeof(uint16_t) * w * 2);
     y = 0;
 
     {
--- a/libass/ass_bitmap.c
+++ b/libass/ass_bitmap.c
@@ -73,8 +73,9 @@ void ass_synth_blur(const BitmapEngine *engine, Bitmap *bm,
         return;
 
     // Apply box blur (multiple passes, if requested)
+    unsigned align = 1 << engine->align_order;
     size_t size = sizeof(uint16_t) * bm->stride * 2;
-    uint16_t *tmp = ass_aligned_alloc(32, size, false);
+    uint16_t *tmp = ass_aligned_alloc(align, size, false);
     if (!tmp)
         return;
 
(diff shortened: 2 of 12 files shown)


MrSmile:
Rasterizer and blur assembly are getting some refurbishment too.

On the side note, what are the thoughts about separating C versions of assembly functions from other code into dedicated folder? Like I've done [here](https://github.com/MrSmile/libass/commit/10e229ce15f218d842be4e233a289dbe423113d6). That should simplify the building process of the future checkasm, especially in 32-bit mode.

TheOneric:
Nice, this now gets rid of all ASM-C discrepancies in my [set of tests using](https://github.com/TheOneric/libass-regression-tests-tmp/runs/2115159036?check_suite_focus=true#step:9:20) using [master](https://github.com/libass/libass/commit/0a3b14a76700b2c1db4d41611199d9aed336c191) C-versions as reference!
Can you add  _fixes #475_  to the `rasterizer: make C and assembly functions bitwise identical` commit, or would you prefer to wait until the `\be` fixes from #400 are merged as well? *(currently there are no `\be`s in my test-set)*

I don't have objections to moving non-ASM versions to a C-subfolder, if it makes building/testing easier.

TheOneric:
Previous comment was to optimistic, while x86_64/x32-ASM and C are now identical, i386-ASM is not:
https://github.com/TheOneric/libass-regression-tests-tmp/runs/2115158918?check_suite_focus=true#step:10:51

MrSmile:
And finally, complete position-independency!

> would you prefer to wait until the `\be` fixes from #400 are merged as well?

It's not relevant since I've rewritten `be_blur()` assembly from scratch.

> However, the i386-ASM still shows some discrepancies to x86_64 and C

Looks like it's not due to the assembly, `--disable-asm` doesn't change anything. I think it's the usual floating-point differences (x87 vs. SSE, optimizations, etc...)

> I don't have objections to moving non-ASM versions to a C-subfolder, if it makes building/testing easier.

My naive attempt to do that induced automake nagging about missing `subdir-objects`.

TheOneric:
> And finally, complete position-independency!

Nice. Building ASM-enabled static executables still results in segfaults in Alpine and Adélie, but dynamically linked builds now work, so the static ones still failing is to blame on their default compiler flags I guess.

> > However, the i386-ASM still shows some discrepancies to x86_64 and C
> 
> Looks like it's not due to the assembly, `--disable-asm` doesn't change anything. I think it's the usual floating-point differences (x87 vs. SSE, optimizations, etc...)

Huh, you're right, the same happens without assembly enabled. Curiously though, running the C version on **real** ARMv7 hardware shows *zero* deviation from amd64's C version.

MrSmile:
> so the static ones still failing is to blame on their default compiler flags I guess.

Incorrect stack alignment perhaps?

> Curiously though, running the C version on **real** ARMv7 hardware shows _zero_ deviation from amd64's C version.

Both have standard 64-bit registers and similar set of operations. On the contrary, i386 uses x87 instruction set with its 80-bit registers. Also I think that architectures with MAD operations can deviate too.

astiob:
> > > However, the i386-ASM still shows some discrepancies to x86_64 and C
> > 
> > Looks like it's not due to the assembly, `--disable-asm` doesn't change anything. I think it's the usual floating-point differences (x87 vs. SSE, optimizations, etc...)
> 
> Huh, you're right, the same happens without assembly enabled. Curiously though, running the C version on **real** ARMv7 hardware shows _zero_ deviation from amd64's C version.

Try adding `-mfpmath=sse` to the i386 compiler flags.

If that fixes it, we could perhaps add this to the `compare` flags automatically somehow.

astiob:
I’m a bit lost in all these PRs. Does this depend on #400 or replace it? Any other PR interdependencies?

TheOneric:
On Tue, Mar 16, 2021 at 15:46:58 -0700, Oleg Oshmyan wrote:
> > > > However, the i386-ASM still shows some discrepancies to x86_64 and C
> > > 
> > >  […] I think it's the usual floating-point differences (x87 vs. SSE, optimizations, etc...)
> 
> Try adding `-mfpmath=sse` to the i386 compiler flags.

`-msse -mfpmath=sse` alone wasn't sufficient, but with
`-march=native -msse -mfpmath=sse` the difference disappeared on my machine.
I can look into what `-march=native` additionally enables tomorrow or so.


> If that fixes it, we could perhaps add this to the `compare` flags automatically somehow.

Once the relevant flags are determined, I can add them to the automated 
builds, if you want to add it to the general libass flags for comapre, checks 
for cpu arch, and ideally also whether sse etc is supported, would be needed.

MrSmile:
> I’m a bit lost in all these PRs. Does this depend on #400 or replace it? Any other PR interdependencies?

Basically, PR #400 (and my continuation of it) consists of the two parts: assembly testing infrastructure and fixes to assembly (and other libass code). So, I've extended #400 to full set of assembly functions and then extracted fixes from it, creating this PR.

TheOneric:
**Re: i386 - amd64/x32/armv7 differences:**
For a default of `-march=i686` adding `-msse -msse2 -mfpmath=sse`reolved all discrepancies. As our ASM needs at least SSE2 anyway, this should be safe to add. See: https://github.com/TheOneric/libass-regression-tests-tmp/runs/2134552179?check_suite_focus=true#step:10:51 *(the riscv64 failure is unrelated and can be safely ignored)*

TheOneric:
> [Regarding the be fixes from #400:] It's not relevant since I've rewritten be_blur() assembly from scratch.

Imo the be_blur rewrite commit message should then mention, that it also fixes a C-ASM difference caused by the lack of f23b9ed64bd4ccf249c686616dd3f51a69d285dc being ported to ASM.

astiob:
I’d appreciate a summary for the less assembler-literate people of what effects the rewrites have. Are they just a stylistic/cosmetic/technical rephrasing of the same code that is essentially equivalent but runs on i386? How do they affect performance?

MrSmile:
> I’d appreciate a summary for the less assembler-literate people of what effects the rewrites have.

Main differences compared to master/checkasm version:
- No special handling of unaligned case (first commit about that).
- Less special code versions and copy-paste (all sse2 and avx2 functions have merged together, same macro has used for `add_bitmaps` and `sub_bitmaps`).
- Vector handling of width tails instead of byte loops.
- Code reorganized to use less registers to simplify porting to x32.

> How do they affect performance?

New version is faster or the same, in some cases (`be_blur`, `mul_bitmaps_avx2`) significantly faster. Also, avx2 versions are always faster than sse2 (at least on my hardware).

I'll add list of changes to relevant commit messages.

TheOneric:
For future reference *(was discussed on IRC prior to merge)*:
On some microarches – only Zen and Zen+ afaik – the avx2 routines can sometimes still be a tad slower than the sse2 version.
Namely on Zen+ `be_blur` is notably slower and `fill_solid32` a tad slower, in all other cases avx2 is either faster or equivalent to sse2. This seems to be due to Zen(+) not having full native AVX2 support.
As this seems to only affect two CPU-generations of one vendor, and the performance difference isn't that big, we decided to leave it as is. Overall this pr should still provide an performance-benefit even on Zen(+).

PR #488, opened 2021-03-23
Print libass version

TheOneric:
This could help us in future bug reports and we already do so for shapers and freetype.
 Also include a guess about the source code version, which may not be entirely reliable, if custom tarballs are involved, commits are added after configure was called or if patches are applied without creating commits. But it may be good enough™ for most cases.
 
 On the other hand this may lead us astray if any of the aforementioned are involved, not sure what's better. What are your thoughts on the "source version guess"?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,5 @@
-AC_INIT(libass, 0.15.0)
+m4_define([LIBASS_VERSION], [0.15.0])
+AC_INIT(libass, LIBASS_VERSION)
 AM_INIT_AUTOMAKE([foreign])
 AC_CONFIG_MACRO_DIR([m4])
 # Disable Fortran checks
@@ -300,6 +301,21 @@ AM_COND_IF([ENABLE_LARGE_TILES], [
 ], [
     AC_DEFINE(CONFIG_LARGE_TILES, 0, [use small tiles])
 ])
+
+## Make a guess about the source code version
+AS_IF([test -d "${srcdir}/.git"], [
+    AC_PATH_PROG([git_bin], [git])
+    AS_IF([test -n "$git_bin"], [
+        srcversion_string="commit: $("$git_bin" -C "$srcdir" describe --tags --long --always --dirty --broken --abbrev=40)"
+    ], [
+        srcversion_string="custom after: LIBASS_VERSION"
+    ])
+], [
+    dnl# Hope no one creates custom tarballs without adjusting the version
+    srcversion_string="tarball: LIBASS_VERSION"
+])
+AC_DEFINE_UNQUOTED([CONFIG_SOURCEVERSION], ["$srcversion_string"],
+                   [string containing info about the used source])
 
 ## Setup output beautifier.
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -29,14 +29,6 @@ jobs:
       - name: checkout code
         uses: actions/checkout@v2
 
-      - name: retrieve caches
-        uses: actions/cache@v2
-        with:
-          path: |
-            $HOME/Library/Caches/Homebrew
-            /usr/local/Homebrew
-          key: ${{ runner.os }}-${{ matrix.os }}-build
-
       - name: install deps
         run: |
           if echo "${{ matrix.os }}" | grep -qE '^macos-' ; then
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -24,6 +24,7 @@
 #include <string.h>
 #include <stdbool.h>
 
+#include "ass.h"
 #include "ass_outline.h"
 #include "ass_render.h"
 #include "ass_parse.h"
@@ -49,6 +50,9 @@ ASS_Renderer *ass_renderer_init(ASS_Library *library)
     FT_Library ft;
     ASS_Renderer *priv = 0;
     int vmajor, vminor, vpatch;
+
+    ass_msg(library, MSGL_INFO, "libass API version: 0x%X", LIBASS_VERSION);
+    ass_msg(library, MSGL_INFO, "libass source: %s", CONFIG_SOURCEVERSION);
 
     error = FT_Init_FreeType(&ft);
     if (error) {


configure.ac
```diff
@@ -301,6 +302,20 @@ AM_COND_IF([ENABLE_LARGE_TILES], [
...
 
+## Make a guess about the source code version
+AS_IF([test -d .git], [
+    AS_IF([which git], [
+        srcversion_string="commit: $(git rev-parse HEAD)"
+    ], [
+        srcversion_string="custom after: LIBASS_VERSION"
+    ])
+], [
+    dnl# Hope noone creates custom tarballs without adjusting the version
```
rcombs, line 313: *no one

configure.ac
```diff
@@ -301,6 +302,20 @@ AM_COND_IF([ENABLE_LARGE_TILES], [
     AC_DEFINE(CONFIG_LARGE_TILES, 0, [use small tiles])
 ])
 
+## Make a guess about the source code version
+AS_IF([test -d .git], [
+    AS_IF([which git], [
+        srcversion_string="commit: $(git rev-parse HEAD)"
```
rcombs, line 308: Does this work out-of-tree?

rcombs (commented):
Couple nits/questions but I like this a lot overall.

configure.ac
TheOneric: thx, fixed.

configure.ac
TheOneric: Nope, it did not. Now it does, thanks.

configure.ac
```diff
@@ -301,6 +302,20 @@ AM_COND_IF([ENABLE_LARGE_TILES], [
     AC_DEFINE(CONFIG_LARGE_TILES, 0, [use small tiles])
 ])
 
+## Make a guess about the source code version
+AS_IF([test -d "${srcdir}/.git"], [
+    AS_IF([which git], [
+        srcversion_string="commit: $(cd "$srcdir"; git rev-parse HEAD)"
```
astiob, line 308: You can also use: `git -C "$srcdir" rev-parse HEAD`

configure.ac
```diff
@@ -301,6 +302,20 @@ AM_COND_IF([ENABLE_LARGE_TILES], [
     AC_DEFINE(CONFIG_LARGE_TILES, 0, [use small tiles])
 ])
 
+## Make a guess about the source code version
+AS_IF([test -d "${srcdir}/.git"], [
+    AS_IF([which git], [
```
astiob, line 307: This should suppress standard output. It seems there’s no standard way to do this other than just `which git >/dev/null`.

(Alternatively, with more effort, this could prettify the output to look like an autoconf check.)

astiob (commented):
Looks good!

Have you considered using `git describe --tags --long --always --dirty --broken` instead of `git rev-parse HEAD`? Perhaps even with `--abbrev=40` to show the full, unabbreviated commit hash to make it easier to identify commits that aren’t in the main repository.

configure.ac
TheOneric: Will `>/dev/null` work for Cygwin builds on Windows?

configure.ac
TheOneric: `git describe` with `--dirty` and `--broken` seems good. It can't do anything about patches added after configuration, but it will catch patches applied before config even without a commit. Changed to: `git -C "$srcdir" describe --tags --long --always --dirty --broken --abbrev=40`

TheOneric:
Switched `which` for `command -v`, which is part of POSIX and more portable than `which`.

As reception has been positive, I'll merge it in a day if no hold or further changes are requested.

configure.ac
astiob: (re: /dev/null) FWIW that’s exactly what I expected, and I fully expect it to work in both Cygwin and MSYS on Windows.

FYI:

Cygwin’s whole goal is to be a full POSIX environment on Windows, so I absolutely expect /dev/null to work there.

MSYS is a more lightweight thing that is mainly a POSIX shell, with an explicit goal to be [just] posixy enough to run build scripts for software originating from the posixland. Precisely for these reasons, I expect /dev/null to work there, too. (Although I use MSYS myself sometimes, my memory isn’t quite fresh enough to assert that I’ve “definitely seen” it work.)

And I imagine the configure script itself expects a POSIX-like shell—even if autoconf does cater for compatibility with exotic and ancient unixes, /dev/null seems like a fairly basic need. Which you’ve now confirmed by saying it does already use /dev/null.

astiob:
Hmm, maybe we should just use [`AC_PATH_PROG`](https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Generic-Programs.html) for Git? For what it’s worth, we’re already using it for NASM, although it seems to me that we’re actually using it in a wrong manner there 😃

TheOneric:
> Hmm, maybe we should just use [`AC_PATH_PROG`](https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Generic-Programs.html) for Git? For what it’s worth, we’re already using it for NASM, although it seems to me that we’re actually using it in a wrong manner there smiley

We're using `AC_CHECK_PROG` for NASM, which also does invoke `AC_SUBST`, which isn't necessary for git. However it's true we don't make any use of `AC_CHECK_PROG`'s outputfile substitution in NASM's case either. `AC_PATH_PROG` doesn't seem to invoke `AC_SUBST`, so it should be suitable for our git use case *(and compared to `command -v` this avoids unlikely but possible false positives with shell functions or aliases named `git` yet not providing `git` iinm)*.
 Re-pushed with `AC_PATH_PROG` instead of `command -v`.

configure.ac
```diff
@@ -301,6 +302,21 @@ AM_COND_IF([ENABLE_LARGE_TILES], [
     AC_DEFINE(CONFIG_LARGE_TILES, 0, [use small tiles])
 ])
 
+## Make a guess about the source code version
+AS_IF([test -d "${srcdir}/.git"], [
+    AC_PATH_PROG([git_bin], [git])
+    AS_IF([test -n "$git_bin"], [
+        srcversion_string="commit: $($git_bin -C "$srcdir" describe --tags --long --always --dirty --broken --abbrev=40)"
```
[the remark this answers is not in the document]
TheOneric: Done, thx.

Issue #98, opened 2014-05-17
Memory font API is semi-broken


astiob:
Only the following results in libass using memory fonts:
1. Call `ass_add_font` for each memory font.
2. Then call `ass_set_fonts(update=1)`.
3. Never call `ass_fonts_update`.

Wrong call order? `process_fontdata` never gets called.
No update=1? The `FcConfigGetFonts` call in `process_fontdata` fails.
Call `ass_fonts_update` afterwards? It rebuilds the font set, effectively removing any memory fonts that have been added to it.

I think this may affect #67 too.

Originally discovered in #97.

astiob:
`ass_fonts_update` and the `update` argument to `ass_set_fonts` are now no-ops (since the introduction of fontselect), which simplifies and unbreaks the API somewhat. `ass_set_fonts` still needs to be called after `ass_add_font`, which also causes #266.

PR #490, opened 2021-03-26
Fix embedded and memory fonts

TheOneric:
With this embedded fonts now work in VLC and our own utilities.
FFmpeg and Aegisub do not properly use `ass_set_extract_fonts`, so it still doesn't work for there *(but patching them should be fairly easy)*.

### DIFF
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -992,7 +992,6 @@ static ASS_FontProvider *
 ass_embedded_fonts_add_provider(ASS_Library *lib, ASS_FontSelector *selector,
                                 FT_Library ftlib)
 {
-    int i;
     ASS_FontProvider *priv = ass_font_provider_new(selector, &ft_funcs, NULL);
     if (priv == NULL)
         return NULL;
@@ -1001,7 +1000,7 @@ ass_embedded_fonts_add_provider(ASS_Library *lib, ASS_FontSelector *selector,
         load_fonts_from_dir(lib, lib->fonts_dir);
     }
 
-    for (i = 0; i < lib->num_fontdata; ++i)
+    for (size_t i = 0; i < lib->num_fontdata; i++)
         process_fontdata(priv, lib, ftlib, i);
 
     return priv;
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -990,7 +990,7 @@ static void process_fontdata(ASS_FontProvider *priv, ASS_Library *library,
  */
 static ASS_FontProvider *
 ass_embedded_fonts_add_provider(ASS_Library *lib, ASS_FontSelector *selector,
-                                FT_Library ftlib)
+                                FT_Library ftlib, size_t *num_emfonts)
 {
     ASS_FontProvider *priv = ass_font_provider_new(selector, &ft_funcs, NULL);
     if (priv == NULL)
@@ -1002,6 +1002,7 @@ ass_embedded_fonts_add_provider(ASS_Library *lib, ASS_FontSelector *selector,
 
     for (size_t i = 0; i < lib->num_fontdata; i++)
         process_fontdata(priv, lib, ftlib, i);
+    *num_emfonts = lib->num_fontdata;
 
     return priv;
 }
@@ -1035,9 +1036,8 @@ struct font_constructors font_constructors[] = {
  * \return newly created font selector
  */
 ASS_FontSelector *
-ass_fontselect_init(ASS_Library *library,
-                    FT_Library ftlibrary, const char *family,
-                    const char *path, const char *config,
+ass_fontselect_init(ASS_Library *library, FT_Library ftlibrary, size_t *num_emfonts,
+                    const char *family, const char *path, const char *config,
                     ASS_DefaultFontProvider dfp)
 {
     ASS_FontSelector *priv = calloc(1, sizeof(ASS_FontSelector));
@@ -1050,7 +1050,7 @@ ass_fontselect_init(ASS_Library *library,
     priv->index_default = 0;
 
     priv->embedded_provider = ass_embedded_fonts_add_provider(library, priv,
-            ftlibrary);
+            ftlibrary, num_emfonts);
 
     if (priv->embedded_provider == NULL) {
         ass_msg(library, MSGL_WARN, "failed to create embedded font provider");
@@ -1133,4 +1133,15 @@ void ass_map_font(const ASS_FontMapping *map, int len, const char *name,
             return;
         }
     }
+}
+
+size_t ass_update_embedded_fonts(ASS_Library *lib, ASS_FontSelector *selector,
+                                 FT_Library ftlib, size_t num_loaded)
+{
+    if (!selector->embedded_provider)
+        return num_loaded;
+
+    for (size_t i = num_loaded; i < lib->num_fontdata; i++)
+        process_fontdata(selector->embedded_provider, lib, ftlib, i);
+    return lib->num_fontdata;
 }
--- a/libass/ass_library.c
+++ b/libass/ass_library.c
@@ -95,24 +95,13 @@ void ass_set_style_overrides(ASS_Library *priv, char **list)
         *q = strdup(*p);
 }
 
-static int grow_array(void **array, int nelem, size_t elsize)
-{
-    if (!(nelem & 31)) {
-        void *ptr = realloc(*array, (nelem + 32) * elsize);
-        if (!ptr)
-            return 0;
-        *array = ptr;
-    }
-    return 1;
-}
-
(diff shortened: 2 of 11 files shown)


libass/ass_fontselect.c
```diff
@@ -1000,8 +1000,9 @@ ass_embedded_fonts_add_provider(ASS_Library *lib, ASS_FontSelector *selector,
         load_fonts_from_dir(lib, lib->fonts_dir);
     }
 
-    for (size_t i = 0; i < lib->num_fontdata; ++i)
+    for (size_t i = lib->num_fontsloaded; i < lib->num_fontdata; ++i)
```
astiob, line 1003: Do I understand correctly that this will always be zero here? The only other place that increments `num_fontsloaded` is in `ass_update_embedded_fonts`, which assumes the embedded font provider already exists.

astiob:
Why do we not default to `extract_fonts` in the first place? Why is it even configurable?…

libass/ass_fontselect.c
TheOneric: My thought was that if `ass_set_fonts` is called a second time *(is that allowed API use?)*, already loaded fonts won't be added a second time. However, looking at it again, the embedded font provider gets recreated so it needs to start from zero regardless.

`ass_update_embedded_fonts` should also check that the provider does exists, in case creation failed.

Re-pushed with both of the above changes.

TheOneric:
> Why do we not default to `extract_fonts` in the first place? Why is it even configurable?…

No idea bout the current default. I'm also not sure, why `ass_clear_fonts` is even public in this specific unsafe form *(and no indication how to use it safely up until now)*.
If a user really wants to strictly control used fonts, disabling embedded fonts might make some sense, though I'm not sure why fonts would need to be controlled that strictly.

libass/ass_library.c
```diff
@@ -95,24 +95,13 @@ void ass_set_style_overrides(ASS_Library *priv, char **list)
...
 void ass_add_font(ASS_Library *priv, const char *name, const char *data, int size)
 {
-    int idx = priv->num_fontdata;
+    size_t idx = priv->num_fontdata;
     if (!name || !data || !size)
         return;
-    if (!grow_array((void **) &priv->fontdata, priv->num_fontdata,
-                    sizeof(*priv->fontdata)))
+    if (!(priv->num_fontdata & 31) &&
+            !ASS_REALLOC_ARRAY(priv->fontdata, priv->num_fontdata + 32))
```
astiob, line 104: By the way, we may want to switch this to exponential growth while we’re at it. I suppose this is well-meant because there are usually few memory fonts, but if there should be a situation that needs many memory fonts, this will be quite slow. And when there are few fonts, we won’t be wasting much memory anyway. If we start at 32 and keep doubling the size, our memory use will be the same until we hit 65 fonts, which is already a bigger number than this is likely to be in practice.

TheOneric: Can we use exponential growth without adding another variable to keep track of current max size? *(not that adding a variable is a problem)*

astiob: Ah, hmm. Well, _technically,_ we can follow the spirit of the current code and use bit arithmetic:

```c
if (!(priv->num_fontdata & priv->num_fontdata-1))
    // num_fontdata is 0 or exactly a power of two, so it's time for a realloc
    ASS_REALLOC_ARRAY(priv->fontdata, FFMAX(32, priv->num_fontdata * 2)) ...
```

TheOneric: Right, with `2 * …` growth all but the first reallocation will happen on a power of two.

libass/ass_fontselect.c
```diff
@@ -1001,7 +1000,7 @@ ass_embedded_fonts_add_provider(ASS_Library *lib, ASS_FontSelector *selector,
         load_fonts_from_dir(lib, lib->fonts_dir);
     }
 
-    for (i = 0; i < lib->num_fontdata; ++i)
+    for (size_t i = 0; i < lib->num_fontdata; ++i)
```
MrSmile, line 1003: While at it, also replace `++i` with `i++` as most developers prefer postfix form.

libass/ass_library.c
```diff
@@ -95,24 +95,13 @@ void ass_set_style_overrides(ASS_Library *priv, char **list)
...
-
 void ass_add_font(ASS_Library *priv, const char *name, const char *data, int size)
 {
-    int idx = priv->num_fontdata;
+    size_t idx = priv->num_fontdata;
     if (!name || !data || !size)
         return;
-    if (!grow_array((void **) &priv->fontdata, priv->num_fontdata,
-                    sizeof(*priv->fontdata)))
+    if (!(idx & 31) && !(idx & (idx - 1)) && // power of two >= 32, or zero --> realloc time
```
MrSmile, line 103: Can be replaced with `!(idx & (idx - 32))`.

libass/ass_library.c
```diff
@@ -135,8 +124,7 @@ void ass_add_font(ASS_Library *priv, const char *name, const char *data, int siz
 
 void ass_clear_fonts(ASS_Library *priv)
 {
-    int i;
-    for (i = 0; i < priv->num_fontdata; ++i) {
+    for (size_t i = 0; i < priv->num_fontdata; ++i) {
```
MrSmile, line 127: `++i` &rarr; `i++`.

MrSmile (commented):
I'm not an expert in the font machinery, so only a couple of superficial comments.

libass/ass_library.c
TheOneric: Done, thx.

libass/ass_fontselect.c
TheOneric: Personally I prefer to only use the postfix form when the *"increment **after** evaluation"* part is required, but that's just a preference and I guess it's true that libass' for loops more often use `i++`.
Done.

libass/ass_fontselect.c
```diff
@@ -1135,3 +1135,12 @@ void ass_map_font(const ASS_FontMapping *map, int len, const char *name,
         }
     }
 }
+
+void ass_update_embedded_fonts(ASS_Library *lib, ASS_FontSelector *selector, FT_Library ftlib)
+{
+    if (!selector->embedded_provider)
+        return;
+    for (size_t i = lib->num_fontsloaded; i < lib->num_fontdata; ++i)
```
astiob, line 1143: If you’re switching the increments to postfix, you’ve missed this one.

libass/ass_fontselect.c
astiob: (FWIW I use prefix in my own non-libass code because it’s the standard type-agnostic default in C++, but indeed libass uses postfix.)

astiob:
Not for this PR, but I imagine that it’s probably possible to switch some (or many) of our other reallocs to this bit-op scheme, too.

libass/ass_fontselect.c
TheOneric: Done, thx.

astiob:
Honestly, I’ve never quite understood (the point of) the library vs renderer distinction. But my impression is the same as yours:

* Creating multiple ASS_Renderers from a single ASS_Library is supported.

* Creating multiple tracks from/for a single ASS_Library/ASS_Renderer is supported.

* Mixing ASS_Library instances… This has gone back and forth between being supported and unsupported.

   Currently (since c22a4ff9a395546637dbe0f1e9d0ee549dd0069a in year 2015), this is unsupported: `ass_start_frame` has an explicit check and balks when given mismatching libraries.

    Some two years earlier, in 2013, 68b0694f09f86e632de55f5d61d52c08f8533372 had explicitly removed this check, and re-adding it may have been a rebasing accident. That commit claimed that, while fonts (and style overrides?) come from the wrong library, otherwise everything “just works”, and indeed this is my impression from our _current_ code.

    Back in 2009, d5b7cb636ccf447543cf2beb4b2d593fa4890570 had explicitly _added_ this check for the first time.

    My first instinct would be to keep this forbidden, but then there’s little actual good reason to do this at the moment.

Historically, `ASS_Library` was introduced in [MPlayer r20493](https://github.com/mpv-player/mpv/commit/efea99c2c9e8ea1f5a96a1cb86aa30ef05b6b828), and looking at that commit, it seems a single, shared `ASS_Library` was to be used player-wide (e. g. by OSD and subtitles alike) and to persist for the whole lifetime of the player, while each component had its own ASS_Renderer. I’m not sure what the idea was that led to this design though.

TheOneric:
New version pushed:

 - Moved font counter into the renderer and renamed to `num_emfonts`.
 - Last commit no longer claims to fix #98, as embedded fonts leaking to other tracks means it is still somewhat broken.
   *This was already the case **before** this pr.*

TheOneric:
Rebased to master.

astiob (approved):

PR #499, opened 2021-04-26
Add support for multiple input directories for regression testing utility

MrSmile:
That simplifies testing in case of multiple sets of target images for the same sub files. Also it's now possible to organize a pack of tests without font duplicating/symlinking trick.

### DIFF
--- a/compare/README.md
+++ b/compare/README.md
@@ -5,11 +5,16 @@ To build a test utility configure libass with the `--enable-compare` flag.
 The utility works with `png` image files so there is external dependency of libpng.
 
 Test program command line:  
-`compare ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:1-8>]`
+`compare ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:1-8>] [-p <pass-level:0-3>]`
 
 * `<input-dir>` is a test input directory, can be several of them;
 * `<output-dir>` if present sets directory to store the rendering results;
-* `<scale>` sets an oversampling factor (positive integer up to 8, default 1).
+* `<scale>` sets an oversampling factor (positive integer up to 8, default 1);
+* `<pass-level>` corresponds to the level of image differences required to pass test:
+  - 0: only `SAME` level accepted, bitwise comparison mode;
+  - 1: `GOOD` level or less required;
+  - 2: `BAD` level or less required, default mode;
+  - 3: `FAIL` level or less required, i. e. any difference accepted, error checking mode.
 
 An input directory consists of font files (`*.ttf`, `*.otf` and `*.pfb`), subtitle files (`*.ass`), and image files (`*.png`).
 All the fonts required for rendering should be present in the input directories as
@@ -62,11 +67,13 @@ Processing 'sub2.ass':
   Time 0:02:33.000 - 0.728 OK
 Only 3 of 4 images have passed test
 ```
-For each target image file the program reports a maximal ratio of the per pixel comparison error to the baseline error scale.
-The baseline error scale tries to take account of neighboring pixels to estimate visibility of an error in that specific pixel location.
-Currently the range of [0&ndash;2.0] is marked as `OK`, the range of [2.0&ndash;4.0] as `BAD` and more than 4.0 as `FAIL`.
-At the end there is a summary of how many images don't fail tests.
-If all images pass that test the program returns 0 otherwise 1.
+For each target image file the program reports a maximal ratio of the per pixel comparison difference to the baseline error scale.
+The baseline error scale tries to take account of neighboring pixels to estimate visibility of a difference in that specific pixel location.
+Zero difference is marked as `SAME` (level 0), nonzero value in the range of [0–2.0] is marked as `GOOD` (level 1),
+in the range of [2.0–4.0] as `BAD` (level 2) and more than 4.0 as `FAIL` (level 3).
+Any problem during processing of a specific image is treated as a level 4 difference.
+If all images have level of difference less or equal to pass level (`-p` switch, default 2) then the test is considered as passed
+and program returns 0. Otherwise program returns maximal level of difference throughout all images (4 in case of error).
 
 Note that almost any type of a rendering error can be greatly exaggerated by the specially tailored test cases.
 Therefore test cases should be chosen to represent generic real world scenarios only.
--- a/compare/README.md
+++ b/compare/README.md
@@ -5,16 +5,16 @@ To build a test utility configure libass with the `--enable-compare` flag.
 The utility works with `png` image files so there is external dependency of libpng.
 
 Test program command line:  
-`compare [-i] <input-dir> [-o <output-dir>] [-s <scale:1-8>]`
+`compare ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:1-8>]`
 
-* `<input-dir>` is a test input directory;
+* `<input-dir>` is a test input directory, can be several of them;
 * `<output-dir>` if present sets directory to store the rendering results;
 * `<scale>` sets an oversampling factor (positive integer up to 8, default 1).
 
 An input directory consists of font files (`*.ttf`, `*.otf` and `*.pfb`), subtitle files (`*.ass`), and image files (`*.png`).
-All the fonts required for rendering should be present in the input directory as
+All the fonts required for rendering should be present in the input directories as
 libass is configured to not use anything external outside of it (`ASS_FONTPROVIDER_NONE`).
-After loading all the fonts in the directory, the test program scans for subtitle files (`<sub_name>.ass`)
+After loading all the fonts in the directories, the test program scans for subtitle files (`<sub_name>.ass`)
 and their corresponding image files in the form of `<sub_name>-NNNN.png`,
 where `NNNN` is an arbitrary number of decimal digits.
 The subtitle file then would be rendered at the time of `NNNN` milliseconds and compared with the loaded image file.
@@ -32,7 +32,7 @@ test/
   sub2-153000.png
 ```
 
-More precisely, the test program (`compare`) would scan the input directory and do the following:
+More precisely, the test program (`compare`) would scan all input directories and do the following:
 1) load all the fonts found (`*.ttf`, `*.otf`, `*.pfb`);
 2) search for subtitle files (`*.ass`);
 3) for every file found (`<sub_name>.ass`) scan for the files with names of `<sub_name>-NNNN.png`;
--- a/compare/compare.c
+++ b/compare/compare.c
@@ -284,13 +284,13 @@ static ASS_Track *load_track(ASS_Library *lib,
                              const char *dir, const char *file)
 {
     char path[4096];
-    snprintf(path, sizeof(path), "%s/%s.ass", dir, file);
+    snprintf(path, sizeof(path), "%s/%s", dir, file);
     ASS_Track *track = ass_read_file(lib, path, NULL);
     if (!track) {
-        printf("Cannot load subtitle file '%s.ass'!\n", file);
+        printf("Cannot load subtitle file '%s'!\n", file);
         return NULL;
     }
-    printf("Processing '%s.ass':\n", file);
+    printf("Processing '%s':\n", file);
     return track;
 }
 
@@ -351,6 +351,7 @@ static bool process_image(ASS_Renderer *renderer, ASS_Track *track,
 
 typedef struct {
     char *name;
+    size_t prefix;
     const char *dir;
     int64_t time;
 } Item;
@@ -384,9 +385,20 @@ static void delete_items(ItemList *list)
 static int item_compare(const void *ptr1, const void *ptr2)
 {
     const Item *e1 = ptr1, *e2 = ptr2;
-    int cmp = strcmp(e1->name, e2->name);
+
+    int cmp_len = 0;
+    size_t len = e1->prefix;
+    if (len > e2->prefix) {
+        cmp_len = +1;
+        len = e2->prefix;
+    } else if (len < e2->prefix) {
+        cmp_len = -1;
+    }
+    int cmp = memcmp(e1->name, e2->name, len);
     if (cmp)
         return cmp;
+    if (cmp_len)
+        return cmp_len;
     if (e1->time > e2->time)
         return +1;
     if (e1->time < e2->time)
@@ -401,11 +413,10 @@ static bool add_sub_item(ItemList *list, const char *dir, const char *file, size
         return false;
 
     Item *item = &list->items[list->n_items];
-    item->name = malloc(len + 1);
+    item->name = strdup(file);
     if (!item->name)
         return out_of_memory();
-    memcpy(item->name, file, len);
-    item->name[len] = '\0';
+    item->prefix = len;
     item->dir = dir;
     item->time = -1;
     list->n_items++;
@@ -438,7 +449,7 @@ static bool add_img_item(ItemList *list, const char *dir, const char *file, size
     item->name = strdup(file);
     if (!item->name)
         return out_of_memory();
-    item->name[pos] = '\0';
+    item->prefix = pos;
     item->dir = dir;
     item->time = 0;
     for (size_t i = first; i < len; i++)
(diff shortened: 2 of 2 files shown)


astiob (commented):
(You may have expected this:) I’m not a fan of the type punning and `argv` abuse. In fact, [the language standard is silent on whether modifying the `argv[i]` pointers is allowed at all](https://stackoverflow.com/q/25737434/865331), even if you disregard type compatibility and aliasing. But even if nothing else, this is just pretty unintuitive code.

By the way (already an issue before this PR): file name extensions should be checked case-insensitively, at the very least for fonts. In my experience, many fonts have all-capitals file names.

MrSmile:
> I’m not a fan of the type punning and `argv` abuse.

I'm not a fan of it either, but also not a fan of dynamic reallocing or double parsing too. Maybe fixed size buffer (~64 items) is good enough? Is there some hard limit on argument count?

> file name extensions should be checked case-insensitively

Is there some standard function for that? Or should I write it from scratch?

TheOneric:
> > I’m not a fan of the type punning and `argv` abuse.
> 
> I'm not a fan of it either, but also not a fan of dynamic reallocing or double parsing too. Maybe fixed size buffer (~64 items) is good enough? Is there some hard limit on argument count?

Not a fan of a fixed size buffer. Why not allocate a buffer of length `(argc+1)/2 + 1` **once**? It's guranteed to be large enough for all pointers to image directories plus terminating `0`.

> > file name extensions should be checked case-insensitively
> 
> Is there some standard function for that?

Not in ISO-C afaik. I think checking all CAPS and all lowercase would also be a good enough™ compromise, that doesn't require a full `strcasecmp` implementation. *(Note that this is not an objection to a full `strcasecmp`.)*

astiob:
> also not a fan of dynamic reallocing or double parsing too

I don’t really see any double-parsing here. And I have no problem with dynamic allocation in this context; I’d just use `ASS_REALLOC_ARRAY`.

> Is there some hard limit on argument count?

I don’t think so. I _hope_ not.

> Is there some standard function for that?

We have `ass_strcasecmp` if ASCII is enough.

astiob:
> Why not allocate a buffer of length `(argc+1)/2 + 1` **once**?

Right, that’s another option.

astiob:
Anyway, I don’t have any comments on the actual meaningful change here. Nothing else has jumped out at me in the code, and I kinda think I understand the idea (and why you’re not just allowing multiple input-dirs: to keep all `.ass` file references unambiguous), but I don’t actually use `compare` myself.

MrSmile:
> Why not allocate a buffer of length `(argc+1)/2 + 1` **once**?

Good idea.

> I’d just use `ASS_REALLOC_ARRAY`.
> We have `ass_strcasecmp` if ASCII is enough.

Both use noninline functions from library without dedicated export directive, can lead to linking problems.

astiob:
> Both use noninline functions from library

Isn’t `compare` linked to `libass` statically?

astiob:
OK, I suppose theoretically a platform might have an actual linking step for `libass.a` that involves a link-time-optimizing compiler that notices the symbols aren’t exported and optimizes things within `libass.a`? I wonder if/how possible/allowed that is.

compare/compare.c
```diff
@@ -442,18 +444,54 @@ static bool add_img_item(ItemList *list, const char *file, int len)
...
+                continue;
+            printf("Cannot load font '%s'!\n", name);
+        } else {
+            continue;
+        }
+        closedir(dir);
+        return false;
+    }
+    closedir(dir);
+    return true;
```
TheOneric, line 487: I find this bit somewhat confusing, the first `closedir; return false;` is dead code after all.
Why not declare `file` first outside of the loop and then just `while(file = readdir(dir))`?

`readdir` is not part of ISO-C99 btw, though for an optional utility like `compare` this probably doesn't matter too much.

I now see `readdir` is also used in `ass_fontselect`...

TheOneric:
> Isn’t compare linked to libass statically?

For the record, I sometimes use(d) dynamically linked `compare` for testing either older libass versions or to verify the build procedure works for both static and dynamic libraries.

rcombs:
You could in theory set `AR="ld" AR_FLAGS="-r -export-symbols [symfile] -o"` or some equivalent and end up with an object file named `.a` that contained the entire library, with only its public symbols exported (the others behave similarly to `static` functions, as if the library was a single big translation unit). This isn't used much to my knowledge, though I'm not entirely sure why; it allows for much better library hygiene.

I usually just use `strcasecmp` with a `#define strcasecmp _stricmp` on win32.

Agreed that a simple `realloc` (it can just be linear +1 every time, this isn't a perf-critical path) would be much cleaner for the arg parsing.

compare/compare.c
MrSmile: It's not dead code but error handling branch. It's executed in case of `add_*_item()` or `load_font()` failure. I can replace it with the usual `goto fail;` trick if it's too hard to understand.

TheOneric: oups, my bad. guess I shouldn't do code reviews late at night.
`goto fail` may be indeed a bit easier to read, but *(now that I'm no longer half-asleep)* the current approach also seems ok.

To elaborate on the `readdir`-part: As it's also used in the lib this isn't a problem. Though even if it wasn't, I wouldn't be opposed to allow using all POSIX-extensions (so also eg `strcasecmp`) in separate, optional utilities like `compare`.
It being used in the core-lib, which is otherwise mostly ISO-C, was a bit surprising to me, but there is no sane way to iterate a directory in pure ISO-C afaik.

MrSmile: I think it's mostly safe to use POSIX functions. And for OSes without (complete) POSIX implementation (Windows) we can use `ass_compat.h` with fallbacks.

MrSmile:
New version without type punning and simplified error handling.

As for case-insensitive comparison, it's as I feared: it can be problematic to use private library symbols. So, should I use `strcasecmp` from POSIX (I already use `readdir` anyway) with its locale complications or write yet another simple implementation?

compare/README.md
```diff
@@ -40,6 +41,9 @@ More precisely, the test program (`compare`) would scan the input directory and
 5) fuzzy compare the rendering result with the corresponding png file;
 6) save the rendering result in the output directory if requested.
 
+Alternatively, image files can be placed into separate image directories, specified by `-t` command line switch.
```
TheOneric, line 44: Nits: “Alternatively” imho sounds as if images would *no* longer be loaded from the input dir when `-t` is passed, which is not the case. “Additionaly” might be better.
Also I think just using Unicode `–` or just ASCII `-` instead of HTML's `&ndash;` would be more readable, but `&ndash;` is already used before this PR.

TheOneric (commented):
> As for case-insensitive comparison, it's as I feared: it can be problematic to use private library symbols. So, should I use strcasecmp from POSIX (I already use readdir anyway) with its locale complications

Since `compare` is not a library you can just use `setlocale(LC_ALL, "C");` from ISO-C99's `locale.h` or use `strcasecmp_l` from POSIX's `strings.h` (included in the standard since Issue 7 from 2008 corresponding to Single-UNIX-Specification v4).
As you noted on IRC however, file insensitive extensions can lead to issues if eg both `a.ext` and `a.EXT` exists, which is probably more of an issue for `ass` and `png` than for fonts.

astiob:
> file insensitive extensions can lead to issues if eg both `a.ext` and `a.EXT` exists, which is probably more of an issue for `ass` and `png` than for fonts.

Perhaps match case-insensitive extensions just for fonts, then. Or fonts and PNGs. This would be worth noting in the readme, but functionally, I think this would be fine. It’s just that fonts specifically tend to come un-renamed from external sources and sometimes do have uppercase extensions.

compare/README.md
astiob: (I agree on `&ndash;`: Markdown is meant to be readable as plain text. But it’s understandable enough, and especially if it’s already there, oh well.)

MrSmile:
Updated version with many improvements:
* specialized image directories replaced by generic input directories;
* new bitwise comparison mode;
* extension case-insensitivity;
* API improvements (libass log in stderr, exit(2) on errors).

I think that second commit "replace str(n)dup" isn't needed, as `strndup` has eliminated in "case-insensitive comparisons" and `strdup` exists under Windows. I'm planning to remove it in the final version if there're no objections.

TheOneric:
> I think that second commit "replace str(n)dup" isn't needed, as strndup has eliminated in "case-insensitive comparisons" and strdup exists under Windows. I'm planning to remove it in the final version if there're no objections.

I have no objection to dropping the `strdup` part, but I think the `strndup` part should be merged into `fix missing mkdir under MinGW` as otherwise it still wouldn't compile under MinGW.

MrSmile:
> otherwise it still wouldn't compile under MinGW.

Only a couple of commits before case-insensitive. I can move MinGW commit later to preserve logic.

TheOneric:
> I can move MinGW commit later to preserve logic.

I'd prefer to have MInGW fixes first and merge them before the rest, so we can enable regression tests on Windows.
So if you were to drop the `strdup` part of commit 3 and squash the `strndup` part of commit 3 into commit 2, I'd support merging the resulting first two commits directly.

TheOneric:
Thanks! First two commits (up to a2e56bf9a36f84e094067cf45d2f1f9dbd57f291) now [compile in Windows](https://github.com/TheOneric/libass-regression-tests-tmp/runs/2535875923?check_suite_focus=true). I think those two can now already be merged, allowing us to enable regular Windows regression tests.
Will look at the rest of this PR soon.

compare/README.md
```diff
@@ -5,11 +5,12 @@ To build a test utility configure libass with the `--enable-compare` flag.
 The utility works with `png` image files so there is external dependency of libpng.
 
 Test program command line:  
-`compare ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:1-8>]`
+`compare ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:0-8>]`
```
TheOneric, line 8: imho strict/detailed mode should be a separate flag.

compare/README.md
```diff
@@ -65,6 +66,7 @@ Only 3 of 4 images have passed test
 For each target image file the program reports a maximal ratio of the per pixel comparison error to the baseline error scale.
 The baseline error scale tries to take account of neighboring pixels to estimate visibility of an error in that specific pixel location.
 Currently the range of [0–2.0] is marked as `OK`, the range of [2.0–4.0] as `BAD` and more than 4.0 as `FAIL`.
+In the bitwise comparison mode (`-s 0`) any error value larger than zero is marked as `FAIL`.
```
TheOneric, line 69: It would be more useful, to preserve the categorisation and let compare return a different value depending on the maximum deviation level in strict/detailed mode. Eg: `0` for flawless, `1` for `OK` (but not all zero), `2` for `BAD`, `3` for `FAIL` and some other larger value for program errors.
This conceptually allows to eg mark CI runs as `neutral` when the result isn't flawless but not yet bad enough for a `FAIL`, once either GHA re-adds neutral success states or we switch the CI provider again.

TheOneric (commented):
btw — no need to change this in this pr — but since `compare` relies on MinGW-fallbacks already anyway, we could also just use `getopt` for parsing the command line flags.

compare/compare.c
```diff
@@ -448,20 +439,88 @@ static bool add_img_item(ItemList *list, const char *file, size_t len)
...
+
+fail:
+    free(pos);
     const char *fmt =
-        "Usage: %s [-i] <input-dir> [-o <output-dir>] [-s <scale:1-8>]\n";
-    printf(fmt, program);
-    return 1;
+        "Usage: %s ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:1-8>]\n";
+    printf(fmt, argv[0]);
+    return NULL;
```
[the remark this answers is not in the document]
MrSmile: This function (`parse_cmdline`) returns `int *` so `NULL` here is correct.

compare/compare.c
```diff
@@ -596,40 +593,50 @@ int main(int argc, char *argv[])
...
-
         total++;
         if (!track)
             continue;
-        char *name = list.items[i].name;
         name[prefix] = '-';  // restore initial filename
-        if (process_image(renderer, track, input, output,
-                          name, list.items[i].time, scale))
+        if (process_image(renderer, track, list.items[i].dir, output,
+                          list.items[i].name, list.items[i].time, scale))
```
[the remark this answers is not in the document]
MrSmile: Ah, It's an artifact from several rebases and shuffling things around. I'll fix it.

compare/README.md
MrSmile: Should I add additional flag for that: `-p <pass-level:0-3>`? Each image gets its own result flag (0:SAME, 1:OK, 2:BAD, 3:FAIL, 4:ERROR) and if that flag is less or equal than pass level then image passes. If all images pass then `compare` returns 0, otherwise returns maximum result flag. So default pass level of 2 is equivalent to the current behavior and pass level 0 is bitwise comparison mode. Also pass level 3 may be useful: it's error checking mode that doesn't care about image comparison results at all.

compare/compare.c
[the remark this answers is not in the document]
TheOneric: Ahh yes, sorry. Probably got confused by the `return 1` above.

compare/README.md
TheOneric: Won't just having different return codes allow all the same functionality as `-p <pass-level>` while being more flexible?

MrSmile: I. e. always have `-p 0`? I can see two problems:
* Program returns nonzero even if all images pass. Not sure how problematic it can be.
* Old format for final summary isn't enough. Well, I can do something about it, for example print counts for all categories.

TheOneric: > I. e. always have -p 0?

No, those return values would only be used if a flag for strict/detailed mode was passed, eg plain `-c` without any arguments  *(it's strict in the sense that anything but flawless returns non-zero, and detailed because the return code specifies how much the worst results deviated)*. If I understood you correctly, the `-p <pass-level>` you proposed only ever returns binary “worst result was below/above specified level”.
With this binary¹ result it would be impossible to classify “flawless” as OK, “not failed” as neutral and “FAIL” as failed in the CI — unless you run the tests twice. Even if the result is non-zero, it is still possible to use this strict mode with `set -e`-scripts, by eg using something like this:
```sh
#/bin/sh
set -e
ret=0
compare . -c $COMP_ARGS || ret="$?"
if [ "$ret" -eq 0 ] ; then
  echo "::set-result=OK"
elif [ "$ret" -le "2" ] ; then
  echo "::set-result=NEUTRAL"
else
  echo "::set-result=FAIL"
fi
```

The `-p <pass-level>` option could be emulated with a simple `test "$ret" -le "$pass_level"`.
```

---

¹: ternary with program error

Comment (MrSmile) [reply compare/README.md:69]: > No, those return values would only be used if a flag for strict/detailed mode was passed, eg plain `-c` without any arguments

No, I mean to replace `-c` with `-p`. So instead of a simple binary flag I'm planning to use more flexible approach.

> If I understood you correctly, the `-p <pass-level>` you proposed only ever returns binary “worst result was below/above specified level”.

I'm planning to return `(worst <= pass_level ? 0 : worst)`, so `-p 0` would be exactly equivalent to `-c`.

Comment (TheOneric) [reply compare/README.md:69]: > I'm planning to return (worst <= pass_level ? 0 : worst), so -p 0 would be exactly equivalent to -c.

Well, then at least for the return code `-p <pass-level>` is fine and basically provides a convenient  "automatic-`(test "$ret" -le "$pass" || exit "$ret")`".

But still a nit: marking everything as `FAIL` in the log mkaes it not impossible, but a tad harder to tell which "failing" tests failed how bad, as you actually need to compare all numbers and lose the classification into groups for everything above `pass-level`. But as I wrote, that's more a nit.

Comment (MrSmile) [reply compare/README.md:69]: I'm planning to keep uniform log independent of `-p` level. Only final summary and exit code would be affected.

Comment (TheOneric) [reply compare/README.md:69]: In that case, the option is fine, but the Documentation in README should state that the return values differ and not just state “error value larger than zero is marked as `FAIL`.”.

File: compare/compare.c [hunk 7]
```diff
@@ -359,23 +360,12 @@ typedef struct {
     Item *items;
 } ItemList;
 
-static bool init_items(ItemList *list)
```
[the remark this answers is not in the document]
MrSmile: Done.

compare/compare.c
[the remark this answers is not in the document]
MrSmile: Done.

compare/README.md
MrSmile: I've created pass-level flag, which encompasses an old strict mode.

compare/README.md
MrSmile: Done.

MrSmile:
New version with separated refactoring commit and command line switch for acceptable level of image differences.

compare/compare.c
```diff
@@ -578,6 +602,16 @@ int main(int argc, char *argv[])
         scale = arg[0] - '0';
     }
 
+    int level = 2;
```
TheOneric, line 605: Nit: imho writing `int level = R_BAD;` would be better

TheOneric (approved):

compare/compare.c
MrSmile: Done.

PR #501, opened 2021-05-01
compare: enable compilation under MinGW/Windows

TheOneric:
**post 0.15.1** patch:

This enables us to run regression tests on MS Windows.
As an optional, non-installed testing utility, compare doesn't need to
have as broad support as libass itself. In the interst of brevity, the
compat fallbacks are kept to the bare minimum to work with MinGW.

`ass_compat.h` is not used to continue to allow dynamic linking for testing purposes *(or linking with a stripped static lib)*.

Since they were touched anyway, includes were reordered, system first, then local. Each group is is sorted alphabetically.

### DIFF
--- a/compare/compare.c
+++ b/compare/compare.c
@@ -16,13 +16,36 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
-#include "image.h"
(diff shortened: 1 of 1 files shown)


TheOneric:
MrSmile already works on this with a cleaner approach.

PR #502, opened 2021-05-03
Add GDI Font Matching Fallback

Ristellise:
This PR attempts to solve Font Manager related issues by introducing a font matching fallback through `CreateFontIndirect`.  
Note that this is a draft PR for the following reasons:
 - [ ] Figure out why does CreateFontFaceFromHdc throws an Access Violation even though SelectObject appears to be valid.
 - [ ] Remove all the `ass_msg` debug logging. 
 - [ ] Check if it really fixes font manager issues.

Currently font managers such as fontbase write directly their values to registry in a way which directwrite is not able to find through their own methods, so when any user tries to load a font throught fontbase for example, the font appears  
To bypass the issue, Aegisub has a GDI font selection patch that overrides libass directwrite font selection behavior.--

This PR's inspiration came mainly from this Chromium [patch](https://codereview.chromium.org/228763008/patch/1/10002).

### DIFF
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -694,7 +694,45 @@ static void match_fonts(void *priv, ASS_Library *lib,
 
     hr = IDWriteGdiInterop_CreateFontFromLOGFONT(provider_priv->gdi_interop, &lf, &font);
     if (FAILED(hr) || !font)
-        return;
+    {
+        ass_msg(lib, MSGL_INFO, "Failed to create iterop for font: \"%s\". Trying GDI fallback.", name);
+        HFONT hf = CreateFontIndirect(&lf);
+        HDC dc = CreateCompatibleDC(NULL);
+        IDWriteFontFace* fontFace;
+        HGDIOBJ response = SelectObject(dc, hf);
+        if (response == NULL)
+        {
+            DeleteDC(dc);
+            DeleteObject(hf);
+            ass_msg(lib, MSGL_INFO, "GDI fallback failed due to SelectObject", name);
+        }
+        hr = IDWriteGdiInterop_CreateFontFaceFromHdc(provider_priv->gdi_interop, dc, &fontFace);
+        if (FAILED(hr) || !fontFace)
+        {
+            ass_msg(lib, MSGL_INFO, "GDI fallback failed due to CreateFontFaceFromHdc.", name);
+            DeleteDC(dc);
+            DeleteObject(hf);
+            return;
+        }
+        IDWriteFontCollection* font_coll = NULL;
+        IDWriteFactory_GetSystemFontCollection(provider_priv->factory, &font_coll, false);
+        if (FAILED(hr))
+        {
+            ass_msg(lib, MSGL_INFO, "GDI fallback failed due to GetSystemFontCollection.", name);
+            DeleteDC(dc);
+            DeleteObject(hf);
+            return;
+        }
+        IDWriteFontCollection_GetFontFromFontFace(font_coll, &fontFace, &font);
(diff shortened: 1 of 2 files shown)


astiob:
Quick question: do you know if there’s a way to reimplement `get_fallback` in GDI? If so, we could replace the entire DirectWrite provider with GDI and gain a native provider for Windows XP for any poor (or proud) souls who might still be using it.

Ristellise:
> Quick question: do you know if there’s a way to reimplement `get_fallback` in GDI? If so, we could replace the entire DirectWrite provider with GDI and gain a native provider for Windows XP for any poor (or proud) souls who might still be using it.

Should be possible, the overall process that Aegisub uses a `EnumFontFamiliesEx` to enumerate through all the font families (with a filter provided with a `LOGFONT` as filter input) and then adding it manually.

astiob:
Uhh, that sounds like preloading the whole font list?

I see Aegisub use `EnumFontFamiliesEx` to find fonts by name, but I don’t see how it helps with font fallback.

Actually, haven’t we broken the DirectWrite provider with fonts of multiple styles? Now `match_fonts` finds only a single font for each name, instead of the whole family. Presumably `EnumFontFamiliesEx` fixes that.

Ristellise:
> so this will make libass work in different way (better or more close to vsfilter) in windows?

In theory it should. I probably need to rethink this when I have the time for it.

astiob:
@Ristellise What’s the current state of this PR? The code seems benign (and has been independently recreated [?] in https://github.com/libass/libass/pull/503#issuecomment-846585523), but you said there was an Access Violation error somewhere. Can we help debug that? Is there a particular way to reproduce it?

Also, are you saying these fonts do actually work in Aegisub’s libass? That is, is it safe to assume they work with `EnumFontFamiliesEx`?

Ristellise:
> @Ristellise What’s the current state of this PR? The code seems benign (and has been independently recreated [?] in [#503 (comment)](https://github.com/libass/libass/pull/503#issuecomment-846585523)), but you said there was an Access Violation error somewhere. Can we help debug that? Is there a particular way to reproduce it?
> 
> Also, are you saying these fonts do actually work in Aegisub’s libass? That is, is it safe to assume they work with `EnumFontFamiliesEx`?

> What’s the current state of this PR?  
Feel free to fork. Currently I'm trying to finish up my fansubbing stuff (Its taking a while), so I dont have motivation to work on it.  
Feel free to close and move it to [#503 (comment)](https://github.com/libass/libass/pull/503#issuecomment-846585523)) if necessary.

>   but you said there was an Access Violation error somewhere...
Specifically the line which `CreateFontFaceFromHdc` is called, throws an Access Violation.

> Is there a particular way to reproduce it?  
Currently building my own libass over at AegisubDC, It might be a bit annoying to setup as it uses Visual Studio to build the entire project.

Another way to reproduce it is to try to load it through a font manager like fontbase and check it out from there, though this method seems a bit unreliable as it sometimes works while sometimes is doesnt. It's really dependent if the user experiences it or not.

> Also, are you saying these fonts do actually work in Aegisub’s libass?  
According to reports I'm getting, yes.

Apache553:
> Specifically the line which `CreateFontFaceFromHdc` is called, throws an Access Violation.

You can refer to [this](https://github.com/Apache553/libass/commit/5848461e2eb447245514ebcc2053b6187d0bd566), in the dwrite_c.h. You lack some dummy vtable entries.

astiob:
Ah, nice, thanks! That would explain it.

Ristellise:
Closing in favor of #503 with a more reasonable approach, refer to latest comments.

PR #506, opened 2021-05-05
Use const char* and ASS_StringView more

rcombs:
- Convert many uses of `char*` to `const char*`
- Fix a few format string bugs
- Use ASS_StringView in a variety of internal `ass.c` and `ass_utils.c` cases
- Change `char*` to `const char*` in several public functions' args
- Improve vsfilter compat in a few subtle ways
- Fix `ass_process_data` not handling lines split across multiple chunks


libass/ass.c
```diff
@@ -485,24 +498,20 @@ static int process_style(ASS_Track *track, char *str)
...
             return -1;
     }
 
-    q = format = strdup(track->style_format);
-    if (!q)
-        return -1;
+    ASS_StringView q = { track->style_format, strlen(track->style_format) }; // format scanning pointer
 
-    ass_msg(track->library, MSGL_V, "[%p] Style: %s", track, str);
+    ass_msg(track->library, MSGL_V, "[%p] Style: %*s", track, (int)p.len, p.str);
```
MrSmile, line 510: Format specifier should be `%.*s` as string length is precision.

libass/ass.c
```diff
@@ -701,9 +701,9 @@ static void event_format_fallback(ASS_Track *track)
 {
     track->parser_priv->state = PST_EVENTS;
     if (track->track_type == TRACK_TYPE_SSA)
-        track->event_format = strdup(ssa_event_format);
+        track->event_format = strdup(ssa_event_format.str);
```
MrSmile, line 704: `ass_copy_string` here too.

libass/ass.c
```diff
@@ -811,7 +809,7 @@ static int process_events_line(ASS_Track *track, char *str)
 
         return process_event_tail(track, event, str, 0);
     } else {
-        ass_msg(track->library, MSGL_V, "Not understood: '%.30s'", str);
+        ass_msg(track->library, MSGL_V, "Not understood: '%*s'", (int)str.len, str.str);
```
MrSmile, line 812: Also `%.*s`. And if we want exact equivalent then it should be `(int) FFMIN(str.len, 30)` too.

libass/ass.c
```diff
@@ -903,25 +898,24 @@ static int process_fonts_line(ASS_Track *track, char *str)
     }
 
     if (!track->parser_priv->fontname) {
-        ass_msg(track->library, MSGL_V, "Not understood: '%s'", str);
+        ass_msg(track->library, MSGL_V, "Not understood: '%*s'", (int)str.len, str.str);
```
MrSmile, line 901: Here too.

libass/ass.c
```diff
@@ -971,54 +965,122 @@ static int process_line(ASS_Track *track, char *str)
...
+
+    // Absurd? Yes.
+    if (str.len >= SIZE_MAX - 1)
+        return -ENOMEM;
+
+    char *buf = malloc(str.len + 1);
+    if (!buf)
+        return -ENOMEM;
+
+    str.str = buf;
```
MrSmile, line 1013: Buffer with garbage. Do you mean `str.str = ass_copy_string(str);`?

libass/ass.c
```diff
@@ -971,54 +965,122 @@ static int process_line(ASS_Track *track, char *str)
...
  * \param data string to parse
  * \param size length of data
 */
-void ass_process_data(ASS_Track *track, char *data, int size)
+void ass_process_data(ASS_Track *track, const char *data, int size)
 {
-    char *str = malloc(size + 1);
-    if (!str)
-        return;
+    ass_msg(track->library, MSGL_DBG2, "Event: %*s", size, data);
```
MrSmile, line 1030: Here too.

libass/ass.c
```diff
@@ -971,54 +965,122 @@ static int process_line(ASS_Track *track, char *str)
...
+
+        if (pos == size) {
+            if (track->process_data_buf_size >= SIZE_MAX - size)
+                goto wipe;
 
-    memcpy(str, data, size);
-    str[size] = '\0';
+            char *newbuf = realloc(track->process_data_buf, track->process_data_buf_size + size);
+            if (!newbuf) {
+wipe:
```
MrSmile, line 1045: Better to move cleanup block at the end of the function.

libass/ass.c
```diff
@@ -1083,47 +1144,41 @@ void ass_process_chunk(ASS_Track *track, char *data, int size,
...
     }
 
-    str = malloc(size + 1);
-    if (!str)
-        goto cleanup;
-    memcpy(str, data, size);
-    str[size] = '\0';
-    ass_msg(track->library, MSGL_V, "Event at %" PRId64 ", +%" PRId64 ": %s",
-           (int64_t) timecode, (int64_t) duration, str);
+    ass_msg(track->library, MSGL_V, "Event at %" PRId64 ", +%" PRId64 ": %*s",
```
MrSmile, line 1150: Here too.

libass/ass_string.c
```diff
@@ -73,3 +77,18 @@ int ass_strncasecmp(const char *s1, const char *s2, size_t n)
     return a - b;
 }
 
+bool ass_sv_iequal_cstr(ASS_StringView s1, const char *s2)
```
MrSmile, line 80: I'm not 100% sure, but isn't it equivalent to `ass_strncasecmp(s2, s1.str, s1.len)`?

libass/ass_types.h
```diff
@@ -207,6 +207,9 @@ typedef struct ass_track {
     ASS_Library *library;
     ASS_ParserPriv *parser_priv;
 
+    char *process_data_buf;
+    size_t process_data_buf_size;
```
MrSmile, line 211: I think it's possible to use `ASS_StringView process_data_buf;` for that.

libass/ass_utils.c
```diff
@@ -150,6 +150,16 @@ void *ass_try_realloc_array(void *ptr, size_t nmemb, size_t size)
     }
 }
 
+bool ass_sv_iequal(ASS_StringView str1, ASS_StringView str2)
```
MrSmile, line 153: I think this two functions is short enough to be inline in `ass_utils.h`.

MrSmile (commented):
Did a superficial review, without diving deep into logic. I'm planning to look more thoroughly later, but here are the first impressions.
* Style: type casts should be with a space `(int) str.len` and postfix `++`/`--` should be preferred when possible.
* `ASS_StringView` is good for mostly immutable strings, but I think that something like `struct arg` with `end` instead of `len` is better for dynamic parsing with moving boundaries. `str.str += n, str.len -= n;` looks quite cumbersome.
* I think that instead of `some_function((ASS_StringView) { buf, size })` it's better to use `some_function(buf, size)` directly and create `ASS_StringView` inside function.

libass/ass_types.h
TheOneric: `ASS_StringView` is an internal struct; `ass_types.h` is a public header, so `ASS_StringView` can't be used here.
However, I think it's possible and would be better to move the bufffer into `parser_priv`.

rcombs: Agreed; moved there.

libass/ass_utils.c
rcombs: They're in the .c to avoid needing to resolve the #includes between ass_utils.h and ass_string.h, but maybe all the ASS_StringView util functions should just move to ass_string.h?

libass/ass_string.c
rcombs: No, that would treat the 2 as equal if `s2` _started_ with `s1`.

libass/ass_drawing.c
```diff
@@ -257,8 +257,8 @@ bool ass_drawing_parse(ASS_Outline *outline, ASS_Rect *cbox,
 
     if (lib)
         ass_msg(lib, MSGL_V,
-                "Parsed drawing with %d points and %d segments",
-                outline->n_points, outline->n_segments);
+                "Parsed drawing with %llu points and %llu segments",
```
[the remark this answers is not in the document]
rcombs: `%zu` doesn't work on some MSVCRT versions, and I wasn't sure if we wanted to support them or not.

libass/ass_font.c
```diff
@@ -474,7 +474,7 @@ int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
         int face_idx;
         ass_msg(font->library, MSGL_INFO,
                 "Glyph 0x%X not found, selecting one more "
-                "font for (%s, %d, %d)", symbol, font->desc.family,
+                "font for (%*s, %d, %d)", symbol, (int)font->desc.family.len, font->desc.family.str,
```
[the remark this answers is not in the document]
rcombs: The old version passed in the struct itself instead of the string pointer, so it definitely wasn't fine. In theory I probably could've just added `.str`, but I think it's better practice to explicitly pass the length whenever possible.

libass/ass_utils.h
```diff
@@ -104,7 +104,11 @@ int numpad2align(int val);
 unsigned ass_utf8_get_char(const char **str);
 unsigned ass_utf8_put_char(char *dest, uint32_t ch);
 void ass_utf16be_to_utf8(char *dst, size_t dst_size, const uint8_t *src, size_t src_size);
-void ass_msg(ASS_Library *priv, int lvl, const char *fmt, ...);
+void ass_msg(ASS_Library *priv, int lvl, const char *fmt, ...)
+#ifdef __GNUC__
+    __attribute__ ((format (printf, 3, 4)));
```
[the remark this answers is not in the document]
rcombs: Whoops! Fixed.

libass/ass.c
```diff
@@ -41,21 +41,37 @@
...
 
-static const char *const ass_style_format =
+static const ASS_StringView ass_style_format = ASS_SV(
         "Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, "
         "OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, "
         "ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, "
-        "Alignment, MarginL, MarginR, MarginV, Encoding";
-static const char *const ass_event_format =
+        "Alignment, MarginL, MarginR, MarginV, Encoding"
+        );
```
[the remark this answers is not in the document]
rcombs: Sure.

libass/ass.c
```diff
@@ -226,11 +243,14 @@ static void set_default_style(ASS_Style *style)
...
-static long long string2timecode(ASS_Library *library, char *p)
+static long long string2timecode(ASS_Library *library, ASS_StringView p)
 {
     int h, m, s, ms;
     long long tm;
-    int res = sscanf(p, "%d:%d:%d.%d", &h, &m, &s, &ms);
+
+    // This is never the last field in a line, so this can't overflow;
+    // TIMEVAL ensures this
+    int res = sscanf(p.str, "%d:%d:%d.%d", &h, &m, &s, &ms);
```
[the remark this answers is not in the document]
rcombs: I don't think `%d` parsing is locale-dependent? That said, it'd be nice to have this move over to something that takes a `StringView`

libass/ass.c
```diff
@@ -386,48 +400,54 @@ static int process_event_tail(ASS_Track *track, ASS_Event *event,
  */
 void ass_process_force_style(ASS_Track *track)
 {
-    char **fs, *eq, *dt, *style, *tname, *token;
+    char **item;
```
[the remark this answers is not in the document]
rcombs: Sure.

libass/ass.c
```diff
@@ -485,24 +498,20 @@ static int process_style(ASS_Track *track, char *str)
         // no style format header
         // probably an ancient script version
         if (track->track_type == TRACK_TYPE_SSA)
-            track->style_format = strdup(ssa_style_format);
+            track->style_format = strdup(ssa_style_format.str);
```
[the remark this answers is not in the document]
rcombs: Ah, yup.

libass/ass.c
rcombs: Oh, whoops, good catch.

libass/ass_drawing.c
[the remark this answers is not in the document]
MrSmile: Both `ll` and `z` introduced in C99 and it's libass requirement anyway. Also a quick search shows that MSVC supports `z` at least since 2015. I think it's safe to use.

libass/ass.c
rcombs: Sure, I guess it's not absurd to limit how much garbage this can spew on corrupt input.

libass/ass.c
rcombs: Whoooooops! Almost; I'd meant to terminate the line with a `\n`. Fixed.

libass/ass.c
rcombs: Sure.

libass/ass_drawing.c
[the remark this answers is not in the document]
rcombs: I've run into issues with this in MSVCRT in the past (where lld worked fine), but we do already have a `%zu` usage in `decode_font`, so sure, fair enough.

libass/ass_utils.c
MrSmile: > maybe all the ASS_StringView util functions should just move to ass_string.h?

Good idea, I should've done that when I introduced `ASS_StringView`. But in that case it's better to move all string-related functions from `ass_utils.h` to `ass_string.h`.

rcombs:
> Style: type casts should be with a space (int) str.len and postfix ++/-- should be preferred when possible.

I think I used postfix ++/-- in cases I introduced here? There were probably a fair few in code that just moved or was tweaked; didn't want to make the PR even bigger than it already was by changing them all.
Did we decide on spaces in casts at some point? I can't say I like that style.

> ASS_StringView is good for mostly immutable strings, but I think that something like struct arg with end instead of len is better for dynamic parsing with moving boundaries. str.str += n, str.len -= n; looks quite cumbersome.

I much prefer having a single type we can use in all of these cases, so we don't need multiple sets of functions to perform similar tasks. That said, it'd probably be useful to have some inline utility routines like `ass_sv_advance(&sv, n)`, `ass_sv_inc(&sv)`, etc. so we don't have to ++/-- directly all the time.

> I think that instead of some_function((ASS_StringView) { buf, size }) it's better to use some_function(buf, size) directly and create ASS_StringView inside function.

Disagree; most of these cases involve passing an already-existing sv, and some are used both ways.

libass/ass_utils.c
rcombs: Moved these to `ass_string.h` as discussed.

rcombs:
Re-pushed:
- Moved ASS_StringView and the directly-related routines to ass_string
- Split out the new generic ASS_StringView utility functions into their own commit
- Added new routines for get/peek from sv, and made use of them (see e.g. `read_digit`)
- Split an unrelated edge-case fix in ass_string into its own commit

astiob (commented):
Nice work!

* Is there any chance of splitting the big commit into parts, with separate commits for the VSFilter-compatible integer parsing and especially for the bufferizing `ass_process_data`? The latter seems like a pretty significant API change, which deserves more visibility. Splitting out also the case-insensitive matches would be completely grand, but I imagine that’d pretty much need a new commit written from scratch.

* The `ass_process_data` change includes another hidden API break. If I’m reading this code correctly, previously, lines could be delimited by NULs, and this is no longer accepted. Of course, we never advertised this. But we never documented how the function _was_ to be used, either. And we still don’t, for that matter.

* See inline comments: the case-sensitiveness changes break embedded fonts in (presumably) very rare cases.

* It seems more sensible to give `process_text` an extra argument controlling whether it processes the last line or leaves it hanging.

* Personally, I find `ass_sv_` a bit too abbreviated, but I’ll get used to it.

* `ass_string_equal` ends up being an outlier in terms of naming. It would make sense to rename `ass_string_equal` to `ass_sv_equal`.

* `equal` vs `iequal` also strikes me as dangerous naming as it is easy both to misread and to mistype (especially since the full function names are long, making the one-char difference easier to miss). But it’s obviously the shortest, so perhaps that’s a danger we are willing to accept and pray that at least for commits, code review will find any mistakes.

libass/ass.c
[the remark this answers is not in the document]
astiob: By the way, did you know VSFilter skips over commas while looking for the colon? :-) And not on an internal copy of the pointer, either: field splitting resumes after reading the centisecond field.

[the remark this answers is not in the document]
MrSmile: We already have custom number parsing functions, so I think it's not hard to rewrite this function to make it `ASS_StringView`-aware and more VSFilter-compatible.

libass/ass.c
```diff
@@ -261,64 +280,64 @@ static long long string2timecode(ASS_Library *library, char *p)
...
             free(target->name); \
             target->name = new_str; \
         }
 
 #define STARREDSTRVAL(name) \
-    } else if (ass_strcasecmp(tname, #name) == 0) { \
-        while (*token == '*') ++token; \
-        char *new_str = strdup(token); \
+    } else if (ASS_SV_IEQ(tname, #name)) { \
+        while (token.len && *token.str == '*') ++token.str, --token.len; \
```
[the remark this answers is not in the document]
MrSmile: Almost all such advancements check string prefix beforehand, something like case-sensitive `gather_prefix()` would be better. Or it should be named `ass_sv_prefix()` and `ass_sv_iprefix()`?

libass/ass.c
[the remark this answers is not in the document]
astiob: Basically just call `parse_int_header` four times. The tricker (but doable) part is to intersperse them with `next_token`-style searches for the colons and the period.

libass/ass.c
```diff
@@ -660,38 +657,37 @@ static inline void check_duplicate_info_line(const ASS_Track *const track,
...
         check_duplicate_info_line(track, SINFO_LANGUAGE, "Language");
-        char *p = str + 9;
-        while (*p && ass_isspace(*p)) p++;
+        vskip_spaces(&str);
         free(track->Language);
-        track->Language = strndup(p, 2);
-    } else if (!strncmp(str, "; Script generated by ", 22)) {
-        if (!strncmp(str + 22,"FFmpeg/Lavc", 11))
+        track->Language = strndup(str.str, FFMIN(str.len, 2));
+    } else if (GATHER_PREFIX(str, "; Script generated by ")) {
```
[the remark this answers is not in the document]
MrSmile: There are many places where case-sensitive `gather_prefix()` would be useful, I think it should be added to `ass_string.h`. Actually it can fully replace `ass_sv_startswith()`.

libass/ass.c
```diff
@@ -261,64 +280,64 @@ static long long string2timecode(ASS_Library *library, char *p)
...
+        while (token.len && *token.str == '*') ++token.str, --token.len; \
+        char *new_str = ass_copy_string(token); \
         if (new_str) { \
             free(target->name); \
             target->name = new_str; \
         }
 
+static float sv_atof(ASS_StringView val)
+{
+    // Header lines always end in a newline or \0, so this can't overflow
```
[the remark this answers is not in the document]
astiob: You make a good point that `strtod` can’t work with unterminated `ASS_StringView`. It seems that at this point we don’t _need_ this, but this might come up again in the future.

If and when we do need `ass_strtod` on potentially unterminated `ASS_StringView`, well, I suppose we’ll have no choice but to adapt `ass_strtod`. But I still think we should prefer to stay away from it until it’s really necessary, as it requires nontrivial effort to understand the current `ass_strtod` code to properly patch it and considerably raises the bar for any future effort to switch to another `strtod` implementation (which is likely to have more complex code than the current one).

The number format (for floating-point numbers) is defined as “Windows `wcstod` and/or `swscanf("%f")`”. Which generally means “standard C89 `strtod`”. And our implementation is just an old but run-of-the-mill `strtod` implementation (whose quality on extreme inputs I’m willing to bet differs from VSFilter’s, as it’s much simpler than `strtod` implementations in any serious C library from this century) taken verbatim from some third-party source with `ctype.h` calls replaced by locale-invariant `ass_` versions and, much later, edited to fix a few generic issues independent of any Windows behaviour. I think the only thing we’ve ever compared to VSFilter is numbers with `d` as the exponent separator, to confirm that we don’t need to add support for them as it isn’t accepted by whatever versions of Microsoft’s C libraries VSFilter binaries use. But certainly, if we ever do replace the implementation, we should check that we don’t introduce extra features like hexadecimal numbers (unless by then they become common in VSFilter binaries).

Floating-point parsing is pretty complicated. I don’t think we can afford to maintain a floating-point parser as a tightly integrated, core part of libass like we maintain our integer parsers, which are trivial in comparison. Or perhaps it’s better to say I hope that we don’t have to. Although I suppose we do maintain a rasterizer and blur, which only one person has command of :-) But here, I think none of us would have (lasting) full grasp of this code.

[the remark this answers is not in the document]
astiob: > our implementation[’s] quality on extreme inputs I’m willing to bet differs from VSFilter’s

For the record, I’m still willing to bet that our results differ, but Microsoft’s `strtod` [wasn’t always perfect](https://www.exploringbinary.com/visual-c-plus-plus-strtod-still-broken/) either :-) No surprises though. This stuff is hard.

libass/ass.c
```diff
@@ -971,54 +961,132 @@ static int process_line(ASS_Track *track, char *str)
...
+    int ret = process_text(track, &str);
+    if (ret < 0)
+        return ret;
+
+    if (!str.len)
+        return 0;
+
+    // Absurd? Yes.
+    if (str.len >= SIZE_MAX - 1)
+        return -ENOMEM;
```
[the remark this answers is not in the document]
rcombs: I `malloc(str.len + 1)`, though?

libass/ass_utils.c
```diff
@@ -249,62 +260,63 @@ uint32_t parse_color_tag(const char *str)
...
 
 // Return a boolean value for a string
-char parse_bool(const char *str)
+// String must be terminated in
+char parse_bool(ASS_StringView str)
 {
-    skip_spaces(&str);
-    return !ass_strncasecmp(str, "yes", 3) || strtol(str, NULL, 10) > 0;
+    vskip_spaces(&str);
+    return ASS_SV_ISTARTSWITH(str, "yes") || parse_int_header(str) > 0;
```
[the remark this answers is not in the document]
rcombs: Yeah, I saw that, but just wanted to keep things simple here for now.

libass/ass.c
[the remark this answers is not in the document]
astiob: Yes. So `str.len == SIZE_MAX - 1` is fine, in the sense that there’s no overflow and the value fits in `size_t`. But the overflow check has `>=` and reports an error.

TheOneric:
> [From [astiob’s review](https://github.com/libass/libass/pull/506#pullrequestreview-655092169):]
> - Is there any chance of splitting the big commit into parts, with separate commits for the VSFilter-compatible integer parsing and especially for the bufferizing ass_process_data? The latter seems like a pretty significant API change, which deserves more visibility. Splitting out also the case-insensitive matches would be completely grand, but I imagine that’d pretty much need a new commit written from scratch.
>
> - The ass_process_data change includes another hidden API break. If I’m reading this code correctly, previously, lines could be delimited by NULs, and this is no longer accepted. Of course, we never advertised this. But we never documented how the function was to be used, either. And we still don’t, for that matter.


Maybe this already is the “pretty significant API change” from the first point, but I didn't spot it being spelt out:
With `ass_process_data`’s signature left unchanged, but now withholding processing of unterminated lines, won’t this break data, which does not end with a terminating newline?

PR #511, opened 2021-05-18
force to select full-matched font when selecting from preloaded fonts

### LINKED ISSUE #509: other-looking subtitles
Players: mpv, mpc-be (with madvr)
https://mega.nz/file/D9JSQCSB#AsAPetej_kUJ5RA-7uyMyzBZeaZNTw6VMlRSPVKeuFs

Screenshots:
mpv (libass, probably 15.1)
image

XYsubfilter (https://github.com/pinterf/xy-VSFilter/releases)
image

XYsubfilter with libass [probably 15.1] (https://github.com/Masaiki/xy-VSFilter/releases/tag/xy_sub_filter_with_libass)
image

Masaiki:
force to load full-matched font with preloaded fonts, and load most similar font with preloaded fonts and system fonts if the previous step doesn't match.
this may fix #509

### DIFF
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -535,7 +535,7 @@ find_font(ASS_FontSelector *priv, ASS_Library *library,
           ASS_FontProviderMetaData meta, bool match_extended_family,
           unsigned bold, unsigned italic,
           int *index, char **postscript_name, int *uid, ASS_FontStream *stream,
-          uint32_t code, bool *name_match)
+          uint32_t code, bool *name_match, bool force_full_match)
 {
     ASS_FontInfo req = {0};
     ASS_FontInfo *selected = NULL;
@@ -562,13 +562,11 @@ find_font(ASS_FontSelector *priv, ASS_Library *library,
                 // If there's a family match, compare font attributes
                 // to determine best match in that particular family
                 score = font_attributes_similarity(font, &req);
-                *name_match = true;
             } else if (matches_full_or_postscript_name(font, fullname)) {
                 // If we don't have any match, compare fullnames against request
                 // if there is a match now, assign lowest score possible. This means
                 // the font should be chosen instantly, without further search.
                 score = 0;
-                *name_match = true;
             }
 
             // Consider updating idx if score is better than current minimum
@@ -600,6 +598,12 @@ find_font(ASS_FontSelector *priv, ASS_Library *library,
         if (selected != NULL)
             break;
     }
+
+    if (force_full_match && score_min != 0)
+        return NULL;
+
+    if (selected != NULL)
+        *name_match = true;
 
     // found anything?
     char *result = NULL;
@@ -666,7 +670,7 @@ static char *select_font(ASS_FontSelector *priv, ASS_Library *library,
 
     result = find_font(priv, library, meta, match_extended_family,
                        bold, italic, index, postscript_name, uid,
-                       stream, code, &name_match);
+                       stream, code, &name_match, true);
 
     // If no matching font was found, it might not exist in the font list
     // yet. Call the match_fonts callback to fill in the missing fonts
@@ -682,7 +686,7 @@ static char *select_font(ASS_FontSelector *priv, ASS_Library *library,
         }
         result = find_font(priv, library, meta, match_extended_family,
                            bold, italic, index, postscript_name, uid,
-                           stream, code, &name_match);
+                           stream, code, &name_match, false);
     }
 
     // cleanup


TheOneric:
Iinm, this would cause regressions when a variant is not attached but also not installed, which can eg happen for unique fonts if the sub-author relies on bold/italics synthesisation. *(I assume VSFilter probably uses the attached font in such a case, but haven't checked (yet))*
Eg the regular variant of a font of family “MySpecialFontFam” is attached to the mkv, but the subs also use a synthesised bold variant. Now, when checking the preloaded fonts, the regular variant of “MySpecialFontFam” is rejected, as the weight differs, and system fonts are queried for a fallback, but they don't contain any member of “MySpecialFontFam” so any unrelated font may be returned. As a result, the appearance will differ more from the desired one than necessary.

Masaiki:
> Iinm, this would cause regressions when a variant is not attached but also not installed, which can eg happen for unique fonts if the sub-author relies on bold/italics synthesisation. _(I assume VSFilter probably uses the attached font in such a case, but haven't checked (yet))_
> Eg the regular variant of a font of family “MySpecialFontFam” is attached to the mkv, but the subs also use a synthesised bold variant. Now, when checking the preloaded fonts, the regular variant of “MySpecialFontFam” is rejected, as the weight differs, and system fonts are queried for a fallback, but they don't contain any member of “MySpecialFontFam” so any unrelated font may be returned. As a result, the appearance will differ more from the desired one than necessary.

the thing you worried about is not to happen, because the behavoir after system fonts are queried is the same with the previous code. I even want to delete the procedure of finding fonts only on preloaded fonts but leave it for saving time (maybe). and my test supports my oppinion.
One thing I do worry about is that it may differ from VSFilter in the choice of additional fonts and system fonts, if both include a font needed in the subtitle. But at least this code makes improvements and fixes #509.

astiob:
Oh, I seem to have missed this.

This is just… wrong. For instance, `\fnArial\b1` will always select Arial Regular. And this doesn’t exactly fix #509, either.

astiob:
Er no, `Arial` was a wrong example. But you know, fonts where the regular variant has the same full name as the family name.

libass/ass_fontselect.c
```diff
@@ -542,17 +542,15 @@ find_font(ASS_FontSelector *priv, ASS_Library *library,
             ASS_FontInfo *font = &priv->font_infos[x];
             unsigned score = UINT_MAX;
 
-            if (matches_family_name(font, fullname)) {
-                // If there's a family match, compare font attributes
-                // to determine best match in that particular family
-                score = font_attributes_similarity(font, &req);
-                *name_match = true;
-            } else if (matches_full_or_postscript_name(font, fullname)) {
```
astiob, line 550: I suppose this would stop being flat-out wrong if you just reverted this logic reordering.

But this would still be no more than a bandaid. But maybe it’s a good bandaid.

astiob:
For reference, what I want to do to _really_ fix lazy-loading issues is:
* Stop adding system fonts to the main font list at all.
* Ask the system font provider for fonts every single time a font is needed, instead of only upon failure to find one.
* Have the font provider return a list of fonts, instead of adding them to the main font list.
* Have the font provider cache the list of fonts for each query.

This will fix #509 by ensuring we query system fonts, but more than that, it will also fix currently-broken scenarios like:
1. We want font Foo.
2. We have no font Foo yet.
3. We ask the system font provider for Foo.
4. It finds a font “Foo”, but also happens to find a font “Foo Bar” with an alternative name of “Bar”.
5. Later, we want font Bar.
6. We already have Bar from the earlier lookup, so we don’t ask for more fonts.
7. Actually, the system contains several other Bars and we’ve selected the wrong one.

Masaiki:
> For reference, what I want to do to _really_ fix lazy-loading issues is:
> 
> * Stop adding system fonts to the main font list at all.
> * Ask the system font provider for fonts every single time a font is needed, instead of only upon failure to find one.
> * Pass the full font details (including weight and slant) in each font provider query.
> * Have the font provider return a list of fonts, instead of adding them to the main font list.
> * Have the font provider cache the list of fonts for each query.
> 
> This will fix #509 by ensuring we always query system fonts, but more than that, it will also fix currently-broken scenarios like:
> 
> 1. We want font Foo.
> 2. We have no font Foo yet.
> 3. We ask the system font provider for Foo.
> 4. It finds a font “Foo”, but also happens to find a font “Foo Bar” with an alternative name of “Bar”.
> 5. Later, we want font Bar.
> 6. We already have Bar from the earlier lookup, so we don’t ask for more fonts.
> 7. Actually, the system contains several other Bars and we’ve selected the wrong one.
> 
> and:
> 
> 1. We have queries for font Foo with several different weights.
> 2. The system has several Foo fonts installed, but one has Foo as its full name and the others as family name.
> 3. The system font provider is GDI, so it is only able to find the full-name Foo if we tell it the exact weight it has.
> 4. We don‘t pass the weight to the font provider, so we never find that font.

I have known what you want to do next, but it seems this PR without reordering can also fix the proposed scenario.

Masaiki:
All of above considered, would this pr (with out reordering) be accepted as a temporary bandaid? Otherwise I will close it and waiting for the better fix.

libass/ass_fontselect.c
```diff
@@ -601,6 +599,12 @@ find_font(ASS_FontSelector *priv, ASS_Library *library,
             break;
     }
 
+    if (force_full_match && score_min != 0)
+        return NULL;
+
+    if (selected != NULL)
+        *name_match = true;
```
astiob, line 606: Is it right that this is moved here, after the `check_glyph`?

Doesn’t this mean that if the named font doesn’t contain the desired glyph, it will be re-added via `match_fonts` every single time when this glyph attempts to be rendered? That seems like it will both be slow and keep consuming infinitely more memory.

How about reverting the `*name_match` assignments and moving the `force_full_match` check inside the `matches_family_name` branch or the inner `for` loop as `if (force_full_match && score) continue;`?

astiob (changes requested):

PR #491, opened 2021-03-27
Core Text: fix font fallback, check allocations, reuse FT_Library

astiob:
Fixes #457.

Referring back to my list of options in https://github.com/libass/libass/issues/457#issuecomment-742155639, I ended up choosing a combination of (1) and (2) to preserve maximum compatibility with Windows font names. After all, there had to be a reason @rcombs changed the font name reads in the first place!

PR #513, opened 2021-05-23
Font fallback: fix Core Text, enable future improvement

astiob:
As well as a couple of small fixes along the way (as often happens).

This preserves strict VSFilter-compatible explicitly-named-font choice on Core Text, DirectWrite and embedded fonts. (Localized names are problematic, but so they are in VSFilter.)

But when font fallback happens, the whole typographic font family will now be considered on Core Text and the best match will be chosen to match the active`\b` weight, even if the fonts’ GDI-compatible family names are different (e. g. Arial vs Arial Black, or Hiragino Sans W0…W9). This both achieves better visuals (although the ASS’s requested `\b` weight may be different from the main font’s actual weight…) and fixes Core Text fallback for good. In fact, this is how Core Text fallback worked before the font choice changes to make it more VSFilter-compatible.

This PR changes nothing on Fontconfig or DirectWrite. (The first version tried to do this also on DirectWrite to improve quality, but it was incomplete and broken. I am now keeping this PR small to focus on fixing the Core Text regression.)

### DIFF
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -129,21 +129,30 @@ static char *get_name(CTFontDescriptorRef fontd, CFStringRef attr)
 static bool fill_family_name(CTFontDescriptorRef fontd,
                              ASS_FontProviderMetaData *info)
 {
-    if (info->n_family)
-        return true;
-
     char *family_name = get_name(fontd, kCTFontFamilyNameAttribute);
     if (!family_name)
         return false;
 
-    info->families = malloc(sizeof(char *));
-    if (!info->families) {
-        free(family_name);
-        return false;
+    info->extended_family = family_name;
+
+    if (!info->n_family) {
+        family_name = strdup(family_name);
+        if (!family_name) {
+            free(info->extended_family);
+            return false;
+        }
+
+        info->families = malloc(sizeof(char *));
+        if (!info->families) {
+            free(family_name);
+            free(info->extended_family);
+            return false;
+        }
+
+        info->families[0] = family_name;
+        info->n_family++;
     }
 
-    info->families[0] = family_name;
-    info->n_family++;
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


astiob:
Er, this isn’t actually enough yet for DirectWrite. Its `match_fonts` does add the whole typographic family when it finds one, but it doesn’t _find_ a family by its typographic family name.

astiob:
I’ve removed the DirectWrite changes from this PR to focus on fixing the Core Text regression. I’m hoping to open a separate DirectWrite/GDI PR shortly to fix the remaining parts of the DirectWrite regression, so I may or may not include this quality fix there.

Can haz review?

libass/ass_coretext.c
```diff
@@ -129,21 +129,30 @@ static char *get_name(CTFontDescriptorRef fontd, CFStringRef attr)
...
-
     char *family_name = get_name(fontd, kCTFontFamilyNameAttribute);
     if (!family_name)
         return false;
 
-    info->families = malloc(sizeof(char *));
-    if (!info->families) {
-        free(family_name);
-        return false;
+    info->extended_family = family_name;
```
TheOneric, line 136: Not sure if that can actually happen, but the `if (info->n_family) return true` guard suggests `fill_family_name` may be called more than once for the same `ASS_FontProviderMetaData` object. Furthermore, iiuc, the result from `get_name` must be `free`d.
If that's the case, won't the `info->extended_family` assignment leak memory on repeated calls?

astiob: The `n->family` guard is for adding a family if `get_font_info` has failed to find one (since #491/https://github.com/libass/libass/pull/491/commits/96caf3e15c1e945607b2eab6517faf197e93c6e7), and the guard is still present, just moved below because now this function also unconditionally adds an `extended_family`.

`extended_family` is freed in `ass_font_provider_free_fontinfo`.

This function is only called once per font from `get_font_info_ct` from `process_descriptors`. I may have forgotten to add a `free` in `process_descriptors` though.

astiob: Added a `free` in `process_descriptors`, thanks.

PR #514, opened 2021-05-24
Cherry-Pick Format fixes

TheOneric:
This just cherry-picks the second and third commit (currently 4893db52623c85d37a7548401fc5738429de5537 and fbb12d1e0a948947e825ac85e758ca3c437bc34e) from rcombs' #506 so they can be included in 0.15.2.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -474,7 +474,7 @@ int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
         int face_idx;
         ass_msg(font->library, MSGL_INFO,
                 "Glyph 0x%X not found, selecting one more "
-                "font for (%s, %d, %d)", symbol, font->desc.family,
+                "font for (%.*s, %d, %d)", symbol, (int) font->desc.family.len, font->desc.family.str,
                 font->desc.bold, font->desc.italic);
         face_idx = *face_index = add_face(fontsel, font, symbol);
         if (face_idx >= 0) {
@@ -491,8 +491,8 @@ int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
             }
             if (index == 0) {
                 ass_msg(font->library, MSGL_ERR,
-                        "Glyph 0x%X not found in font for (%s, %d, %d)",
-                        symbol, font->desc.family, font->desc.bold,
+                        "Glyph 0x%X not found in font for (%.*s, %d, %d)",
+                        symbol, (int) font->desc.family.len, font->desc.family.str, font->desc.bold,
                         font->desc.italic);
             }
         }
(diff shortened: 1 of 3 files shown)


rcombs (approved):

libass/ass_font.c
```diff
@@ -474,7 +474,7 @@ int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
         int face_idx;
         ass_msg(font->library, MSGL_INFO,
                 "Glyph 0x%X not found, selecting one more "
-                "font for (%s, %d, %d)", symbol, font->desc.family,
+                "font for (%.*s, %d, %d)", symbol, (int)font->desc.family.len, font->desc.family.str,
```
MrSmile, line 477: It's not a bug (and cannot crash) as `ASS_Font`'s `family` is always zero-terminated (and used as such [here](https://github.com/libass/libass/blob/master/libass/ass_fontselect.c#L696), for example). Commit message should be changed accordingly. Also space after `(int)`.

rcombs: No, it's a bug, because it tries to print `font->desc.family` instead of `font->desc.family.str`.

TheOneric: Added spaces after `(int)`, otherwise unchanged.
Imho it's better to use `%.*s` here — even if `%s` also works — to foster consistency and avoid having to wonder if/why the currently used `ASS_StringView` is zero-terminated. At those places, where zero-termination for `ASS_StringView` is known to exist and used, we currently are and imho should be explicit about it, see here
https://github.com/libass/libass/blob/b9f34688f6ef1988136ba5e90a91ae8ac9b8a55f/libass/ass_fontselect.c#L696
and here
https://github.com/libass/libass/blob/b9f34688f6ef1988136ba5e90a91ae8ac9b8a55f/libass/ass_render.c#L1185

MrSmile (approved):

libass/ass_utils.h
```diff
@@ -104,7 +104,11 @@ int numpad2align(int val);
 unsigned ass_utf8_get_char(char **str);
 unsigned ass_utf8_put_char(char *dest, uint32_t ch);
 void ass_utf16be_to_utf8(char *dst, size_t dst_size, uint8_t *src, size_t src_size);
-void ass_msg(ASS_Library *priv, int lvl, const char *fmt, ...);
+void ass_msg(ASS_Library *priv, int lvl, const char *fmt, ...)
+#ifdef __GNUC__
+    __attribute__ ((format (printf, 3, 4)))
+#endif
+;
```
astiob, line 111: Nit: can’t this go before `void ass_msg` to avoid leaving the semicolon all alone on a line?

TheOneric: Seems like it. Both gcc's and clang's doc list examples of function attributes *before* the function and it builds fine (and warns about mistakes) with both gcc and clang for me. Moved it.

astiob (approved):

Issue #126, opened 2014-08-08
Process \clip using subtraction blending?


rcombs:
Subtraction blending can be around twice as fast as multiplication blending on SSE2/AVX2 (especially so on AVX2), and will be at least somewhat faster in scalar code. Scripts where clipping takes significant CPU time aren't common, but do occur. We could probably process \clip as subtraction blending (like \iclip) by rasterizing the clip bitmap inverted and making a few tweaks to the clip handling code.

rcombs:
Ended up doing the inverse, see #521

PR #521, opened 2021-06-14
Make equations for \clip and \iclip complimentary

MrSmile:
There should be `clip_func(src, clip) = iclip_func(src, 255 - clip)`, but our blending equations don't follow that. Also our clip function `sub_bitmaps()` is incorrect in case of half-transparent source and can lead to aliasing. So I've fixed it at the cost of a slight performance drop.

### DIFF
--- a/libass/ass_bitmap.c
+++ b/libass/ass_bitmap.c
@@ -395,16 +395,14 @@ void ass_add_bitmaps_c(uint8_t *dst, intptr_t dst_stride,
     }
 }
 
-void ass_sub_bitmaps_c(uint8_t *dst, intptr_t dst_stride,
-                       uint8_t *src, intptr_t src_stride,
-                       intptr_t width, intptr_t height)
+void ass_imul_bitmaps_c(uint8_t *dst, intptr_t dst_stride,
+                        uint8_t *src, intptr_t src_stride,
+                        intptr_t width, intptr_t height)
 {
-    short out;
     uint8_t* end = dst + dst_stride * height;
     while (dst < end) {
         for (unsigned j = 0; j < width; ++j) {
-            out = dst[j] - src[j];
-            dst[j] = FFMAX(out, 0);
+            dst[j] = (dst[j] * (255 - src[j]) + 255) >> 8;
         }
         dst += dst_stride;
         src += src_stride;
--- a/libass/ass_bitmap.h
+++ b/libass/ass_bitmap.h
@@ -69,7 +69,7 @@ typedef struct {
     FillGenericTileFunc fill_generic;
 
     // blend functions
-    BitmapBlendFunc add_bitmaps, sub_bitmaps;
+    BitmapBlendFunc add_bitmaps, imul_bitmaps;
     BitmapMulFunc mul_bitmaps;
 
     // be blur function
--- a/libass/ass_func_template.h
+++ b/libass/ass_func_template.h
@@ -34,9 +34,9 @@ void DECORATE(fill_generic_tile32)(uint8_t *buf, ptrdiff_t stride,
 void DECORATE(add_bitmaps)(uint8_t *dst, intptr_t dst_stride,
                            uint8_t *src, intptr_t src_stride,
                            intptr_t width, intptr_t height);
-void DECORATE(sub_bitmaps)(uint8_t *dst, intptr_t dst_stride,
-                           uint8_t *src, intptr_t src_stride,
-                           intptr_t width, intptr_t height);
+void DECORATE(imul_bitmaps)(uint8_t *dst, intptr_t dst_stride,
+                            uint8_t *src, intptr_t src_stride,
+                            intptr_t width, intptr_t height);
 void DECORATE(mul_bitmaps)(uint8_t *dst, intptr_t dst_stride,
                            uint8_t *src1, intptr_t src1_stride,
                            uint8_t *src2, intptr_t src2_stride,
@@ -105,7 +105,7 @@ const BitmapEngine DECORATE(bitmap_engine) = {
 #endif
 
     .add_bitmaps = DECORATE(add_bitmaps),
-    .sub_bitmaps = DECORATE(sub_bitmaps),
+    .imul_bitmaps = DECORATE(imul_bitmaps),
     .mul_bitmaps = DECORATE(mul_bitmaps),
 
     .be_blur = DECORATE(be_blur),
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -731,9 +731,9 @@ static void blend_vector_clip(ASS_Renderer *render_priv, ASS_Image *head)
 
             // Blend together
             memcpy(nbuffer, abuffer, ((ah - 1) * as) + aw);
-            render_priv->engine->sub_bitmaps(nbuffer + atop * as + aleft, as,
-                                             bbuffer + btop * bs + bleft, bs,
-                                             w, h);
+            render_priv->engine->imul_bitmaps(nbuffer + atop * as + aleft, as,
+                                              bbuffer + btop * bs + bleft, bs,
+                                              w, h);
         } else {
             // Regular clip
             if (ax + aw < bx || ay + ah < by || ax > bx + bw ||
--- a/libass/x86/blend_bitmaps.asm
+++ b/libass/x86/blend_bitmaps.asm
@@ -60,21 +60,18 @@ SECTION .text
 %endmacro
 
 ;------------------------------------------------------------------------------
-; BLEND_BITMAPS 1:add/sub
+; ADD_BITMAPS
 ; void add_bitmaps(uint8_t *dst, intptr_t dst_stride,
-;                  uint8_t *src, intptr_t src_stride,
-;                  intptr_t width, intptr_t height);
-; void sub_bitmaps(uint8_t *dst, intptr_t dst_stride,
 ;                  uint8_t *src, intptr_t src_stride,
 ;                  intptr_t width, intptr_t height);
 ;------------------------------------------------------------------------------
 
-%macro BLEND_BITMAPS 1
+%macro ADD_BITMAPS 0
 %if ARCH_X86_64
-cglobal %1_bitmaps, 6,8,3
+cglobal add_bitmaps, 6,8,3
     DECLARE_REG_TMP 7
 %else
-cglobal %1_bitmaps, 5,7,3
+cglobal add_bitmaps, 5,7,3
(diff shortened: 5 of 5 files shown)


astiob (approved):
Nice!

But why do we even use `(src * clip + 255) / 256` to begin with? 🤯

rcombs:
I'll note that our current behavior seems to match vsfilter more closely: https://github.com/Cyberbeing/xy-VSFilter/blob/fc01a8da5ea6af9091aaab839bc62dc94a90094e/src/subtitles/RTS.cpp#L1028

astiob:
> our current behavior seems to match vsfilter more closely

I think the code you linked shows this PR’s _new_ behaviour matches more closely. Isn’t that code building a clip mask, not applying it? Note how there’s no multiplication there. So it’s using the original mask for `\clip` and subtracting the mask from all-ones for `\iclip`. Which is exactly what this PR emulates.

TheOneric (approved):
looks good
I assume the different function signature is why `mul_bitmaps` and `imul_bitmaps` can't share a common macro in the assembly as `sub_bitmaps` and `add_bitmaps` did previously?

TheOneric:
> But why do we even use (src * clip + 255) / 256 to begin with? :exploding_head:

Because we actually use `>> 8` for `/ 256` and an addition and bit shift is (probably) faster than an integer or float division.
As the error is `0 ≤ (255^2 - src * clip)/(255 * 256) ≤ 255/256` this will at most lead to different rounding. With a positive error, and integer division and bitshift always rounding down, this might overall even happen to be closer to the exact decimal value than a naive integer `(src * clip) / 255`

libass/ass_bitmap.c
```diff
@@ -395,16 +395,14 @@ void ass_add_bitmaps_c(uint8_t *dst, intptr_t dst_stride,
...
+                        uint8_t *src, intptr_t src_stride,
+                        intptr_t width, intptr_t height)
 {
-    short out;
     uint8_t* end = dst + dst_stride * height;
     while (dst < end) {
         for (unsigned j = 0; j < width; ++j) {
-            out = dst[j] - src[j];
-            dst[j] = FFMAX(out, 0);
+            dst[j] = (dst[j] * (uint8_t) ~src[j] + 255) >> 8;
```
rcombs, line 405: Any reason not to use `255 - src[j]` here? It compiles to the same thing, and imo it's clearer.

rcombs:
@astiob ah, so it does; this is what I get for trying to check vsfilter behavior on mobile.

One nit posted about manually performing an optimization that the compiler's going to do anyway, but otherwise LGTM.

MrSmile:
As there is interest, I've used [this](https://github.com/libass/libass/files/6659297/clip.txt) sub file at 1280&times;720. Relevant parts of the resulting images:
clip

- ref — reference &times;64 oversampled image;
- old — current master;
- fast — subtraction instead of multiplication as per #126 from my [alternative branch](https://github.com/MrSmile/libass/commits/uniform-clip-fast);
- best — this PR, multiplication instead of subtraction.

Red arrows mark aliasing problems.
</details>

libass/ass_bitmap.c
MrSmile: Done.

rcombs:
Thanks much for the demo! Agreed that this looks much better and is definitely the way to go. Changeset LGTM.

astiob:
> and bit shift is (probably) faster than an integer or float division

Division by a constant is optimizable, so I actually checked division by 255 last night, but, sadly, it does require an extra multiplication. Effectively, it would halve the speed of this blending, by going from 1 multiplication to 2.

> As the error is `0 ≤ (255 * 255 - src * clip)/(255 * 256) ≤ 255/256`, this will at most lead to different rounding.

This expression ignores the floor/ceiling in our current formula. But you’re right anyway: plotting/bruteforcing the difference between the exact value of src×clip÷255 and our current rounded value of ⌈src×clip÷256⌉ shows the maximum absolute difference is indeed 255/256 (when src=clip=1).

A plot of ⌈src×clip÷256⌉−src×clip÷255 against src×clip looks like a fixed-width descending band starting at 0 for 0 and 255/256 for 1, with the lower envelope tending from 0 towards −255/256 and the upper envelope from 255/256 towards 0 for increasing values of the argument. These limits are never reached, because the argument src×clip is a product of two integers and its possible values become increasingly scarce as src and clip increase, so starting at src×clip≈58000, the upper envelope stays at 0 and the lower envelope steadily climbs up to 0 in a smooth concave curve.

It’s not quite uniformly distributed and the mean error is 247/1024 (for uniform independent src and clip), but that’s not the end of the world.

> this might overall even happen to be closer to the exact decimal value than a naive integer `(src * clip) / 255`

(For what it’s worth, I wouldn’t even consider an unrounded division.)

MrSmile:
> Division by a constant is optimizable, so I actually checked division by 255 last night, but, sadly, it does require an extra multiplication.

Actually, `(mul + 127) / 255 == ((mul + 128) + (mul + 128) / 256) / 256`, so multiplication can be replaced with shift and add. Not sure that would be faster though. But &plusmn;1 error at the clip boundary is absolutely invisible, so the only strict requirement is that for `clip == 255` holds `dst == src`.

astiob:
> Actually, `(mul + 127) / 255 == ((mul + 128) + (mul + 128) / 256) / 256`

Nice! I’m not going to insist that we use this, but for my own education, how did you come up with this expression?

MrSmile:
> how did you come up with this expression?

Basic outline is from `x * (256.0 / 255.0) ≈ x + (x >> 8)`, after that some trial and error to get additional offsets right.

astiob:
> You forgot the `+ 255`

I didn’t. The +255 applies ceiling.

> As the rounding only happens in the last step _(int-division || bit-shift)_, it's not that relevant. Since as shown, the error with both expression having infinite precision is always smaller than `1`,

I don’t follow. If a>b, a−floor(b) > a−b, so a bound on a−b does not tell you a bound on a−floor(b).

astiob:
> But it would be better to compare the accurately rounded current version to the best we could actually achieve with our integer byte values for alpha, which would be [an exact] calculation whose result is converted to an integer via unbiased rounding

Right, that’s also a good metric. But src×clip is not uniformly distributed and does not fully cover the 0…65025 range. Maybe I’m forgetting something else, but my calculations still show a mean error of −247/1024 for uniformly distributed, independent src and clip.

astiob:
> my calculations still show a mean error of 247/1024 for uniformly distributed, independent src and clip.

Or 1555/8789 (which is smaller, about 0.177) for a value of src×clip uniformly distributed among all possible (deduplicated) products.

Anyway, I think we all agree that this is mainly of academic interest, right? and we’re happy to keep the approximate formul with the bit shift?

TheOneric:
> I think we all agree that this is mainly of academic interest, right? and we’re happy to keep the approximate formula with the bit shift?

yup.

> > You forgot the + 255
>
> The +255 applies ceiling, so I folded it into an actual ceiling operation.

Oups, I missed that; sorry. You're right that this gives the same result. But in that case … why does your graph look so different? It's especially noticeable at the upper end, when I use the ceil, the graph is identical to the image I posted before.

> I don’t follow. If a>b, then a−floor(b) > a−b, so a bound on a−b does not tell you a bound on a−floor(b).

That's correct,but two things: First, we do know more than just `a>b`, we know the exact differences and its bounds; meaning we could derive a (worse) bound iinm. Second, this is comparing to something which can't be output anyway, as the alpha precision is limited to 8-bit integers.
Especially, since the decimal error is larger than zero and smaller than 1, and since **truncation** only happens in the last step, we know, the difference will only be ±1 (so only rounding differs) to the most accurate we can output (and doesn't overflow a byte). (So just by that it already isn't *too* bad)

astiob:
> why does your graph look so different? It's especially noticeable at the upper end, when I use the ceil, the graph is identical to the image I posted before.
> Did you already remove all values which can't be produced by `clip*src`?

Yep! That’s exactly it. (It’s kinda impressive how it affects the very upper end.) I did also make a graph (that I didn’t show) that just scanned all arguments from 0 to 65025, and it looked much like your first graph, so it’s probably safe to assume that this is the only difference.

> since the decimal error is larger than zero and smaller than 1, and since **truncation** only happens in the last step, we know, the difference will only be ±1 (so only rounding differs) to the most accurate we can output

Right… I’m slowly getting it. Please bear with me just a little more. 😅

I think I see how this happens if we assume the same kind of rounding (e. g. truncation) applies to both, or if the smaller value is rounded up and the bigger value down. If their pre-rounding difference is less than 1, then it can’t get any bigger than 1 post rounding.

But if—as in the “best we could ever do using integers without dithering” scenario—we’re rounding the higher of the two values to the nearest integer, which may be rounding up, while we’re always rounding the lower of the two values down, can’t we—in the general case—get a post-rounding difference of 2?

But looking at your fully-unrounded graph again, I realize this doesn’t apply to the specific situation we were discussing, as (clip×src+255)÷256 is actually never smaller than clip×src÷255, so if I had noticed this, then indeed, the difference between floor((clip×src+255)÷256) and round(clip×src÷255) could be analytically understood to be no bigger than 1.

TheOneric:
> But src×clip is not uniformly distributed and does not fully cover the 0…65025 range.

Sure, that's right, I just wasn't too concerned about this significantly changing the result. But now here's the data for this compared to unbiased rounding. FIrst the graph:
opti-int_to_rounded__sparse

The weighted arith. average of the differences to biased rounding is `247/1024 ≈ 0.241`,
There are `3793` cases of the value being one higher and `683` of them being one lower and `13102` where they are identical *(25% of deduplicated values differ by one)*.
There are a total of `18760` **non-deduplicated** `(src, clip)` combinations for which the values differ by one *(29% differ by one)*.


> But if—[…]–we’re rounding the higher of the two values to the nearest integer, which may be rounding up, while we’re always rounding the lower of the two values down, can’t we—in the general case—get a post-rounding difference of 2?

If the larger value where to be rounded up than yes; eg if the difference were `0.6` and we'd have `[2.9, 3.5]` then: `round(3.5) = 4` and `floor(2.9)=2`. But as you already realised this doesn't happen here.

rcombs:
Does anyone have any objection to merging this now, then?

TheOneric:
No objections from me

MrSmile:
I was sure that current plan is to make 0.15.2 with font system related fixes, but if it's not, I can merge this.

rcombs:
If we wanna do a patch release branching off a previous release, we can do that; we don't have to defer merging other stuff.

PR #524, opened 2021-06-20
ci/gha: add musl build

TheOneric:
Inspired by discussion in #523, here it is.
GHA's native Docker support isn't suitable for this, because it doesn't handle mixed non-Docker and Docker job variants well.
The `tail -f /dev/null` ~~hack~~method to keep the container running *(ignoring all zombie processes)* was ~~stolen~~learned from GHA's native Docker setup procedure.

The Docker setup relies on the current default of having the working dir and cloned repos somewhere in `/home/runner/work` and everything in the container will be executed as root.

### DIFF
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+    name: build(${{ matrix.docker_image || matrix.os }}, ${{ matrix.cc }})
     strategy:
       matrix:
         os: [ubuntu-18.04, macos-10.15]
@@ -24,6 +25,11 @@ jobs:
           - os: ubuntu-18.04
             cc: clang
             do_coverity: yes
+          # Build in an Alpine Docker image
+          - os: ubuntu-latest
+            cc: gcc
+            docker_image: alpine:latest
+            shell: '/usr/bin/docker exec dockerciimage sh -e {0}'
           # Add an additional Windows build (MinGW-gcc via MSYS2) with no extras
           - os: windows-2019
             cc: gcc
@@ -37,6 +43,17 @@ jobs:
       - name: checkout code
         uses: actions/checkout@v2
 
+      - name: Start Docker
+        if: matrix.docker_image
+        shell: bash
+        run: |
+          # Note: With this setup everything inside the container will be run as root
+          docker pull "${{ matrix.docker_image }}"
+          docker create --name dockerciimage \
+            -v "/home/runner/work:/home/runner/work" --workdir "$PWD"  \
+            --entrypoint "tail" "${{ matrix.docker_image }}" "-f" "/dev/null"
+          docker start dockerciimage
+
       - name: Setup MSys2
         uses: msys2/setup-msys2@v2
         if: matrix.os == 'windows-2019'
@@ -46,7 +63,7 @@ jobs:
 
       - name: install deps
         run: |
-          case "${{ matrix.os }}" in
+          case "${{ matrix.docker_image || matrix.os }}" in
             macos-*)
               #brew update
               # fontconfig, freetype, autoconf and libtool are preinstalled
@@ -60,6 +77,11 @@ jobs:
                       automake autoconf nasm make \
                       $pre-libtool $pre-pkg-config $pre-gcc \
                       $pre-fribidi $pre-freetype $pre-harfbuzz $pre-fontconfig
+              ;;
+            alpine:*)
+              apk add nasm ${{ matrix.cc }} musl-dev \
+                      make automake autoconf libtool pkgconf \
+                      fontconfig-dev freetype-dev fribidi-dev harfbuzz-dev
               ;;
             *)
               sudo apt-get update #&& sudo apt-get upgrade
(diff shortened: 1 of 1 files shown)


rcombs:
Looks fine to me, though I don't really know enough about the GHA details going on here to be particularly confident.

TheOneric:
To explain how GHA executes job steps:

Steps using `run:` save the commands specified in `run:` to a temporary file 
(in `/home/runner/work/_temp/`) and then call an interpreter on this file.
There's a predefined set of shells which can be used by a name alone. However 
it's also possible to define custom shells, using `{0}` as a placeholder for 
the file name. For more details, see:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

In our case we override GHA's platform dependent shell default with 
`matrix.shell` if existing and `bash` otherwise. For the Docker job this
was specified such as to execute `sh -e {0}` inside the Docker container.
The Docker init and cleanup steps override the job-wide default for 
their own step to always use `bash`. They are executed before/after all
other `run:` steps.

Similarly, the Windows build sets the default shell to `msys {0}`
where this MSys-shell is installed and setup at the beginning by the
external ***@***.***` action.

Steps invoking an external action via `uses:` are not affected
by the selected shell default.

.github/workflows/ghci.yml
```diff
@@ -46,7 +63,11 @@ jobs:
 
       - name: install deps
         run: |
-          case "${{ matrix.os }}" in
+          pkgflavour="${{ matrix.docker_image }}"
+          if [ -z "$pkgflavour" ] ; then
+            pkgflavour="${{ matrix.os }}"
+          fi
+          case "$pkgflavour" in
```
astiob, line 70: In `jobs.build.name`, you use `${{ matrix.docker_image || matrix.os }}`. Can’t that be used here as well?

.github/workflows/ghci.yml
```diff
@@ -24,6 +25,11 @@ jobs:
           - os: ubuntu-18.04
             cc: clang
             do_coverity: yes
+          # Build in an Alpine Docker image
+          - os: ubuntu-latest
+            cc: gcc
+            docker_image: alpine:latest
```
astiob, line 31: Any reason the new build uses Ubuntu and Alpine “latest” but all the others use specific versions?

astiob:
Small typos in the commit message:

> Since GHA only has native support for glibc-linux use ALpine via Docker.

`Alpine`, not `ALpine`.

> Eventhough, GHA does have native Docker support, its

Even though GHA does…

.github/workflows/ghci.yml
TheOneric: Yup, it can and now does, thx.

TheOneric:
> Small typos in the commit message: […]

Fixed, thanks!

.github/workflows/ghci.yml
TheOneric: The idea of the fixed versions was, to keep testing with older but sill support libs *(assuming breaking changes in our dependencies are rare, but accidentally using new APIs is more likely)*. This is somewhat limited by GHA apparently only ever supporting two versions at a time, dropping support for older ones *(meaning we need to change fixed versions every 2 years, or every 4 years if we skip a version)*. Also, originally there were considerations about the no longer existing cache not dealing well with a sudden OS version change.

For Docker builds, the underlying Ubuntu image doesn't matter, we only want to run Docker with it, so “ubuntu-latest” is fine.
Regarding Alpine, “latest” points to the last stable version and I wasn't sure how long each version is supported, but had the impression it is moving relatively fast and thus used “latest”.
Checking the details: a new stable is released every 6 months and while Alpine's “main” repo of each version is supported for 2 years, but support for the “community” repo ends with the next release *(for comparison: Debian has 3 years of full and additional 2 years of LTS support (+ELTS); Ubuntu LTS overall 6 years)*. I'm not sure what sets those two repos apart, some fairly common things like KDE, GNOME and some GUI programs like eg GIMP are in “community”. But it appears like everything we need for the build is in “main”. It would also be possible to follow the latest stable of the same major version *(since Alpine switched from uClibc to musl in 2014, the major version is 3)*.
If you prefer we could track a specific Alpine release or the major version, but on the other hand there might also be a benefit in having *both* builds in an more conservative, older environment (fixed version Ubuntu,MacOS&Windows build) **and** a build with new-ish dependencies (“latest” Alpine build).

astiob: Thanks for the explanation. I think that’s fine, and indeed this may even provide a benefit as you mentioned.

.github/workflows/ghci.yml
astiob: Just a note: you can now inline `pkgflavour` if you’d like.

.github/workflows/ghci.yml
```diff
@@ -161,3 +184,10 @@ jobs:
 
             exit $exit_code
           fi
+
+
```
astiob, line 188: There’s a double blank line here. Might be an accident.

astiob (approved):

.github/workflows/ghci.yml
TheOneric: The Coverity step above already uses single blank lines to segment its shell code, so I used two blank lines to better set the steps apart. Maybe, we could move the Coverity logic into it's own script file and call that from the yml, but where should it be placed?

.github/workflows/ghci.yml
TheOneric: Done, thanks.

PR #526, opened 2021-07-01
use win API to open file.

lifegpc:
This support both ANSI and UTF8 filenames.

### DIFF
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -176,57 +176,57 @@ static ASS_FontProviderFuncs ft_funcs = {
  * \return int 1 for success, 0 for failure
  */
 static int load_fonts_from_dir_win32(ASS_Library *library, const char *dir, UINT codePage) {
-	// The dir name must ending with \*, D:\temp not work, D:\temp\* works well.
-	size_t dir_len = strlen(dir);
-	char* tdir = malloc(dir_len + 3);
-	strcpy(tdir, dir);
-	strcat(tdir, "\\*");
-	WIN32_FIND_DATAW wfdata;
-	HANDLE whFind;
-	wchar_t *wdir;
-	int wlen;
+    // The dir name must ending with \*, D:\temp not work, D:\temp\* works well.
+    size_t dir_len = strlen(dir);
+    char* tdir = malloc(dir_len + 3);
+    strcpy(tdir, dir);
+    strcat(tdir, "\\*");
+    WIN32_FIND_DATAW wfdata;
+    HANDLE whFind;
+    wchar_t *wdir;
+    int wlen;
 
-	wlen = MultiByteToWideChar(codePage, 0, tdir, -1, NULL, 0);
-	if (wlen > 0) {
-		wdir = (wchar_t*)malloc(sizeof(wchar_t) * wlen);
-		if (MultiByteToWideChar(codePage, 0, tdir, -1, wdir, wlen)) {
-			whFind = FindFirstFileW(wdir, &wfdata);
-			if (whFind != INVALID_HANDLE_VALUE) {
-				do {
-					char *filename;
-					int fn_len;
-					fn_len = WideCharToMultiByte(CP_UTF8, 0, wfdata.cFileName, -1, NULL, 0, NULL, FALSE);
-					if (fn_len > 0) {
-						filename = (char*)malloc(fn_len);
-						if (WideCharToMultiByte(CP_UTF8, 0, wfdata.cFileName, -1, filename, fn_len, NULL, FALSE)) {
-							if (filename && filename[0] == '.') {
-								free(filename);
-								continue;
-							}
-							size_t buffsize = strlen(filename) + strlen(tdir) + 2;
-							char *fullname = (char*)malloc(buffsize);
-							snprintf(fullname, buffsize, "%s/%s", dir, filename);
-							size_t bufsize = 0;
-							ass_msg(library, MSGL_INFO, "Loading font file '%s'", fullname);
-							void *data = read_file(library, fullname, &bufsize);
-							if (data) {
-								ass_add_font(library, filename, data, bufsize);
-								free(data);
-							}
-							free(fullname);
-						}
-						free(filename);
-					}
-				} while (FindNextFileW(whFind, &wfdata) != 0);
-				FindClose(whFind);
-				free(tdir);
-				free(wdir);
-				return 1;
-			}
-		}
-		free(wdir);
-	}
-	return 0;
+    wlen = MultiByteToWideChar(codePage, 0, tdir, -1, NULL, 0);
+    if (wlen > 0) {
+        wdir = (wchar_t*)malloc(sizeof(wchar_t) * wlen);
+        if (MultiByteToWideChar(codePage, 0, tdir, -1, wdir, wlen)) {
+            whFind = FindFirstFileW(wdir, &wfdata);
+            if (whFind != INVALID_HANDLE_VALUE) {
+                do {
+                    char *filename;
+                    int fn_len;
+                    fn_len = WideCharToMultiByte(CP_UTF8, 0, wfdata.cFileName, -1, NULL, 0, NULL, FALSE);
+                    if (fn_len > 0) {
+                        filename = (char*)malloc(fn_len);
+                        if (WideCharToMultiByte(CP_UTF8, 0, wfdata.cFileName, -1, filename, fn_len, NULL, FALSE)) {
+                            if (filename && filename[0] == '.') {
+                                free(filename);
+                                continue;
+                            }
+                            size_t buffsize = strlen(filename) + strlen(tdir) + 2;
+                            char *fullname = (char*)malloc(buffsize);
+                            snprintf(fullname, buffsize, "%s/%s", dir, filename);
+                            size_t bufsize = 0;
+                            ass_msg(library, MSGL_INFO, "Loading font file '%s'", fullname);
+                            void *data = read_file(library, fullname, &bufsize);
+                            if (data) {
+                                ass_add_font(library, filename, data, bufsize);
+                                free(data);
+                            }
+                            free(fullname);
+                        }
+                        free(filename);
+                    }
+                } while (FindNextFileW(whFind, &wfdata) != 0);
+                FindClose(whFind);
+                free(tdir);
+                free(wdir);
+                return 1;
+            }
+        }
+        free(wdir);
+    }
+    return 0;
 }
 #endif
 
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -31,7 +31,11 @@
 #include <limits.h>
 #include <ft2build.h>
 #include <sys/types.h>
+#if defined(_WIN32) && !defined(__CYGWIN__)
+#include <windows.h>
+#else
 #include <dirent.h>
+#endif
 #include FT_FREETYPE_H
 #include FT_SFNT_NAMES_H
 #include FT_TRUETYPE_IDS_H
@@ -165,6 +169,7 @@ static ASS_FontProviderFuncs ft_funcs = {
 
 static void load_fonts_from_dir(ASS_Library *library, const char *dir)
 {
+#if !defined(_WIN32) || defined(__CYGWIN__)
     DIR *d = opendir(dir);
     if (!d)
         return;
@@ -185,6 +190,80 @@ static void load_fonts_from_dir(ASS_Library *library, const char *dir)
         }
     }
     closedir(d);
+#else
+    // The dir name must ending with \*, D:\temp not work, D:\temp\* works well.
+    size_t dir_len = strlen(dir);
+    char* tdir = malloc(dir_len + 3);
+    strcpy(tdir, dir);
+    strcat(tdir, "\\*");
+    WIN32_FIND_DATAW wfdata;
+    HANDLE whFind;
+    wchar_t *wdir;
+    int wlen;
+    
+    wlen = MultiByteToWideChar(CP_UTF8, 0, tdir, -1, NULL, 0);
+    if (wlen > 0) {
+        wdir = (wchar_t*) malloc(sizeof(wchar_t) * wlen);
+        if (MultiByteToWideChar(CP_UTF8, 0, tdir, -1, wdir, wlen)) {
+            whFind = FindFirstFileW(wdir, &wfdata);
+            if (whFind != INVALID_HANDLE_VALUE) {
+                do {
+                    char *filename;
+                    int fn_len;
+                    fn_len = WideCharToMultiByte(CP_UTF8, 0, wfdata.cFileName, -1, NULL, 0, NULL, FALSE);
+                    if (fn_len > 0) {
+                        filename = (char*) malloc(fn_len);
+                        if (WideCharToMultiByte(CP_UTF8, 0, wfdata.cFileName, -1, filename, fn_len, NULL, FALSE)) {
+                            if (!strcmp(filename, ".") || !strcmp(filename, "..")) {
+                                free(filename);
+                                continue;
+                            }
+                            size_t buffsize = strlen(filename) + strlen(tdir) + 2;
+                            char *fullname = (char*) malloc(buffsize);
+                            snprintf(fullname, buffsize, "%s/%s", dir, filename);
+                            size_t bufsize = 0;
+                            ass_msg(library, MSGL_INFO, "Loading font file '%s'", fullname);
+                            void *data = read_file(library, fullname, &bufsize);
+                            if (data) {
+                                ass_add_font(library, filename, data, bufsize);
+                                free(data);
+                            }
+                            free(fullname);
+                        }
+                        free(filename);
+                    }
+                } while (FindNextFileW(whFind, &wfdata) != 0);
+                FindClose(whFind);
+                free(tdir);
+                free(wdir);
+                return;
+            }
+        }
+        free(wdir);
+    }
+    WIN32_FIND_DATAA fdata;
+    HANDLE hFind;
+    hFind = FindFirstFileA(tdir, &fdata);
+    if (hFind == INVALID_HANDLE_VALUE) {
+        free(tdir);
+        return;
+    }
+    do {
+        if (!strcmp(fdata.cFileName, ".") || !strcmp(fdata.cFileName, "..")) continue;
+        char *fullname = (char*) malloc(dir_len + MAX_PATH + 2);
+        snprintf(fullname, dir_len + MAX_PATH + 2, "%s/%s", dir, fdata.cFileName);
+        size_t bufsize = 0;
+        ass_msg(library, MSGL_INFO, "Loading font file '%s'", fullname);
+        void *data = read_file(library, fullname, &bufsize);
+        if (data) {
+            ass_add_font(library, fdata.cFileName, data, bufsize);
+            free(data);
+        }
+        free(fullname);
+    } while (FindNextFileA(hFind, &fdata) != 0);
+    FindClose(hFind);
+    free(tdir);
+#endif
 }
 
 /**
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -167,6 +167,62 @@ static ASS_FontProviderFuncs ft_funcs = {
     .destroy_font      = destroy_font_ft,
 };
 
+#if defined(_WIN32) && !defined(__CYGWIN__)
+static int load_fonts_from_dir_win32(ASS_Library *library, const char *dir, UINT codePage) {
+	// The dir name must ending with \*, D:\temp not work, D:\temp\* works well.
+	size_t dir_len = strlen(dir);
+	char* tdir = malloc(dir_len + 3);
+	strcpy(tdir, dir);
+	strcat(tdir, "\\*");
+	WIN32_FIND_DATAW wfdata;
+	HANDLE whFind;
+	wchar_t *wdir;
+	int wlen;
+
+	wlen = MultiByteToWideChar(codePage, 0, tdir, -1, NULL, 0);
+	if (wlen > 0) {
+		wdir = (wchar_t*)malloc(sizeof(wchar_t) * wlen);
+		if (MultiByteToWideChar(codePage, 0, tdir, -1, wdir, wlen)) {
+			whFind = FindFirstFileW(wdir, &wfdata);
+			if (whFind != INVALID_HANDLE_VALUE) {
+				do {
+					char *filename;
+					int fn_len;
+					fn_len = WideCharToMultiByte(CP_UTF8, 0, wfdata.cFileName, -1, NULL, 0, NULL, FALSE);
+					if (fn_len > 0) {
+						filename = (char*)malloc(fn_len);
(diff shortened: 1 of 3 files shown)


TheOneric (changes requested):
This is insufficient, as it doesn't deal with directory traversal in `ass_fontselect.c`'s `load_fonts_from_dir` *(called when the public `ass_set_fonts_dir` was invoked)*.
Some more feedback and questions:

 - the macro check must also consider old Cygwin, i.e. use `#if defined(_WIN32) && !defined(__CYGWIN__)` instead of `#ifdef _WIN32` and analogous for the negation
 - The documentation of related public API must be updated to mention the accepted encodings and their priority **on Microsoft platforms**
 - iirc we agreed in the end that despite the API-break it's probably better to give UTF-8 precedence *(your patch gives ANSI preference)*. This also means you must update `LIBASS_VERSION` in `ass.h` to reflect the API-break.
 - your avoiding a race condition by reusing the stream from `_(w)open`; but why can't you just use `(w)fopen` directly instead of doing the round-trip to `_(w)open` and `_(w)fdopen`? *(iinm this would also allow us to drop the `io.h` and `fcntl.h` includes)*
 
 For reference [here](https://github.com/libass/libass/issues/500#issuecomment-842541104) was a proof-of-concept patch, which also does deal with the dir-traversal. *However*, it also has issues. Summarising the long discussion about it: It also gave ANSI preference, it has a race condition in MS-file-opening *(which you're already avoiding)*, could get a better bound for allocating conversion buffers *(you already do that)* and uses `wreaddir` which is provided by MinGW and some independent implementations, but not by MS. As suggested by astiob, it would be better to use [“`_wfindnext` and friends”](https://docs.microsoft.com/en-us/cpp/c-runtime-library/filename-search-functions?view=msvc-160).
 *(Also in case you were to use this patch as base, you might want to tidy up the comments to remove me venting my frustration about MS' overly complex and broken file and codepage API)*

lifegpc:
> This is insufficient, as it doesn't deal with directory traversal in `ass_fontselect.c`'s `load_fonts_from_dir` _(called when the public `ass_set_fonts_dir` was invoked)_.

I see, `compare/compare.c` also need same patch. I already do a ANSI [patch](https://github.com/lifegpc/libass-MSVC/commit/d1f75bb308cff63457abd22cf8f00f7456693be5#diff-3c78801ab1df4d116480b2868bf6194caaebeb05b59016bbf1bc60bf5756c73cR490) before. ( I try to add MSVC support by using cmake build system)

TheOneric:
> I see, compare/compare.c also need same patch.

`compare` is a useful utility for us to do eg regression testing, but **not** included in the library or even build by default.
Relying on a `readdir` implementation from MinGW or others and being restricted to ASCII (and 8.3) filenames on Microsoft platforms isn't really a problem for this purposes. But ofc, feel free to port it anyway.

> I try to add MSVC support by using cmake build system.

Are you aware of [ShiftMediaProject's libass-repo](https://github.com/ShiftMediaProject/libass)? It contains the last libass-release with MSVC-project files.

lifegpc:
> Are you aware of [ShiftMediaProject's libass-repo](https://github.com/ShiftMediaProject/libass)? It contains the last libass-release with MSVC-project files.

Yea. I see. But I have issue to build libass with that project files.

lifegpc:
image
Looks well now, just need some more works

lifegpc:
Should I update libass API Version to `0x1502000` ?

libass/ass_fontselect.c
```diff
@@ -185,6 +190,80 @@ static void load_fonts_from_dir(ASS_Library *library, const char *dir)
         }
     }
     closedir(d);
+#else
+    // The dir name must ending with \*, D:\temp not work, D:\temp\* works well.
+    size_t dir_len = strlen(dir);
+    char* tdir = malloc(dir_len + 3);
+    strcpy(tdir, dir);
+    strcat(tdir, "\\*");
```
TheOneric, line 198: Will this also work when the path already ends in either a `/` or `\`?

libass/ass_fontselect.c
```diff
@@ -185,6 +190,80 @@ static void load_fonts_from_dir(ASS_Library *library, const char *dir)
...
+            whFind = FindFirstFileW(wdir, &wfdata);
+            if (whFind != INVALID_HANDLE_VALUE) {
+                do {
+                    char *filename;
+                    int fn_len;
+                    fn_len = WideCharToMultiByte(CP_UTF8, 0, wfdata.cFileName, -1, NULL, 0, NULL, FALSE);
+                    if (fn_len > 0) {
+                        filename = (char*) malloc(fn_len);
+                        if (WideCharToMultiByte(CP_UTF8, 0, wfdata.cFileName, -1, filename, fn_len, NULL, FALSE)) {
+                            if (!strcmp(filename, ".") || !strcmp(filename, "..")) {
```
TheOneric, line 217: This is different from the POSIX implementation which ignores all files/dirs starting with a dot. Same further below.

libass/ass_fontselect.c
```diff
@@ -185,6 +190,80 @@ static void load_fonts_from_dir(ASS_Library *library, const char *dir)
...
+                free(tdir);
+                free(wdir);
+                return;
+            }
+        }
+        free(wdir);
+    }
+    WIN32_FIND_DATAA fdata;
+    HANDLE hFind;
+    hFind = FindFirstFileA(tdir, &fdata);
```
TheOneric, line 246: I suspect this will be prone to the exact same errors as the `readdir` implementation, i.e. not all files in the dir can be read with some just not being listed at all and others being listed *(with a fallback name?)* but not being able to be opened, see #500.
Ideally, if UTF-8 conversion or opening fails, the path should be tried to be converted *from ANSI* to MS' “wide char” and tried to be opened via the same “wide char” version as above. This would also allow for less code duplication in the Microsoft-API-path.

Qutoing a relevant section of astiob's comment from #500:
> `         return MultiByteToWideChar(CP_ACP, 0, fname_raw, -1,`
>
> The file path encoding can be CP_OEM, not only CP_ACP, depending on the current value of `AreFileApisANSI`.
>
> There’s also this comment in FFmpeg that says CP_ACP conversions are forbidden in Windows Store. And judging by VLC’s patches to libass, VLC is (or wants to be) in the Windows Store, so they might appreciate us not requiring them to add yet another patch. I haven’t been able to find other sources corroborating this comment’s claim, however.

With the mentioned FFmpeg comment reading:
> […], but doing explicit conversions with CP_ACP is allegedly forbidden in windows store apps (or windows phone), […]

Which also doesn't sound to confident in this restriction actually being a thing. Do you know more about this?
If `CP_ACP` works everywhere you'll "only" have to consider `AreFileApisANSI` for the conversion from "ANSI/OEM", otherwise more ifdeffery it is \o/

TheOneric (changes requested):
Thanks!

> Should I update libass API Version to 0x1502000 ?

No, this would signify a new release `0.15.2`. I'm actually not sure if we have a convention to distinguish breaking and non-breaking API-version changes in-between releases. For now just increase the current value by one.

libass/ass_fontselect.c
lifegpc: No. [MS document](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstfilew#parameters) say:
> This parameter should not be NULL, an invalid string (for example, an empty string or a string that is missing the terminating null character), or end in a trailing backslash (\).

libass/ass_fontselect.c
lifegpc: I can implement a new function with a CodePage parameter. then call that function in `load_fonts_from_dir`

libass/ass_fontselect.c
lifegpc: I am sorry I misunderstand it.
But it works:
image

libass/ass_fontselect.c
lifegpc: Now I put CP_OEMCP before CP_ACP

TheOneric: That's not good, it always tries both OEMCP and ACP and potenttially loads multiple dirs  instead of only the inteded one as *all* encodings are *always* tried.
Ignoring the alleged WinRT restriction for now, here's what you probably want to do according to my understanding:

 1) try decoding as UTF-8
     - if fails, goto `2)`
     - if succesful try to open with `FindFirstFileW`
       - if fails goto `2)`
       - if succesful goto `4)`
  2) check [`AreFileApisANSI()`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-arefileapisansi),
      if true use `CP_ACP` in `3)` for `cur_cp`, otherwise use `CPOEMCP`
  3) try decoding as `cur_cp`
    - if fails abort
    - if succesful try opening with `FindFirstFileW`
       - if fails abort
  4) traverse the opened directory with `FindNextFileW`

TheOneric:
Re CP_ACP conversion; to all interested or knowledgeable in this:
I've found the where the FFmpeg comment comes from: https://lists.libav.org/pipermail/libav-devel/2014-November/065294.html
However fwiw I also can't find this restriction mentioned in MS docs or anywhere else. Was this changed in the meantime or can someone test if this still applies or if WinRT just always returns `false` for `AreFileApisANSI()`?

astiob:
Looks like [`AreFileApisANSI`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-arefileapisansi) itself is not available on WinRT: its “minimum supported client/server” lines say “desktop apps only”.

But even if we assume file APIs are always ANSI on WinRT, that doesn’t help too much: the purported problem is that we can’t _convert_ the path, not that we don’t know what encoding it is in. (Unless conversion from `CP_ACP` fails but we’re able to resolve `CP_ACP` to a specific encoding and convert from that successfully.)

lifegpc:
I am confused now. :(

astiob:
I haven’t reviewed the whole PR yet, but a few other notes:

We should set `MB_ERR_INVALID_CHARS` for `MultiByteToWideChar`. Otherwise, it happily converts invalid bytes into U+FFFD REPLACEMENT CHARACTER (or drops them in ancient Windows versions). mpv doesn’t set it, but it’s presumably safe because all of its filenames originate from wchars explicitly converted to UTF-8. FFmpeg does set it for its UTF-8 conversion (and it does not do any ACP/OEM conversions). I think we should set it for all conversions.

We should double-check all APIs for their availability on WinRT.

> I've found the where the FFmpeg comment comes from: https://lists.libav.org/pipermail/libav-devel/2014-November/065294.html

Speaking of this thing—and FWIW the code is still there in FFmpeg master—it also claims that `MoveFileExA` is unavailable on Windows Phone and “for Windows Store apps, it is available but not allowed by the app certification kit”. Yet if we just go by the “minimum supported” platforms in Microsoft’s docs, this function is supposedly available for “UWP apps”.

And https://github.com/mpv-player/mpv/commit/4637b029cdd168d4196f5ab69fa5f91556ee5d11 changed a `FindFirstFileW` call into `FindFirstFileExW` without comment, but again MSDN says `FindFirstFileW` is available in UWP apps.

It seems there’s still something I don’t understand about this and something additional that we need to take into account.

Frederisk:
If you are trying to use Win32 APIs for Windows Phone, [this](https://docs.microsoft.com/en-us/previous-versions/windows/apps/jj662956(v=vs.105)#list-of-supported-win32-apis-on-windowsphone8) may be useful, and there is [another](https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-apis) for UWP.

Due to the strict security policy of Windows RT, it can be noted that most of the APIs will not be available. Maybe you need to implement file read and write functions with [`ReadFile`](https://docs.microsoft.com/zh-tw/windows/win32/api/fileapi/nf-fileapi-readfile) and [`WriteFile`](https://docs.microsoft.com/zh-tw/windows/win32/api/fileapi/nf-fileapi-writefile) for Windows RT.

astiob:
Thanks for the links, especially the one for Windows Phone 8!

…Looking through the linked Windows Phone docs, it seems neither GDI nor DirectWrite are available at all. How do people do fonts/text/graphics on this platform at all?! But I suppose libass could still run without a system font provider.

> Due to the strict security policy of Windows RT, it can be noted that most of the APIs will not be available.

We’re aware of that in principle, but we’re finding it hard to figure out exactly what is available and what isn’t!

astiob:
I’m just guessing, but it looks like Windows Phone 8 APIs are in `WINAPI_FAMILY_PHONE_APP` per `winapifamily.h` checks. So what I was doing until now wasn’t necessarily wrong; I just looked only at “desktop apps” (Win32) and “PC apps” (WinRT) but completely ignored “phone apps” (also WinRT? but different API subset).

Seeing as it’s all about _Universal_ Windows Platform now, Windows Phone docs are tucked away in a corner and there’s no mention of phone apps in the main docs, is it safe to assume modern Windows-based phones (if any) are supposed to run “PC” apps rather than “phone” apps?

astiob:
Hm, no. I must’ve misread something. APIs absent from Windows Phone 8 such as `MoveFileExA` and `FindFirstFileW` are not marked in any way in that SDK.

Frederisk:
> If you are trying to use Win32 APIs for Windows Phone, [this](https://docs.microsoft.com/en-us/previous-versions/windows/apps/jj662956(v=vs.105)#list-of-supported-win32-apis-on-windowsphone8) may be useful, and there is [another](https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-apis) for UWP.
> 
> 
> 
> Due to the strict security policy of Windows RT, it can be noted that most of the APIs will not be available. Maybe you need to implement file read and write functions with [`ReadFile`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile) and [`WriteFile`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile) for Windows RT.

There is a way:
```cpp
#ifdef IOWIN32_USING_WINRT_API

bool open_file(const void* path, FILE* fp) {

  HANDLE hFile = NULL;
  CREATEFILE2_EXTENDED_PARAMETERS extParams = {0};
  extParams.dwFileAttributes = FILE_ATTRIBUTE_NORMAL;
  extParams.dwSize = sizeof(cf2ex);

#ifdef UNICODE
  hFile = CreateFile2((LPCTSTR)path, GENERIC_READ | GENERIC_WRITE, FILE_SHARE_READ, OPEN_EXISTING, &extParams);
#else
  WCHAR pathW[FILENAME_MAX+0x201];
  MultiByteToWideChar(CP_ACP, 0, (const char*)filename, -1, filenameW, FILENAME_MAX + 0x200);
  hFile = CreateFile2((LPCTSTR)pathW, GENERIC_READ, FILE_SHARE_READ, OPEN_EXISTING, &extParams);
#endif

if(hFile == NULL || hFile == INVALID_HANDLE_VALUE) return false;

  fp = _fdopen(_open_osfhandle((intptr_t)hFile, _O_RDONLY), "rb");
  if(fp == NULL) {
    CloseHandle(hFile);
    return false;
  }

  return true;
}
```

This is just a simple idea and may need to be modified to ensure completeness.

astiob:
Thanks a lot for your continued help!

Regarding Windows Phone specifically, after a brief discussion on IRC, it seems we’re inclined to not add explicit support for it. Several reasons:

* we’re confused and scared (at least I am);
* it appears to be abandoned by Microsoft (so much that MSDN no longer mentions it in current API docs);
* it seems to need a lot of extra code like the code just posted due to missing/forbidden APIs (?);
* some things that we kinda need appear to simply blow up at runtime as evidenced [by that FFmpeg patch](https://lists.libav.org/pipermail/libav-devel/2014-November/065294.html);
* it has no font APIs;
* and while FFmpeg and mpv seem to have gone out of their way to add Phone-compatible code, we’re not aware of actual attempts to use libass on Windows Phone (and given its EOL status, this will only get less likely over time).

But if there’s any gross misunderstanding here or otherwise good reason to support Windows Phone 8, I’d love to be educated; and, of course, patches are always welcome.

Modern UWP supposedly has `fopen` and `_wfopen`, so it would seem that we don’t need to use `CreateFile2`. Or do we? (How do I find out for sure? That’s what’s frustrating me.)

But they’re listed among [extension APIs](https://docs.microsoft.com/en-us/uwp/win32-and-com/win32-extension-apis) as “added in 10.0.10240”. But so is the whole rest of the standard C library, and we’re certainly not going to stop using the whole standard C library. So are these APIs available in older Windows versions? Does this just mean the app needs to include its own CRT as opposed to one coming preinstalled in the system, so we can still assume they exist? Is it safe to assume `fopen` and `_wfopen` exist all the way back to Windows 8.1?

(Wikipedia tells me Windows 8 no longer receives even third-party app updates, as in Microsoft has frozen its Windows Store, so presumably it’s impossible for new libass to run on Windows 8 WinRT in any case.)

(And if they’re available on all PC WinRT and all modern WinRT devices, does that mean they’re just unavailable on Windows Phone 8 specifically?)

astiob:
Now, even if we forget about Windows Phone for a moment, I’m still worried by this FFmpeg comment:

> for Windows Store apps, [`MoveFileExA`] is available but not allowed by the app certification kit

How can we find out what’s allowed by this certification kit and what isn’t? I’d like to avoid using APIs that are “available” on UWP according to MSDN and the Windows 10 SDK only to find out later that our code can’t actually be used in Microsoft Store apps due to failing certification checks.

Frederisk:
It is worth mentioning that native Windows Store Apps (Windows 8) and UWP Apps have very strict restrictions on accessible files. The program can only access Application install directory(readonly), Application data locations, Removable devices, User's Downloads folder, locations [declared by the APP profile](https://docs.microsoft.com/zh-tw/windows/uwp/packaging/app-capability-declarations) (such as [music, documents, pictures](https://docs.microsoft.com/zh-tw/windows/uwp/files/quickstart-managing-folders-in-the-music-pictures-and-videos-libraries), and access more through [Picker](https://docs.microsoft.com/zh-tw/windows/uwp/files/quickstart-using-file-and-folder-pickers).

Obviously, this program did not actually use WinRT API but tried to use Win32 API to access files. According to [this documentation](https://docs.microsoft.com/zh-tw/windows/uwp/files/file-access-permissions), programs using Win32 API can only read and write files in Application install directory, Application data locations. I am not sure if this restriction is what you expect or can accept.🤔

Frederisk:
> How can we find out what’s allowed by this certification kit and what isn’t? I’d like to avoid using APIs that are “available” on UWP according to MSDN and the Windows 10 SDK only to find out later that our code can’t actually be used in Microsoft Store apps due to failing certification checks.

[This](https://docs.microsoft.com/en-us/windows/uwp/packaging/app-capability-declarations) is a list about capability declarations. And [this](https://docs.microsoft.com/en-us/windows/uwp/debug-test-perf/windows-app-certification-kit-tests) a way to test your application by yourself. [https://docs.microsoft.com/en-us/windows/uwp/debug-test-perf/windows-app-certification-kit-tests

Frederisk:
> We’re confused and scared (at least I am);

Although WinRT and UWP have Win32 and C++ std interfaces, in fact this is a completely different system, which is isolated in the environment separately, and not all functions are available. Or you choose to use some hacking methods or ［Desktop Bridge](https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-extensions).

> it appears to be abandoned by Microsoft (so much that MSDN no longer mentions it in current API docs);

Windows Phone (the last version  is Windows 10 Mobile v1709) has indeed [been abandoned](https://support.microsoft.com/en-us/windows/windows-10-mobile-end-of-support-faq-8c2dd1cf-a571-00f0-0881-bb83926d05c5), and I also do not recommend that you pay too much for this outdated platform.

> it seems neither GDI nor DirectWrite are available at all. How do people do fonts/text/graphics on this platform at all?! But I suppose libass could still run without a system font provider.

Using font-related APIs in UWP is indeed very challenging.  The actual method is often to use d3d11_1.h, dwrite.h, and d2d1_1.h. If you use the .NTE platform, you need to introduce wrl.h and CRL support. [Here](https://github.com/Frederisk/Frederisk.RtfEditor/tree/master/Frederisk.RtfEditor.DirectXWrapper) is a UWP-based device font query category library I wrote.  It is part of my sample program for a rich text editor.

I am glad that my knowledge of WinRT and UWP can help you.😄

astiob:
> Using font-related APIs in UWP is indeed very challenging.

In UWP, there’s DirectWrite. What I was saying there is that on Windows Phone 8 (according to your link), there’s nothing!

Frederisk:
> > Using font-related APIs in UWP is indeed very challenging.
> 
> 
> 
> In UWP, there’s DirectWrite. What I was saying there is that on Windows Phone 8 (according to your link), there’s nothing!

Hmm. It's seems DirectWrite already exists on API of Windows 8 Apps. According to the table of contents of [_Programming Windows 6th Edition_ Chapter 15](https://ptgmedia.pearsoncmg.com/images/9780735671768/samplepages/9780735671768.pdf). But I don't remember the specific situation of Windows Phone clearly. These information are difficult to obtain now.

TheOneric:
> @lifegpc:
> I am confused now. :(

Understandable,  I'm also confused about those various restrictions `^^

To recap: We now decided to only support “Desktop” and modern “UWP” Windows variant's for the time being.
As `AreFileAPIsANSI()` is only available on “Desktop”, I checked [MS' `fopen` docs](https://docs.microsoft.com/en-gb/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-160) which claim that on non-“Desktop” the encoding is always `CP_ACP`.
So the procedure for directories remains almost the same as suggested before:

 1) try decoding as UTF-8
     - if fails, goto `2)`
     - if succesful try to open with `FindFirstFileW`
       - if fails goto `2)`
       - if succesful goto `4)`
 2) do:
    ```C
      UINT cur_cp = CP_ACP;
      #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
      if (!AreFileApisANSI()) cur_cp = CP_OEMCP;
      #endif
    ```
 3) try decoding as `cur_cp`
    - if fails abort
    - if succesful try opening with `FindFirstFileW`
       - if fails abort
 4) traverse the opened directory with `FindNextFileW`

This should be able to provide what we need.
Since we just now also gained  UWPish-CI tests we should now hopefully be able to catch any undocumented(or hard to discover) further restrictions should they exists. *(Please rebase (and ideally squash) your patches so the new CI actually runs)*

---

In addition to that, let me remind you of astiob's previous comment about the conversion flags, which needs to be addressed as well:
> We should set `MB_ERR_INVALID_CHARS` for `MultiByteToWideChar`. Otherwise, it happily converts invalid bytes into
+U+FFFD REPLACEMENT CHARACTER (or drops them in ancient Windows versions).

TheOneric:
@Frederisk Since I noticed you adding commits: the current and closest to fully correct version of the Microsoft-path stuff is #573.

Frederisk:
Well, I saw that this problem has not been solved recently, so after getting in touch with the author of this branch, I tried to update the content, and I have already written the UWP support part, and plan to add it later.

I am not sure whether the new PR is easy to adjust, I may need to understand this part that you mentioned later.

TheOneric:
Superseded by #573.

PR #536, opened 2021-08-07
[RFC] documentation for manual struct edits

TheOneric:
There was previously no documentation at all and to me no clear assumption which could be derived from the code. So this will define the API itself and not just be documentation of existing assumptions. I'd appreaciate if everyone could give some feedback on the restrictions layed out by these patches.

Also as a side-product, I noticed that `treat_fontname_as_pattern`/`treat_family_as_pattern` which was introduced in 256c11324911da42ff829d9cb6906186b5173464 no longer seems to do anything, though a few vestiges remain in ass_render.c and the (very outdated) documentation of `ass_font_select`. Before these emains are removed completly, can someone verify it is actually doing nothing and supposed to be like this?


libass/ass_types.h
```diff
@@ -98,32 +98,32 @@ typedef struct ass_library ASS_Library;
...
     int MarginL;
     int MarginR;
     int MarginV;
     int Encoding;
-    int treat_fontname_as_pattern;
-    double Blur;
-    int Justify;
+    int treat_fontname_as_pattern; // does nothing (left in place for ABI-compatibility)
+    double Blur; // sets a default \blur for the event, same values as \blur
+    int Justify; // sets text justification independent of event alignment, use ASS_JUSTIFY_*
```
rcombs, line 126: Semicolons rather than commas here

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
 
 #include <stdint.h>
 
+/**
+ * GENERAL NOTE regarding the definitions exposed by this header
+ *
+ * The main use case for this is _reading_ the track fields, especially
+ * track->YCbCrMatrix, to correctly display the rendering results.
+ *
+ * Furthermore the exposed definitions also open up the possibility to _modify_
```
rcombs, line 31: Would be worth sprucing up the grammar here a bit: "Furthermore", "e.g.", "well-versed", "internals,", "API breaks"…

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ * eg creation of intermediate ASS-text buffers when creating dynamic events.
+ * This is an advanced use case and should only be done when well versed in ASS
+ * and aware of the effects and legal values of _all_ fields of the structs.
+ * The burden of sanitising and correctly initialising fields is then also
+ * placed on the API-user.
+ * By nature of direct struct modifications working closer to library internals
+ * workflows that make use of this possibility are also more likely to be
+ * affected by future API-breaks than those which do not.
+ *
+ * To avoid desynchronisation with internal states there are some restrictions
```
rcombs, line 42: "states,",

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ * on when and how direct struct modification can be performed.
+ * Ignoring them may lead to undefined behaviour. See the following listing:
+ *
+ *  - Manual struct edits and track-modifying (including modification to the
+ *    event and style elements of the track) API-calls cannot be freely mixed:
```
rcombs, line 47: "API calls"

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ *    - Before manual changes are performed it is allowed to call any such API,
+ *      unless the documentation of the API-call says otherwise.
```
rcombs, line 49: "function"

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ * workflows that make use of this possibility are also more likely to be
+ * affected by future API-breaks than those which do not.
+ *
+ * To avoid desynchronisation with internal states there are some restrictions
+ * on when and how direct struct modification can be performed.
+ * Ignoring them may lead to undefined behaviour. See the following listing:
+ *
+ *  - Manual struct edits and track-modifying (including modification to the
+ *    event and style elements of the track) API-calls cannot be freely mixed:
+ *    - Before manual changes are performed it is allowed to call any such API,
```
rcombs, line 48: "performed,"

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ *      unless the documentation of the API-call says otherwise.
+ *    - After manual changes were performed, no track-modifying API must be
```
rcombs, line 50: "changes have been performed"; "API may be"

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ *      invoked any more, except for ass_track_set_feature and ass_flush_events.
```
rcombs, line 51: drop "any more"

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ *  - After rendering started existing array members, eg members of events, and
```
rcombs, line 52: "After the first call to ass_render_frame, existing array members (e.g. members of events) and non-array track fields (e.g. PlayResX or event_format) must not be modified."

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ *    all non-array track fields, like eg PlayResX or event_format, must not
+ *    be modified any more.
+ *    Adding _new_ members to arrays and updating
+ *    the corresponding counter remains allowed.
```
rcombs, line 56: Wrap?

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ *  - Adding and removing members to array fields, like events or styles,
+ *    must be done through the corresponding API-call, eg ass_alloc_event.
```
rcombs, line 58: "API function, e.g."

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ *    Also see the documentation of these API-calls.
```
rcombs, line 59: "See also the documentation of these functions."

libass/ass_types.h
```diff
@@ -22,6 +22,52 @@
...
+ *
+ * A non-exhaustive list of examples of track-modifying API-functions:
+ *  ass_process_data, ass_process_codec_private,
+ *  ass_process_chunk, ass_read_styles, ...
+ *
+ * Direct struct modification can be done safely, but it is also easy to
+ * miss an initialisation or violate the restrictions introducing bugs
```
rcombs, line 66: "violate these restrictions, thus introducing bugs…"

libass/ass.h
```diff
@@ -597,6 +599,10 @@ int ass_alloc_event(ASS_Track *track);
  * \param track track
  * \param sid style id
  * Deallocates style data. Does not modify track->n_styles.
+ * Freeing a style without subsequently setting track->n_styles
+ * to a value smaller or equal to the freed style id before
```
rcombs, line 603: "less than or equal to"

libass/ass.h
```diff
@@ -605,6 +611,10 @@ void ass_free_style(ASS_Track *track, int sid);
  * \param track track
  * \param eid event id
  * Deallocates event data. Does not modify track->n_events.
+ * Freeing an event without subsequently setting track->n_events
+ * to a value smaller or equal to the freed event id before rendering
+ * the next frame or calling ass_flush_events is undefined behaviour.
```
rcombs, line 616: just say calling any other libass API function on the track

libass/ass.h
```diff
@@ -564,6 +564,7 @@ ASS_Track *ass_new_track(ASS_Library *);
  * headers. (ass_process_chunk() will not change any of the flags.)
  * Additions to ASS_Feature are backward compatible to old libass releases (ABI
  * compatibility).
+ * Calling this after rendering of the track started is disallowed.
```
rcombs, line 567: "after calling ass_render_frame"?

libass/ass_types.h
```diff
@@ -98,32 +98,32 @@ typedef struct ass_library ASS_Library;
 
 /* ASS Style: line */
 typedef struct ass_style {
-    char *Name;
-    char *FontName;
+    char *Name;     //must be a valid non-NULL string pointer
+    char *FontName; //must be a valid non-NULL string pointer
```
rcombs, line 102: Specify how these should be allocated? afaik they have to come from the same copy of `malloc` as libass uses (which has some implications on windows)

Also, worth noting that while they can't be null, they can point to empty strings.

libass/ass_types.h
```diff
@@ -242,9 +242,9 @@ typedef struct ass_track {
     int PlayResY;
     double Timer;
     int WrapStyle;
-    int ScaledBorderAndShadow;
-    int Kerning;
-    char *Language;
+    int ScaledBorderAndShadow; // 0 or 1 (boolean)
+    int Kerning; // 0 or 1 (boolean)
+    char *Language; // zero-terminated ISO-639-1 codes
```
rcombs, line 247: this can be null, right? also, singular "code"

rcombs (changes requested):
Just bikeshedding on phrasing and such; conceptually I like the clarifications.

libass/ass_types.h
TheOneric: thanks, removed the plural. Yes, NULL is allowed here.

libass/ass_types.h
```diff
@@ -52,32 +99,32 @@ typedef struct ass_library ASS_Library;
...
-    double ScaleY;
+    int Bold;      // 0 or 1 (boolean)
+    int Italic;    // 0 or 1 (boolean)
+    int Underline; // 0 or 1 (boolean)
+    int StrikeOut; // 0 or 1 (boolean)
+    double ScaleX; // positive with 1.0 representing 100%
+    double ScaleY; // positive with 1.0 representing 100%
     double Spacing;
     double Angle;
     int BorderStyle;
```
CastagnaIT, line 117: it would be useful to document BorderStyle as well, and i could suggest to add a borderstyle enum values or something like constants to be used to set this property

TheOneric: The `BorderStyle` values are identical to ASS + libass' ASS extensions. I don't think we need to repeat ASS(-extension) spec/behaviour in each field's doc-comment.

TheOneric:
Added
```
 * Additionally if styles a re freed such that an for an event ev in track->events
 * ev->Style >= track->n_styles will also result in undefined behaviour.
```
to the documentation of `ass_free_style`.

TheOneric:
Rebased, added changes to docs of `ass_render_frame`'s `detect_change` parameter and applied the “api: do not leak deprecation macros”-commit to master to hopefully avoid anyone starting to use those.

TheOneric:
If anyone is planning to take (another) look at this, say so now and I'll wait for whatever amount of time is required. Otherwise I'll assume implicit approval and merge this in ~2 days.

PR #542, opened 2021-09-01
Use a libtool-wrapper script for NASM

### LINKED ISSUE #539: Building ASM on OpenBSD: error on -fPIC, despite -prefer-non-pic
I have had the ASM disabled in the OpenBSD port for quite awhile as I couldn't get it to build. I tried again with 0.15.1, but it is still not building Ok.

I am looking around with libtool trying to figure out where the -fPIC -DPIC was added. It's the -fPIC that is causing the error.

```
/usr/local/bin/libtool  --tag=CC --mode=compile nasm -DARCH_X86_64=1 -f elf64 -Dprivate_prefix=ass -I./ -o x86/rasterizer.lo x86/rasterizer.asm -prefer-non-pic
libtool: compile:  nasm -DARCH_X86_64=1 -f elf64 -Dprivate_prefix=ass -I./ x86/rasterizer.asm  -fPIC -DPIC -o x86/.libs/rasterizer.o
nasm: fatal: unrecognised output format `PIC' - use -hf for a list
Type nasm -h for help.
```

TheOneric:
Closes #539; NetBSD was also affected.

Patch was tested on the following systems *(`y`: test successful; `-`: not tested,)*
|     *OS*     |  *Family*  |  ix86 | amd64 |
| ------------ |:----------:|:-----:| -----:|
| Alpine 3.14  | Linux-musl |   y   |   -   |
| Debian 11    | Linux-glibc|   y   |   y   |
| Haiku R1B3   |  BeOS-like |   y   |   y   |
| NetBSD 9.2   |    *BSD    |   -   |   y   |
| OpenIndiana  |   illumos/Solaris  |  y   |   y   |

@brad0 tested an earlier version on OpenBSD and our GHA CI will test on MacOS and Windows.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,5 @@
 ACLOCAL_AMFLAGS = -I m4
-EXTRA_DIST = libass.pc.in Changelog
+EXTRA_DIST = libass.pc.in Changelog ltnasm.sh
 
 pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libass.pc
--- a/libass/Makefile.am
+++ b/libass/Makefile.am
@@ -12,7 +12,7 @@ nasm_verbose_ = $(nasm_verbose_$(AM_DEFAULT_VERBOSITY))
 nasm_verbose_0 = @echo "  NASM    " $@;
 
 .asm.lo:
-	$(nasm_verbose)$(LIBTOOL) $(AM_V_lt) --tag=CC --mode=compile $(AS) $(ASFLAGS) -I$(srcdir)/ -o $@ $< -prefer-non-pic
+	$(nasm_verbose)$(LIBTOOL) $(AM_V_lt) --tag=CC --mode=compile $(top_srcdir)/ltnasm.sh $(AS) $(ASFLAGS) -I$(srcdir)/ -o $@ $<
 
 SRC_INTEL = x86/rasterizer.asm x86/blend_bitmaps.asm x86/be_blur.asm x86/blur.asm x86/cpuid.asm \
             x86/cpuid.h
--- a/ltnasm.sh
+++ b/ltnasm.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+# Translate libtool supplied C-compiler options for NASM.
+# libtool treats NASM like the C compiler, and may supply -f… options
+# which are interpreted as the output file format by NASM, causing errors.
+# Notably libtool will set -DPIC -fPIC and -fno-common;
+# we want to use -DPIC by translating it to -DPIC=1, but remove everything else
+#
+# Theoretically the way the filtering is done here in a plain POSIX shell script,
+# does mess up if there were spaces in any argument. However this will never happen
+# since neither our filenames nor options do not contain spaces and source paths
+# are not allowed to contain spaces by configure.
+
+cmd=""
+while [ "$#" -gt 0 ] ; do
+    case "$1" in
+        # NASM accepts both -f format and -fformat,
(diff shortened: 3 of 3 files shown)


brad0:
Ya, I noticed NetBSD had ASM disabled for their package and came across a bug report from 7 years ago..

http://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=48703

 > /bin/sh ../libtool --quiet --mode=compile "yasm"      -DARCH_X86_64=1 -m amd64     -o x86/blend_bitmaps.lo x86/blend_bitmaps.asm -prefer-non-pic
 > libtool: compile: unable to infer tagged configuration
 > libtool: compile: specify a tag with `--tag'

TheOneric:
@brad0 fyi: we'll soon release 0.15.2, but I decided to delay this patch to after 0.15.2, because I'm still a bit worried about spaces and **if** they can occur in arguments afterall, having been in master for several months will hopefully allow us to catch and fix it before a release, instead of introducing a regression shortly before a release. Unfortunately this also means you and the pkgsrc-maintainers will have to backport the patch for just this one more version if enabling assembly.

---

> [pkgsrc] bug report from 7 years ago
> http://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=48703

Heh, the SunOS part of that bugreport got fixed 5 years ago when VLC upstreamed their libass-patches. If the linked bugreport would have been forwarded, both issues could already have been fixed 7 years ago.

brad0:
> @brad0 fyi: we'll soon release 0.15.2, but I decided to delay this patch to after 0.15.2, because I'm still a bit worried about spaces and **if** they can occur in arguments afterall, having been in master for several months will hopefully allow us to catch and fix it before a release, instead of introducing a regression shortly before a release. Unfortunately this also means you and the pkgsrc-maintainers will have to backport the patch for just this one more version if enabling assembly.

It's all good. The small patch is fine in our ports tree for now and is what we're using to ship libass with -current, soon to be our 7.0 release.

> Heh, the SunOS part of that bugreport got fixed 5 years ago when VLC upstreamed their libass-patches. If the linked bugreport would have been forwarded, both issues could already have been fixed 7 years ago.

Downstream projects are not always good at making sure upstream projects are even aware of issues. Doesn't mean a suitable patch exists as is, but at least make the project aware of an issue as a start.

astiob (approved):
LGTM.

> any opinions on #542, in particular where to place ltnasm.sh?

No strong opinion, but I think all the rest of build stuff is in the repository root, both the original source in Git and the files generated by Autotools, except the standard macros copied into `m4/`. (Which are for libtool… I wonder if there’s any relation between those files and what you’re trying to do/fix here.) So it could make sense to keep this in the root, too.

TheOneric:
Thanks for the review.
I don't have a strong opinion on where to place it either. Here's the options I considered:
 - If we had more build-releated scripts like these, `tools` probably would make sense to not clutter the root, but for `ltnasm.sh` alone it might actually be worse than just putting it into root *(but maybe future things could be put there)*.
  - After `autoreconf` there's also a generated/copied `ltmain.sh`, which on one hand fits well with `ltnasm.sh`, but on the other hand might make it seem like `ltnasm.sh` is also generated, if it's placed in root *(at most a minor cosmetic concern though)*.
  - Since this will only ever be used by x86-assembly, it might also make sense to put it into `libass/x86/`.

I'd be fine with any of these options and don't really have a preference between them.

TheOneric:
Moved `ltnasm.sh` to the repo-root, will merge when all CI-runs succeeded.

brad0:
I am impressed that you tested Haiku.

PR #543, opened 2021-09-01
Don’t try to look up glyphs for invisible control characters

astiob:
Supersedes and closes #508.

Implements stupid simple logic that seems to match Uniscribe/VSFilter somewhat, but perhaps not fully. At the very least, this should be a clear improvement over our current behaviour.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -486,13 +486,6 @@ int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
         return 0;
     }
 
-    // try with the requested face
-    if (*face_index < font->n_faces) {
-        face = font->faces[*face_index];
-        index = FT_Get_Char_Index(face, ass_font_index_magic(face, symbol));
-    }
-
-    // not found in requested face, try all others
     for (i = 0; i < font->n_faces && index == 0; ++i) {
         face = font->faces[i];
         index = FT_Get_Char_Index(face, ass_font_index_magic(face, symbol));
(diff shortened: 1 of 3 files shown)


astiob:
@avih Could you test this and confirm that it’s good enough for your use case?

avih:
> @avih Could you test this and confirm that it’s good enough for your use case?

Seem to be working.

The last commit - display missing glyph index at the warning - works (I applied it first to still see how the warning looks like), and display in mpv e.g.
```
[osd/libass] fontselect: failed to find any fallback with glyph 0x2060 for font: (Arial, 400, 0)
```

The other (first 5) commits make the warning not show at all - work as expected.

TheOneric (commented):
Code seems fine to me and printing which symbol couldn't be located is also a good thing.
As for the actual logic of just using the preceding fontface or zero, I'm trusting your finding that this (mostly) matches VSFilter/Uniscribe behaviour and will continue to render visible ""invisble character"".


One question:
> bidi splitting after script splitting can create awkward short runs,
> and in general, script splitting seems problematic in itself.

This is only about non-VSF-compatible shaping, right? Also I may misremember, but I seem to recall you mentioning that some (or all?) HarfBuzz-APIs only work on constant scripts, so is it even possible for us to avoid script splitting?

astiob:
> This is only about non-VSF-compatible shaping, right?

Not necessarily. #insufficientdata

> Also I may misremember, but I seem to recall you mentioning that some (or all?) HarfBuzz-APIs only work on constant scripts, so is it even possible for us to avoid script splitting?

That’s right. Take that more as “there are fundamental limits to how well we can shape” than as “we’re using HarfBuzz insufficiently well”.

astiob:
Sorry for the noise. I pushed branches in the wrong order the first time around, so GitHub lost all commits in this PR. Had to reopen the PR and repush a few things to knock it back into shape.

PR #546, opened 2021-09-11
FIX: ScaledBorderAndShadow=no not working

passerbya:
FIX: ScaledBorderAndShadow=no not working

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -979,7 +979,9 @@ static void init_font_scale(ASS_Renderer *render_priv)
         render_priv->border_scale =
             font_scr_h / render_priv->track->PlayResY;
     else
-        render_priv->border_scale = render_priv->blur_scale;
+        render_priv->border_scale = 1;
+    if (!settings_priv->storage_height)
+        render_priv->blur_scale = render_priv->border_scale;
 
     if (render_priv->state.apply_font_scale) {
         render_priv->font_scale *= settings_priv->font_size_coeff;


astiob:
No, this is working as intended. When the user resizes the video player window (including going into/out of full-screen or moving a full-screen player between different-resolution screens), the whole user-visible picture must always be rescaled together, regardless of `ScaledBorderAndShadow` or any other flags or API usage. This is what `blur_scale` achieves, by scaling from video storage resolution to target display resolution. This is what users expect, and this matches all other ASS renderers.

I can only guess, but ***if*** your issue is that you’re actually seeing borders scaled from PlayResX/Y to video resolution, then that means you’re using a video player that does not invoke `ass_set_storage_size`. This needs to be fixed in the video player; it cannot be fixed in libass.

passerbya:
This feature is incompatible with VSFilter,  when user's ass generates by software base on VSFilter or libass versions prior to 0.15, will show different effects.

astiob:
No, this is precisely compatible with VSFilter (any VSFilter variant in DirectShow, and XySubFilter, and with some exceptions MPC-HC’s embedded VSFilter since many years ago). VSFilter compatibility is the main goal of our current implementation of ScaledBorderAndShadow, `border_scale` and `blur_scale`, and this has been well tested. If you’re seeing different results in your libass-equipped player, then most likely it doesn’t call `ass_set_storage_size`, which is a problem in the player and must be solved in the player. Without the storage size, libass simply can’t do this fully correctly.

If you still think there’s an incompatibility, please show a full example and we will investigate.

astiob:
A few questions to help diagnose this quicker:

* Are you seeing the problem on a particular file? Can you provide that file?
* What video player software are you using?
* Have you (actually) compared to a VSFilter-equipped player?
* Have you tried playing the same file in the [mpv](https://mpv.io/) player? It uses libass and properly calls `ass_set_storage_size`.

PR #549, opened 2021-09-14
Use x86 asm code if we’ve built it, regardless of C compiler

astiob:
We detect x86* and enable building asm code in `configure`. However, we don’t actually use that code unless we detect x86* via the C compiler’s predefined macros.

Perhaps the best thing to do would be to set a `CONFIG_X86` flag in `configure` and check that. Shall we do that?

If not, here’s a patch that checks more macros to add support for:
  * Microsoft Visual C++ (`_M_IX86`/`_M_X64`)
  * Intel C++ Compiler for Windows (`_M_IX86`/`_M_X64`)
  * Oracle Developer Studio (contrasts `__i386__` with `__x86_64` and sets `__i386` on both)

Subject to C99/C11/dependency support, this also supports:
  * Digital Mars (`_M_IX86`, no 64-bit target)
  * Watcom C/C++ (`_M_IX86`, no 64-bit target; defines `_M_IX86` also for 16-bit, but the `CONFIG_ASM` guard will block that)
  * Stratus VOS/OpenVOS (`__i386`, no 64-bit target)

Inspired by https://github.com/ShiftMediaProject/libass/commit/9a92b600710e37beeb5aff55b5e5c8162c270d38 mentioned in https://github.com/libass/libass/pull/330#issuecomment-903386563.

### DIFF
--- a/libass/ass_bitmap.c
+++ b/libass/ass_bitmap.c
@@ -42,7 +42,7 @@
 #undef ALIGN
 #undef DECORATE
 
-#if (defined(__i386__) || defined(__x86_64__)) && CONFIG_ASM
+#if CONFIG_ASM && ARCH_X86
 
 #define ALIGN           4
 #define DECORATE(func)  ass_##func##_sse2
--- a/configure.ac
+++ b/configure.ac
@@ -322,6 +322,9 @@ AM_CONDITIONAL([DIRECTWRITE], [test "x$directwrite" = xtrue])
 ## Define C Macros not relating to libraries
 AM_COND_IF([ASM], [
     AC_DEFINE(CONFIG_ASM, 1, [ASM enabled])
+    AM_COND_IF([X86], [
+        AC_DEFINE(ARCH_X86, 1, [targeting a 32- or 64-bit x86 host architecture])
+    ])
 ], [
     AC_DEFINE(CONFIG_ASM, 0, [ASM enabled])
 ])
--- a/configure.ac
+++ b/configure.ac
@@ -214,34 +214,34 @@ can_asm=false
 AS_IF([test "x$enable_asm" != xno], [
     AS_CASE([$host],
         [i?86-*], [
-            INTEL=true
             AS=nasm
+            X86=true
             BITS=32
             BITTYPE=32
             ASFLAGS="$ASFLAGS -DARCH_X86_64=0"
         ],
         [x86_64-*-gnux32|amd64-*-gnux32], [
             AS=nasm
-            INTEL=true
-            X64=true
+            X86=true
+            X86_64=true
             BITS=64
             BITTYPE=x32
             ASFLAGS="$ASFLAGS -DARCH_X86_64=1"
         ],
         [x86_64-*|amd64-*], [
             AS=nasm
-            INTEL=true
-            X64=true
+            X86=true
+            X86_64=true
             BITS=64
             BITTYPE=64
             ASFLAGS="$ASFLAGS -DARCH_X86_64=1"
         ],
         [ # default
-            INTEL=false
+            X86=false
             AC_MSG_NOTICE([Assembly optimizations are not yet supported on this architecture; disabling.])
         ]
     )
-    AS_IF([test "x$INTEL" = xtrue], [
+    AS_IF([test "x$X86" = xtrue], [
         AC_CHECK_PROG([nasm_check], [$AS], [yes])
         AS_IF([test "x$nasm_check" != xyes], [
             AC_MSG_WARN(nasm was not found; ASM functions are disabled.)
@@ -306,8 +306,8 @@ AC_SUBST([PKG_REQUIRES_PRIVATE], [${pkg_requires}])
 
 ## Setup conditionals for use in Makefiles
 AM_CONDITIONAL([ASM], [test "x$can_asm" = xtrue])
-AM_CONDITIONAL([INTEL], [test "x$INTEL" = xtrue])
-AM_CONDITIONAL([X64], [test "x$X64" = xtrue])
+AM_CONDITIONAL([X86], [test "x$X86" = xtrue])
+AM_CONDITIONAL([X86_64], [test "x$X86_64" = xtrue])
 
 AM_CONDITIONAL([ENABLE_LARGE_TILES], [test "x$enable_large_tiles" = xyes])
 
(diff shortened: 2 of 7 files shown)


TheOneric:
This would break `tcc`, which defines `__i386`, `__i386` and `i386` only on 32bit and only `__x86_64__` on 64-bit x86.
Given that and since it seems the macro-checks are all only interested if we are on *any* x86-arch, I think it makes more sense to utilise a config.h macro.

astiob:
Thanks for the extra compiler info!

I’ve switched to a config.h macro. Two notes:
* Meson, Visual Studio projects etc. will need to adjust.
* I don’t like the name. See commit message.

Bonus note: our `AC_DEFINE` and `config.h` are a mess. Each option only ever gets _one_ description in `config.h.in`, but we try to use different descriptions in different branches of `configure`. This doesn’t work as we expect, and only the last description ends up used. The options and their descriptions also appear even if the option is never defined by a particular `configure` run (in which case the macro is commented out). So all in all, each description should sufficiently explain what the macro is for and what each of its possible values means. Further, standard Autoconf checks write (almost all) their descriptions as full sentences with a leading uppercase letter and a full stop, in imperative mood, like so: “Define to 1 if you have [foo].” or “Define to [the foo of your bar].” and we could do well to follow suit. _Finally,_ config.h contains some predefined package version macros that we could use in lieu of our custom `LIBASS_VERSION` and URL macros that we’re not filling.

TheOneric:
> Bonus note: our AC_DEFINE and config.h are a mess. Each option only ever gets one description in config.h.in, but we try to use different descriptions in different branches of configure.

The autoconf-docs don't explicitly point it out, but it makes sense considering the “description is put into the generated config.h.in as the comment before the macro define.” which happens _before_ `configure` gets executed. We only use different descriptions in case of CoreText, DirectWrite and `LARGE_TILES`.

> So all in all, each description should sufficiently explain what the macro is for and what each of its possible values means.

I think the description is intended to be used exactly like this anyway :)

> Finally, config.h contains some predefined package version macros that we could use in lieu of our custom LIBASS_VERSION and URL macros that we’re not filling.

`LIBASS_VERSION` is part of the API, so imho it makes sense to use it internally as well and I don't think there's a file which includes `ass.h` just for `LIBASS_VERSION`, or is there one?

TheOneric (approved):
I agree, that the name isn't ideal; but otherwise the change seems good to me.

astiob:
> `LIBASS_VERSION` is part of the API, so imho it makes sense to use it internally as well and I don't think there's a file which includes `ass.h` just for `LIBASS_VERSION`, or is there one?

Oh, no, I don’t mean the C macro from `ass.h`. I mean the M4 macro that also happens to be named `LIBASS_VERSION` that we define in `configure.ac` to support `CONFIG_SOURCEVERSION`. It’s not a big deal, but we could instead do something like:

```m4
## Make a guess about the source code version
AS_IF([test -d "${srcdir}/.git"], [
    AC_PATH_PROG([git_bin], [git])
    AS_IF([test -n "$git_bin"], [
        srcversion_string="\"commit: $("$git_bin" -C "$srcdir" describe --tags --long --always --dirty --broken --abbrev=40)\""
    ], [
        srcversion_string="\"custom after: \" VERSION"
    ])
], [
    dnl# Hope no one creates custom tarballs without adjusting the version
    srcversion_string="\"tarball: \" VERSION"
])
AC_DEFINE_UNQUOTED([CONFIG_SOURCEVERSION], [$srcversion_string],
                   [string containing info about the used source])
```

astiob:
How’s this? I’ve pushed a new patch series that renames our existing `INTEL` and `X64` flags to `X86` and `X86_64` and accordingly uses `CONFIG_X86` for the new config.h macro.

libass/ass_bitmap.c
```diff
@@ -42,7 +42,7 @@
 #undef ALIGN
 #undef DECORATE
 
-#if (defined(__i386__) || defined(__x86_64__)) && CONFIG_ASM
+#if CONFIG_ASM && CONFIG_X86
```
MrSmile, line 45: I think it's better to mention that `CONFIG_X86` implies `CONFIG_ASM` and simplify condition. Also, do we really need `CONFIG_ASM`? Currently it's synonym to `CONFIG_X86` but even in the future we'd use something like `CONFIG_ARM` instead.

astiob: Today, `CONFIG_X86` (or the corresponding Automake check) does not actually imply `CONFIG_ASM` if we found NASM but it turned out to be unusable or too old. But we could, of course, change that.

Using `CONFIG_<arch>` to mean “<arch> _asm_ is enabled” seems unintuitive to me. But we could, of course, rename this to `CONFIG_<arch>_ASM`.

Beyond that, I think it’s a coin toss. Maybe we’ll have some use for an arch-independent `CONFIG_ASM` section some day, maybe not.

TheOneric: Fwiw having `CONFIG_ASM` distinct from the arch-macros seems more intuitive to me, but I don't care strongly about it.

TheOneric (approved):
The last commit still has the note: “Add CONFIG_INTEL […] This is a misnomer […]”, otherwise looking good.

astiob:
Thanks; fixed the commit message.

libass/ass_bitmap.c
MrSmile: FYI: corresponding macro in [checkasm](https://github.com/libass/libass/pull/400/commits/c8c46a34115b18f6d7e341875441c4d797f90b37#diff-e3e4c2b9b20b19afe5d5359e81ef7f883576b28089ca959cdc811e8918842686L118) is called `ARCH_X86`. We can adopt that naming scheme to reduce future changeset.

astiob: I don’t see `ARCH_X86` in the checkasm PR any more. I do see `ARCH_X86_64`, but it already exists, and it’s an assembler-only macro. So in a sense the `ARCH_` and `CONFIG_` macros are parallel: one for assembler and one for C. Maybe not optimal, but the prospect of using a non-`CONFIG_` name for just one out of all configure macros feels weird.

MrSmile: It's in the original [checkasm](https://github.com/libass/libass/commit/3906012d9cf4afc19e522a7de749847657d0fbeb#diff-e3e4c2b9b20b19afe5d5359e81ef7f883576b28089ca959cdc811e8918842686L118), it got removed later in the PR.

TheOneric (approved):

MrSmile (approved):

PR #551, opened 2021-09-16
Configure: sourceversion in git submodules

TheOneric:
Noticed JavascriptSubtitlesOctopus' libass builds didn't set the commit as source version; while at it also pulled in the config enhancements from #486.
Note, that even with this applied current JSO doesn't yet set the source version due to it copying the libass submodule to a different location for building, breaking the relative link to the real git-directory. This will be dealt with in a JSO-patch.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -333,7 +333,7 @@ AM_COND_IF([ENABLE_LARGE_TILES], [
 ])
 
 ## Make a guess about the source code version
-AS_IF([test -d "${srcdir}/.git"], [
+AS_IF([test -e "${srcdir}/.git"], [
     AC_PATH_PROG([git_bin], [git])
     AS_IF([test -n "$git_bin"], [
         srcversion_string="commit: $("$git_bin" -C "$srcdir" describe --tags --long --always --dirty --broken --abbrev=40)"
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -108,6 +108,9 @@ AS_IF([test "x$enable_fontconfig" != xno], [
         fontconfig=true
     ], [
         fontconfig=false
+        AS_IF([test "x$enable_fontconfig" = xyes], [
+            AC_MSG_ERROR([Couldn't locate fontconfig required for requested feature.])
```
astiob, line 112: How about: “fontconfig support was requested, but it was not found.”

Also, the `--help` message claims the default is `--enable-fontconfig=enabled`, which is especially confusing after this change. Set that to `check`?

Also, CoffeeFlux is gonna hate you for changing configure logic yet again :-) But this seems sensible to me.

TheOneric: Fixed the help string and changed the failure messages.

TheOneric:
New version pushed; does now also fail if `--enable-asm[=yes]` was passsed but assembly is not supported on the architecture or a recent enough NASM couldn't be located. This effectively incorporates rcombs' 484ed335953c887a842038e8fb4b4721a5dec4de from #400.

astiob (approved):

Issue #545, opened 2021-09-05
Persian/Arabic lowercase letters problem


Bayany:
It converts them to uppercase letters and ruins the sentence
![image](https://user-images.githubusercontent.com/60580365/132121925-e1cbffc3-8677-4498-bde9-0c62873070fe.png)

Subtitle:
```
Dialogue: 1,0:00:00.57,0:00:01.15,OP - Alok,,0,0,0,fx,{\fscx0\fscy0\an5\3c&H0E0E10&\pos(836,85.9999)\t(0,627,\fscx100\fscy100)\fad(250,0)}ب‍
Dialogue: 1,0:00:00.60,0:00:01.42,OP - Alok,,0,0,0,fx,{\fscx0\fscy0\an5\3c&H0E0E10&\pos(820,85.9999)\t(0,841,\fscx100\fscy100)\fad(250,0)}‍ه‍
Dialogue: 1,0:00:00.62,0:00:01.83,OP - Alok,,0,0,0,fx,{\fscx0\fscy0\an5\3c&H0E0E10&\pos(796,85.9999)\t(0,740,\fscx100\fscy100)\fad(250,0)}‍م
Dialogue: 1,0:00:00.65,0:00:02.25,OP - Alok,,0,0,0,fx,{\fscx0\fscy0\an5\3c&H0E0E10&\pos(760.999,85.9999)\t(0,182,\fscx100\fscy100)\fad(250,0)}ب‍
Dialogue: 1,0:00:00.67,0:00:02.45,OP - Alok,,0,0,0,fx,{\fscx0\fscy0\an5\3c&H0E0E10&\pos(746.999,85.9999)\t(0,990,\fscx100\fscy100)\fad(250,0)}‍گ‍
Dialogue: 1,0:00:00.70,0:00:02.70,OP - Alok,,0,0,0,fx,{\fscx0\fscy0\an5\3c&H0E0E10&\pos(730,85.9999)\t(0,108,\fscx100\fscy100)\fad(250,0)}‍‍و
Dialogue: 1,0:00:00.72,0:00:02.87,OP - Alok,,0,0,0,fx,{\fscx0\fscy0\an5\3c&H0E0E10&\pos(698.999,85.9999)\t(0,642,\fscx100\fscy100)\fad(250,0)}ب‍
[...]

Issue #498, opened 2021-04-25
Memleaks on shaping related failures


TheOneric:
If the below patch isn't a flawed way to test this, the existing fail-cases in `ass_render_event` suffer from memory leaks.
Depending on the failure point I get a few kB or ~½MB of leaked memory per test file.

They can be applied individually and tested with eg ASAN:
```diff
diff --git a/libass/ass_render.c b/libass/ass_render.c
index b91bdd7..be76834 100644
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2657,7 +2657,8 @@ ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
             resolve_base_direction(render_priv->state.font_encoding));
     ass_shaper_find_runs(render_priv->shaper, render_priv, text_info->glyphs,
             text_info->length);
-    if (!ass_shaper_shape(render_priv->shaper, text_info)) {
+    //if (!ass_shaper_shape(render_priv->shaper, text_info)) {
+    if (true) {
         ass_msg(render_priv->library, MSGL_ERR, "Failed to shape text");
         free_render_context(render_priv);
         return false;
diff --git a/libass/ass_shaper.c b/libass/ass_shaper.c
index 4665e7a..7982f03 100644
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -671,9 +671,10 @@ static bool shape_harfbuzz(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
         }
 
         int offset = i;
-        hb_font_t *font = get_hb_font(shaper, glyphs + offset);
-        if (!font)
+        //hb_font_t *font = get_hb_font(shaper, glyphs + offset);
+        //if (!font)
             return false;
+            hb_font_t *font = NULL; //to make compiler happy
         int run_id = glyphs[offset].shape_run_id;
         int level = shaper->emblevels[offset];
 
```

TheOneric:
For reference, below are the memleaks as listed by ASAN using ART's crash-test files. The above diff is applied on top of 1140b6b885c89d37eef13dc1f31f144e9a76a4d7 and comments are toggled as needed. I'm not sure how pointers held in cache can leak, as to my understanding all cache items are supposed to get cleaned up at the end.

If forcing a "fail" during `shape_harfbuzz` *(and not replacing the `ass_shaper_shape` call in `ass_render_event`)* around ½MB leaks:
```
[CRASH-TEST]: ./generic
  From:    0.00s
[...]

PR #553, opened 2021-09-18
Pass ZWJ/ZWNJ to HarfBuzz across shape run boundaries

astiob:
ZWJ and the characters next to it may be separated into different shape runs by bidi, script or font-fallback splitting. Nevertheless, ZWJ must affect Arabic-script characters on both sides of itself. This is noted in the description of the [Unicode bidirectional algorithm](https://unicode.org/reports/tr9/#Joiners) and matches the observed behavior of VSFilter (GDI/Uniscribe).

Fixes https://github.com/libass/libass/issues/545.

Unlike ZWJ, I have not been able to produce a test for ZWNJ that would tell me whether VSFilter does this for ZWNJ as well. It shouldn’t matter for Arabic as the boundaries are non-joining by default, but it might make a difference for Indic scripts or others. It seems to make sense, so in absence of evidence to the contrary, this commit applies the same treatment to ZWNJ as well.


astiob:
Repushed with extra explanation in the commit message.

libass/ass_shaper.c
```diff
@@ -682,9 +691,19 @@ static bool shape_harfbuzz(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
                 level == shaper->emblevels[i + 1])
```
TheOneric, line 691: Unrelated to your patch: It seems to me the `return false` 6 lines prior leaks memory as it doesn't destroy `buf`, doesn't it? *(reminding me of #498 which this appears to be a part of but not all of it)*

libass/ass_shaper.c
```diff
@@ -682,9 +691,19 @@ static bool shape_harfbuzz(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
...
+                is_shaping_control(glyphs[i + 1].symbol))
+            trail_context = 1;
+
         hb_buffer_pre_allocate(buf, i - offset + 1);
-        hb_buffer_add_utf32(buf, shaper->event_text + offset, i - offset + 1,
-                0, i - offset + 1);
+        hb_buffer_add_utf32(buf,
+                shaper->event_text + offset - lead_context,
+                i - offset + 1 + lead_context + trail_context,
+                lead_context, i - offset + 1);
```
TheOneric, line 706: Maybe I'm misunderstanding the harfbuzz-API, but shouldn't the last parameter, item length, be `i - offset + 1 - trail_context` to not include the trail in the buffer, but only for context? *(as done for `lead_context`)*

astiob: `i - offset + 1` is the length of the item. Compare to `i - offset + 1 + lead_context + trail_context` in the line above, which is the length of the item _plus_ context.

The code above peeks at the leading and trailing context without incrementing `i`.

The previous argument, `lead_context`, is the offset from the start of the “text” (the pointer in the second argument) to the start of the item. The whole “text” in this case is the item plus both leading and trailing context.

libass/ass_shaper.c
astiob: Yes, looks likely! Nice catch; I’ll make a fix.

astiob: I’m going through our HarfBuzz code and fixing memory checks and other leaks like this. I may end up touching this line, so I think it’s best to split that into a follow-up PR if this one gets merged soon.

TheOneric (approved):

Issue #554, opened 2021-09-24
Crash if FT_Get_Postscript_Name returns NULL


chosungmann:
Unfortunately, `FT_Get_Postscript_Name`, which is supposed to return the PostScript name of a given font, seems to return `NULL` for some fonts. This leads to crashes because [`ass_face_open()`](https://github.com/libass/libass/blob/master/libass/ass_font.c#L133) operates under the assumption that the PostScript name is not `NULL`. For example, [at this line](https://github.com/libass/libass/blob/master/libass/ass_font.c#L143), a given font cannot be created even if `face->num_faces` has a value greater than `1` due to the fact that `postscript_name` is `NULL`.

I attached the problematic font for testing: [test.zip](https://github.com/libass/libass/files/7221881/test.zip).

#### Result from [the master branch](https://github.com/libass/libass/tree/0915955733bd236ecc44645ee968fb7a55ad5079)

```
➜  test git:(vrew-3124) ✗ ./test image.png subtitle.ass 12
test.c: Available font providers (4): 'None', 'Autodetect', 'CoreText', 'Fontconfig'.
libass: libass API version: 0x1502000
libass: libass source: commit: 0.15.2-9-g03e73dcdb439c0497b44e46d4ba9f92f57c78204
libass: Raster: FreeType 2.11.0
libass: Shaper: FriBidi 1.0.10 (SIMPLE) HarfBuzz-ng 2.9.1 (COMPLEX)
libass: Initialized
libass: Using font provider coretext
libass: File size: 885
libass: [0x7fede8c0dc70] Style: Default,Cafe24 Ohsquare air,52,&Hffffff,&Hffffff,&H0,&H0,0,0,0,0,100,100,0,0,1,1,0,2,10,10,10,0
libass: Track treated as legacy ffmpeg sub.
libass: Added subtitle file: 'subtitle.ass' (2 styles, 5 events)
libass: fontselect: (Cafe24 Ohsquare air, 400, 0) -> /Users/${user_id}/Library/Fonts/Cafe24Ohsquareair.ttf, -1, (none)
libass: Font face with no charmaps
[1]    6614 segmentation fault  ./test image.png subtitle.ass 12
➜  test git:(vrew-3124) ✗ 
```

```
(lldb) bt all
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x50)
  * frame #0: 0x000000010031c41a libfreetype.6.dylib`FT_Request_Size + 50
[...]

PR #555, opened 2021-09-24
fontselect: handle the case where FT_Get_Postscript_Name returns NULL

chosungmann:
Unfortunately, FT_Get_Postscript_Name, which is supposed to return the
PostScript name of a given font, seems to return NULL for some fonts. This
leads to crashes because ass_face_open() operates under the assumption that
the PostScript name is not NULL. This patch address the issue by using an
empty string when FT_Get_Postscript_Name returns NULL.

BUG=#554


astiob:
Crash fixed by b6240cdaef1b51a9add42a4bdfa72064297dbc7d, and desired behaviour for the specific sample (which _accidentally_ worked as long as the PostScript was anything but NULL—now this is intentional) restored by 2b30c69ca3a0c0767dac4f3486c9182d2a79cb38.

Once again thanks for the detailed report!

PR #557, opened 2021-09-25
Follow-up for strikeout/underline commit

MrSmile:
The most strange for me was `FT_Get_Glyph()` call that looks totally unnecessary.
I've removed it in second commit, but maybe there's some hidden meaning in it?

On a side note: maybe we should extend lines by half pixel to mask gaps?

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -461,15 +461,10 @@ int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
  * \brief Get a glyph
  * \param ch character code
  **/
-FT_Glyph ass_font_get_glyph(ASS_Font *font, int face_index, int index,
-                            ASS_Hinting hinting)
+bool ass_font_get_glyph(ASS_Font *font, int face_index, int index,
+                        ASS_Hinting hinting)
 {
-    int error;
-    FT_Glyph glyph;
-    FT_Face face = font->faces[face_index];
-    int flags = 0;
-
-    flags = FT_LOAD_NO_BITMAP | FT_LOAD_IGNORE_GLOBAL_ADVANCE_WIDTH
+    FT_Int32 flags = FT_LOAD_NO_BITMAP | FT_LOAD_IGNORE_GLOBAL_ADVANCE_WIDTH
             | FT_LOAD_IGNORE_TRANSFORM;
     switch (hinting) {
     case ASS_HINTING_NONE:
@@ -485,28 +480,18 @@ FT_Glyph ass_font_get_glyph(ASS_Font *font, int face_index, int index,
         break;
     }
 
-    error = FT_Load_Glyph(face, index, flags);
+    FT_Face face = font->faces[face_index];
+    FT_Error error = FT_Load_Glyph(face, index, flags);
     if (error) {
         ass_msg(font->library, MSGL_WARN, "Error loading glyph, index %d",
                 index);
-        return 0;
+        return false;
     }
-    if (!(face->style_flags & FT_STYLE_FLAG_ITALIC) &&
-        (font->desc.italic > 55)) {
+    if (!(face->style_flags & FT_STYLE_FLAG_ITALIC) && (font->desc.italic > 55))
         FT_GlyphSlot_Oblique(face->glyph);
-    }
-
-    if (font->desc.bold > ass_face_get_weight(face) + 150) {
+    if (font->desc.bold > ass_face_get_weight(face) + 150)
         ass_glyph_embolden(face->glyph);
-    }
-    error = FT_Get_Glyph(face->glyph, &glyph);
-    if (error) {
-        ass_msg(font->library, MSGL_WARN, "Error loading glyph, index %d",
-                index);
-        return 0;
-    }
-
-    return glyph;
+    return true;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


libass/ass_font.c
```diff
@@ -620,3 +523,87 @@ void ass_font_clear(ASS_Font *font)
...
+
+/**
+ * \brief Convert glyph into ASS_Outline according to decoration flags
+ **/
+bool ass_get_glyph_outline(ASS_Outline *outline, int32_t *advance,
+                           FT_Face face, FT_Glyph glyph, unsigned flags)
+{
+    int32_t y_scale = face->size->metrics.y_scale;
+    int32_t adv = flags & DECO_ROTATE ? face->glyph->linearVertAdvance : glyph->advance.x;
+    adv = ((adv >> 9) + 1) >> 1;  // cannot overflow
```
astiob: Is this an inlined version of `d16_to_d6`? I’d prefer that to be used explicitly and, if overflow is a concern, `d16_to_d6` itself to be adjusted to use this code.

MrSmile: I'm not sure if `face->glyph->linearVertAdvance + 512` can overflow. `glyph->advance.x + 512` certainly cannot following mention of it in `FT_Get_Glyph()` docs. Actually, after this PR it will be the only use of `d16_to_d6()` so maybe better to remove `d16_to_d6()` then?

astiob: My thinking is that the conversion functions make intent obvious and allow us to centralize any rounding/overflow fixes (like now). I now see that they probably apply in the pos/size/desc conversions below, too.

libass/ass_outline.c
```diff
@@ -24,45 +24,75 @@
...
+    outline->n_segments = outline->max_segments = 0;
+}
+
+/*
+ * \brief Initialize ASS_Outline and allocate memory
+ */
+bool outline_alloc(ASS_Outline *outline, size_t max_points, size_t max_segments)
+{
+    assert(max_points && max_segments);
+    outline->points = malloc(sizeof(ASS_Vector) * max_points);
```
TheOneric, line 46: Pre-existing issue, but this will allocated to few memory if `max_points > SIZE_MAX / sizeof(ASS_Vector)`, causing subsequentmemory- overreads.
Using one of `ASS_REALLOC_ARRAY` or `ass_(try_)realloc_array` would avoid that by returning `NULL` or not modifying the (`NULL`-initialised) pointer instead.

MrSmile: In all use cases `max_points` is up to factor of 2 larger than previously existing point array size. So it's physically impossible to hit that limit. I can add `assert(max_points <= SIZE_MAX / sizeof(ASS_Vector));` to catch incorrect input due to overflows, but real handling of that case looks like overkill for me.

TheOneric: Is this previously existing point array size guaranteed to always be `< SIZE_MAX / sizeof(ASS_Vector)`? Regardless, it might still be good to have such an overflow check, assert or a comment addressing why it can't happen to aid readability.

Assuming this is never supposed to happen, I don't see much of a difference between an manual `assert` crashing and using `ass_realloc_array` failing the allocation. I'd be fine with either one and would prefer it over a silent assumption of overflow being impossible.

libass/ass_font.c
MrSmile: Replaced with `d16_to_d6()` call.

libass/ass_outline.c
MrSmile: Added a check for reliability.

libass/ass_font.c
```diff
@@ -620,3 +523,86 @@ void ass_font_clear(ASS_Font *font)
...
+    int32_t y_scale = face->size->metrics.y_scale;
+    int32_t adv = flags & DECO_ROTATE ? face->glyph->linearVertAdvance : glyph->advance.x;
+    *advance = adv = d16_to_d6(adv);
+
+    int n_lines = 0;
+    int32_t line_y[2][2];
+    if (adv > 0 && (flags & DECO_UNDERLINE)) {
+        TT_Postscript *ps = FT_Get_Sfnt_Table(face, ft_sfnt_post);
+        if (ps && ps->underlinePosition <= 0 && ps->underlineThickness > 0) {
+            int64_t pos  = ((int64_t) ps->underlinePosition  * y_scale + 0x8000) >> 16;
```
astiob, line 529: FWIW this differs from `FT_MulFix`: this rounds half towards +infinity, whereas `FT_MulFix` rounds it away from zero. But we probably don’t mind.

MrSmile: That's why I replaced `FT_MulFix`: its code looks like a case of heavy overengineering (they tried to combat implementation defined right shifts?), so I decided to use simple and consistent formula coupled with correct overflow handling.

astiob (approved):

libass/ass_font.c
```diff
@@ -620,3 +523,86 @@ void ass_font_clear(ASS_Font *font)
...
+            int64_t size = ((int64_t) ps->underlineThickness * y_scale + 0x8000) >> 16;
+            pos = -pos - (size >> 1);
+            if (pos >= -OUTLINE_MAX && pos + size <= OUTLINE_MAX) {
```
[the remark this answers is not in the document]
MrSmile: Should not as both `pos` and `size` have at least 16 padding bits due to shift.

libass/ass_font.c
astiob: To be clear, I don’t mind this. My only lament is that `FT_MulFix(a, b)` _reads_ much nicer than `((int64_t) a * b + 0x8000) >> 16`.

---

The main point of `FT_MulFix` is outdated portability: FreeType works on platforms that don’t have native int64 (and in case the compiler emulates int64 but is bad at it, they provide faster assembler implementations).

I assume the half-away-from-zero rounding is a conscious choice; indeed, I would normally do the same.

I don’t see any difference in overflow handling though. Do you want to elaborate?

(And I’m still not a fan of assuming arithmetic right shift, but it’s a given for this PR.)

libass/ass_font.c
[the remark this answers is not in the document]
astiob: Oh, they’re still `int64_t`. I didn’t pay attention.

libass/ass_font.c
MrSmile: There's a possibility of [silent truncation](https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/src/base/ftcalc.c#L243) in case of `sizeof(long) == 4`.

astiob: Ah. Right. This goes together with my other comment: I didn’t notice that you’re storing the result in `int64_t`.

Good.

<i>(Ironically, you’ll also notice that the implementation you linked does in fact assume arithmetic right shift, although the pure-32-bit-C implementation refuses to do it.)</i>

TheOneric (approved):

PR #558, opened 2021-09-26
Fix buffer overrun in rasterizer

MrSmile:
Found buffer overrun (since assembly overhaul, I think) in rasterizer, but instead of an easy fix decided to go the long way and complete longstanding TODO.


astiob:
May I ask a stupid question?

My intuitive understanding of tiled rendering is that tiles are disjoint rectangles of whole pixels. Why should there be any difference, then, whether addition or max is used to combine them?

MrSmile:
It's about mechanism to combat different winding direction: I'm generating two different border outlines with positive and negative offset, rasterize them independently and then blend. That final blending should be max-blending and not addition.

astiob (approved):
Ah, that thing! Right. LGTM (including the new commit).

astiob:
Ah: could you perhaps clarify just how big the buffer overrun was, and whether it was read-only or write as well?

TheOneric (approved):

PR #559, opened 2021-09-26
BitmapEngine refactoring + new assembly

MrSmile:
This is a part of my version of the checkasm branch. I decided to split it into several PRs for ease of reviewing. This part is mostly preliminary refactoring.


libass/ass_utils.c
```diff
@@ -31,41 +31,6 @@
...
-    ass_get_cpuid(&eax, &ebx, &ecx, &edx);
-    if (!(ecx & (1 << 27))) // not OSXSAVE
-        return 0;
-    uint32_t misc = ecx;
-    ass_get_xgetbv(0, &eax, &edx);
-    if ((eax & 0x6) != 0x6)
-        return 0;
-    eax = 0;
-    ass_get_cpuid(&eax, &ebx, &ecx, &edx);
-    return (ecx & 0x6) == 0x6 ? (misc >> 28) & 0x1 : 0; // check high bits are relevant, then AVX support
```
TheOneric, line 57: There seems to be no equivalent of the `(ecx & 0x6) == 0x6` in the replacement code. However afaict the new version is correct and the old incorrect. On my Zen+-CPU, the old `has_…` functions actually claim it would only supported SSE2, while the new `ass_get_cpu_flags` correctly identifies SSE2, AVX and AVX2 to all be supported. Does this also affect other µarches?
The `ass_get_cpu_flags` commit-message should document the fix.

TheOneric (commented):
Not an expert in assembly and the related code , but afaict it seems good, except for one documentation issue.

libass/ass_utils.c
MrSmile: First and third commits have taken verbatim from @rcombs [checkasm](https://github.com/libass/libass/pull/400/commits) branch (third had merge conflict but only with context and not actual changeset) so I've not looked deeply into them. I'll dive deeper when I have time.

MrSmile: Actually this CPU-flag business is orthogonal to moving code to separate directory. So I should postpone it to an actual checkasm PR.

astiob: Just my 2¢, perhaps irrelevant: I remember x86 CPUID checks being tricky in the sense that some extensions may be supported by the CPU but not by the OS, in which case they mustn’t be used. Could it be that the 0x6 mask checked something along these lines? At any rate, whenever anyone gets around to re-checking this, they need to be _careful._

MrSmile:
New version. Changes:
 - Removed CPU-flag commits;
 - Massive refactoring of `ass_be_blur_c()`: I think new version is simpler to understand and it's actually faster for me (~20%).

libass/ass_utils.c
TheOneric: I tested this on some more CPUs, the results are that the new (now removed) version was correct in all tests, the old version wrong for all AMD CPUs. In more detail the old version reports (putting what of the tested the architecture actually supports in paranthises):
 - AMD Zen+ (SSE2, AVX, AVX2): **SSE2**
 - AMD Puma+ (SSE2, AVX): **SSE2**
 - Intel Kaby Lake (SSE2, AVX, AVX2): **SSE2, AVX, AVX2**
 
 ---
 
 [Reading up](https://en.wikipedia.org/w/index.php?title=CPUID&oldid=1040897815#EAX=0:_Highest_Function_Parameter_and_Manufacturer_ID) on what info CPUID-leaf 0 contains, `ebx`, `edx` and `ecx` simply contains a 12byte SCII-string of the manufacturer name. The `(ecx & 0x6) == 0x6` is a spot-check which is true for `"GenuineINtel"` and false for `"AuthenticAMD"`. Thus preventing all AMD, Via, Cyrix, etc and also many Hypervisors, VMs and emulators — *even if running on a physical Intel CPU* — from making use of AVX and AVX2.
 As stated before, I'm no expert in x86(-assembly), but this seems undesireable to me. At least the CPUID-page of the English Wikipedia does not note any difference in the encoding of any of the feature flags we check between different manufacturers.

TheOneric:
>     * Massive refactoring of `ass_be_blur_c()`: I think new version is simpler to understand and it's actually faster for me (~20%).

Can you clarify which CFLAGS you used and if the ~20% come from isolated (eg checkasm) test of the function or if it is the  overall improvement for an ASS-sample *(in which case: what sample and what test procedure)*?

MrSmile:
It's from the timing of that function in the checkasm branch. I've used it to check if my refactoring have negative impact on performance. I've used `CFLAGS=-g -Wall -fno-omit-frame-pointer -Ofast -march=native` for that particular test, but as checkasm timings are quite noisy it's not a reliable performance test.

libass/ass_utils.c
TheOneric: > […] So I should postpone it to an actual checkasm PR.

Since this blocks non-Intel- and virtual-CPUs from making use of our AVX2-routines and after discussion with rcombs on IRC, there appears to be no need for such a (broad) block, I'd prefer to have this fixed soon'ish, though not necessarily (but possibly) in this PR.
Whether to do this independent or together (but documented) with the BitmapEngine-change, I have no opinion on.
Same goes for whether to check for and block some of the very first AVX (but iinm not AVX2) CPUs by AMD like [FFmpeg partially does](https://git.ffmpeg.org/gitweb/ffmpeg.git/blob/1ad4782d2a17d9fed38c0356f1c9f790973eec01:/libavutil/x86/cpu.c#l191) (those *may be* slower with AVX than SSE2).

MrSmile:
Okay, reintroduced CPU-flag back, but refactored: as flags now aren't used for CPU only, keeping it in `ass_cpu.h` doesn't make sense, so I merged them into `ass_bitmap_engine.h`. Also removed flags unused by libass. There's `ASS_CPU_FLAG_X86_AVX512ICL` left as it used by checkasm. Is it better to remove it too?

astiob:
I see you’ve replaced some `intptr_t` by `ptrdiff_t` in the prototype of `be_blur`. I’ve asked about `intptr_t` before, and the answer was that it’s intentionally used as even though it’s not the right _C_ type, it’s the safest type for the C–asm interop, because the C compiler/library might define `ptrdiff_t` and `size_t` as integers of different width than the native register/pointer width, which the assembler code expects. [x264’s own x86 asm code](https://code.videolan.org/videolan/x264/-/tree/master/common/x86) does this. If nothing else, pairing `ptrdiff_t` with `uintptr_t` seems weird to me.

astiob (commented):
The same applies to `{add,imul,mul}_bitmaps` (indeed, that’s the bulk of the change) and to the existing rasterizer & blur functions that already use `ptrdiff_t` with `uintptr_t`: my unsophisticated understanding is that they shouldn’t. The primary objective is to ensure the two types can be added together in the asm code, so they should form a signed/unsigned pair in C. The secondary objective is that their width must reliably match _something_ on the asm side, currently the native register size, for which `intptr_t` is deemed safest.

The “deduplicate…” commit makes the compiler see some function declarations twice, right? This doesn’t hurt anything, of course; but just as an idea, maybe the template file should be split?

Other than this, LGTM. And none of these comments is strictly a blocker.

libass/C/blend_bitmaps.c
```diff
@@ -22,51 +22,48 @@
...
-    unsigned out;
-    uint8_t* end = dst + dst_stride * height;
+    uint8_t *end = dst + dst_stride * height;
     while (dst < end) {
-        for (unsigned j = 0; j < width; ++j) {
-            out = dst[j] + src[j];
-            dst[j] = FFMIN(out, 255);
+        for (uintptr_t x = 0; x < width; x++) {
+            unsigned out = dst[x] + src[x];
+            dst[x] = out < 255 ? out : 255;
```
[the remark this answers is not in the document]
MrSmile: Ah, it's an early change when I tried to decouple checkasm from the whole libass. I've failed at that in the end, so I can revert this line.

MrSmile:
About `ptrdiff_t`: I'm fine either way and simply changed for consistency. Personally I think that it's fine to rely on pointer, `ptrdiff_t` and `size_t` being the same size as it's only matter for architectures where we use assembly (currently only x86). We can try to guard from insane compilers that have `sizeof(ptrdiff_t) > sizeof(intptr_t)` in x86 (I'm pretty sure that such doesn't exist) but we're screwed anyway as we don't have guarantee that conversion from `ptrdiff_t` to `inptr_t` is safe in that case. On the other hand, using `ptrdiff_t` is safer for C-versions as it's guaranteed to work in any compiler/architecture.

> The “deduplicate…” commit makes the compiler see some function declarations twice, right?

I have conflicting feelings about that too as some compilers throw a bunch of warnings for that. I don't like many file approach as that doesn't help understanding. Maybe include guards for parts of `ass_func_template.h` would work?

> Personally, I’m not too big a fan of the capitalized `C` directory name. I’d prefer lowercase `c`.

Okay, I'll change to lowercase.

TheOneric:
Re `intptr_t`, `ptrdiff_t`, etc

> it's fine to rely on pointer, ptrdiff_t and size_t being the same size as it's only matter for architectures where we use assembly

Per C-standard all or some of `ptrdiff_t` and `(u)intptr_t` could in theory be *larger* than the pointer size, since it's only required to correctly be converted back and forth; unless I missed it, the stqandard doesn't say much about if conversion back to pointers is still safe if the value got modified. In fact with a flat memory model, it would make sense for `ptrdiff_t` to be one bit larger than the pointer size and ISO-C99 requires a minimum of `32767 = 2^15 -1` for `INTPTR_MAX` but a minimum of `65535 = 2^16-1` for `ptrdiff_t`.
However, I guess it is unlikely for `sizeof((u)intptr_t) != sizeof(void*)` to  actually happen.

Practically, on systems (yes also x86 systems) with segmented memory models, `size_t` and `ptrdiff_t` is or can be *smaller* than the pointer size. *(Is this safe for passing to asm, or will the free part of the register contain junk?)*
And ofc, we may gain assembly for more architectures in the future.

TheOneric (commented):
Just some nits and questions.

libass/ass_bitmap_engine.h
```diff
@@ -87,6 +87,7 @@ enum {
     ASS_CPU_FLAG_X86_AVX512ICL = 0x04,  // placeholder for checkasm
     ASS_CPU_FLAG_MAX,
     ASS_CPU_FLAG_ALL = ((ASS_CPU_FLAG_MAX & ~1ull) << 1) - 1,
+    ASS_FLAG_LARGE_TILES       = 0x08,
```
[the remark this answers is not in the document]
astiob: That reminds me: why is it `ASS_CPU_FLAG_MAX & ~1ull` rather than the simpler `ASS_CPU_FLAG_MAX - 1ull`?

> Do we actually need a CPU-specfic `ASS_CPU_FLAG_ALL`, or could we use a general `ASS_FLAG_ALL`-mask instead?

Well, the way this is used in this PR is that `ASS_CPU_FLAG_ALL | optionally, depending on configure flags, ASS_FLAG_LARGE_TILES` are enabled. But I’m not sure what the goal is.

libass/C/be_blur.c
```diff
@@ -24,69 +24,51 @@
 #include <stdint.h>
 
 
+static inline uint16_t sum_swap(uint16_t *ptr, uint16_t val)
+{
+    uint16_t sum = *ptr + val;
+    *ptr = val;
+    return sum;
+}
+
```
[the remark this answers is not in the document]
MrSmile: I dislike both `sum_swap` and `sum_copy`. Maybe something with intent and not description, like `running_sum` or `pair_sum`, would be better?

libass/ass_bitmap_engine.h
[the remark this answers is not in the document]
MrSmile: > each addition of a new CPU-related flag will also need to modify the values of non-CPU flags

Good catch. I'll add some room for them, i.&nbsp;e. `ASS_FLAG_LARGE_TILES = 0x10000` or something.

> Do we actually need a CPU-specfic `ASS_CPU_FLAG_ALL`, or could we use a general `ASS_FLAG_ALL`-mask instead?

No, it's exactly what is needed. It's intention is "use all CPU features available". All other flags are configuration, we want to enable them individually.

> For my own understanding, can you elaborate on why this is needed?

Look into my last [checkasm](https://github.com/MrSmile/libass/commits/checkasm) branch, specifically [here](https://github.com/MrSmile/libass/commit/83a8f169481e3a5c2a73de93ee51db62565f3471#diff-b878f925440d48c88a81a7564472974abb7170db8259751a8d5de111c3833accR285).

> That reminds me: why is it `ASS_CPU_FLAG_MAX & ~1ull` rather than the simpler `ASS_CPU_FLAG_MAX - 1ull`?

I think I'll remove `ASS_CPU_FLAG_X86_AVX512ICL`, `ASS_CPU_FLAG_MAX` and redefine `ASS_CPU_FLAG_ALL` to a fixed number like `0xFFFF`.

MrSmile:
> And ofc, we may gain assembly for more architectures in the future.

We can handle platform-specific types in the corresponding assembly anyway, as I've done with `uint32_t` and `uint64_t` parameters of `fill_halfplane_tile`. Actually I'm more and more inclining to replace `uintptr_t` with `size_t` too: it's a correct type and guaranteed to work for C fallback functions. All architecture-specific parts could be handled in assembly.

astiob:
TL;DR: Hopefully this doesn’t currently matter. I do prefer `size_t` in C. But technically `uintptr_t` seems safer for our x86 asm code and for likely future non-x86 code; and x264 uses `intptr_t` for their many-arch asm.

But FWIW FFmpeg (after a cursory look only) seems to use `int` (!) and `ptrdiff_t` for strides in cross-platform interfaces and explicit, conditionally-defined `int{32,64}_t` in x86-specific internal interfaces.

---

I absolutely agree that _in pure C,_ `size_t` and optionally `ptrdiff_t` are the right types. This was precisely my instinct when I first saw the types in our first assembler functions, too.

But the thing that the x264 people (?) realized is, that’s not quite the goal. (And we can, of course, disagree and do our own thing. But they do have more experience supporting assembler on more architectures, so the argument is worth hearing.)

Their advantage in C comes from the fact libc defines them to its preferred types for the current platform. They’re expected to be optimal in some sense, but it’s inherently an arbitrary choice left to the whims of libc and/or the compiler.

What we want for assembler, instead, is a type that is most easily predictable across C libraries and compilers on each architecture, and ideally suited for indexing pointers on as many architectures as possible (via simple addition or indirect addressing constructs). Hence, `intptr_t`: while the C environment has freedom to impose constraints on object size and pointer differences may need an extra sign bit, the conversion of pointers to integers is, at least on flat-address-space architectures, quite a bit more straightforward and unambiguous and generally makes pointer arithmetic trivial.

Of course, each architecture’s assembler code still ultimately needs to ensure it handles the types, regardless of which types we choose. But keep in mind that C types aren’t necessarily decided by the architecture, but rather by the C library and/or compiler.

What our current assembler code (and likely future code) does is just add integers to pointers, so what it really wants to take is “integers the same size as byte pointers”. `intptr_t` is the closest thing that C has.

At the end of the day, it’s still only a guess, and an unconventional (or evil) compiler might even define _all_ of the standard types to different widths and still conform to the C standard. But as long as we stick to standard types: if we’re gonna assume we know the exact size of `size_t` and `ptrdiff_t`, then they become no more “correct” than e. g. `intrptr_t`; and if we want to leave the libc/compiler some wiggle room, then `intptr_t` seems the least likely to be different from assembler-code pointers.

(Indeed, the most fool-proof way would be to define custom fixed-width types based on the architecture… or on `sizeof(void*)`, given that we’re passing actual C pointers to assembler functions, too, and assuming they Just Work.)

And of course, I concur that I don’t have actual examples of platforms where this choice even matters, and I _hope_ they don’t exist (or we don’t/won’t have assembler code for them), although I don’t know such details of many architectures to begin with. Ultimately, this is about minimizing the already-minuscule risk on currently supported platforms and making the code maximally future-proof for when we want to add others.

---

Where `size_t` and/or `ptrdiff_t` might actually be _better_ is on weird architectures where they might be _bigger_ than `intptr_t`: we have no assembler code for them, but the C code would fail to run properly with `intptr_t` because the arguments would get truncated/UB’d. But if we were to use the unsigned `uintptr_t` exclusively, then I think such an architecture can’t be constructed without violating the C standard. Even if `size_t` and `ptrdiff_t` are wider _types,_ I think all their nonnegative _values_ should always fit in `uintptr_t`.

x264 and our existing code use signed `intptr_t` for strides/sizes. I suppose this is technically insufficient, in case the OS allows an object to occupy more than half of the entire address space and the stride’s conversion to `intptr_t` overflows.

libass/ass_bitmap_engine.h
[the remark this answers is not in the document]
TheOneric: > I'll add some room for them, i. e. `ASS_FLAG_LARGE_TILES = 0x10000` or something. 

ISO-C99 only guarantees enums to hold the same values as `int` and `int`'s maximum value must only be at least `2^15 - 1`, so `0x10000` is technically UB, depending on the target platform. *(POSIX raises the minimal acceptable value to `2^31 - 1`)*
Iinm currently all of libass' enum values are below `2^15 - 1`.

>  Look into my last checkasm branch, specifically here.

Thanks.

libass/C/be_blur.c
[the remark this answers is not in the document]
TheOneric: `running_sum` seems good to me.

MrSmile:
Okay, I've looked into a couple of projects with assembly, my findings:
* dav1d uses `ptrdiff_t` and `size_t`;
* I'm inclining to say that ffmpeg mainly uses `ptrdiff_t`, but there's also `intptr_t` even in the same functions as `ptrdiff_t`;
* x264 and x265 use `intptr_t` for strides;
* absolutely all projects have `int` parameters in assembly functions.

We rely on platform ABI for assembly interop and sizes of `ptrdiff_t`/`size_t` are a part of ABI too, so I incline to the pure C approach and the latest of the projects (dav1d) agrees with me.

libass/C/be_blur.c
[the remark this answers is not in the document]
MrSmile: Sum of sequence til some point is usually called running sum and it's not a case here. I think
```c
static inline uint16_t pair_sum(uint16_t *prev, uint16_t next)
```
is better. Or maybe `cur` or `val` instead of `next`.

libass/ass_bitmap_engine.h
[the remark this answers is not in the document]
astiob: Standard reminder that the rest of libass assumes `int` holds 32 bits anyway, as much as I don’t like it. But this has been improving with MrSmile’s PRs that adopt fixed-width types, and it shouldn’t hurt to make this enum more compliant ahead of time—but only if we think it leaves enough bits for future expansion, of course.

astiob:
Thanks for the research!

> We rely on platform ABI for assembly interop and sizes of `ptrdiff_t`/`size_t` are a part of ABI too

This is a very good point! But I’m not sure how true this premise is. The ABI defines what an `int` is, but not necessarily the C typedefs. For example, I’ve just checked the System V x86 psABIs (used by x86 UNIXes) (links: [x86-64](https://gitlab.com/x86-psABIs/x86-64-ABI), [i386](https://gitlab.com/x86-psABIs/i386-ABI/-/tree/hjl/x86/master)), and they do define `size_t` (only since 2002) but say nothing about either `ptrdiff_t` or `[u]intptr_t`. Microsoft’s [x64 docs](https://docs.microsoft.com/en-us/cpp/build/x64-software-conventions?view=msvc-160) don’t seem to mention _any_ of these types. On the other hand, it’s true that we do check the whole platform triple in `configure` and it just feels intuitively reasonable to assume a particular triple implies a particular `size_t` and `ptrdiff_t`, much like our C code assumes some Microsoft things when targeting Windows. Still, I don’t have confidence that this intuition is correct.

If we do settle on assuming particulars of `size_t` and `ptrdiff_t`, I think we should migrate away from `intptr_t` completely, rather than stay with the weird mix of `ptrdiff_t` and `uintptr_t`.

BTW, what’s the motivation for `ptrdiff_t` in particular? Why not `size_t` for everything?

---

Speaking of ABI, I believe our assembler code is technically broken on x32 (the x86-64 ILP32 ABI) and sort of only works by luck.

The assembler code assumes all input pointers and sizes are full 64-bit registers, as on x86-64. On the C side, pointers and sizes are 32-bit. The psABI says _pointer_ arguments must be zero-extended in function calls, so those are fine. But it says nothing about integers. Indeed, a test on Godbolt confirms GCC ensures zero extension for pointer but not for integer arguments.

Our code apparently works in practice (as evidenced by libass-tests), but only because the arguments are `mov`’d into 32-bit registers, which implicitly zero-extends them. But—as much as it’s hard to imagine this happening—they could potentially form as the truncation of 64-bit values, in which case the register’s upper half would contain garbage.

MrSmile:
> I think we should migrate away from `intptr_t` completely

Yes, that's the plan.

> BTW, what’s the motivation for `ptrdiff_t` in particular? Why not `size_t` for everything?

Strides technically can be negative. We don't currently use such bitmap-reversing technique but why not have support for it for free? Also stride is the difference of pointers to adjacent bitmap rows hence `ptrdiff_t`.

libass/C/be_blur.c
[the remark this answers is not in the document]
TheOneric: `pair_sum` reads like just the sum of two numbers to me; while the function does return the sum of two numbers, it also sets the value of the first arg to the second arg's. How about `sum_history` or `sum_memory` as it records/memorises the last summand? Or `sum_moving` because the value of the pointer argument moves through all used summand?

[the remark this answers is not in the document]
astiob: `sliding_sum`? `windowed_sum`?

MrSmile:
Improved version: directory renamed to lowercase `c`, `intptr_t/uintptr_t` replaced with `ptrdiff_t/size_t`, other issues addressed. There's one issue left: how to get rid of duplication declaration warning.

TheOneric:
> The “deduplicate…” commit makes the compiler see some function declarations twice, right?
>
> I have conflicting feelings about that too as some compilers throw a bunch of warnings for that. I don't like many file approach as that doesn't help understanding. Maybe include guards for parts of ass_func_template.h would work?

The lowest effort workaround would be adding `-Wno-redundant-decls` to  `AM_CFLAGS` in `libass/Makefile.am`, but I'm not a fan of that. Declaration guards specific to the macros values, like `#if !LARGE_TILES && !defined(WIDE_C)` for `DECORATE`, `#if !LARGE_TILES` for `DEC_A` and `#ifndef WIDE_C` for `DEC_T` should work. Splitting the declaration headers may be a cleaner solution though.

Apart from the redundant declarations the current version seems good to me.

MrSmile:
Actually, `-Wredundant-decls` is not enabled by default by `-Wall` or `-Wextra`, we're adding it manually [here](https://github.com/libass/libass/blob/master/libass/Makefile.am#L3). So removing it isn't quite a hack, only a profit/loss concern.

TheOneric:
> we're adding it [`-Wredundant-decls`] manually here. So removing it isn't quite a hack, only a profit/loss concern.

I still think this would be a hack. What's with guarding the definitions behind a to be externally defined macro as I suggested before? This would keep the  logic for what we need or don't need outside the header and also avoids redundant declarations. Something like this:
```C
#ifndef OMIT_T_DECLS
void DEC_T(fill_solid_tile)(uint8_t *buf, ptrdiff_t stride, int set);
// ...
#endif

#ifndef OMIT_A_DECLS
void DEC_A(stripe_unpack)(int16_t *dst, const uint8_t *src, ptrdiff_t src_stride,
                          size_t width, size_t height);
// ...
#endif

// ...
```

MrSmile:
> What's with guarding the definitions behind a to be externally defined macro as I suggested before?

It's a leek of logic innards from `ass_func_template.h` to the outside. Or violation of encapsulation. Its use in `ass_bitmap_engine.c` is already quite complex, additional bookkeeping enlarges that code 2&ndash;3 times. I've considered another approach: isolate all bookkeeping in `ass_func_template.h` instead&mdash;create preprocessor switch for `PREFIX` and create include guards individually for each case. But due to insufficiencies of C preprocessor it's quite long and repetitive code too.

On the other hand, redundant declarations cannot lead to any real bugs and security vulnerabilities, it's a purely cosmetic deficiency. And we have a ton of redundant `#include` directives too, which, while cosmetic, can lead to longer compilation times. Anyway, I feel that trading logic for cosmetic isn't a right move.

rcombs:
Compilation times are just about on the bottom of the list of things I'm worried about for this project, which builds from clean in under 10 wallclock seconds for me.

TheOneric:
> > What's with guarding the definitions behind a to be externally defined macro as I suggested before?
> 
> It's a leek of logic innards from ass_func_template.h to the outside.

`ass_func_template.h` already requires and uses externally defined macros to change the output (`DECORATE`, `WIDE_C`, ...); I don't see why having one or two additional macros to control which definitions are generated from the template is a “leak of logic innards” but those others aren't.
While technically redundant declarations are harmless if there signatures really match; in practice ending up with redundant declarations is most likely an indication that something is off so I do no want to globally disable warnings about it.

MrSmile:
> `ass_func_template.h` already requires and uses externally defined macros to change the output

They're parameters that define output `BitmapEngine` struct, not some internal prototypes.

Anyway, I've adopted your suggestion, but in a modified way: I've merged the whole `ass_func_template.h` into `ass_bitmap_engine.c` (so there're no local innards anymore) and replaced multiple include trick with macros. I think the result is even simpler than before.

rcombs:
Can we please stop building new stuff around these static-data `ass_bitmap_engine_*` structs that expect every routine to always be implemented for every extension. I removed that on my branch deliberately.

MrSmile:
I think some simple ifdeffery can easily circumvent that. Or you can directly append the whole struct definition directly, should be more or less equivalent to your dynamic approach.

rcombs:
Why, though? Every single other open-source multimedia project with substantial assembly code I'm aware of uses the dynamic-vtable approach to at least some extent, and even those that set vtable contents in large groups (e.g. rav1e) have separate lists for each supported ISA extension rather than building their infrastructure around the assumption that everything will always be implemented everywhere. FFmpeg, x264, x265, dav1d, libyuv (which I'd never even heard of before)… Even libvpx and libaom do it, though they use a monstrous perl script to generate everything. Why are we doing something that no other project does, that makes it harder to add support for new platforms, and doesn't make any internal sense?

MrSmile:
I've shown how to make a separate list for ISA, so I don't see the problem. Instead, your dynamic approach looks harder for me, as it's a ton of copy-paste distributed through several files (you currently have [a error](https://github.com/libass/libass/pull/400/commits/07baea17d927f2d15ce2ff5dd6d12fe10e40d36c#diff-0c1187bf4a1bb9e2aced66eb83fd6c20235736fd7766209033b32fb3e0576601R100) in that copy-paste).

Actually, there's another approach: designated initializers can repeat and only last matters. So it can be done this way:
```c
const BitmapEngine ass_bitmap_engine_neon = {
    BIMAP_ENGINE(4, 16, 4, 16, c),
    .add_bitmaps    = ass_add_bitmaps_neon,
    .imul_bitmaps   = ass_imul_bitmaps_neon,
    .mul_bitmaps    = ass_mul_bitmaps_neon,
    .be_blur        = ass_be_blur_neon,
};
const BitmapEngine ass_bitmap_engine_lt_neon = {
    BIMAP_ENGINE(4, 16, 5, 32, c),
    .add_bitmaps    = ass_add_bitmaps_neon,
    .imul_bitmaps   = ass_imul_bitmaps_neon,
    .mul_bitmaps    = ass_mul_bitmaps_neon,
    .be_blur        = ass_be_blur_neon,
};

```

MrSmile:
Updated and rebased. Main changes:
- adopted new `ARCH_*` configure flags;
- set copyright for all new files to "libass contributors".

Also I've picked @rcombs NEON assembly into my [checkasm](https://github.com/MrSmile/libass/commits/checkasm) branch, so there's no problem in my static `BitmapEngine` approach. I don't have M1, so I've used qemu + schroot, and it looks like a chunk of checkasm is missing due to that triggering another branch. I've ended up doing additional copy-paste from dav1d to fix that, but the result is quite ugly and incomplete. Also, there's a bug in `add_bitmaps_neon`, which I've fixed.

TheOneric:
For the record:
I'm not convinced splitting of the C implementations is worth the trouble and duplicated object file names, but do not consider this a blocker.
Regarding the approach to the BitmapEngine though, I'd really prefer if a consensus could be reached before anything is merged on a “didn't hear feedback in X days” basis.

From a look at it, I'm inclined to agree with the dynamic engine approach being more flexible and easier to work with while requiring seemingly less code *(assuming I'm notmisjudging the impact the additional tile stuff etc would require with the dynamic approach)*. **But** as stated before I'm no expert in the ASM infrastructure and would much rather see a consensus being reached by @MrSmile and @rcombs, who are familiar with it.

MrSmile:
> I'm not convinced splitting of the C implementations is worth the trouble and duplicated object file names

I've added `c_` prefix to avoid duplication, so that problem should be solved. As for splitting, after trying to set aarch64 virtualisation, I'm more and more inclined to maximally separate checkasm from the rest of libass, as third-party dependencies are a real pain.

> I'm inclined to agree with the dynamic engine approach being more flexible and easier to work with while requiring seemingly less code

Ehm, [dynamic approach](https://github.com/libass/libass/commit/07baea17d927f2d15ce2ff5dd6d12fe10e40d36c#diff-f8a9c505294aeb2047680b1bc4097c32b19b786e81b287a7dc6716a1bd18ba81R37-R74) and equivalent [static code](https://github.com/MrSmile/libass/commit/ac1489249c51d38a5c389fbd30aa03e26e35f303#diff-f8a9c505294aeb2047680b1bc4097c32b19b786e81b287a7dc6716a1bd18ba81R49-R79). Also, the same static code is reused for SSE2/AVX2, while for dynamic it's a whole ton of [error-prone copy-paste](https://github.com/libass/libass/commit/07baea17d927f2d15ce2ff5dd6d12fe10e40d36c#diff-0c1187bf4a1bb9e2aced66eb83fd6c20235736fd7766209033b32fb3e0576601R35-R105). And it's all without complications for tile size and blur stripe width. So I don't see any sort of benefit of dynamic approach. @rcombs had concern about integrating partial NEON assembly into infrastructure, so I've tried and implemented it without problems.

TheOneric:
Basically, rcombs’ comments [[1](https://github.com/libass/libass/pull/559#issuecomment-986183696), [2](https://github.com/libass/libass/pull/559#issuecomment-986188703)] *(and #400 still being open)* read to me like they were intended to be a blocker and I'd prefer an unresolved blocker to not just be ignored. Maybe I'm misunderstood the comments though. At any rate, it would be good if rcombs could comment to either clarify if she considers this PR ok to merge or if not to discuss and reach a consensus with you. If you two agree on this, I have no issue.

---

> Also, the same static code is reused for SSE2/AVX2, while for dynamic it's a whole ton of [error-prone copy-paste](https://github.com/libass/libass/commit/07baea17d927f2d15ce2ff5dd6d12fe10e40d36c#diff-0c1187bf4a1bb9e2aced66eb83fd6c20235736fd7766209033b32fb3e0576601R35-R105).

If it’s covering mostly the same subset, we could still utilise macros to cut down on copy-paste (#400 doesn't try to).

> rcombs had concern about integrating partial NEON assembly into infrastructure, so I've tried and implemented it without problems.

My interpretation of rcombs comment is not that integration with static infrastructure would be impossible, but more complicated since eventhough not all functions are implemented, the init still has to deal with *all* functions and their variants.

As I understand, which approach is better depends on what expectations we have for our future assembly. If we'd expect to gain many new asm implementations and all of them to (be able to meaningfully) cover all functions, then yes your static approach is better. Otherwise though consider this hypothetical example:
Consider there were three ISA-extensions whose performance when all are available would be expected to be `A > B > C`, but presence of any one of them does not allow any conclusions about the presence of the other two. Now also consider they all cover distinct but overlapping strict subsets of the bitmap functions. *(Maybe they are so specialised to only make sense for a couple functions.)* Then with static initialisers we'd need one initialiser for each combination, i.e. 9 different initialisers — and they wouldn't as well macroïse as x86 currently does. With the dynamic approach we’d only need 3 `if` clauses each only adding those functions implemented by its ISA-extension.

I do not know what is more likely to happen and to my understanding rcombs and you disagree on this.

rcombs:
I think the changes on this branch have only made the unsuitability of the static-`BitmapEngine` architecture clearer. It now has even more layers of differentiation, including ones that are ultimately determined by configuration flags that are entirely unrelated to CPU feature support, requiring even more `BITMAP_ENGINE` invocations and code paths in the init function. It also introduces duplicate copies of several asm function that are completely identical to each other.

There's also code that apparently-deliberately produces different output depending on which CPU features are available (which is what checkasm is supposed to prevent in the first place), which is handled by implementing separate C versions that match the behavior of the assembly (what?).
If we really want to have quality-speed tradeoffs, we should be following the same model other open-source multimedia projects do: provide a separate vtable entry for each version, treat them as separate functions (each of which may or may not have assembly implementations making use of particular CPU features), and call the appropriate vtable entry at runtime. Don't swap to an entirely different copy of the vtable (and introduce substantial additional complexity in checkasm) for the sake of a configuration flag that only affects some of the functions.

Also, yes, if we ever have support for ISA extensions that don't strictly require all previous extensions to exist, the static approach would quickly produce a completely untenable combinatorial explosion. This would be the case if we supported a number of ARMv7 extensions, or any of several more recent x86 ones (especially those introduced by AMD).

I don't buy the fragility issue either; the issues on my branch were introduced entirely because I was reverting the entire vtable system at once and hadn't yet tested on all supported architectures. Proper checkasm testing on all platforms would have caught them well before merge. In normal conditions, you'd only regularly be making changes to code for the platform you're building and testing on (or to C), so any similar error would present itself immediately (as it does in _every other major open-source multimedia project_).

MrSmile:
**TheOneric**:

> If it’s covering mostly the same subset, we could still utilise macros to cut down on copy-paste

Yes, It's possible and I suspect that final result would be not that much different from my variant.

> My interpretation of rcombs comment is not that integration with static infrastructure would be impossible, but more complicated since eventhough not all functions are implemented, the init still has to deal with _all_ functions and their variants.

If we start from macroized version of dynamic approach, then maybe we can save a couple of lines. But certainly not in the case with a separate file for each architecture.

> Otherwise though consider this hypothetical example

Yes, combinatorial explosion is the case where dynamic approach is better, but I doubt that we hit that limit in practice. Currently it looks like premature optimization for me.

**rcombs**:

> There's also code that apparently-deliberately produces different output depending on which CPU features are available

It's not technically different output (in mathematical sense), it's different data layout for intermediate blur values, depending on vector register size.

> provide a separate vtable entry for each version, treat them as separate functions

Obviously that would be excess complexity in the case such functions are drop-in replacements in the same algorithm. And checkasm complications aren't serious given sufficient instrumentation.

> Also, yes, if we ever have support for ISA extensions that don't strictly require all previous extensions to exist, the static approach would quickly produce a completely untenable combinatorial explosion.

We don't even have SSSE3 versions, which is significant improvement over SSE2, so I doubt we hit that limitation any time soon. But point taken anyway. (Actually, I've tried to write assembly for a wide range of x86 architectures, so adding SSSE3 should be fairly trivial. Maybe now is a good moment for introducing SSSE3 flag?)

> In normal conditions, you'd only regularly be making changes to code for the platform you're building and testing on (or to C)

That maybe true for video codec development, where a set of assembly functions is basically fixed by the codec specification. But in libass, change of the rendering algorithm can lead to a change of required assembly. I've already hit that several times, so that's why I want checkasm to be a simple binary without any dependencies, friendly to the simplest forms of architecture emulation and debugging.

------------------
Okay, I understand that the main issue is potential combinatorial explosion. In that case I'll make a dynamic approach variant (with all necessary macro instrumentation and support of different alignments and tile sizes) to do the more relevant complexity comparison.

MrSmile:
I've implemented dynamic approach too and can say that there's no principal difference between code complexity: [static variant](https://github.com/MrSmile/libass/blob/checkasm/libass/ass_bitmap_engine.c), [dynamic variant](https://github.com/MrSmile/libass/blob/checkasm-dynamic/libass/ass_bitmap_engine.c). Now I understand why **rcombs** was so concerned about different data layout and forced alignment to 32 in her checkasm branch: it's an area where dynamic approach doesn't work. If there's some global property of all assembly functions (in our case it's data layout, i.&nbsp;e. alignment) you cannot simply reuse some assembly functions from another branch, you have to choose only compatible variants. For static variant you have to manually specify all functions anyway, so it's not a problem, but for dynamic you have to introduce some hacks to synchronize different branches. Dropping vector size dependency isn't a solution as it can lead to performance degradation especially in case of small bitmaps, so dynamic approach has a limited usefulness. I think that even in the dynamic case it's better to not try to share code between different vector register types and fill all the fields of `BitmapEngine` struct completely (sorta static approach). Maybe I should try to implement SSSE3 branch as it uses the same registers as SSE2, so dynamic approach should shine there without any hacks.

MrSmile:
Okay, now there's also [SSSE3 branch](https://github.com/MrSmile/libass/blob/checkasm-ssse3/libass/ass_bitmap_engine.c) (relevant functions are 10&ndash;20% faster than SSE2 on my processor). Dynamic approach, indeed, can save a dozen of lines in that case. So, @rcombs and @TheOneric, can you say if that variant of dynamic approach is good enough in your opinion?

rcombs:
This approach looks much better! The macros help simplify common groupings very nicely. I just have a couple issues left: 
- It'd be cleaner in checkasm (and also probably in general) to have `ASS_FLAG_LARGE_TILES` and `ASS_FLAG_WIDE_STRIPE` be separate from the CPU-flag mask. Instead, just have separate `BitmapEngine` members for the `tile16` and `tile32` or blur 16/32-alignment variants, and use the appropriate pointer at runtime. If a given extension doesn't provide one of the variants, we can just leave that member pointed at the C version. A given `BitmapEngine` member should always point to a function with the same observable behavior.
- We should fill in `engine` with the C variants first, then fill in per-arch functions as applicable. This ensures that we can add new functions without having to worry about the code paths for each architecture.

MrSmile:
> A given `BitmapEngine` member should always point to a function with the same observable behavior.

It's virtually impossible: even seemingly equivalent functions have different alignment requirements. Use of a common denominator for alignment can hurt performance, especially in the case of many small letter bitmaps. Also, separate `BitmapEngine` members would not be cleaner in checkasm: current [clean](https://github.com/MrSmile/libass/blob/checkasm-ssse3/checkasm/blur.c#L175) [loops](https://github.com/MrSmile/libass/blob/checkasm-ssse3/checkasm/rasterizer.c#L287) would be replaced by ugly copy-paste or unreadable macros. And certainly they wouldn't be cleaner in libass, as no code is always better.

> We should fill in `engine` with the C variants first, then fill in per-arch functions as applicable.

Only if all variants are binary compatible, which I think is possible only in simple cases. So long as there's incompatibility, I think it's safer to keep generic case at the end. My [previous variant](https://github.com/MrSmile/libass/blob/checkasm-dynamic/libass/ass_bitmap_engine.c) have it at the start and it's quite ugly.

rcombs:
The point of checkasm is to ensure that all versions of a function are binary-compatible. Output should _never_ vary depending on architecture. If we aren't able to provide identical behavior for a given arch, we need to provide a runtime flag to enable the degraded behavior, and provide that same functionality in at least C. If our existing code has arch-specific output differences (which apparently is the case), that's a bug that needs to be fixed.

MrSmile:
> The point of checkasm is to ensure that all versions of a function are binary-compatible.

It is not. Its point is to provide an easy way to validate assembly. We should provide C reference versions of all assembly functions and checkasm will compare their output. Where do the function pointers reside is outside of the scope of checkasm, it uses string identifiers to associate equivalent functions.

> Output should _never_ vary depending on architecture.

libass output _could_ vary depending on architecture due to performance reasons. I'm pretty sure that it'll be almost impossible to force any future GPU renderer output the same result. And of course, internal data layout can vary greatly even with the same output.

> If we aren't able to provide identical behavior for a given arch, we need to provide a runtime flag to enable the degraded behavior, and provide that same functionality in at least C.

Yes, I'm doing exactly that in all my checkasm branches. It's quite simple and doesn't have significant impact on code complexity. On the other hand, introducing all possible function variants into `BitmapEngine` would have much larger impact.

MrSmile:
Updated PR to accommodate the new unified Makefile approach. Changes:
- libass sources sorted to keep related files together;
- `c_` prefixes have been removed, as there's no warning about collisions;
- adopted dynamic vtable approach.

On the side note, do we really need all that `SRC_*` variables in `Makefile_library.am`? They separate source lists from including condition and make file longer without any benefits. Maybe it's better to eliminate them altogether?

libass/Makefile_library.am
```diff
@@ -34,11 +34,16 @@ libass_libass_la_SOURCES = \
...
+    libass/ass_rasterizer.h libass/ass_rasterizer.c \
     libass/ass_render.h libass/ass_render.c libass/ass_render_api.c \
     libass/ass_bitmap_engine.h libass/ass_bitmap_engine.c \
+    libass/c/rasterizer.c \
+    libass/c/blend_bitmaps.c \
+    libass/c/be_blur.c \
+    libass/c/blur.c \
     libass/wyhash.h
 
+libass_libass_la_CPPFLAGS = -I$(top_srcdir)/libass
```
TheOneric, line 46: Since now all programs/libraries have this set, how about we move this to toplevel `AM_CPPFLAGS` instead, dropping (these,) test, profile and compare’s `_CPPFLAGS` and change `fuzz_fuzz_CPPFLAGS` to `= $(AM_CPPFLAGS) $(FUZZ_CPPFLAGS)`?

libass/ass_bitmap_engine.h
```diff
@@ -31,13 +31,13 @@ typedef void FillGenericTileFunc(uint8_t *buf, ptrdiff_t stride,
...
-                           uint8_t *src1, intptr_t src1_stride,
-                           uint8_t *src2, intptr_t src2_stride,
-                           intptr_t width, intptr_t height);
+typedef void BitmapBlendFunc(uint8_t *dst, ptrdiff_t dst_stride,
+                             const uint8_t *src, ptrdiff_t src_stride,
+                             size_t width, size_t height);
+typedef void BitmapMulFunc(uint8_t *dst, ptrdiff_t dst_stride,
+                           const uint8_t *src1, ptrdiff_t src1_stride,
+                           const uint8_t *src2, ptrdiff_t src2_stride,
+                           size_t width, size_t height);
```
TheOneric, line 40: There was some discussion about the `intptr_t` to `size_t`/`ptrdiff` change. Could you include a short recap of the considerations that went into this change and why it is better in the commit message?

TheOneric (commented):
Mostly just looked at the changes this time and as before I’m still no assembly expert.

> On the side note, do we really need all that SRC_* variables in Makefile_library.am?

We could just conditionally append those directly to `libass_libass_la_SOURCES`, sure.

---

I don't recall where the discussion here and on IRC regarding the newly introduced C versions for checkasm left off. If it’s resolved now and there aren't any more concerns raised  in the next few days or so, great! (cc @MrSmile @rcombs) If it’s still up for debate, I believe at least everything up to and including `be_blur: reorganize code` (0c5617d6c01c8bda62c352fc2eea1e9fbc325b0b) should be uncontroversial and could be merged soon?
*(Personally, comparing assembly version of the same function to different reference implementations strikes me as a bit odd, but I also have barely a clue about x86 and the mentioned data layout problems, so this doesn't count for much)*

libass/Makefile_library.am
MrSmile: After that old warning came back. Looks like I should revert `c_` prefixes too.

MrSmile: Done.

libass/ass_bitmap_engine.h
MrSmile: Done.

MrSmile:
> I don't recall where the discussion here and on IRC regarding the newly introduced C versions for checkasm left off.

That discussion is currently in progress (and I think, this thread is the main place of that discussion).
There're no completely new C versions in this PR either, the only difference is that versions previously available through preprocessor now would be accessible at runtime. Even better: there's less actual code due to elimination of copy-paste in rasterizer. As for partial merge, this PR does not change vtable organization, so there's no particular commit that's a good stopping point. Similarly as with dynamic vtable, current master has static vtable and PR doesn't change that save for the last commit.

> Personally, comparing assembly version of the same function to different reference implementations strikes me as a bit odd

It's not the same function. For example, rasterizer can work with 16x16 or 32x32 tiles (or any other POT size if it matters). Function that fills 16x16 tile is clearly different from function that fills 32x32 (even if their code is generated by the same macro), so to check assembly we have to compare versions for the same tile size. **rcombs** thinks that if we have different functions we should introduce separate entries for them in `BitmapEngine`, I think that it's a detrimental complication. Basically, there's only one advantage of such approach:
- in `ass_bitmap_engine_init()` it's possible to fill C versions first and then proceed with assembly versions;

but there're several disadvantages:
- unnecessary bloat of `BitmapEngine` struct with completely unused fields;
- unnecessary logic to choose correct version inside rasterizer (and blur);
- current version can support 8x8 or 64x64 tiles seamlessly, version with separate functions would require a corresponding change of `BitmapEngine` struct.

TheOneric:
> It's not the same function. For example, rasterizer can work with 16x16 or 32x32 tiles (or any other POT size if it matters). Function that fills 16x16 tile is clearly different from function that fills 32x32 (even if their code is generated by the same macro), so to check assembly we have to compare versions for the same tile size. 

I fell like the following was discussed before, but can't find it right now. If you don't mind, could you (re-)explain?
With the current and imminent implementations, ignoring hypothetical future GPU-implementations and such as well as the known x87-SSE float or possible system math routine precision differences *(which aren't resolved either way)*: why can't we just run the smaller sized version multiple times in checkasm to compare to versions of larger tile sizes?
The listed sizes are all squares with edges being a power of two, so they'll fit nicely and iiuc the results of filling an area via any of the versions should always result in bit-identical final output. This would not only avoid unused entries in `BitmapEngine`, but also unused functions being present in the library at all.
For benchmarks there’d be a slight overhead from looping over the actual implementation, but I suspect that this should be small and when the public library routines are used there’d would be more loop iterations too *(though the fixed base cost is likely spread over more iterations than it would be in checkasm)*.

MrSmile:
> why can't we just run the smaller sized version multiple times in checkasm to compare to versions of larger tile sizes?

For blur it's possible and I've done exactly that before all that checkasm business. But without exact match it's further in the range of unit—integration tests: there's more additional code for format conversions, some small details can be hidden by that indirection, in the case of error it's much harder to debug (I cannot simply move step-by-step and compare intermediate results of equivalent functions). Having done both variants I can say that approach from this PR is much better. And for rasterizer it's simply impossible, as with larger size there's a loss of precision (it's essentially the reason why I've kept configure switch instead of sticking to one optimal size).

> This would not only avoid unused entries in `BitmapEngine`, but also unused functions being present in the library at all.

As I understand, the whole point of **rcombs** disapproval is the lack of those unused entries. And if unnecessary functions in the library is a problem then my approach where checkasm uses sources directly instead of the whole library can solve it.

TheOneric:
> And if unnecessary functions in the library is a problem

To clarify, I don't think this is necessarily a problem with the current amount at least. I only mentioned it because it seemed like a logical step further from the stated benefit of “no unused `BitmapEngine` entries”.

> But without exact match it's further in the range of unit—integration tests: there's more additional code [...]

To ensure my understanding is correct: this doesn't affect current or imminent assembly implementations, right?

> And for rasterizer it's simply impossible, as with larger size there's a loss of precision (it's essentially the reason why I've kept configure switch instead of sticking to one optimal size).

As above: is this about a mismatch between an `--enable-large-tiles` and `--disable-large-tiles` build, or a mismatch *within* a build with assembly?

MrSmile:
> To ensure my understanding is correct: this doesn't affect current or imminent assembly implementations, right?

Additional code? It would go into testing infrastructure and not into libass directly. Also it make harder to write new assembly due to a more complicated debug.

> As above: is this about a mismatch between an `--enable-large-tiles` and `--disable-large-tiles` build, or a mismatch _within_ a build with assembly?

It's mismatch between enable & disable, not with assembly for the same tile size (there were differences previously, but already fixed). So it's impossible to check function for large tiles with 4 calls of function for small tiles.

TheOneric:
Thanks. For the parts I'm qualified to comment on, there are no more blocking issues as far as I am concerned.

rcombs:
I'm fine with merging everything through 9aab3057518cdb7fd5c2b67688ea9f212cb86108, at least.

I continue to take issue with code that produces different output at runtime depending on CPU architecture, and with test infrastructure that reinforces that. Again, if we want to provide an option to produce potentially-degraded output, it should be behind a runtime flag separate from the CPU-capabilities bitmask.

MrSmile:
> I continue to take issue with code that produces different output at runtime depending on CPU architecture, and with test infrastructure that reinforces that.

It is out of scope of this PR or even checkasm. If I remember right, my rasterizer PR was committed by you, and variants with different tile sizes were in plain sight. Why it was OK then and a problem now? It's quite nontrivial to implement 17-tap vertical blur efficiently due to its memory access pattern, so I've tailored intermediate data layout for it, hence dependency on vector register size. Again, blur PR was open for anyone interested but no complains arise. Performance was good too.

If the rendering algorithms are acceptable, then where's a problem? Is it exposing additional C variants for the purpose of testing? Or the layout of `BitmapEngine` struct? I want to see how do you suggest to test the full set of current assembly functions.

Even in the purity standpoint, virtual table is about applying conceptually the same action on objects of _different_ types. So different input/output of the equivalent bitmap functions is perfectly fine. Moreover, the same algorithms work even on GPU, so in the future `Bitmap` can become an opaque handle, referencing some region of GPU texture, and bitmap functions will operate on it.

> Again, if we want to provide an option to produce potentially-degraded output, it should be behind a runtime flag separate from the CPU-capabilities bitmask.

It's possible to introduce another flag parameter into `ass_bitmap_engine_init()` function, but I think it's unnecessary code bloat. Also availability of such additional options can depend on CPU capabilities (for example, `LARGE_TILES` do not make sense without assembly). So it would make sense for `ass_get_cpu_flags()` to produce a list of available rendering options with conjunction with pure CPU flags. It's currently not doing that and unsupported options are ignored in `ass_bitmap_engine_init()`, but maybe I should change that (and also rename `ass_get_cpu_flags()` for clarity).

rcombs:
Just to clarify, I don't have any objection to the variations in intermediate layout with identical output; it's a little funky, but it can be handled by test routines (i.e. testing at the level above where those intermediates exist). I also don't have any objection to modes with different render output, only to the output silently varying at runtime depending on the processor we're running on.

MrSmile:
But currently there's no such dependency of the output (see automatic tests) and this PR doesn't change that.

> testing at the level above where those intermediates exist

It's a bad idea: I used to test blur in that way and can say that approach from this PR with dedicated C functions for every possible layout is much better (at the cost of small runtime overhead).

MrSmile:
Moved NEON assembly (and while at it, SSSE3 too) from my checkasm branch into this PR per **rcombs** request. I think, the only missing part is [an adequate configurations script](https://github.com/libass/libass/pull/559/commits/ccafccc5ff638f9903b542178de792ae95b8466b#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R323). Maybe we should create an one-line assembly test similarly to x86?

libass/aarch64/be_blur.S
```diff
@@ -0,0 +1,150 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include "asm.S"
```
TheOneric, line 19: It appears you forgot to actual include `libass/aarch64/asm.S` in a commit; also see the `distcheck` failure in CI.

configure.ac
```diff
@@ -360,6 +372,12 @@ AM_COND_IF([ASM], [
     AM_COND_IF([X86], [
         AC_DEFINE(ARCH_X86, 1, [targeting a 32- or 64-bit x86 host architecture])
     ])
+    AM_COND_IF([X86_64], [
+        AC_DEFINE(ARCH_X86_64, 1, [targeting a 64-bit x86 host architecture])
+    ])
```
TheOneric, line 377: This C macro appears to be wholly unused, so adding it only bloats `configure.ac`.

libass/ass_bitmap_engine.h
```diff
@@ -84,6 +84,8 @@ enum {
     ASS_CPU_FLAG_X86_SSE2      = 0x0001,
     ASS_CPU_FLAG_X86_SSSE3     = 0x0002,
     ASS_CPU_FLAG_X86_AVX2      = 0x0004,
+#elif ARCH_AARCH64
+    ASS_CPU_FLAG_ARM_NEON      = 0x0001,
 #endif
```
TheOneric, line 89: This means everytime the `ASS_CPU_*` flags are used they need to be guarded with architecture macros. It might be better to just always define the enums, so not everything needs to be littered with architecture macro checks?

configure.ac
```diff
@@ -233,6 +234,9 @@ AS_IF([test "x$enable_require_system_font_provider" != xno  dnl
 
 
 # Locate and configure Assembler appropriately
+X86=false
+X86_64=false
+AARCH64=false
```
TheOneric, line 239: Rather than tracking each CPU Arch in its own boolean variable, imho it would be better to replace them with a unified `cpu_family` variable and replace the `    AS_IF([test "x$X86" = xtrue], [...])     AS_IF([test "x$AARCH64" = xtrue], [...])` with an `AS_CASE([$cpu_family], ...)`.

TheOneric (commented):
Here’s an alternative concept for `configure.ac` as suggested below. It might still need some refinement regarding e.g. the `cpu_family` names: [559_aarch-configure.tar.gz](https://github.com/libass/libass/files/10050313/559_aarch-configure.tar.gz).
Also, should we forward `ASFLAGS` (nasm) to `CCASFLAGS` and /or vice versa, or keep them separate?

libass/aarch64/be_blur.S
MrSmile: It has added by checkasm commit, so it got lost after shuffling. Fixed.

configure.ac
MrSmile: It would be needed by `checkasm.c`. Moved to relevant checkasm commit.

libass/ass_bitmap_engine.h
MrSmile: They have always used in already guarded architecture-specific context, so it's not a problem. I think it's better to not introduce unrelated flags, as it can help to detect configuration errors.

configure.ac
MrSmile: Adopted your (slightly modified) patchset.

MrSmile:
Used variant of `configure.ac` suggested by **TheOneric**, but assembly problem still stands: is it OK to enable assembly unconditionally for aarch64, or there should be some check? I think we should at least detect a big-endian ARM and disable assembly for it.

rcombs:
Is any of the ASM actually endianness-dependent? It doesn't look to me like it should be, though it's possible I'm missing something. I checked ffmpeg, and it doesn't seem to vary (or disable) any aarch64 assembly based on endianness.

MrSmile:
I've used `ldr` instead of `ld1` which is endiannes-dependent for horizontal shifts like that of blur. Compilers use `ldr`, so I think there can be performance penalty in the use of `ld1`. Also there's some hack in rasterizer, which should be written differently for the big-endian case. All-in-all, there's no problem in supporting big-endian as is, it's adequate configure check and testing environment that's missing.

rcombs:
The tests I've seen from [Firestorm](https://dougallj.github.io/applecpu/firestorm-simd.html) and [Icestorm](https://dougallj.github.io/applecpu/icestorm-simd.html), which are major targets for this code, indicate no difference at all in performance between LDR and LD1.

That said, a configure-time endianness check does sound worthwhile (it'd simplify the libpng stuff as well); we can use `AC_C_BIGENDIAN` for that.

MrSmile:
> no difference at all in performance between LDR and LD1.

In that case I probably should replace `ldr` with `ld1`, it would be simpler for macro arguments too. Hope it's the same for common ARM Cortex-* too. I can live without check and eliminate rasterizer hack, but without a way to verify correctness it probably would be broken for big-endian anyway. Does anybody know a simple way to setup a big-endian ARM environment?

TheOneric:
> Does anybody know a simple way to setup a big-endian ARM environment?

There are aarch64be binary pkgsrc packages for NetBSD [here](https://cdn.netbsd.org/pub/pkgsrc/packages/NetBSD/aarch64eb/) and supposedly there are also some [aarch64eb install image](https://wiki.netbsd.org/ports/evbarm/) though I can’t immediately spot a link to it *(but to multipe big-endian 32-bit ARM variants)*. If you want to pursue that route, you could ask on libera.chat#NetBSD or the user mailing list.

Alternatively you could cross-compile musl, expat, fribidi, FreeType, HarfBuzz, libpng once for aarch64be Linux and use user-mode `qemu-aarch64_be-static` to test a freshly cross-compiled libass. On a aarch64el host this is presumably performant. *(NetBSD doesn’t have user-mode qemu afaict)*

> All-in-all, there's no problem in supporting big-endian as is, it's adequate configure check and testing environment that's missing.

Since we use GNU Assembler syntax, won’t GAS define the the same `__BYTE_ORDER__` preprocessor macros as gcc for detecting endianess? If so, I’m guessing `llvm-as` will as well since it wants to be an (almost) drop-in replacement. Are there any other GAS-compatible assemblers and does any of these not define those macros?

libass/aarch64/asm.S
```diff
@@ -0,0 +1,140 @@
...
+#if HAVE_AS_FUNC
+        .func \name
+#endif
+    .endif
+\name:
+#if ARCH_AARCH64
+    .if \export
+         AARCH64_VALID_CALL_TARGET
+    .endif
+#endif
```
TheOneric, line 85: Is this file also supposed to be used with theoretical future 32-bit ARM assembly?

MrSmile: **rcombs**' version of this file doesn't work well with checkasm for my aarch64 setup, so I've copied [the relevant bunch of code from dav1d](https://code.videolan.org/videolan/dav1d/-/blob/master/src/arm/asm.S#L215) as-is. Actually it's not a production solution--I probably should use original dav1d file complete with their copyright header and do slight modifications on it. Alternatively it could be written from scratch, but that can lead to substantial consequences for checkasm.

MrSmile:
> Alternatively you could cross-compile musl, expat, fribidi, FreeType, HarfBuzz, libpng once for aarch64be Linux and use user-mode `qemu-aarch64_be-static` to test a freshly cross-compiled libass.

Most dependencies wouldn't be needed as I have a version of checkasm without full libass dependency. But while I indeed have `qemu-aarch64_be-static`, there's no easy way to install relevant `aarch64_be` cross compiler.

> Since we use GNU Assembler syntax, won’t GAS define the the same `__BYTE_ORDER__` preprocessor macros as gcc for detecting endianess?

Oh, good idea! And it indeed works! Therefore problem with endian-dependent hacks is solved.

TheOneric:
Found out how to use NetBSD for aarch64eb. There isn’t a traditional installer ISO, but an image file ready to flash on the storage of a physical device or to use with qemu: http://nycdn.netbsd.org/pub/NetBSD-daily/HEAD/202211211350Z/evbarm-aarch64eb/binary/gzimg/

Instructions how to boot it with `qemu-syste-aarch64` can be found [here](https://wiki.netbsd.org/ports/evbarm/qemu_arm/) and the mentioned `QEMU_EFI.fd` [here](https://releases.linaro.org/components/kernel/uefi-linaro/latest/release/qemu64/QEMU_EFI.fd), but may need some adjusting to give the VM internet access. IN my case, using a bridge created via `virsh`, I ended up `allow virbr0`ing the interface in `/etc/qemu/bridge.conf` and using `-netdev bridge,id=hn0,br=virbr0 -device virtio-net-pci,netdev=hn0,id=nic1`.

Once booted up with internet access, login as `root` (no password by default) and run:
```
arm64# export PKG_PATH=http://ftp.netbsd.org/pub/pkgsrc/packages/NetBSD/aarch64eb/current_2022Q1/All
arm64# pkg_add -v pkgin
arm64# echo ftp://ftp.fr.netbsd.org/pub/pkgsrc/packages/NetBSD/aarch64eb/current_2022Q1/All > /usr/pkg/etc/pkgin/repositories.conf
arm64# pkgin update
arm64# pkgin install mozilla-rootcerts openssl git autoconf automake libtool-base pkg-config freetype fribidi harfbuzz fontconfig png
arm64# mozilla-rootcerts install
```

And you’re ready to clone and build libass and use libass-tests.
Current HEAD a6c8d1a5dbe7c3479a8b48b6018004b947e0bf7c fails all regression tests and since the host triplet is: `aarch64_be-unknown-netbsd9.99`, I needed to adjust the check in `configure.ac`.

Speaking of alternative aarch64 ABIs, there’s aarch64 ILP32 similar to x32, which is supposedly used in embedded devices and some Apple watches. I’m not sure if the aarch64-ilp32 ever got upstreamed to Linux though and have no idea how to test this. It’s probably fine to ignore it unless someone using and able to test it shows up.

libass/aarch64/asm.S
MrSmile: Removed 32-bit support for now.

MrSmile:
Changes:
* added support for big-endian 64-bit ARM;
* added aarch64_be and darwin flag in configure;
* revamped `asm.S` with original dav1d copyright header.

libass/aarch64/asm.S
```diff
@@ -0,0 +1,280 @@
...
+        sub             \rd, \rd, -(\offset)
+  .else
+        adrp            \rd, \val+(\offset)
+        add             \rd, \rd, :lo12:\val+(\offset)
+  .endif
+#elif defined(PIC)
+        adrp            \rd, \val+(\offset)
+        add             \rd, \rd, :lo12:\val+(\offset)
+#else
+        ldr             \rd, =\val+\offset
```
TheOneric, line 275: Is this `ldr` also correct on BE?

MrSmile: Should be, as there's no anything endian-specific in [docs](https://sourceware.org/binutils/docs/as/ARM-Opcodes.html). `ldr` (or `ldp`, `str`, `stp`) can lead to problems only in case of loading several values into a single register. Usually it's `ldr qN` for vector register load, but I also have pairs of 32-bit or 16-bit values.

configure.ac
```diff
@@ -258,6 +259,11 @@ AS_IF([test "x$enable_asm" != xno], [
             BITTYPE=64
             ASFLAGS="$ASFLAGS -DARCH_X86_64=1"
         ],
+        [aarch64-*|aarch64_be-*|arm64-*|armv8-*], [
```
TheOneric, line 262: What platforms use a `arm64-` or `armv8-*` host triplet and how do they denote big endian?

MrSmile: Removed for now.

PR #560, opened 2021-09-26
Cosmetic: refactor track features to bitflags

TheOneric:
The idea was proposed and first implemented in #481 and is supposed to scale better when adding future boolean-features. If we want to add non-boolean features at some point this might need to be changed again.

The implementation differs from #481's and also does not include a `ass_track_get_feature_state` API, as there were open question regarding handling of meta-features and about future non-boolean features. As long as we only have boolean features and all are initialised to false, I don't think `ass_track_get_feature_state` is strictly necessary yet.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -1535,17 +1535,33 @@ fail:
 
 int ass_track_set_feature(ASS_Track *track, ASS_Feature feature, int enable)
 {
+    if (feature >= sizeof(track->parser_priv->feature_flags) * CHAR_BIT || feature < 0)
+        return -1;
+
+    // all supported non-meta features
+    static const uint32_t supported =
+#ifdef USE_FRIBIDI_EX_API
+        FEATURE_MASK(ASS_FEATURE_BIDI_BRACKETS) |
+#endif
+        0;
+    uint32_t requested = 0;
+
     switch (feature) {
     case ASS_FEATURE_INCOMPATIBLE_EXTENSIONS:
-        //-fallthrough
-#ifdef USE_FRIBIDI_EX_API
-    case ASS_FEATURE_BIDI_BRACKETS:
-        track->parser_priv->bidi_brackets = !!enable;
-#endif
-        return 0;
+        requested = supported;
+        break;
     default:
-        return -1;
+        if (!(FEATURE_MASK(feature) & supported))
+            return -1;
+        requested = FEATURE_MASK(feature);
     }
+
+    if (enable)
+        track->parser_priv->feature_flags |= requested;
+    else
+        track->parser_priv->feature_flags &= ~requested;
+
+    return 0;
 }
 
 /**
--- a/libass/ass_utils.h
+++ b/libass/ass_utils.h
@@ -48,6 +48,8 @@
 
 #define ASS_PI 3.14159265358979323846
 
+#define FEATURE_MASK(feat) (((uint32_t) 1) << (feat))
+
 #if (defined(__i386__) || defined(__x86_64__)) && CONFIG_ASM
 int has_sse2(void);
 int has_avx(void);
--- a/libass/ass_priv.h
+++ b/libass/ass_priv.h
@@ -61,9 +61,7 @@ struct parser_priv {
     // tracks [Script Info] headers set by the script
     uint32_t header_flags;
 
-#ifdef USE_FRIBIDI_EX_API
-    bool bidi_brackets;
-#endif
+    uint32_t feature_flags;
 };
 
 #endif /* LIBASS_PRIV_H */
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2883,7 +2883,7 @@ ass_start_frame(ASS_Renderer *render_priv, ASS_Track *track,
     ass_shaper_set_level(render_priv->shaper, render_priv->settings.shaper);
 #ifdef USE_FRIBIDI_EX_API
     ass_shaper_set_bidi_brackets(render_priv->shaper,
-            track->parser_priv->bidi_brackets);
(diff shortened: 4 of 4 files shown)


libass/ass.c
```diff
@@ -1535,17 +1535,33 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
+
+    // all supported non-meta features
+    static const uint32_t supported = 0
 #ifdef USE_FRIBIDI_EX_API
-    case ASS_FEATURE_BIDI_BRACKETS:
-        track->parser_priv->bidi_brackets = !!enable;
+        | FEATURE_MASK(ASS_FEATURE_BIDI_BRACKETS)
 #endif
-        return 0;
+        ;
```
MrSmile, line 1546: This lone `;` is hurting my eyes. Maybe simple:
```c
    uint32_t supported = 0;
#ifdef USE_FRIBIDI_EX_API
    supported |= FEATURE_MASK(ASS_FEATURE_BIDI_BRACKETS);
#endif
```
would be better?

libass/ass.c
```diff
@@ -1535,17 +1535,33 @@ ASS_Track *ass_new_track(ASS_Library *library)
...
-        return -1;
+        if (!(FEATURE_MASK(feature) & supported))
+            return -1;
+        requested = FEATURE_MASK(feature);
     }
+
+    if (enable)
+        track->parser_priv->feature_flags |= requested;
+    else
+        track->parser_priv->feature_flags &= ~(requested);
```
MrSmile, line 1562: Is there any meaning in additional brackets?

MrSmile (approved):

libass/ass.c
TheOneric: No, probably a leftover from the macro-version; removed them now.

libass/ass.c
TheOneric: That would be more repetitive than the current version, so I think the current approach is better for scaling with future additions. How about adding either a trailing `| 0` or enclosing the whole expression in cosmetic parantheses `()` to avoid a freestanding `;`? WIth the parantheses it would end in `);` similar to how structs end in `};`.

MrSmile: I think that dynamic approach is better for scaling instead (it can even support runtime feature detection).
Otherwise I think that `0;` is better than `);`.

TheOneric: Now using `0;`.

The availability of currently planned features are all determined at build-time, if we get runtime-dependent features in the futures we can still add them with `if (…) supported |= …`.
If we we're always using `supported |=` it would be a lot more repetitive though — or a bit more repetitive and less consistent imho; compare:
```C
// supported |= for build-time-determined features
    uint32_t supported = FEATURE_MASK(ASS_FEATURE_UNCOND1)
        | FEATURE_MASK(ASS_FEATURE_UNCOND2);
#ifdef USE_FRIBIDI_EX_API
    supported |= FEATURE_MASK(ASS_FEATURE_BIDI_BRACKETS);
#endif
    if (runtime_check()) {
        supported |= FEATURE_MASK(ASS_FEATURE_RUNTIME1);
    }


// Single expression for build-time-determined and unconditional features
    uint32_t supported =
#ifdef USE_FRIBIDI_EX_API
        FEATURE_MASK(ASS_FEATURE_BIDI_BRACKETS) |
#endif
        FEATURE_MASK(ASS_FEATURE_UNCOND1) |
        FEATURE_MASK(ASS_FEATURE_UNCOND2);
    if (runtime_check()) {
        supported |= FEATURE_MASK(ASS_FEATURE_RUNTIME1);
    }
```

libass/ass_utils.h
```diff
@@ -48,6 +48,8 @@
 
 #define ASS_PI 3.14159265358979323846
 
+#define FEATURE_MASK(feat) (1ULL << (feat))
```
astiob, line 51: Seems a bit odd that all the variables are `uint32_t` but this is an (unsigned) `long long`.

TheOneric: The idea was, that we won't need to change the macro in the future; unsigned values gets safely truncated to the range of the variable. `uint32_t` matches the type of the existing header-bitflags.
Would you prefer the macro to use `((uint32_t)1 << (feat))` or the variables to use `unisgned long long`?

astiob: That was indeed my guess as to the intention, and I can sympathize with it. On the other hand, this may be a little confusing, and it seems a little useless to micro-optimize this one line given that three other lines name the type and therefore would need to be changed anyway.

Ironically, I think I wouldn’t have even nit-picked this if the bit-shift was written inline, but when it’s factored out to a named macro, it somehow feels like it should closely match to the feature-mask variables.

I don’t have too strong of an opinion here. Really, I’d be fine with something even more basic such as `unsigned int`. But if you ask me, `(uint32_t)1` seems like a better option simply because it wastes less memory/registers.

At any rate, feel free to commit/merge any version of this if you want to.

TheOneric: `unsigned int` *(at least 16 bits)* may be smaller than `uint32_t`, so `1u << (feat)` doesn't seem like a good idea to me *(`unsigned long long` is at least 64 bits per ISO-C99)*.

astiob: For one thing, we aren’t even using 16 bits now.

For another, this is your scheduled reminder that libass assumes `int` fits 32 bits. Or do we think all instances have been patched out by MrSmile’s various type-changing commits?

TheOneric: > libass assumes int fits 32 bits

But why add new code relying on this?

Anyway, I changed the macro to use `((uitn32_t)1)`. ANother approach would have been to use `unisgned long long` everywhere but the flag variable, so we'd only need to change one place, perhaps when/if we ever need to bump the limit.

PR #561, opened 2021-09-29
Bunch of miscellaneous commits from checkasm branch

MrSmile:
Another part of checkasm branch. Unrelated to checkasm but useful.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -275,7 +275,6 @@ AS_IF([test "x$enable_asm" != xno], [
                     ))
                 ]
             )
-            ASFLAGS="$ASFLAGS -Dprivate_prefix=ass"
             AC_MSG_CHECKING([if $AS supports vpmovzxwd])
             echo "vpmovzxwd ymm0, xmm0" > conftest.asm
             AS_IF([$AS conftest.asm $ASFLAGS -o conftest.o >conftest.log 2>&1], [
--- a/.gitignore
+++ b/.gitignore
@@ -18,9 +18,9 @@ Makefile.in
 /stamp-h1
 
 # Test programs
-/test/test*
-/compare/compare*
-/profile/profile*
+/test/test
+/compare/compare
+/profile/profile
 
 # pkgconfig
 /libass.pc
--- a/libass/Makefile.am
+++ b/libass/Makefile.am
@@ -12,7 +12,7 @@ nasm_verbose_ = $(nasm_verbose_$(AM_DEFAULT_VERBOSITY))
 nasm_verbose_0 = @echo "  NASM    " $@;
 
 .asm.lo:
-	$(nasm_verbose)$(LIBTOOL) $(AM_V_lt) --tag=CC --mode=compile $(top_srcdir)/ltnasm.sh $(AS) $(ASFLAGS) -I$(srcdir)/ -o $@ $<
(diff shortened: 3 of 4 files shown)


configure.ac
```diff
@@ -289,6 +289,9 @@ AS_IF([test "x$enable_asm" != xno], [
             ])
             rm conftest.asm conftest.o > /dev/null 2>&1
         ])
+    ], [
+        AC_MSG_NOTICE(Assembly optimizations are not yet supported on this architecture; disabling.)
+        enable_asm=no
```
TheOneric, line 294: This was effectively already done in f975de5d96aafd804c66ea79f0780ce2b30e1928 and can be dropped from this patch series; look ~50 lines up at the initial architecture check.

MrSmile: Ah, I have a feeling that I've already seen that somewhere, but rebase went off without a hitch so I thought that it's from some unmerged PR.

TheOneric (approved):
Nit: Imho it would be good to add “preparation for checkasm” or something to the commit-message of the first commit as it's otherwise not obvious moving it is a good idea (and the follow up commits don't give context either).
Otherwise LGTM

MrSmile:
No, it's not needed for checkasm. It's simply better to not pollute dynamic flags with static define, without which libass simply cannot compile.

TheOneric:
`checkasm.asm` uses [its own prefix](https://github.com/libass/libass/pull/400/commits/b0e795e718f3052ce4df4fe37d12d3e4f5707c5c#diff-6393d0147105cdd7fe9765e62d58a588551b08f0f48061603b2b9dbf80a30efaR26). I'm not sure how NASM handles multiple defines; perhaps it just silently overrides previous versions, but that at least gives a motivation for why the `private_prefix` definition is moved. Ignoring checkasm I see no reason why putting `private_prefix=ass` into `libass/Makefile.am` is supposed to be better, than defining it together with all other assembler flags.
But that's just bikeshedding over a small mit, if you disagree fell free to leave it as is.

MrSmile:
> perhaps it just silently overrides previous versions

Looks like it: I've removed this commit from the checkasm branch and didn't see any difference. Maybe we should use checkasm way to define this prefix too? I'll look into other projects with `x86inc.asm` then.

TheOneric:
To clarify: With checkasm with it's different `private_prefix` in mind, I do think it is better style, to move `-Dprivate_prefix=ass` out of the global `ASFLAGS` even if perhaps not strictly required for it to compile; only in isolation it's not clear why it's moved. But that's still just nit-picking and not something crucial.

MrSmile:
About `private_prefix`: x264 defines it in the command line like after this PR, dav1d uses a meson tool to add define into the assembly compile line (dunno which automake analog it corresponds better), ffmpeg uses `%define` in an assembly file. It's all other the place and I don't have a strict personal preference here, so maybe I should drop that commit too?

TheOneric:
> About private_prefix: […] It's all other the place and I don't have a strict personal preference here, so maybe I should drop that commit too?

Well, I can see the point in moving the `private_prefix` definition, even if not strictly necessary, to not create the wrong impression checkasm is also using this prefix; but I don't have any strong preference on this.

MrSmile:
I think at least the first two commits can be merged. Also I can add a fix for AVX2 check here, as another checkasm-inspired commit.

TheOneric:
The comment regarding the commit-message of the prefix-move was just a nit, not a blocker, so we can merge all and I also added the AVX-fix.

Issue #478, opened 2021-02-24
Script resolution questions


xwlcn:
subtitle content:
```
[Script Info]
; Script generated by Aegisub 3.2.2
; http://www.aegisub.org/
Title: sub
Original Script: FIXSubtitleGroup
Synch Point: 1
ScriptType: v4.00+
WrapStyle: 2
Timer: 100.0000
PlayResX: 384
PlayResY: 288
ScaledBorderAndShadow: no

[Aegisub Project Garbage]
Last Style Storage: Default
Video File: ?dummy:23.976000:36000000:1920:1080:25:101:87:
Video AR Value: 1.777778
Video Zoom Percent: 0.718750
Active Line: 9
Video Position: 1078

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Default,华文黑体,15,&H00FFFFFF,&H00000000,&H00FFFFFF,&H80000000,-1,0,0,0,100,100,0,0,1,0.1,0.4,8,10,10,20,1
Style: 特效,微软雅黑,20,&H00FFFFFF,&HFF0000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,0,0,2,10,10,10,1
Style: 注释,微软雅黑,18,&H00DEDEDE,&HFF000000,&H00000000,&H00000000,0,0,0,0,99.9998,100,0,0,1,1.5,0,8,0,0,5,1
Style: 歌词,微软雅黑,13,&H00DEDEDE,&H00000000,&H00000000,&H00000000,0,0,0,0,99.9998,100,0,0,1,1,0,7,15,5,20,1
Style: 特效 - 背景,微软雅黑,20,&H00FFFFFF,&HFF0000FF,&H00000000,&H00000000,0,0,0,0,99.9998,100,0,0,3,1,0,2,10,10,10,1

[Events]
Format: Layer, Start, End, Style, Name, MarginL, MarginR, MarginV, Effect, Text
Dialogue: 0,0:00:44.94,0:00:48.42,特效,,0,0,0,,{\pos(12,26.5)\a4\p1\fscx30\fscy40\c&H392810&}m 0 0 l 250 0 l 250 344 l 0 344 l 0 0
[...]

PR #563, opened 2021-10-03
fix .ass scaled with PlayResX

dyphire:
fix .ass scaled with PlayResX
It would fix [#171](https://github.com/libass/libass/issues/171), [#478](https://github.com/libass/libass/issues/478) and mpv [#9247](https://github.com/mpv-player/mpv/issues/9247)

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2008,6 +2008,11 @@ static bool parse_events(ASS_Renderer *render_priv, ASS_Event *event)
         info->border_y = render_priv->state.border_y;
         info->hspacing = render_priv->state.hspacing;
         info->bold = render_priv->state.bold;
+        double dar = ((double) render_priv->orig_width) / render_priv->orig_height;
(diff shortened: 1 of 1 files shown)


astiob:
At the very least:
* This breaks text except with `\fscx100\fscy100` (which isn’t necessary for text).
* This doesn’t fix vector drawings with `\fscx100\fscy100` (which is allowed for vector drawings).
* This breaks vector drawings with `\fax`.
* This doesn’t fix vector drawings with `\fay`.
* This doesn’t fix `\xshad`.
* This doesn’t fix `\xbord`.
* This allows division by zero that could be avoided, although we do currently also divide by the same things in other places.

<i>This is also a verbatim copy of https://github.com/libass/libass/issues/478#issuecomment-785646103 without credit.</i>

For an incomplete (?) proper fix, see e. g. https://github.com/astiob/libass/commit/a550fc5115bd36cb9d0510976d7357b609410645.

dyphire:
I see
I hope it can be solved soon

PR #568, opened 2021-10-26
Add CIFuzz

AdamKorcz:
Adds CIFuzz to Libass's OSS-fuzz integration. This will run Libass's fuzzer when pull requests are made. 
Currently the time is set to 600 seconds, but this can be changed.

The documentation for CIFuzz is available here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,30 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer: [address, undefined]
+    steps:
+    - name: Build Fuzzers (${{ matrix.sanitizer }})
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libass'
+        language: c++
+        sanitizer: ${{ matrix.sanitizer }}
+    - name: Run Fuzzers (${{ matrix.sanitizer }})
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'libass'
+        language: c++
+        fuzz-seconds: 600
+        sanitizer: ${{ matrix.sanitizer }}
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: ${{ matrix.sanitizer }}-artifacts
+        path: ./out/artifacts


TheOneric:
Merging this should probably be postponed until the current (likely bogus) memleak reports have been fixed: https://github.com/google/oss-fuzz/issues/6440#issuecomment-927143116 . @astiob: It seems like the source and fix for those reports hasn't been identified yet, or did it just move to a new issue?
I'm assuming (once the new bogus samples become public) this would cause all CIFuzz runs to just fail.

Apart from that, the current CI run of the PR failed with a timeout(?) on a sample. Reading the workflow file it seemed to me like the sample was supposed to be uploaded, yet the artifact-archive is empty. Can you explain when/which samples get uploaded as build artificats?
Does this only run against the known corpus to test for regressions or does it also create new samples by mutation? If the latter and the samples get uploaded freely accesible to all GitHub-users, couldn't this find bugs unrelated to the PR and in the worst case publicly expose a pre-existing security-bug before we had any chance to fix it? (Of course, in that case it could also have been found by others already by fuzzing...)

astiob:
> It seems like the source and fix for those reports hasn't been identified yet, or did it just move to a new issue?

It hasn’t; correct. But:

> I'm assuming (once the new bogus samples become public) this would cause all CIFuzz runs to just fail.

If I understand the docs right, CIFuzz should only report on _new_ failures introduced by each individual PR:

> If CIFuzz finds a crash, it reports the crash only if both of following are true:
> * The crash is reproducible (on the PR/commit build).
> * The crash does not occur on older OSS-Fuzz builds. (If the crash does occur on older builds, then it was not introduced by the PR/commit being tested.)

---

> Apart from that, the current CI run of the PR failed with a timeout(?) on a sample. Reading the workflow file it seemed to me like the sample was supposed to be uploaded, yet the artifact-archive is empty.

Yeah, that’s weird. The fuzzer said it timed out, the reproducer ran and said it timed out, then the upload stage ran… and successfully uploaded “0 files”.

---

> Does this only run against the known corpus to test for regressions or does it also create new samples by mutation?

I’m only guessing, but I assume it does run mutation, and that’s precisely why the job has a time allotment:

> If CIFuzz doesn’t find a crash during the allotted time, the CI test passes (green check).

But as the starting point, it uses the standard corpora and recent public OSS-Fuzz testcases:

> CIFuzz uses 30 day old/public regressions and corpora from OSS-Fuzz. This makes fuzzing more effective and gives you regression testing for free.

TheOneric:
Reading the last paragraph of the CIFuzz-docs (again), it does seem to me like the failure from the previous CI-Run should have been uploaded as a GitHubActions-Artificat, yet it didn't. If that wasn't a bug, this lack of understanding how CIFuzz works is imho the main blocker for a merge. @AdamKorcz: Could you explain what's up with this?

On another note: according to the docs, *“The default [sanitiser] is 'address'.”*; we probably also want to enable `undefined`.

AdamKorcz:
@TheOneric I will follow up later this week.

AdamKorcz:
Fixed the recently broken build: https://github.com/google/oss-fuzz/pull/6888

CIFuzz also had some recent fixes to its infrastructure, so I suggest we test again since it by the looks of it seems the testcase should have been uploaded. 
Can you confirm that the test case was not uploaded at oss-fuzz.com? I don't have access to the libass dashboard myself.

> Does this only run against the known corpus to test for regressions or does it also create new samples by mutation?

CIFuzz will start with the existing corpus and run mutations based on that also noted by @astiob 


> If the latter and the samples get uploaded freely accesible to all GitHub-users, couldn't this find bugs unrelated to the PR and in the worst case publicly expose a pre-existing security-bug before we had any chance to fix it? (Of course, in that case it could also have been found by others already by fuzzing...)

Theoretically, yes. However it is highly unlikely that the fuzzer will find a bug in 10 minutes that hundreds/thousands of hours of continuous fuzzing has not found.

astiob:
> Can you confirm that the test case was not uploaded at oss-fuzz.com?

At the very least, nothing on oss-fuzz.com jumps out at me as related to this PR’s CI runs.

Since we’re no longer seeing a failure here, I suppose we may as well merge now? And hope that the artefact gets uploaded when it becomes relevant, or that we can debug the non-upload then.

TheOneric:
> Fixed the recently broken build: google/oss-fuzz#6888

Thanks!

---

I force-pushed to squash the commits and fix the indents; only cosmetics no other changes.

The linked CIFuzz docs suggest to me, `language` should be `c++`. Although libass is pure C, the fuzzing-consumer program is (currently) C++ and [marked such in `project.yaml`](https://github.com/google/oss-fuzz/blob/1c4bcf6267f1524df21f467525fe1c512cdc4249/projects/libass/project.yaml#L2) *(this might change with a fix for the mysterious bogus(?) memory leaks)*. Interestingly it seems to work anyway.
 
.

Otherwise, **not a blocker**, but a noteworthy observation from the last three CIFuzz runs and a question:  
Fuzzing aborts if a sample causes crahes or trips the sanitisers, *even if the bug is already known to OSS-Fuzz*. In the latter case it doesn't result in a failure-status for a CI run, but it cuts short the time spent fuzzing, likely reducing some of its usefulness.
   - For the last three runs the `undefined` sanitiser only spent **6 to 8 seconds** actually fuzzing, before encountering a known case — incidentally(?) it seems to be the same case everytime.
   - `address` spent only around a minute fuzzing before it encountered a known timeout.

On a related note: can the fuzzing be told to ignore timeouts? OR at least make CIFuzz use a significantly larger timeout value than OSSFuzz? If the fuzzer is unlucky, the generated ASS can get expensive to render and may exceed the limit during regular operation. I haven't checked all timeout-samples ever reported by OSSFuzz to libass, but anecdotally the timeout-sample I recentishly tested was such a false positive.

jonathanmetzman:
There was an issue with reporting timeouts that has been fixed. In fact timeouts won't even be reported by default anymore.
See: https://github.com/google/oss-fuzz/issues/6703

TheOneric:
Thanks for the info @jonathanmetzman, though when I rerun the CI job yesterday it still aborted early on the very same samples as before and rerunning now (after changing language to `c++`) now *both* `address` and `undefined` abort on timeouts before even just a minute of fuzzing. *(See: https://github.com/libass/libass/runs/4411037399?check_suite_focus=true)*

Is the “ignore timeouts” already included in what CIFuzz uses? And is it really intended that CIFuzz aborts completely on the first failing sample even if this is a known failure? As it stands now, those abortions prevent CIFuzz from testing anything but the first couple of samples, likely meaning most newly introduced failures cannot be noticed by it.

Of course, this is no longer something requiring changes on the content of this PR anymore, but rather on CIFuzz itself.

PR #569, opened 2021-10-28
use utf16le encoded path to open file or dir on windows.

Masaiki:
An alternative implementation of #526, some suggestions from #526 are referenced.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -1235,20 +1235,27 @@ out:
 }
 #endif                          // ICONV
 
-/**
- * \brief read file contents into newly allocated buffer
- * \param fname file name
- * \param bufsize out: file size
- * \return pointer to file contents. Caller is responsible for its deallocation.
- */
-char *read_file(ASS_Library *library, char *fname, size_t *bufsize)
+FILE *ass_open_file(const char *path)
+{
+#if defined(_WIN32) && !defined(__CYGWIN__)
+    wchar_t *wpath = ass_convert_to_utf16le(path, CP_UTF8, NULL);
+    if (wpath) {
+        FILE *f = _wfopen(wpath, L"rb");
+        free(wpath);
+        if (f)
+            return f;
+    }
+#endif
+    return fopen(path, "rb");
+}
+
+static char *read_opened_file(ASS_Library *library, const char *fname, FILE *fp, size_t *bufsize)
 {
     int res;
     long sz;
     long bytes_read;
     char *buf;
 
-    FILE *fp = fopen(fname, "rb");
     if (!fp) {
         ass_msg(library, MSGL_WARN,
                 "ass_read_file(%s): fopen failed", fname);
@@ -1292,6 +1299,28 @@ char *read_file(ASS_Library *library, char *fname, size_t *bufsize)
         *bufsize = sz;
     return buf;
 }
+
+/**
+ * \brief read file contents into newly allocated buffer
+ * \param fname file name
+ * \param bufsize out: file size
+ * \return pointer to file contents. Caller is responsible for its deallocation.
+ */
+char *read_file(ASS_Library *library, char *fname, size_t *bufsize)
+{
+    FILE* fp = ass_open_file(fname);
+    return read_opened_file(library, fname, fp, bufsize);
+}
+#if defined(_WIN32) && !defined(__CYGWIN__)
+char *read_wide_named_file(ASS_Library *library, wchar_t *wpath, size_t *bufsize)
+{
+    FILE *fp = _wfopen(wpath, L"rb");
+    char *fname = ass_convert_from_utf16le(wpath, CP_UTF8, NULL);
+    char *res = read_opened_file(library, fname ? fname : "(conversion failed)", fp, bufsize);
+    free(fname);
+    return res;
+}
+#endif
 
 /*
  * \param buf pointer to subtitle text in utf-8
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -31,7 +31,11 @@
 #include <limits.h>
 #include <ft2build.h>
 #include <sys/types.h>
+#if defined(_WIN32) && !defined(__CYGWIN__)
+#include <windows.h>
+#else
 #include <dirent.h>
+#endif
 #include FT_FREETYPE_H
 #include FT_SFNT_NAMES_H
 #include FT_TRUETYPE_IDS_H
@@ -166,6 +170,113 @@ static ASS_FontProviderFuncs ft_funcs = {
     .destroy_font      = destroy_font_ft,
 };
 
+#if defined(_WIN32) && !defined(__CYGWIN__)
+/*
+ * Output parameters are only valid if the call was sucessful, ignore them otherwise.
+ * \param path          input; path of the dir to open, may be UTF-8 or the codepage used by fopen
+ * \param file_context  output
+ * \param find_handle   output; needs to be closed
+ * \param out_length    output
+ * \return wide-version of path or NULL on failure; needs to be freed if non-NULL
+ */
+static wchar_t *win32_open_dir(const char *path, WIN32_FIND_DATAW *file_context, HANDLE *find_handle, size_t *out_length)
+{
+    size_t tmp;
+
+    if (!path || !(tmp = strlen(path)) || tmp > SIZE_MAX - 3)
+        return NULL;
+
+    char* valid_search_dir = malloc(tmp + 3);
+    if (!valid_search_dir)
+        return NULL;
+    memcpy(valid_search_dir, path, tmp);
+    switch (valid_search_dir[tmp - 1]) {
+    case '\\':
+    case '/':
+        break;
+    default:
+        valid_search_dir[tmp++] = '\\';
+    }
+    valid_search_dir[tmp++] = '*';
+    valid_search_dir[tmp] = '\0';
+
+    UINT cur_cp = CP_UTF8;
+    size_t wlen;
+    wchar_t *wpath = ass_convert_to_utf16le(valid_search_dir, cur_cp, &wlen);
+    if (!wpath || (*find_handle = FindFirstFileExW(wpath, 0, file_context, 0, NULL, 0)) == INVALID_HANDLE_VALUE) {
+        free(wpath);
+        cur_cp = CP_ACP;
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+        if (!AreFileApisANSI()) cur_cp = CP_OEMCP;
+#endif
+        wpath = ass_convert_to_utf16le(valid_search_dir, cur_cp, &wlen);
+        if (wpath) {
+            *find_handle = FindFirstFileExW(wpath, 0, file_context, 0, NULL, 0);
+            if (*find_handle == INVALID_HANDLE_VALUE) {
+                free(wpath);
+                wpath = NULL;
+                goto clean;
+            }
+        } else
+            goto clean;
+    }
+
+    // path_utf8 ends with "\*" or "/*"
+    wlen -= 2;
+    wpath[wlen] = '\0';
+    if (out_length)
+        *out_length = wlen;
+
+clean:
+    free(valid_search_dir);
+    return wpath;
+}
+
+static void load_fonts_from_dir(ASS_Library *library, const char *dir)
+{
+    WIN32_FIND_DATAW find_file_context;
+    HANDLE find_file_handle;
+    size_t wdirlen;
+
+    wchar_t *wdir = win32_open_dir(dir, &find_file_context, &find_file_handle, &wdirlen);
+    if (!wdir)
+        return;
+
+    char *dir_utf8 = ass_convert_from_utf16le(wdir, CP_UTF8, NULL);
+
+    size_t namemax = 0;
+    wchar_t *wnamebuf = NULL;
+    do {
+        if (find_file_context.cFileName[0] == L'.')
+            continue;
+        size_t namelen = wdirlen + wcslen(find_file_context.cFileName) + 2u;
+        if (namelen < 2 || namelen - 2 < wdirlen)
+            continue;
+        if (namelen > namemax) {
+            size_t newlen = FFMAX(2048, namelen + FFMIN(256, SIZE_MAX - namelen));
+            if (ASS_REALLOC_ARRAY(wnamebuf, newlen))
+                namemax = newlen;
+            else
+                continue;
+        }
+        swprintf(wnamebuf, namemax, L"%ls\\%ls", wdir, find_file_context.cFileName);
+        char *file_name_utf8 = ass_convert_from_utf16le(find_file_context.cFileName, CP_UTF8, NULL);
+        ass_msg(library, MSGL_INFO, "Loading font file '%s\\%s'",
+                dir_utf8 ? dir_utf8 : "(conversion failed)",
+                file_name_utf8 ? file_name_utf8 : "(conversion failed)");
+        size_t bufsize = 0;
+        void *data = read_wide_named_file(library, wnamebuf, &bufsize);
+        if (data) {
+            ass_add_font(library, file_name_utf8, data, bufsize);
+            free(data);
+        }
+        free(file_name_utf8);
+    } while (FindNextFileW(find_file_handle, &find_file_context));
+    free(dir_utf8);
+    free(wnamebuf);
+    FindClose(find_file_handle);
+}
+#else
 static void load_fonts_from_dir(ASS_Library *library, const char *dir)
 {
     DIR *d = opendir(dir);
@@ -202,6 +313,7 @@ static void load_fonts_from_dir(ASS_Library *library, const char *dir)
     free(namebuf);
     closedir(d);
 }
+#endif
 
 /**
  * \brief Create a bare font provider.
--- a/libass/ass_utils.c
+++ b/libass/ass_utils.c
@@ -511,3 +511,34 @@ ASS_Style *lookup_style_strict(ASS_Track *track, char *name, size_t len)
     return NULL;
 }
 
+#if defined(_WIN32) && !defined(__CYGWIN__)
+wchar_t *ass_convert_to_utf16le(const char *str, UINT in_code_page, size_t *out_strlen)
+{
+    int wlen = MultiByteToWideChar(in_code_page, MB_ERR_INVALID_CHARS, str, -1, NULL, 0);
+    if (!wlen)
+        return NULL;
+    wchar_t *wstr = malloc(sizeof(wchar_t) * wlen);
+    if (!MultiByteToWideChar(in_code_page, MB_ERR_INVALID_CHARS, str, -1, wstr, wlen)) {
+        free(wstr);
+        return NULL;
+    }
+    if (out_strlen)
+        *out_strlen = wlen - 1;
+    return wstr;
+}
+
+char *ass_convert_from_utf16le(const wchar_t *wstr, UINT out_code_page, size_t *out_wstrlen)
+{
+    int len = WideCharToMultiByte(out_code_page, 0, wstr, -1, NULL, 0, NULL, NULL);
+    if (!len)
+        return NULL;
+    char *str = malloc(len);
+    if (!WideCharToMultiByte(out_code_page, 0, wstr, -1, str, len, NULL, NULL)) {
+        free(str);
+        return NULL;
+    }
+    if (out_wstrlen)
+        *out_wstrlen = len - 1;
+    return str;
+}
+#endif
--- a/libass/ass_utils.h
+++ b/libass/ass_utils.h
@@ -213,4 +213,10 @@ static inline int mystrtoi32(char **p, int base, int32_t *res)
     return *p != start;
 }
 
+#if defined(_WIN32) && !defined(__CYGWIN__)
+#include <windows.h>
+wchar_t *ass_convert_to_utf16le(const char *str, UINT in_code_page, size_t *out_strlen);
+char *ass_convert_from_utf16le(const wchar_t *wstr, UINT out_code_page, size_t *out_wstrlen);
+#endif
+
 #endif                          /* LIBASS_UTILS_H */
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -24,7 +24,7 @@
 #include <stdarg.h>
 #include "ass_types.h"
 
-#define LIBASS_VERSION 0x01502000
+#define LIBASS_VERSION 0x01502001
 
 #ifdef __cplusplus
 extern "C" {
@@ -258,6 +258,10 @@ void ass_library_done(ASS_Library *priv);
  * Optional directory that will be scanned for fonts recursively.  The fonts
  * found are used for font lookup.
  * NOTE: A valid font directory is not needed to support embedded fonts.
+ * On Microsoft Windows, when using WIN32-APIs,
+ * fonts_dir must either be in UTF-8 or the encoding accpeted by fopen
+ * with UTF-8 taking precedence if both versions are valid and exist.
+ * On all other systems there is no need for special considerations like that.
  *
  * \param priv library handle
  * \param fonts_dir directory with additional fonts
@@ -474,7 +478,7 @@ void ass_get_available_font_providers(ASS_Library *priv,
  * If the requested fontprovider does not exist or fails to initialize, the
  * behavior is the same as when ASS_FONTPROVIDER_NONE was passed.
  * \param config path to fontconfig configuration file, or NULL.  Only relevant
- * if fontconfig is used.
+ * if fontconfig is used. The encoding must match the one accepted by fontconfig.
  * \param update whether fontconfig cache should be built/updated now.  Only
  * relevant if fontconfig is used.
  *
@@ -664,6 +668,10 @@ void ass_flush_events(ASS_Track *track);
  * \param fname file name
  * \param codepage encoding (iconv format)
  * \return newly allocated track or NULL on failure
+ * NOTE: On Microsoft Windows, when using WIN32-APIs,
+ * fname must either be in UTF-8 or the encoding accpeted by fopen
+ * with UTF-8 taking precedence if both versions are valid and exist.
+ * On all other systems there is no need for special considerations like that.
 */
 ASS_Track *ass_read_file(ASS_Library *library, char *fname,
                          char *codepage);
@@ -683,6 +691,10 @@ ASS_Track *ass_read_memory(ASS_Library *library, char *buf,
  * \param fname file name
  * \param codepage encoding (iconv format)
  * \return 0 on success
+ * NOTE: On Microsoft Windows, when using WIN32-APIs,
+ * fname must either be in UTF-8 or the encoding accpeted by fopen
+ * with UTF-8 taking precedence if both versions are valid and exist.
+ * On all other systems there is no need for special considerations like that.
  */
 int ass_read_styles(ASS_Track *track, char *fname, char *codepage);
 
--- a/libass/ass_library.h
+++ b/libass/ass_library.h
@@ -39,5 +39,8 @@ struct ass_library {
 };
 
 char *read_file(struct ass_library *library, char *fname, size_t *bufsize);
+#if defined(_WIN32) && !defined(__CYGWIN__)
+char *read_wide_named_file(struct ass_library *library, wchar_t *wpath, size_t *bufsize);
+#endif
 
 #endif                          /* LIBASS_LIBRARY_H */


libass/ass.c
```diff
@@ -1235,6 +1235,31 @@ static char *sub_recode(ASS_Library *library, char *data, size_t size,
 }
 #endif                          // ICONV
 
+FILE* ass_fopen(const char* _FileName, const char* _Mode) {
+#if defined(_WIN32) && !defined(__CYGWIN__)
+    wchar_t* wpath, * wmode;
+    UINT cur_cp = CP_UTF8;
```
MrSmile, line 1241: I think we've already decided that `MultiByteToWideChar` with `CP_UTF8` is not enough: we need WTF-8 to ensure safe round-trip of any wide sequence, even broken UTF-16.

libass/ass.c
```diff
@@ -1235,6 +1235,31 @@ static char *sub_recode(ASS_Library *library, char *data, size_t size,
...
+    FILE* f = NULL;
+    wpath = ass_convert_to_utf16le(_FileName, cur_cp), wmode = ass_convert_to_utf16le(_Mode, cur_cp);
```
MrSmile, line 1243: Use of an universal conversion function for the file mode looks like a great overkill for me. I've used this code in a similar context:
```c
wchar_t wmode[4] = L"\0b\0";
while (*mode) switch (*mode++) {
    case 'r':  wmode[0] = L'r';  break;
    case 'w':  wmode[0] = L'w';  break;
    case 'a':  wmode[0] = L'a';  break;
    case '+':  wmode[2] = L'+';  break;
}
```
But even that is unneeded for libass, we only use "rb". So it's better to drop the second argument of `ass_fopen()` completely.

libass/ass.c
```diff
@@ -1235,6 +1235,31 @@ static char *sub_recode(ASS_Library *library, char *data, size_t size,
...
+    if (!wpath || !wmode || !(f = _wfopen(wpath, wmode))) {
+        free(wpath), free(wmode);
+        cur_cp = CP_ACP;
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+        if (!AreFileApisANSI()) cur_cp = CP_OEMCP;
+#endif
+        wpath = ass_convert_to_utf16le(_FileName, cur_cp), wmode = ass_convert_to_utf16le(_Mode, cur_cp);
+        if (!wpath || !wmode || !(f = _wfopen(wpath, wmode))) {
```
MrSmile, line 1251: Better to use `fopen()` directly instead of manually emulating it.

libass/ass_fontselect.c
```diff
@@ -166,7 +166,64 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+            size_t newlen = FFMAX(2048, namelen + FFMIN(256, SIZE_MAX - namelen));
+            if (ASS_REALLOC_ARRAY(namebuf, newlen))
+                namemax = newlen;
+            else
+                goto clean_d_name;
+        }
+        snprintf(namebuf, namemax, "%s/%s", dir, d_name);
+        size_t bufsize = 0;
+        ass_msg(library, MSGL_INFO, "Loading font file '%s'", namebuf);
+        void* data = read_file(library, namebuf, &bufsize);
```
MrSmile, line 214: At this point we know that `namebuf` is UTF-8 file name, but `read_file()` also tries to open it as ANSI. That can potentially lead to undesirable side effects so it's better to avoid it. Also there's unnecessary UTF16→UTF8→UTF16 round-trip here, maybe it's worth a try to eliminate it.

MrSmile (changes requested):
Nice try, but this subject has a lot of hidden complications that should be addressed.
I also have a bunch of style complaints, but decided not to voice it yet.

libass/ass.c
Masaiki: I admit this and will change it in a later commit.

libass/ass.c
Masaiki: later commit, too.

libass/ass.c
Masaiki: [here](https://github.com/libass/libass/pull/526#issuecomment-873599277) referenced that MB_ERR_INVALID_CHARS is needed. So back the flag to 0 ?

MrSmile: AFAIK, `MultiByteToWideChar` cannot do the thing at all, so a custom conversion function is needed.

libass/ass_fontselect.c
Masaiki: this also bothered me, but most time it won't fallback to ascii.
also, converting between utf16 and utf8 is lossless, though little inefficient.

libass/ass.c
Masaiki: just like sub_recode?

MrSmile: 'sub_recode()` uses iconv, which doesn't support WTF-8 too.

TheOneric:
Some remarks, not a full review:
 - The commits need to be squashed.
 - Also needs to update all docs relating to filepaths explaining the special encoding requirements or pointing out the lack of it for the fontconfig-configfile
  - #526 used MS native `FindFirstFileW` for directory traversal; this would be better for MSVC which doesn't provide a `dirent.h`
  - There are some style issues, eg libass uses `snake_case` for variable names and putting multiple statements on the same line like `free(), free();` is discouraged iirc.
  
WTF-8 being required to really work for all files was discussed on IRC; that's because MS' allows file-names to contain invalid UTF16-sequences. However, imho it's probably “good enough” for now to start with UTF8.

Masaiki:
> * [use win API to open file. #526](https://github.com/libass/libass/pull/526) used MS native `FindFirstFileW` for directory traversal; this would be better for MSVC which doesn't provide a `dirent.h`  
one thing i found is that libass in vcpkg or https://github.com/ShiftMediaProject/libass both provide a dirent.h, one [here](https://github.com/tronkko/dirent) and one [here](https://github.com/ShiftMediaProject/libass/blob/master/SMP/dirent.h) they use FindNextFileW inside and work well with _wopendir and _wreaddir

TheOneric:
> one thing i found is that libass in vcpkg or https://github.com/ShiftMediaProject/libass both provide a dirent.h, one here and one here they use FindNextFileW inside and work well with _wopendir and _wreaddir

Exactly, they need to and do provide an additional header to get a `_wreaddir`/`_wopendir`-wrapper around `FindNextFileW`. Since we'll do a special-ifdeffery only for Windows' file-access anyway, ideally we would not require such a  wrapper, but just use `FindNextFileW` directly.

libass/ass_fontselect.c
```diff
@@ -166,7 +168,91 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+    HANDLE find_file_handle;
+    UINT cur_cp = CP_UTF8;
+
+    char* valid_search_dir;
+    int tmp;
+
+    if (!dir || !(tmp = strlen(dir)))
+        return;
+
+    valid_search_dir = (char*)malloc(sizeof(char) * (tmp + 2));
```
TheOneric, line 185: `sizeof(char)` is per definition always `1`, so not needed. The cast is also superfluous in C.

Also `tmp + 2` may result in undefined behaviour if `tmp > INT_MAX - 2`, and in practice worse than a sigend overflow, is that the following character accesses like `valid_search_dir[tmp - 1]` may read out of bounds.

libass/ass_fontselect.c
```diff
@@ -166,7 +168,91 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+    if (!dir_utf16le || (find_file_handle = FindFirstFileW(dir_utf16le, &find_file_context)) == INVALID_HANDLE_VALUE) {
+        free(dir_utf16le);
+        cur_cp = CP_ACP;
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+        if (!AreFileApisANSI()) cur_cp = CP_OEMCP;
+#endif
+        dir_utf16le = ass_convert_to_utf16le(valid_search_dir, cur_cp);
+        if (dir_utf16le) find_file_handle = FindFirstFileW(dir_utf16le, &find_file_context);
+    }
+    if (find_file_handle == INVALID_HANDLE_VALUE) {
```
TheOneric, line 209: If both conversions failed, this will access uninitialised memory.

libass/ass_fontselect.c
```diff
@@ -166,7 +168,91 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+    wchar_t* dir_utf16le = ass_convert_to_utf16le(valid_search_dir, cur_cp);
+
+    if (!dir_utf16le || (find_file_handle = FindFirstFileW(dir_utf16le, &find_file_context)) == INVALID_HANDLE_VALUE) {
+        free(dir_utf16le);
+        cur_cp = CP_ACP;
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+        if (!AreFileApisANSI()) cur_cp = CP_OEMCP;
+#endif
+        dir_utf16le = ass_convert_to_utf16le(valid_search_dir, cur_cp);
+        if (dir_utf16le) find_file_handle = FindFirstFileW(dir_utf16le, &find_file_context);
```
TheOneric, line 207: The handle opened by `FindFirstFileW` never gets [closed](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findclose).

TheOneric (changes requested):
For readability it would probably be good to refactor the directory-handle opening into its own function.

TheOneric:
Also this seems to be missing an include, see CI-failure.

Masaiki:
> For readability it would probably be good to refactor the directory-handle opening into its own function.

If you use a new seperate function to implement it, its parameters bother me a little bit, so I haven't changed this part.

Masaiki:
it seems this two lines  https://github.com/libass/libass/blob/ac2ddef8841aa2ff37ca9b83f19f15092d0cacc6/libass/ass_fontselect.c#L181-L182
skip the current dir, previous dir and hidden files
should i change the windows version of `load_fonts_from_dir` to windows specific implementation in my pr

TheOneric:
> [regarding skipping dot-files]
> should i change the windows version of load_fonts_from_dir to windows specific implementation (using FILE_ATTRIBUTE_DIRECTORY and FILE_ATTRIBUTE_HIDDEN) in my pr

I'm leaning towards leaving the dot-file logic as is, as this makes it easier for people who share or synchronise their font-dirs between different operating systems *(e.g. if the whole mpv-config-dir is shared or synchronised)* and matches current behaviour. But that's not a strong preferences.

> If a new seperate function is used to implement it, its parameters bother me a little bit,

How about this?
```C
/*
 * Output parameters are only valid if the call was sucessful, ignore them otherwise.
 * \param path          input; path of the dir to open, may be UTF-8 or the codepage used by fopen
 * \param file_context  output
 * \param find_handle   output; needs to be closed
 * \param path_utf8     output; utf-8 encoded version of path; needs to be freed
 * \return true if succesful, false otherwise
 */
 static bool microsoft_open_dir(char* path, WIN32_FIND_DATAW file_context, HANDLE find_handle, char* path_utf8);
 // alternatively use wchar_t* wpath as output
```

> I have noticed that the maximum path length is approximately 32 767, so should it be further limited by the maximum value.

It's not limited like that. While checking previously if the handle needs to be closed, I stumbled upon docs that told me there is a special path-prefix for surpassing the old limit and since some WIndows10-update that limit can also be  surpassed without using that special path prefix.

---

Also documentation for the encoding is still missing. Eg for `ass_set_fonts_dir`, `ass_read_file` and `ass_read_styles` something like this:
```diff
+ * NOTE: On Microsoft Windows, when using WIN32-APIs,
+ * fonts_dir must either be in UTF-8 or the encoding accpeted by fopen
+ * with UTF-8 taking precedence if both versions are valid and exist.
+ * On all other systems there is no need for special considerations like that.
```
And for `ass_set_fonts`:
```diff
  * behavior is the same as when ASS_FONTPROVIDER_NONE was passed.
  * \param config path to fontconfig configuration file, or NULL.  Only relevant
- * if fontconfig is used.
+ * if fontconfig used. The encoding must match the one accepted by fontconfig.
  * \param update whether fontconfig cache should be built/updated now.  Only
  * relevant if fontconfig is used.
```

libass/ass_utils.c
```diff
@@ -511,3 +511,26 @@ ASS_Style *lookup_style_strict(ASS_Track *track, char *name, size_t len)
     return NULL;
 }
 
+#if defined(_WIN32) && !defined(__CYGWIN__)
+wchar_t* ass_convert_to_utf16le(const char* data, UINT in_code_page) {
+    int len_utf16le = MultiByteToWideChar(in_code_page, MB_ERR_INVALID_CHARS, data, -1, NULL, 0);
+    if (!len_utf16le) return NULL;
```
TheOneric, line 517: Per libass' style, control flow statements like `break`, `return`, etc should always be placed on their own line.

TheOneric (changes requested):
Mostly style and other nits. I'm also not completely happy with the wording of the documentation I proposed; if there are better suggestion I'd love to read them.
Also, please squash all commits into one.

Apart from those nits, the remaining issue is whether to keep the conversion round trip or not and it would probably best if someone more familiar with Microsoft's APIs could take a look at the patch.

libass/ass_fontselect.c
MrSmile: > Ideally there'd be no roundtrip conversion, yes, but then we'd need another `read_file` for Microsoft's `wchar_t`.

We can move `fopen()` outside of `read_file()` and make something like `read_opened_file()`.

> we pass it as UTF-8 and UTF-8 is the first one to be tested.

But opening can fail even if filename decoded correctly (access problems). And this can trigger incorrect second choice. I think there should be at least some flag to skip that second opening attempt.

> As long as we don't support WTF-8 anyway,

Implementing WTF-8 support isn't that hard so I don't see the benefit of postponing it.

TheOneric: Right, opening can also fail if the file exists (or it could have been deleted in the meantime). So I'm revising my previous statement: It should use either use `wchar_t*` directly or add WTF-8-support. (If `read_file` is split though, I think this should by a separate *preceding* commit and also add an `ass_` prefix to the new functions.)

> Implementing WTF-8 support isn't that hard so I don't see the benefit of postponing it.

If someone is volunteering to do that, sure.

TheOneric:
As this is a (breaking) API-change, it also needs to bump `LIBASS_VERSION` in `ass.h`.

libass/ass_fontselect.c
Masaiki: About the problem that well-decoded filename failed to be opened because of some access problems, would "checking the file exsiting before opening it“ be a solution for the problem?
codes like this
```cpp
bool ass_file_exsits(const wchar_t *path) {
    DWORD file_attributes = GetFileAttributesW(path);
    if (file_attributes == INVALID_FILE_ATTRIBUTES)
    {
        switch (GetLastError())
        {
        case ERROR_FILE_NOT_FOUND:
        case ERROR_PATH_NOT_FOUND:
            return false;
        default:
            return true;
        }
    }
    return true;
}
```

TheOneric (commented):
Remaining issues:
 - the opening `{` and closing `}` offunctions should be on their own lines (style)
 - bump `LIBASS_VERSION`
 - don't attempt to open file-paths we got from `wchar_t` as ANSI
 - possibly: use WTF-8 instead of UTF-8, requiring custom conversion function to and from ill-formed UTF-16 and to valid UTF-8 for logging  [(WTF-8 docs here)](https://simonsapin.github.io/wtf-8/#decode-ill-formed-utf-16)
 - ideally: review by someone familiar with Microsoft's APIs

libass/ass.c
```diff
@@ -1235,6 +1235,21 @@ static char *sub_recode(ASS_Library *library, char *data, size_t size,
...
+#if defined(_WIN32) && !defined(__CYGWIN__)
+    wchar_t *path_utf16le = ass_convert_to_utf16le(path, CP_UTF8);
+    FILE *f = NULL;
+    if (!path_utf16le || !(f = _wfopen(path_utf16le, L"rb"))) {
+        free(path_utf16le);
+        return fopen(path, "rb");
+    }
+    free(path_utf16le);
+    return f;
+#else
```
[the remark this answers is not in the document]
MrSmile: `_wfopen()` is a MS-specific variant of `fopen()` so it shouldn't require any extra headers.

MrSmile (changes requested):
Current Windows version of `load_fonts_from_dir()` is equivalent (with ton of copy-paste) to the old `load_fonts_from_dir()` with custom `readdir` implementation. So we should either make it more specific (skip ANSI opening path etc.) or separate differences into `readdir`.

Also I think that names such as `path_utf8` and `path_utf16le` are cumbersome (especially second), better to use simple `path` and `wpath` correspondingly.

I'm planning to look into WTF conversion functions later (find or reinvent implementation), but sadly we need `MultiByteToWideChar()` for `CP_ACP`/`CP_OEMCP` too.

libass/ass_fontselect.c
```diff
@@ -166,6 +170,111 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+        if (namelen < 2 || namelen - 2 < wdirlen)
+            continue;
+        if (namelen > namemax) {
+            size_t newlen = FFMAX(2048, namelen + FFMIN(256, SIZE_MAX - namelen));
+            if (ASS_REALLOC_ARRAY(wnamebuf, newlen))
+                namemax = newlen;
+            else
+                continue;
+        }
+        swprintf(wnamebuf, namemax, L"%s\\%s", wdir, find_file_context.cFileName);
```
TheOneric, line 262: The format specifiers are wrong, they need to be `%ls` since the arguments are wide/long strings, not regular `char*` ones. *(And if they were `char*`, you'd need to check for `swnprintf`-failures during conversion.)*

libass/ass.c
```diff
@@ -1235,29 +1235,36 @@ static char *sub_recode(ASS_Library *library, char *data, size_t size,
...
-                "ass_read_file(%s): fopen failed", fname);
+            "ass_read_file(%s): fopen failed", fname);
         return 0;
     }
     res = fseek(fp, 0, SEEK_END);
     if (res == -1) {
         ass_msg(library, MSGL_WARN,
-                "ass_read_file(%s): fseek failed", fname);
+            "ass_read_file(%s): fseek failed", fname);
         fclose(fp);
```
TheOneric, line 1268: Don't add arbitrary whitespace changes. Another instance a bit further below.

libass/ass.c
```diff
@@ -1235,29 +1235,36 @@ static char *sub_recode(ASS_Library *library, char *data, size_t size,
...
+        free(wpath);
+        if (f)
+            return f;
+    }
+#endif
+    return fopen(path, "rb");
+}
+
+static char* read_opened_file(ASS_Library *library, char *fname, FILE *fp, size_t *bufsize)
 {
```
TheOneric, line 1253: Misplaced `*`.

libass/ass_fontselect.c
```diff
@@ -31,7 +31,11 @@
 #include <limits.h>
 #include <ft2build.h>
 #include <sys/types.h>
+#if defined(_WIN32) && !defined(__CYGWIN__)
+#include <Windows.h>
```
TheOneric, line 35: The previously used lowercase variant is better as it matches the preexisting `#include <windows.h>` and `#include <winapifamily.h>` from `config.ac` and the DirectWrite font provider. (Several instances of this)

libass/ass_fontselect.c
```diff
@@ -166,6 +170,111 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+/*
+ * Output parameters are only valid if the call was sucessful, ignore them otherwise.
+ * \param path          input; path of the dir to open, may be UTF-8 or the codepage used by fopen
+ * \param file_context  output
+ * \param find_handle   output; needs to be closed
+ * \param path_utf8     output; utf-8 encoded version of path; needs to be freed
+ * \return true if succesful, false otherwise
+ */
+static wchar_t *win_open_dir(const char *path, WIN32_FIND_DATAW *file_context, HANDLE *find_handle, size_t *out_length)
+{
```
TheOneric, line 183: The documentaion no longer matches the signature and implementation.

Also, instead of using a return value **and additionally** an output-argument, it's better to use a single `ASS_StringView` from `ass_utils.h`, as MrSmile suggested.
The latter applies to all function changed this way.

TheOneric (changes requested):
Two of the previously mentioned issues persist:
 * the opening `{` and closing `}` of functions should be on their own lines (style)
 * need to bump `LIBASS_VERSION` as it is an API-cahnge *(a -break even)*
 
 And some new issues appeared with the latest changes:

libass/ass_fontselect.c
```diff
@@ -166,6 +170,107 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+
+    size_t dirlen = strlen(dir_utf8);
+    size_t namemax = 0;
+    char *namebuf = NULL;
+    do {
+        const wchar_t *d_name_utf16le = find_file_context.cFileName;
+        char *d_name_utf8 = ass_convert_from_utf16le(d_name_utf16le, CP_UTF8);
+        if (d_name_utf8[0] == '.')
+            continue;
+        size_t namelen = dirlen + strlen(d_name_utf8) + 2u;
```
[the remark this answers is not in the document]
TheOneric: The `u` is copied from the POSIX version. Theoretically it is possible to have `sizeof(size_t) + 1 = sizeof(int)`. In this case if `2` without `u` were used all summands are auto-"promoted" to a *signed* `int` and UB can oocur via signed overflow if `dirlen + strlen(d_name_utf8) >= 2 * SIZE_MAX - 1`. With `u` all types remain unsigned and a possible overflow is well defined. *(Possibly but unconfirmed, the `s390`(≠`s390x`) platform with its logically 31bit-pointers might be affected by this.)*
Ofc, for all currently known MS Windows platforms this shouldn't be an issue, but leaving the `u` doesn't negatively impact anything and matches the POSIX-version.

libass/ass_fontselect.c
```diff
@@ -166,6 +170,107 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+#if defined(_WIN32) && !defined(__CYGWIN__)
+/*
+ * Output parameters are only valid if the call was sucessful, ignore them otherwise.
+ * \param path          input; path of the dir to open, may be UTF-8 or the codepage used by fopen
+ * \param file_context  output
+ * \param find_handle   output; needs to be closed
+ * \param path_utf8     output; utf-8 encoded version of path; needs to be freed
+ * \return true if succesful, false otherwise
+ */
+static bool microsoft_open_dir(const char *path, WIN32_FIND_DATAW *file_context, HANDLE *find_handle, char **path_utf8) {
```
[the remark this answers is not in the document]
TheOneric: I slightly prefer the `microsoft` prefix, as the API might also apply to differently named/marketed systems like “Xbox system software” or the OSs might be renamed in the future as happened with MacOS/OSX twice, while the company name changing is less likely.
But I'd be fine with a `windows`-prefix too.
However, I do not like `win_` as it is too short; out of context it wouldn't immediately be clear to me what `win_` stands for.

libass/ass_fontselect.c
astiob: Microsoft’s `swprintf` uses `%s` for wide strings. But `%ls` should also work.

libass/ass_fontselect.c
astiob: And it’s not just for superficial consistency: when cross-compiling on Linux, the names are usually case-sensitive and only the lowercase version exists.

libass/ass_fontselect.c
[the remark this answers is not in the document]
astiob: The platform(s) this code is aimed at (and the guard macro) is named “Win32” and a possible second platform is named “Windows Runtime”/“Universal Windows Platform”; and Microsoft has also had (POSIX subsystem/Services for UNIX/Interix) and currently has (CBL-Mariner) other platforms that are more compatible with POSIX and lack the Win32 APIs that this code uses.

I think the most straightforward, honest and appropriate name tag might be “win32”.

libass/ass_fontselect.c
TheOneric: Oups, forgot this is a `wchar_t*`, so `ASS_StringView doesn't work (except for the conversion **to** UTF-8). And declaring a new Windows-exclusive type isn't too pretty either; nvm then.

Masaiki:
i don't know should "bump LIBASS_VERSION" be done by me, done whatever.

libass/ass_utils.h
```diff
@@ -213,4 +213,10 @@ static inline int mystrtoi32(char **p, int base, int32_t *res)
     return *p != start;
 }
 
+#if defined(_WIN32) && !defined(__CYGWIN__)
+#include <windows.h>
```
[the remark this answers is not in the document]
TheOneric: Are you suggesting to put the defines into `ass_compat.h` or do you want to create a new header for (currently) only those two functions?

libass/ass_fontselect.c
```diff
@@ -166,6 +170,111 @@ static ASS_FontProviderFuncs ft_funcs = {
...
 };
 
+#if defined(_WIN32) && !defined(__CYGWIN__)
+/*
+ * Output parameters are only valid if the call was sucessful, ignore them otherwise.
+ * \param path          input; path of the dir to open, may be UTF-8 or the codepage used by fopen
+ * \param file_context  output
+ * \param find_handle   output; needs to be closed
+ * \param out_length    output
+ * \return utf-16le version of path; needs to be freed
```
TheOneric, line 180: `\return wide-version of path or NULL on failure; needs to be freed if non-NULL`

libass/ass_fontselect.c
```diff
@@ -166,6 +170,111 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+ */
+static wchar_t *win32_open_dir(const char *path, WIN32_FIND_DATAW *file_context, HANDLE *find_handle, size_t *out_length)
+{
+    UINT cur_cp = CP_UTF8;
```
TheOneric, line 184: Nit: imho it would be better to move this closerto its usage, eg after `size_t wlen;`

libass/ass_fontselect.c
```diff
@@ -166,6 +170,111 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+    wchar_t *wpath = ass_convert_to_utf16le(valid_search_dir, cur_cp, &wlen);
+    if (!wpath || (*find_handle = FindFirstFileExW(wpath, 0, file_context, 0, NULL, 0)) == INVALID_HANDLE_VALUE) {
+        free(wpath);
+        cur_cp = CP_ACP;
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+        if (!AreFileApisANSI()) cur_cp = CP_OEMCP;
+#endif
+        wpath = ass_convert_to_utf16le(valid_search_dir, cur_cp, &wlen);
+        if (wpath) {
+            if ((*find_handle = FindFirstFileExW(wpath, 0, file_context, 0, NULL, 0)) == INVALID_HANDLE_VALUE) {
```
TheOneric, line 214: Nit: This assignment would imho be better moved out of the `if`

libass/ass_fontselect.c
```diff
@@ -166,6 +170,111 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+                free(wpath);
+                wpath = NULL;
+                goto clean;
+            }
+        }
+        else
+            goto clean;
```
TheOneric, line 221: Style Nit: Imho the `else` is better placed at the same line as the closing `}` from the `if`-block

libass/ass_fontselect.c
```diff
@@ -166,6 +170,111 @@ static ASS_FontProviderFuncs ft_funcs = {
...
+}
+
+static void load_fonts_from_dir(ASS_Library *library, const char *dir)
+{
+    WIN32_FIND_DATAW find_file_context;
+    HANDLE find_file_handle;
+    wchar_t *wdir = NULL;
+    size_t wdirlen;
+
+    if (!(wdir = win32_open_dir(dir, &find_file_context, &find_file_handle, &wdirlen)))
```
TheOneric, line 242: Nit: Again, I'd prefer the assignment to be outside the `if`

libass/ass_utils.c
```diff
@@ -511,3 +511,33 @@ ASS_Style *lookup_style_strict(ASS_Track *track, char *name, size_t len)
...
+    wchar_t *wstr = malloc(sizeof(wchar_t) * wlen);
+    if (!MultiByteToWideChar(in_code_page, MB_ERR_INVALID_CHARS, str, -1, wstr, wlen)) {
+        free(wstr);
+        return NULL;
+    }
+    if (out_length)
+        *out_length = wlen - 1;
+    return wstr;
+}
+char *ass_convert_from_utf16le(const wchar_t *wstr, UINT out_code_page, size_t *out_length)
```
TheOneric, line 529: Style Nit: two functions should be separated by an empty line

TheOneric (commented):
Thanks for your continued effort on this :)
There are a few more issues I didn't notice before and a couple of nits.

MrSmile:
I've tried an alternative approach to this problem (beware, not tested): https://github.com/MrSmile/libass/commits/win-unicode
I've created a custom `readdir` wrapper with all OS-specific parts mostly isolated in a single file. That wrapper is also generic enough to use it for `compare` in the future. What do you think, should I finalize it into a PR?

TheOneric:
> I've tried an alternative approach to this problem (beware, not tested)

I see the appeal of isolating this, but at least in its current form I find this harder to read than the approach taken here in #569. Also (apart fro mnneding testing) your commit lacks the `LIBASS_VERSION`-bump, documentation and at first glance `enlarge_size_wtf16to8` seems suspicious with it's magic `3`, but I'd need to take a closer look at it *(apart from future changes, where updating the magic value might be forgotten, isn't the current max already `4`?)*.
Also, while there's some appeal to splitting it out, we (hopefully) will never need this OS-specifc stuff anywhere else but in the places covered by the current #569, so I don't think it's strictly necessary.

If readability improves, the missing bits are added and it passes testing and review, your approach could be technically superior. But I'd also be mostly fine with #569 as is. For both though, a review from someone familiar with Microsoft-APIs *(I'm not)* would be preferable.

.

> That wrapper is also generic enough to use it for compare in the future.

To be honest, I do not care if `compare` — not part of the library, but a utility for testing which already needs a specific directory layout and filenames anyway —  can access non-ASCII filenames on Windows. But if you want it to handle this or have a usecase for it, then sure.

MrSmile:
> `enlarge_size_wtf16to8` seems suspicious with it's magic `3`

I've simply followed all branches of `convert_wtf16to8()` and got the worst case (see relevant comments `1w -> 1`, `1w -> 2`, `2w -> 4` and `1w -> 3` there).

Okay, I'll convert this experiment into a proper PR then.

astiob (commented):
A few comments on the Windows side. They also apply to MrSmile’s version.

By the way, thanks for making paths such a mess, Microsoft.

Besides my inline comments, our behaviour could potentially be improved further, but this may be seen as an extra improvement out of scope of the current issue/PRs:

* We can hit the path length limit. To make this less likely, we _could_ check whether the path does not start with `\\?\` (which is a request that the path be left alone as far as possible), normalize the path and prepend our own `\\?\` or even `\\.\` if that actually works here.

* Microsoft’s docs make a good point that’s relevant on all platforms: as working directories are process-wide, we should avoid reusing relative paths, lest another thread change the working directory mid-`load_fonts`. On Windows (as far as I can tell), as well as on older UNIXes, the only alternative is to convert the input path to absolute as soon as we can and proceed with absolute paths. On modern UNIX, we could use `openat` etc., which are even better. Note that if we were to prepend `\\?\` on Windows, we’d have to convert the path to absolute for that already.

Issue #461, opened 2020-12-10
Support \kt tag


rcombs:
This exists in xy-vs and it's unclear why we don't handle it… but I'm not entirely sure how it's used. Corresponding xy-vs code: https://github.com/Cyberbeing/xy-VSFilter/blob/ff77719517bc0c1144a75c7b2c1e6268d582614a/src/subtitles/RTS.cpp#L2509-L2517

astiob:
It’s part of ASS2, which we don’t currently support any of. Neither does Aegisub, although, of course, it renders \kt when VSFilter is selected.

astiob:
As for its use, it lets you specify when a karaoke syllable starts as an absolute time (during the event’s duration) instead of as an implicit sum of all preceding karaoke tags.

TheOneric:
If I didn't miss something our prior discussions and findings about v4++ aren't summarised anywhere on the issue tracker yet; let's use this issue for v4++ as a whole.

All new features are:
 - `\kt` as discussed above.
 - Splitting the `MarginV` property of styles and events into two distinct for top and bottom respectively
 - Adding a new Style property internally called `relativeTo`. It does … *something*   in guliverkli and guliverkli2, but afaict is just ignored after parsing in xy-VSFilter.

Since libass parses Format lines *(and non-standard lines affect the defaults)*, we need to know the canonical field names. The internal names of VSF do not always match the in-document names. VSFilter itself only creates v4 and v4+ but no v4++ files, thus not contributing any hint. Also while Aegisub for a while supported read/write of v4++ files, it never wrote `Format:` lines for v4++ files.

TheOneric:
Regarding `relativeTo` (originally intended) purpose and the `Format:` field names, here's an excerpt from asa’s wiki (see: https://web.archive.org/web/20071018035454/http://asa.diac24.net/ASS2 ):

> ASS2 is a minor extension to the [ASS](https://web.archive.org/web/20071018035454/http://asa.diac24.net/ASS) subtitle format. It was introduced in [VSFilter SVN revision 453](https://web.archive.org/web/20071018035454/http://svn.sourceforge.net/viewcvs.cgi/guliverkli/trunk/guliverkli/src/subtitles/STS.cpp?r1=423&r2=453) (November 2005), but did not become public until being noticed in MPC somewhen around March 2006 (cf. [http://forum.doom9.org/showthread.php?t=108356](https://web.archive.org/web/20071018035454/http://forum.doom9.org/showthread.php?t=108356))
[...]

Issue #121, opened 2014-07-21
Clip is incorrectly applied before shifting lines for collisions


tgoyne:
VSFilter: ![dummy_002_0](https://cloud.githubusercontent.com/assets/408944/3639774/21a9a024-1089-11e4-9885-b38801a38f1a.png)

libass: ![dummy_001_0](https://cloud.githubusercontent.com/assets/408944/3639773/19696822-1089-11e4-9066-849cbebbba4e.png)

Script: https://gist.github.com/tgoyne/e5b398428755a66d6edb

Not really very important as clipping multiple lines without explicitly setting their positions rarely makes any sense.

TheOneric:
Same happens for out of frame clipping, without any explicit `\clip` or `\iclip` tags, see #391 (which I only now realised is related to this). I'll close #391 in favour of this.

PR #570, opened 2021-11-03
RFC: render: support explicitly tagging a line as TS or dialogue

rcombs:
Some sub authors had commented on wanting to be able to use e.g. clips without making libass think a line was dialogue, or tag a margin-positioned line as TS if it's expected to align with explicitly-positioned ones. This is a simple implementation of that.

This is done via a comment at the start of the line, and only has meaning when selective overrides are enabled, so it should be safe to add.

Do we use `explicit` for anything that this might give unexpected results for? (i.e. anything where this might cause changes in rendering even without selective overrides enabled?)

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1060,8 +1060,14 @@ init_render_context(ASS_Renderer *render_priv, ASS_Event *event)
     render_priv->state.effect_skip_timing = 0;
 
     apply_transition_effects(render_priv, event);
-    render_priv->state.explicit = render_priv->state.evt_type != EVENT_NORMAL ||
-                                  event_has_hard_overrides(event->Text);
+
+#define EXPLICIT_DIALOGUE_MARKER "{dialogue}"
+#define EXPLICIT_TS_MARKER "{ts}"
+
+    render_priv->state.explicit = !strncmp(event->Text, EXPLICIT_DIALOGUE_MARKER, sizeof(EXPLICIT_DIALOGUE_MARKER) - 1) ? 0 :
+                                  !strncmp(event->Text, EXPLICIT_TS_MARKER, sizeof(EXPLICIT_TS_MARKER) - 1) ? 1 :
+                                  (render_priv->state.evt_type != EVENT_NORMAL ||
+                                   event_has_hard_overrides(event->Text));
 
     reset_render_context(render_priv, NULL);
     render_priv->state.alignment = render_priv->state.style->Alignment;


astiob:
> Do we use `explicit` for anything that this might give unexpected results for? (i.e. anything where this might cause changes in rendering even without selective overrides enabled?)

Yes, and so does VSFilter: for repositioning due to collisions and for screen clipping.

rcombs:
Huh, I couldn't find an equivalent of our full list in vsfilter (with clips, p, etc…).

astiob:
OK, maybe not for collisions.

> Huh, I couldn't find an equivalent of our full list in vsfilter (with clips, p, etc…).

See here for XySubFilter:

https://github.com/pinterf/xy-VSFilter/blob/1f557059965668a9309008e0f5296425eced3295/src/subtitles/RTS.cpp#L1765-L1829

https://github.com/pinterf/xy-VSFilter/blob/1f557059965668a9309008e0f5296425eced3295/src/subtitles/RTS.cpp#L1956-L1971

For MPC-HC, I couldn’t find an equivalent at first glance, so I’m not sure what happens in MPC-HC.

astiob:
As a concept, this has been desired for a long time and could be useful in the (more distant) future for more things, such as rendering & blending TS at video resolution and text at display resolution. I think this was even written into that next-gen subtitle format wishlist on our wiki.

In more concrete matters, why special comment instead of proper new tag?

rcombs:
Mostly because I realized that we don't actually need a compat-break for this, and adding entire new tags feels much more like breaking compat and expecting people to target libass in particular (whereas this is intended to just be a non-normative hint for override cases, which already don't provide any compatibility guarantees).

TheOneric:
Have you considered using Style names as a hint instead, similar to what #213 does *(but with different names, also names for dialogue and not using `explicit`)*, or do you specifically want to mark some lines of the same style as a type and others as dialogue?

rcombs:
This is a new feature for future files; I'm open to other implementation options. If we use style/actor/effect, we should consider whether we match on substring, prefix, or entire string.

Making it a comment was partly a psychological thing, yeah. Making it a tag might be preferable, though; what would we want to call it?

witchymary:
I was discussing this with other fansubbers, as there is great interest in something like this being implemented. We thought that a good solution would be to just use the Script Info section. For example:

```
[Script Info]
; Script generated by Aegisub 9214, Daydream Cafe Edition [Shinon]
; http://www.aegisub.org/
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
PlayResX: 1920
PlayResY: 1080
YCbCr Matrix: TV.709
DialogueStyle: Dialogue, Dialogue - Italics, Dialogue - Top
```
Or maybe:
```
[Script Info]
; Script generated by Aegisub 9214, Daydream Cafe Edition [Shinon]
; http://www.aegisub.org/
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
PlayResX: 1920
PlayResY: 1080
YCbCr Matrix: TV.709
DialogueStyle: Dialogue
DialogueStyle: Dialogue - Italics
DialogueStyle: Dialogue - Top
SignsStyle: Signs
```

Thoughts? I find it less cumbersome and cleaner than the other suggested implementations.

astiob:
That seems kinda perverse to me. If it’s a style property, it should be part of the Style line.

Can we just reuse the existing ASS2 relativeTo field? IIRC MPC-HC is the only one that currently honours it, but what it does is probably quite similar to what libass does to dialogue lines with margin use enabled.

> use e.g. clips in dialogue without making libass think a line was TS

How would this even work with margins?

astiob:
> Can we just reuse the existing ASS2 relativeTo field? IIRC MPC-HC is the only one that currently honours it, but what it does is probably quite similar to what libass does to dialogue lines with margin use enabled.

FWIW, searching for `relativeTo` through MPC-HC’s RTS.cpp suggests that it does indeed do almost the same as current libass with margins on, the only difference being that it scales each dimension (X and Y) independently as display area size ÷ storage/play resolution, whereas libass attempts a letterbox/pillarbox fit of the aspect-ratio-preserving rectangle and then shifts (translates) events along the longer dimension depending on their Alignment property.

We could change this in libass or in MPC-HC to make them match each other, or just declare margins explicitly “implementation-defined behaviour”, and then finish https://github.com/libass/libass/issues/461 and simply advise subbers to start using ASS2.

TheOneric:
> > use e.g. clips in dialogue without making libass think a line was TS
>
> How would this even work with margins?

I sure hope this is not what is being abused here, but if like in #121 the shift into margins happens after clipping it will currently produce reliable results. (But we already know we do the wrong thing here.)
I guess if we really wanted to, we could treat shifting into margins as a separate later step after shifting for collisions and clipping. But this would mean there can be seemingly odd collision shift as e.g. an `\an2` and `an6` event may collide on video res, but not anymore after expanding into playback margins. (And I currently can’t imagine a useful case for this)

> then finish https://github.com/libass/libass/issues/461 and simply advise authors to start using ASS2.

I agree `relativeTo` is the ideal solution, but to keep expectations realistic making `relativeTo` usable requires work beyond libass and likely will take a while.
 - #461 requires an ABI break (and would be much simpler if we’d also break style-struct API) which is the reason why https://github.com/libass/libass/pull/651 got split from `\kt` and indefinitely delayed
 - Aegisub can’t read v4++ files (anymore, it used to once)
 - potentially mkvtoolnix, ffmpeg etc might also need patches to be able to mux v4++ files

Mikewando:
I think there are broadly situations outside of rendering which are interested in whether a line is dialogue or typesetting that should be considered. I would prefer not to tie whether a line is dialogue or typesetting to how it is rendered.

Personally I have some automated timing scripts which want to treat dialogue and typesetting differently. But there are other situations I could imagine like running a spellcheck only on dialogue. Today these cases would have to rely on heuristics and can't be confidently run across scripts where style names, actors, etc are inconsistent. A flag on the style would I think be a reasonable to avoid relying on heuristics, but they don't really want to change how the script is rendered.

So maybe a new style field is appropriate and `relativeTo` can be kept as a rendering-only field instead.

astiob:
I’m not sure it would be wise to extend the _format_ for something that isn’t meant to affect rendering. ASS does, of course, already have the `Actor`/`Name` field that does nothing, but is it really wise to add even more?

Besides, is it reasonable to expect that an extra Boolean toggle would be enough to cover all possible use cases authors have for splitting lines into groups? Even for the spellcheck example, you may very well have signs that are full words and you do want to include in the checking. Within a single script or closely related group of scripts where you have exact knowledge of how the scripts were authored, you can, as mentioned, group lines by style, by effect, by actor, by special comment. If you have a large collection of disparate scripts, it’s hard for me to imagine where having a “typesetting vs dialogue, oh, but not for rendering” split would be useful. Not to mention that if it doesn’t affect rendering, it has no more enforced consistency than the Actor field, so it _will_ be filled with garbage data.

TheOneric:
> the main reason why this came up right now was the use case of blending subtitles in players. [...] it's been proposed to blend captions onto the video at the video's resolution (which is usually the intended usage anyway) while dialogue is still rendered at the player's resolution.

This fits well with the already existing `relativeTo`, anything blended into the video is by necessity video-relative and cannot use margins, i.e. `relativeTo=1`. And anything which *needs* to be relative to the video is exceedingly likely to be intended to coverup/replace/extend/modify a feature of the video and thus could be rendered+blended in video resolution.

> Of course this needs additional concessions [...]

Adding this new ordering into current format versions *could* lead to problems, since it will look different depending on whether the player chose to implement the split blending or blends everything at the same time.
I guess it would be possible to avoid the higher rasterising cost for types if libass only internally first rasterisers them at storage res and then upscales the bitmap before passing them out, but this will probably not look as good blending into the raw video frame and the upscaling costs for each bitmap might negate the gains.

But either way, for such split blending to be possible we’d need at least and ABI (with split rendering being opt-in) if not even an API break in libass and modifications to all libass-using players which want/need to support this.

> Of course, having a more robust way to decide which events should be affected by style overrides is another big application. This feels equivalent to the first use case, but I'm sure one could also find an example where the two differ...

I guess an iconic phrase, attack name, or similar might appear in normal dialogue and for fun get effects and features (sparks, flames, or whatever) added. In this case it doesn't need to be in any specific position relative to a video feature, but style overrides would turn the effects and added features into a mess or at least misplace them.  
However, to bring text and additional features together most likely explicit positioning is required and tbh I don’t know/remember how `relativeTo` interacts with explicit positioning. If positioning enforces video-relative this might be moot, but if the position is allowed to be shifted in a predictable way, this might be a case of “use margins, but no style overrides”.  
It’s likely rare though.

> Aegisub Extradata

For the spellcheck example this does seem like an appropriate place to me. E.g.the script could default to using `relativeTo` unless the style is explicitly in/excluded by some Extradata header, or just exclusively use an Extradata header.

astiob:
> if libass only internally first rasterisers them at storage res and then upscales the bitmap before passing them out, but this will probably not look as good blending into the raw video frame and the upscaling costs for each bitmap might negate the gains.

For reference: this is exactly what XySubFilter does (for all lines) when “Render To Original Video Size” is enabled in its preferences. And it does indeed not look very good when paired with a video renderer that uses a high-quality upscaler. (XySubFilter’s upscaler is bilinear or similar.) Personally, I was excited to discover this checkbox when I went back to Windows, but I believe I eventually disabled it for this very reason.

Split-resolution rendering may not need an ABI break, but I think that may end up being the least of our concerns. More importantly, video players will need significant adjustments to actually implement the split blending.

(Side note: personally, this is exactly the primary motivation why I’ve wanted a dialogue/typesetting split for years <sub>(and I did mention it in https://github.com/libass/libass/pull/570#issuecomment-958583431)</sub>. Style overrides and repositioning into margins are features that I don’t normally use myself but that I always have in mind because other people use them and because, unlike split-resolution rendering, existing renderers actually support them.)

> tbh I don’t know/remember how `relativeTo` interacts with explicit positioning

Not in any special way. Exactly as it says on the tin, `relativeTo==0` makes all positions relative to the target display area, whether explicit or implicit. As mentioned, MPC-HC scales the horizontal positions & sizes and separately the vertical positions & sizes from the corresponding storage/script dimension to the corresponding display-area dimension, stretching the storage/script pixel grid and distorting its aspect ratio if appropriate.

Normal dialogue with special effects is a real pain in this context, let me tell you. Thanks for mentioning it. On the one hand, these effects tend to be based around elements in the video (e. g. unusual margins or clipping to match on-screen elements or a `\move` to evade present-in-original hardsubs), and even if not, tend to require (or at least be easier) explicit positioning to make various layers match up. On the other hand, if the viewer overrides dialogue styles and/or enables margins, having one positioned, styled line really jumps out. Ideally, you’d rather just have all of this line’s styling be dropped and any decorative elements dropped entirely. But even a typesetting/dialogue Boolean toggle isn’t enough to express this.

clsid2:
MPC-HC ISR also has option to configure sub rendering size similar to XySubFilter.
So subs can be rendered at e.g. 1080p and result is then upscaled to video display size (e.g. 4k). For such high resolution that still gives decent quality.
So that is something that a libass API user can already do. But this is something that is done independent of the complexity of individual subtitle lines. So not ideal if only small percentage of subtitles renders too slow.
So a possible libass option could be to internally render only lines with very complex typesetting at half the requested resolution. And then upscale before delivery to API user. Using exactly half size allows for best upscaling quality with a frame doubling algo.
API user can then choose to activate that option when target resolution is high.
You could even make this behavior (optionally) dynamic by measuring the time it takes to render a complex sub. If it is longer than video frame duration, then halve resolution (if not done already) for subsequent complex sub lines.
Or maybe even dynamically skip costly effects such as blurring if render time is too slow.

I am curious about the libass performance differences for different target resolutions for a complex sub sample. And which effects have greatest performance impact.

astiob:
To be clear, split-resolution rendering is not just about performance; it’s also about quality. Complex typesetting routinely uses large amounts of layers that effectively amount to bitmap drawings, and they cannot be correctly rescaled at non-integer scaling rates in VSFilter/libass rendering pipelines (which rasterize each event separately and blend the bitmaps in sequence), producing visual artefacts such as discoloured edges between the original pixels or distorted shapes (where the edges coincided with physical pixel edges at the original video resolution but involve fractional pixels at display resolution).

> MPC-HC ISR also has option to configure sub rendering size similar to XySubFilter.

Oh? I don’t see one in the usual Options → Subtitles menu. Where is it?

clsid2:
Options > Subtitles > maximum texture resolution
(the resolutions shown are actually just pixelcounts internally, it maintains AR of original target size)
(default is 1440p, but with an internal special case to then use 1080p for 4k)

Yes, I know it impacts quality. Hence the idea to only do it at high res. Rendering a bitmap at exactly half width/height (so actually quarter size) should allow for a good upscaling result.

astiob:
> Options > Subtitles > maximum texture resolution

Thanks. I can’t say I’m fully sure what this does, but it definitely doesn’t sound like the same as XySubFilter’s, which is to use the original video resolution, whatever it happens to be for each video you’re watching, rather than impose a fixed upper limit.

> Rendering a bitmap at exactly half width/height

of what, though? To achieve correct rendering, you must render and blend at video resolution or an integer multiple of it, not the display resolution. And this requires changes in the video renderer, not just in libass.

astiob:
By the way, if all the discussions here and in #297 lead to this requiring a major revamp of ASS… or if we ever do produce a major revamp of ASS regardless of this… then I think it would make sense to make typesetting use YCbCr directly rather than RGB with a hackish matrix header.

astiob:
Also, especially since handling this properly would eventually require video player pipelines to be adjusted anyway, it might make sense to just force typesetting and dialogue to be separate subtitle tracks that are displayed simultaneously. This would complicate external subtitles, though. But on the bright side, this would help cases like dual-audio releases where signs should probably be visible regardless of audio language but the user likely wants to hide the dialogue subtitles when watching the dub.

rcombs:
Not fully; in practice, typesetting sometimes needs adjustments to work with dubs anyway (like, if terminology is different in the sub vs dub tracks, if a different song translation is used, etc). I wouldn't want to make players have to support selecting separate TS and dialogue tracks; that's a bunch of extra UI work that, in practice, isn't going to happen.

astiob:
> Modern scripts should just require to match video, and there is no need to worry about colorspaces.

So you’re agreeing with me: scripts need to use video’s colorspace, which is YCbCr.

PR #574, opened 2021-11-26
ci/gha: add CIFuzz integration

DavidKorczynski:
This adds CI integration with CIFuzz. CIFuzz will run the libass OSS-Fuzz fuzzers for a short period of time for each PR. The benefit is that regressions and bugs has a chance of being captured early in the development process.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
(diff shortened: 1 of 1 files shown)


astiob:
Thanks! This is much appreciated, but someone else was faster in #568. 😅 The current form of #568 further adds the `undefined` sanitizer, so could you perhaps look over that and share any comments? We also delayed merging that because we saw a CI failure on GitHub that didn’t upload any GitHub artefact, but it’s no longer failing, so we can’t continue debugging that.

DavidKorczynski:
Ah! Thanks, and my bad! I will leave Adam to it then - he showed me how to do this so he should be able to carry that one through

Issue #386, opened 2020-04-01
libass doesn't know how to handle OpenType Variable Fonts


sheik124:
https://docs.microsoft.com/en-us/typography/font-list/bahnschrift

This is the font in reference, couldn't include the TTF since Microsoft doesn't seem to allow redistribution. If you have a current install of Windows 10, you have Bahnschrift.ttf already.

Noticed this behavior in MPV and Plex, did some homework and found out they use libass.

Bahnschrift (and other variable fonts) have variable weight and width. Some also include "preset" combinations of weight/width - e.g. regular Bahnschrift is 400 wt 100 wd, but Bahnschrift Condensed is 400 wt 75 wd. If I pick "Bahnschrift Condensed" as a font, MPV and Plex can't display it properly, with MPV showing the regular version, and Plex showing a misscaled bigger version of regular Bahnschrift, even with Bahnschrift.ttf embedded. XySubFilter (or perhaps LAV Splitter?) can properly find the font and select "Condensed", whether it's only attached to the MKV or installed on the system.

Current workaround was to use FontLab 7 (trial, paid) to split Bahnschrift.ttf into independent .otf fonts based on weight. If I pick the new "Bahn-Cond" font I made and attach bond-cond.otf, libass _can_ display it properly.

I don't know if it has anything to do with Aegisub 3.2.2's font name handling.

TheOneric:
To ease testing, since Microsofts font is hard to get:
Here's an open-source OpenType Varaible Font variable in width and weight and width predefined/named Variations: https://github.com/googlefonts/Inconsolata
Another one would be: https://github.com/jenskutilek/homecomputer-fonts
Or adobes testing prototype for variable fonts: https://github.com/adobe-fonts/adobe-variable-font-prototype

Using Inconsolata-VF.ttf, I tested the following ASS event:
```
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\fs96}{\bord6\b0} Test\N  {\fnInconsolata Extra-light}Test {\fnInconsolata Light}Test {\fnInconsolata}Test {\fnInconsolata\b1}Test{\b0} {\fnInconsolata Black}Test\N{\fnInconsolata} {\b200} Test {\b300}Test {\b400}Test {\b700}Test {\b900}Test
```

First line is for comparsion with default font, second line attempts to use named weight variations and third selects the same weight variations via `{\bxxx}`:
[...]

Issue #578, opened 2021-12-15
Can't find 'Cascadia Code' with PostScript name


chosungmann:
We can't use the font named [Cascadia Code](https://github.com/microsoft/cascadia-code). Specifying the font as a full name or a PostScript name leads to the fallback font being selected all the time. So I did some digging and it turns out fonts in the form of so-called Multiple Masters always seems to give the problematic result. For the Multiple Masters, each named instance has its own PostScript name, so I think we have to consider it when comparing PostScript names.

I attached the subtitle and the fonts I used for testing: [test.zip](https://github.com/libass/libass/files/7717094/test.zip)

**Result from [the master branch](https://github.com/libass/libass/tree/7732119e53ef235942b9f4820f70a72c1d6045c2):**

```console
test.c: Available font providers (4): 'None', 'Autodetect', 'CoreText', 'Fontconfig'.
libass: libass API version: 0x1502000
libass: libass source: commit: 0.15.2-30-gc75017b4e7ad686009c0ac14b7d9b22db4811aa1
libass: Raster: FreeType 2.11.0
libass: Shaper: FriBidi 1.0.11 (SIMPLE) HarfBuzz-ng 3.1.2 (COMPLEX)
libass: Initialized
libass: Using font provider coretext
libass: File size: 889
libass: [0x600002cac000] Style: Default,CascadiaCodeItalic-Bold,52,&Hffffff,&Hffffff,&H0,&H0,0,0,0,0,100,100,0,0,1,1,0,2,10,10,10,0
libass: Track treated as legacy ffmpeg sub.
libass: Added subtitle file: 'subtitle.ass' (2 styles, 5 events)
libass: fontselect: Using default font family: (CascadiaCodeItalic-Bold, 400, 0) -> /System/Library/Fonts/Helvetica.ttc, -1, Helvetica
libass: fontselect: (CascadiaCodeItalic-Bold, 400, 0) -> /System/Library/Fonts/Helvetica.ttc, -1, Helvetica
```

**Result from [the master branch](https://github.com/libass/libass/tree/7732119e53ef235942b9f4820f70a72c1d6045c2) with #579**

```console
test.c: Available font providers (4): 'None', 'Autodetect', 'CoreText', 'Fontconfig'.
libass: libass API version: 0x1502000
libass: libass source: commit: 0.15.2-30-gc75017b4e7ad686009c0ac14b7d9b22db4811aa1
libass: Raster: FreeType 2.11.0
[...]

PR #579, opened 2021-12-15
[Abandoned] Add support for OpenType Variable Fonts

chosungmann:
For the Multiple Masters, each named instance has its own PostScript name,
so we have to consider it when comparing PostScript names.

BUG=#578

### DIFF
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -668,8 +668,11 @@ static bool matches_full_or_postscript_name(ASS_FontInfo *f,
 
     if (check_postscript(f))
         return matches_postscript_name;
-    else
-        return matches_fullname;
+
+    if (matches_postscript_name)
+        return matches_postscript_name;
+
+    return matches_fullname;
 }
 
 /**
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -22,6 +22,7 @@
 #include <inttypes.h>
 #include <ft2build.h>
 #include FT_FREETYPE_H
+#include FT_MULTIPLE_MASTERS_H
 #include FT_SYNTHESIS_H
 #include FT_GLYPH_H
 #include FT_TRUETYPE_TABLES_H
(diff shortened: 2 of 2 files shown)


libass/ass_fontselect.c
```diff
@@ -668,8 +668,11 @@ static bool matches_full_or_postscript_name(ASS_FontInfo *f,
 
     if (check_postscript(f))
         return matches_postscript_name;
-    else
-        return matches_fullname;
+
+    if (matches_postscript_name)
+        return matches_postscript_name;
```
astiob, line 673: In particular, this doesn’t seem right. Here, `matches_postscript_name` is very intentionally disregarded if `check_postscript()` is not true; otherwise, this could be just `return matches_postscript_name || matches_fullname`.

This needs to be tested against VSFilter (I’ll do it if nobody else does before me) using the fonts from #386 and #578, because there’s a small chance that Windows handles variable fonts differently, but for normal fonts, the very point of this function is to match
* _either_ the PostScript name (and ignore the full name) for CFF-outline fonts (which use Adobe’s font driver on Windows)
* _or_ the full name (and ignore the PostScript name) for TrueType fonts (which use Microsoft’s font driver on Windows).

astiob (commented):
Thanks!

This will probably (eventually) fix #386, which #578 seems to be a duplicate of.

Naming note: according to at least [FreeType docs](https://freetype.org/freetype2/docs/reference/ft2-multiple_masters.html#ft_set_named_instance), fonts that bore the actual “Multiple Master” name did _not_ have named instances. However, OpenType Variable Fonts do, and that’s what Cansadia Code is.

At first glance, the additional FreeType code looks pretty nice. This can serve as a good start! But I think it isn’t yet integrated correctly into libass’s font selection machinery. Things that worry me:
* Full-names should probably also be compared, e. g. #386 has an example with Inconsolata that uses names that don’t look like valid PostScript names.
* PostScript vs full names, keeping in mind VSFilter compatibility. See inline comment.
* This patch only modifies a single branch of `ass_face_open`. That code is only used to open system fonts found by Core Text. This means that this patch allows one to use variable system fonts on macOS, but not system fonts on Windows and not embedded/attached/`ass_set_fonts_dir` fonts on any platform.

PR #580, opened 2021-12-15
ci/gha: bump to ubuntu-latest and dedicated "oldlib"-build

TheOneric:
First commit as suggested by astiob, though I'm not sure how far back the extra `--dirty` step works.
The "oldlib" container has almost the minimum version of all direct dependencies:
  - FreeType2 `2.3.6`(external) - `9.17.3`(internal)
  - HarfBuzz `1.2.3`
  - Fontconfig `2.10.92`
  - FriBidi `0.19.1` *(I couldn't find a tarball for `0.19.0` and there's no tag for it either; though the diff shows configure.ac used 0.19.0 at some point)*

As well as:
  - automake 1.11
  - autoconf 2.64
  - gcc 6.3
  - Debian Stretch's default version of `libpng` and `libexpat`, the latter being a transitive dependency

`docker_fetch` and `docker_image` allow to easily extend this to more purpose-built containers both from the same or other build-repo.
The build-recipe and binary for the container can be found here: https://github.com/TheOneric/libass-containers

### DIFF
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -36,6 +36,12 @@ jobs:
             cc: gcc
             docker_image: alpine:latest
             shell: '/usr/bin/docker exec dockerciimage sh -e {0}'
+          # Add docker-build with minimum version of dependencies
+          - os: ubuntu-latest
+            cc: gcc
+            docker_image: oldlibs
+            docker_pullprefix: 'ghcr.io/theoneric/libass-containers/'
+            shell: '/usr/bin/docker exec dockerciimage sh -e {0}'
           # Add a Windows build (MinGW-gcc via MSYS2) with no extras
           - os: windows-2019
             msystem: MINGW32
@@ -64,11 +70,13 @@ jobs:
         if: matrix.docker_image
         shell: bash
         run: |
-          # Note: With this setup everything inside the container will be run as root
-          docker pull "${{ matrix.docker_image }}"
+          # Note: Many containers default to the root user
+          docker pull "${{ matrix.docker_pullprefix }}${{ matrix.docker_image }}"
           docker create --name dockerciimage \
             -v "/home/runner/work:/home/runner/work" --workdir "$PWD"  \
-            --entrypoint "tail" "${{ matrix.docker_image }}" "-f" "/dev/null"
+            --entrypoint "tail" \
+            "${{ matrix.docker_pullprefix }}${{ matrix.docker_image }}" \
+            "-f" "/dev/null"
           docker start dockerciimage
 
       - name: Setup MSys2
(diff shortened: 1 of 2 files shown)


astiob:
Thanks!

I recall GitHub had its own Docker registry. I don’t know any details, but have you considered using that for your image?

FWIW, as far as I can tell, a 0.19.0 release never existed, and it was an internal version number during the development leading up to the release of 0.19.1 since the very first FriBidi commit in CVS (and now, Git). In particular, some of the stuff we use appeared during that development phase, so a 0.19.0 minimum requirement technically isn’t accurate. But as far as releases go, it doesn’t matter.

<i>Nitpicking: the official spelling is “FriBidi” with lowercase “d”. Indeed, it’s “GNU FriBidi”, but hey.</i>

TheOneric:
> FWIW, as far as I can tell, a 0.19.0 release never existed, and it was an internal version number during the development [...] In particular, some of the stuff we use appeared during that development phase, so a 0.19.0 minimum requirement technically isn’t accurate.

Should we bump the minimum version declared in `configure.ac` to `0.19.1` then?

> I recall GitHub had its own Docker registry. I don’t know any details, but have you considered using that for your image?

I did not know about that. It could potentially simplify the process if it doesn't need to rely on release assets as storage. However, looking at what's available to non-paying users on the [billing info](https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages) it isn't clear to me how the storage used is counted. Uncompressed the container image takes up ~350MiB, compressed it varies between 130MiB and 65MiB based on compression algorithm and level.
Non-paying users get “500MB” of storage and lmited traffic, depending on the employed and (not-)counted compression and how easy it is to clear out everything but the latest revision of a container this may already be or quickly become problematic if more container-varaints are added. *(To avoid containers being unavailable temporarily, I'd want to stay below half the storage so a new version of all can be uploaded and only *after this* delete the old versions)*.

astiob:
> Should we bump the minimum version declared in `configure.ac` to `0.19.1` then?

Sounds good to me.

> looking at what's available to non-paying users on the [billing info](https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages)

Hmm, I think it says “GitHub Packages usage is free for public packages. For private packages, [the tables below apply].”

If still in doubt, can we, uhh, try? If your current GitHub account is free, it should be safe to try to exceed the limits and see what happens.

TheOneric:
> Hmm, I think it says “GitHub Packages usage is free for public packages.

Oups, I missed that line. In that case it should work and I can look into using it instead of (ab)using release assets as storage.

TheOneric:
Now using GH's container registry and bumped minimum FriBidi to `0.19.1`.

Issue #493, opened 2021-04-01
Ease FriBidi linking on Windows


TheOneric:
As we discussed on IRC around a month ago, we might want to always define `FRIBIDI_LIB_STATIC` so no one needs to care if libass is linked against static or shared FriBidi *(only relevant for Windows builds)*.
Just to make sure we don't forget about it, here's an issue and a patch. The patch is completely untested and I have no intention to build anything on Windows.

If someone wants to do that, this patch would need to be tested on Windows with both static and dynamic FriBidi before it can be merged. Alternatively we could decide it's better to not hardcode workarounds like this and continue to leave this up to builders to specify which FriBidi is used *(just as it's required for all other libs/executables linking to FriBidi)*.

*The `#ifdef _WIN32` guard may not even be required, at least it seems to work on my machine *(but with a FriBidi predating https://github.com/fribidi/fribidi/commit/46ec6f04820d97159c1008539ed5fc846eefb671)**

```patch
From 5b22ee07dcadbec8fa3fd907a48ccf7deadffdb5 Mon Sep 17 00:00:00 2001
From: Oneric <oneric@oneric.stub>
Date: Thu, 1 Apr 2021 22:06:41 +0200
Subject: [PATCH] Workaround Windows build failures

Because FriBidi exports a variable its headers need to set dllimport
to work with dynamic builds. However, linking against static FriBidi
on Windows will fail when dllimport being set in the header. To allow
use of static FriBidi it is possible to surpress dllimport by defining
the 'FRIBIDI_LIB_STATIC' macro.
FriBidi tries to automatically set this macro in it's pc file, but the
used Cflags.private field is not a standard pkgconfig field and only
supported by one of the many pkgconfig implementations.

However, functions do not require dllimport to be set to work with both
dynamic and static builds. Since libass does only use FriBidi functions,
just always define 'FRIBIDI_LIB_STATIC' on Windows.
---
 libass/ass_compat.h | 6 ++++++
 1 file changed, 6 insertions(+)

diff --git a/libass/ass_compat.h b/libass/ass_compat.h
index ae16bb7..c1930bb 100644
--- a/libass/ass_compat.h
+++ b/libass/ass_compat.h
@@ -26,6 +26,12 @@
 #define inline __inline
 #endif
 
+// Workaround build failures on Windows with static FriBidi.
+// Possible because we only use FriBidi functions, not variables.
+#if defined(_WIN32) && !defined(FRIBIDI_LIB_STATIC)
[...]

PR #581, opened 2021-12-31
autotools: Fix static library for mingw

Biswa96:
This change defines FRIBIDI_LIB_STATIC while building static library
if fribidi present. Otherwise, libass static library imports fribidi
APIs with dllimport attribute.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -77,6 +77,12 @@ PKG_CHECK_MODULES([FRIBIDI], [fribidi >= 0.19.1], [
     CFLAGS="$CFLAGS $FRIBIDI_CFLAGS"
     LIBS="$LIBS $FRIBIDI_LIBS"
     AC_DEFINE(CONFIG_FRIBIDI, 1, [found fribidi via pkg-config])
+    if test "$enable_static" = "no"; then
+        static_fribidi=no
+    else
+        static_fribidi=yes
+    fi
+    AM_CONDITIONAL([STATIC_FRIBIDI], [ test "$static_fribidi" = yes ])
 ])
 
 PKG_CHECK_MODULES([HARFBUZZ], [harfbuzz >= 1.2.3], [
--- a/libass/Makefile.am
+++ b/libass/Makefile.am
@@ -1,3 +1,11 @@
+if STATIC_FRIBIDI
+extra_CPPFLAGS = -DFRIBIDI_LIB_STATIC
+else
+extra_CPPFLAGS =
+endif
+
+AM_CPPFLAGS = $(extra_CPPFLAGS)
+
 AM_CFLAGS = -std=gnu99 -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter \
             -Werror-implicit-function-declaration -Wstrict-prototypes        \
             -Wpointer-arith -Wredundant-decls -Wno-missing-field-initializers\


astiob:
Thanks! See also: #493. @TheOneric Could you remind me where that ended up at?

astiob:
If I understand this patch correctly, it adds `FRIBIDI_LIB_STATIC` if static libass is enabled, but it takes effect even if shared libass is also enabled and affects the shared libass as well (because the object files are not compiled separately for static and shared). At that point, it seems that this has few advantages over simply defining the macro _always_ as proposed in #493. Technically shared-only libass builds would be slightly more efficient in calling FriBidi with this patch than with #493, but this efficiency difference hardly seems worth anything.

Could you perhaps test #493 with static and shared libass and confirm whether it works?

Biswa96:
Yes, it also works. But as you said, it will be always defined which also removes the `__declspec(dllimport)` keyword from all fribidi function declarations irrespective static or shared build.

TheOneric:
> Could you remind me where that [#493] ended up at?

The patch was never actually tested, we weren't sure if defining `FRIBIDI_LIB_STATIC` or `FRIBIDI_EXTERN` would be better and we hope `pkg-config` will adopt the `Cflags.private`-patch as a cleaner solution with more portable pc-files and avoiding the need for a patch on our side. I did not get a response from a pkg-config maintainer yet though.

TheOneric:
The proposed patch is wrong. `-DFRIBIDI_LIB_STATIC` should depend on the nature of the used *FriBidi* library being linked to, whether or not *libass* is being built shared or static bears no relevance to my knowledge. If FriBidi's `declspec(dllimport)` were needed for libass' use case, defining `-DFRIBIDI_LIB_STATIC` as done here would prevent usage of any static libass build with a dynamic FriBidi *(which is done in parts our CI)* and also as noted by astiob, prevent shared libass built together with the static variant *(the default behaviour)* from being used with a shared FriBidi. Furthermore duplicating logic from our dependency into our own build is undesirable since prone to failure with older or future versions of the dependency.

In fact, FriBidi does make an attempt to get `-DFRIBIDI_LIB_STATIC` defined automatically, by adding `CFLAGS.private: -DFRIBIDI_LIB_STATIC` to its pc-file. The issue are `Cflags.private` being specific to the `pkgconf`-implementation and the capitalisation being non-standard relying on another `pkgconf`-specific property: case-insensitive field names. Also the compiler-flags are only used if the `--static` option is passed to pkgconf (which you should do when aiming for fully-static builds).  
If this is your issue, the preferred solution is to get the original `pkg-config` to adopt `Cflags.private` or similar *(which will with high certainty lead to OpenBSD's `pkg-config` and likely other maintained implementation to follow suit)* and get FriBidi to standardise the capitalisation of the field.  
There's a patch by me to add `Cflags.private` to the original `pkg-config`, see [pkg-config!13](https://gitlab.freedesktop.org/pkg-config/pkg-config/-/merge_requests/13). If you're already using `--static` in MSys2 you may adopt this as a distro-patch in MSys2 (if required together with [pkg-config!15](https://gitlab.freedesktop.org/pkg-config/pkg-config/-/merge_requests/15) to fix builds from git).  
Note that — assuming not defining `-DFRIBIDI_LIB_STATIC` were relevant — you'd still need to build libass twice if you want a (shared) version which can link to shared FriBidi and a (static) version which can link to static FriBidi. For libass specifically this might be avoidable, see next paragraph, but in general that will be required for any library with dependecies with an impactful `Cflags.private` field.

As astiob mentioned, the `dllimport` might not be relevant for libass. If you are interested in working around the need for either a pkg-config patch or two builds for different FriBidi versions on libass side, see [libass#493](https://github.com/libass/libass/issues/493). For one of 493's patches (eg defining `FRIBIDI_EXTERN` instead of `FRIBIDI_LIB_STATIC`) to be considered to applied, it needs to be tested on MS Windows to build and also to work once linked in a binary (e.g. with our [regression tests](https://github.com/libass/libass-tests)) for all combinations of static and shared libass and FriBidi. Ideally, but not necessarily with *both* current FriBidi and FriBidi version 1.0.9 *(before the export behaviour was changed)*.

---

> Yes, it also works. But as you said, it will be always defined which also removes the __declspec(dllimport) keyword from all fribidi function declarations irrespective static or shared build.

See https://github.com/libass/libass/issues/493

astiob:
Hang on. For one thing, IIRC MSYS uses pkgconf nowadays, which already supports that field, for this very reason. For another, how is pkgconf supposed to know whether FriBidi is static or not? _We_ are the ones asking it for `--libs` and adding or not adding `--static`, but we don’t know which FriBidi library files are actually available, and neither does pkgconf. Surely our best choice of `--static` is/would be no different from the patch that was proposed here.

Anyway, I think #493 should just be merged.

astiob:
Indeed, the linked MSYS [issue](https://github.com/msys2/MINGW-packages/issues/8817)/[PR](https://github.com/msys2/MINGW-packages/pull/8818) suggests a pure static build already works, so I assume `--static` is being passed and pkgconf is picking up the `Cflags.private` as expected.

TheOneric:
> For another, how is pkgconf supposed to know whether FriBidi is static or not?

Someone wanting to build a fully static libass can(should) use eg `PKG_CONFIG="/usr/bin/pkg-config --static" ./configure` which will cause everything to use the flags appropiate for static-building.  
If only certain dependencies are desired to be used statically, it's more complicated. If `--static` is not specified but only a static library of a dependency is available it will fail if this dependecy has unique dependencies — even without a `Cflags.private` field. Common build systems and pkg-config implementations are poorly equipped for user-selectable mixed static/shared builds. *(In autoconf the program-author can specifcally select static flags by using `PKG_CHECK_MODULES_STATIC` instead of `PKG_CHECK_MODULES`)*.

rcombs:
Shouldn't fribidi emit the `-D` in `cflags` (not just `cflags.private`) if built static(-only)?

astiob:
> Shouldn't fribidi emit the `-D` in `cflags` (not just `cflags.private`) if built static(-only)?

Right. Probably. Maybe it does.

I guess I misunderstood Oneric’s reasoning for why the proposed patch is wrong, and now that I understand it better, I disagree with it (and I’m frustrated by the harshness of reality).

Anyway, let’s just merge #493 (which makes it work in all cases) and stop worrying about this.

---

> Someone wanting to build a fully static libass can(should) use eg `LDFLAGS="-static" PKG_CONFIG="/usr/bin/pkg-config --static" ./configure` which will cause everything to use the flags aproppiate for static-building. So for a fully static build the user/builder can/should set `--static`, not our configure script.

I guess this is an actual reason why `Cflags.private` isn’t “the solution”. Thanks again, Microsoft. You simply _can’t_ make a standalone library if it depends on anything else. In the general case, each library needs to know the static/dynamic choices in the entire application’s dependency tree. And the specific case where you _can_ is if you never export anything but functions and never use `dllimport`, which is precisely what we pretend FriBidi does in #493 (and what we do ourselves in libass).

Earlier, it sounded like it all hinged on `Cflags.private`, but now it seems like that’s not it. It already has them, they already work, but that’s not enough. And even if the `Cflags.private` weren’t there, if the user is gonna have to add custom build-system flags anyway, they might as well add `FRIBIDI_LIB_STATIC` directly.

FWIW I think normally people do expect—I would—that the binaries built by `./configure --enable-static` are suitable for a fully static build, without any extra arcane incantations. Much like how `--enable-shared` is expected to be suitable for a fully dynamic build. (As you point out, on Windows, _neither_ can be used in a mixed build in the general case.) So I would agree with using `FRIBIDI_LIB_STATIC` for `--enable-static` if it mattered; but the problem would be `--enable-static --enable-shared`.

Looking again at our Autoconf-generated `configure`, I don’t see any `--static` flags for pkg-config, so I assume my earlier conclusion that it was already used given `--disable-shared` was wrong. I guess it’s because `pkg-config --static` never affected static output libraries before pkgconf added `Cflags.private` and because there’s no obvious reason to want it except as a kludge for this `dllimport` stuff on Windows. That MSYS PR must’ve not actually succeeded in fixing their static libass/FriBidi build, which must’ve prompted this libass PR.

TheOneric:
> Shouldn't fribidi emit the -D in cflags (not just cflags.private) if built static(-only)?

FriBidi is being built static-only; iiuc it's using the regular fribidi package from MSys2, which ships both static and shared with one pc-file.

> Anyway, let’s just merge #493 (which makes it work in all cases)

If it gets actually tested before, sure.

> the problem would be `--enable-static --enable-shared`.

This is the default, without any flags needing to get passed.

> I guess I misunderstood Oneric’s reasoning for why the proposed patch is wrong.

Let me rephrase my argument and trim the additional digressions.

If we assume both omitting and setting `-DFRIBIDI_LIB_STATIC` depending on the FriBidi-build used is important, then the patch is wrong because it will create a shared libass library which can't link to shared fribidi – by default even and because furthermore it makes it *impossible* to build a static libass which can link to a shared FriBidi *(like our CI does)* without patching our build system. In this case a least-bad solution relates to the other bits omitted this time.

If only omitting is a problem but never setting the flag (as speculated in #493), then the patch is wrong because with `--disable-static` a shared libass-library is produced which cannot link to a static FriBidi, eventhough we expect both static and shared FriBidi to work and additionally the patch is overly complex about setting `-DFRIBIDI_LIB_STATIC`. In this case #493 would be the correct solution.

Issue #575, opened 2021-12-01
Rendering: the font cannot show


hooke007:
### Screenshots

mpv
![Snipaste_2021-12-01_19-44-18](https://user-images.githubusercontent.com/41094733/144228762-83fdbd3b-a47c-423a-aa1c-9cc97a1ad06e.jpg)
vsfilter

### Description of the issue

The Japanese words cannot be rendered correctly, just showed a point.
Check the specific font below, sub should start from 00:01:30.

### libass version

API version: 0x1502000 commit: f8c71aba5d14afcd6759507bad8a25037fb854b1

### ASS Sample

sub&font
[sample.zip](https://github.com/libass/libass/files/7633413/sample.zip)

### Special Fonts

I uploaded or linked to the required font

### System Information

win11
i7 10875H

### Log

![Snipaste_2021-12-01_19-29-49](https://user-images.githubusercontent.com/41094733/144227455-cad22c57-f9c6-41b4-8c57-3eab35b83399.jpg)

[mpv.log](https://github.com/libass/libass/files/7633453/mpv.log)

astiob:
Thanks for the detailed report!

I believe this is the same issue as #531: the font has no Unicode names or cmaps but instead has Shift-JIS ones, which libass doesn’t support (yet).

hooke007:
Thanks, I'll follow that issue and close it.

Issue #550, opened 2021-09-16
Take track language into account when performing font fallback


rcombs:
When rendering converted SRT (and also ASS without attached fonts, though I'm not entirely clear on how common this is for CJK), we'll pretty much always need to perform fallback for CJK characters. Currently, this means that we'll get a font for whichever language the font provider gives us, which may not be suitable for the subtitles' language due to Han unification. See for instance U+7D9A, which renders as <span lang="ja">続</span>, but <span lang="zh">続</span> in Chinese (note the top-left radical).

We can resolve this by taking the script's language (whether set by the script itself via our extension `Language` field, or by the consuming application via `track->Language`) when performing font fallback. On macOS, this means using `CTFontCreateForStringWithLanguage` instead of `CTFontCreateForString` (I tested a mockup of this and it works as expected); in Fontconfig, I believe it means setting `FC_LANG` on the `FcPattern` in `cache_fallbacks` (though we're currently explicitly _unsetting_ that for reasons I don't fully understand); on Windows, I believe it means passing the locale arg to `IDWriteFactory_CreateTextFormat`.

Does anyone know how this works in VSFilter? I'm guessing you ultimately get Arial Unicode MS, but what character variants within that font do you end up with? Is it just the OS language? If that's not a CJK language, what do you get?

I'm fairly comfortable differing from VSFilter here, especially if the VSFilter behavior is user-locale-dependent; I figure letting the script or calling application choose a language is probably a better guess than the user locale. I'd imagine the calling application behavior would be to pass in the language associated with the stream in the input file (or sidecar filename), or if that's not set, the UI language (which the calling application may be in a better position to know than us).

astiob:
First: when `Language` is set, I agree we should take it into account for font fallback (I’ve thought about this myself).

When it isn’t:

VSFilter’s effective font fallback seems to be locale-dependent, indeed. With a non-CJK locale, I’m getting a Japanese font, but I don’t know if this can be relied upon. The same font is selected in pure-GDI runs and in runs with forced complex shaping via Uniscribe (CJK ideographs alone don’t trigger it automatically).

[...]

PR #583, opened 2022-01-10
font: use iconv to handle non-unicode cmap microsoft fonts.

### LINKED ISSUE #531: Support non-Unicode Microsoft cmaps
problem occured on two traditional Chinese font. libass found the fonts, but can't render the charactors.
image
[logs](https://paste.ubuntu.com/p/zS6frpS7MW/) show that graphs of the charcodes coundn't be found.
[here](https://drive.google.com/file/d/1dTtgYB_TKwGZnLZc8VswXRSjtwQF-SUQ/view?usp=sharing) are the ass files and font files

Apache553:
To resolve #531.
In case of non-Unicode cmap fonts, convert unicode code point(a UCS-4 code unit) to corresponding multibyte charset bytes, then assemble those bytes into a uint32_t, assuming those bytes present a prefix-zero-bytes-omitted big-endian integer. Freetype can make use of the converted code point to find correct glyph by non-Unicode cmap.

It introduces a dependency of `libiconv`. On Windows, `WideCharToMultiByte` is available to avoid libiconv if needed.

And I have just tested this patch on samples from #531 and #575, which covers BIG5 and SHIFT-JIS. For other encodings, I think it should work but I can't confirm. So further testing is needed.

#531 testing screenshot: 
fix1

#575 testing screenshot: 
fix2

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -26,6 +26,7 @@
 #include FT_GLYPH_H
 #include FT_TRUETYPE_TABLES_H
 #include FT_OUTLINE_H
+#include FT_TRUETYPE_IDS_H
 #include <limits.h>
 
 #include "ass.h"
@@ -34,6 +35,156 @@
 #include "ass_fontselect.h"
 #include "ass_utils.h"
 #include "ass_shaper.h"
+
+/**
+ *  Get the mbcs codepoint from the output bytes of iconv/WideCharToMultiByte,
+ *  by treating the bytes as a prefix-zero-byte-omitted big-endian integer.
+ */
+static inline uint32_t pack_mbcs_bytes(const char *bytes, size_t length)
+{
+    uint32_t ret = 0;
+    for (size_t i = 0; i < length; ++i) {
+        ret <<= 8;
+        ret |= (uint8_t) bytes[i];
+    }
+    return ret;
+}
+
+/**
+ * Convert a UCS-4 code unit to a packed uint32_t in given multibyte encoding.
+ *
+ * We don't exclude Cygwin for Windows since we use WideCharToMultiByte only,
+ * this shall not violate any Cygwin restrictions on Windows APIs.
+ */
+#if defined(_WIN32)
+
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+
+static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
+{
+    // map freetype encoding to windows codepage
+    UINT codepage;
+    switch (encoding) {
+    case FT_ENCODING_MS_SJIS:
+        codepage = 932;
+        break;
+    case FT_ENCODING_MS_GB2312:
+        codepage = 936;
+        break;
+    case FT_ENCODING_MS_BIG5:
+        codepage = 950;
+        break;
+    case FT_ENCODING_MS_WANSUNG:
+        codepage = 949;
+        break;
+    case FT_ENCODING_MS_JOHAB:
+        codepage = 1361;
+        break;
+    default:
+        return codepoint;
+    }
+
+    WCHAR input_buffer[2];
+    size_t inbuf_size;
+
+    // encode surrogate pair, assuming codepoint > 0 && codepoint <= 10FFFF
+    if (codepoint >= 0x10000) {
+        // surrogate pair required
+        inbuf_size = 2;
+        input_buffer[0] = 0xD7C0 + (codepoint >> 10);
+        input_buffer[1] = 0xDC00 + (codepoint & 0x3FF);
+    } else {
+        inbuf_size = 1;
+        input_buffer[0] = codepoint;
+    }
+
+    // do convert
+    char output_buffer[2];
+    BOOL conversion_fail;
+    int output_length = WideCharToMultiByte(codepage, WC_NO_BEST_FIT_CHARS, input_buffer, inbuf_size,
+                                          output_buffer, sizeof(output_buffer), NULL, &conversion_fail);
+    if (output_length == 0 || conversion_fail)
+        return codepoint;
+
+    return pack_mbcs_bytes(output_buffer, output_length);
+}
+#elif defined(CONFIG_ICONV)
+
+#include <iconv.h>
+
+static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
+{
+    typedef struct { const char *names[5]; } EncodingList;
+
+    EncodingList encoding_list;
+
+    // map freetype encoding to iconv encoding
+    switch (encoding) {
+    case FT_ENCODING_MS_SJIS:
+        encoding_list = (EncodingList) {{"CP932", "SHIFT_JIS", NULL}};
+        break;
+    case FT_ENCODING_MS_GB2312:
+        encoding_list = (EncodingList) {{"CP936", "GBK", "GB18030", "GB2312", NULL}};
+        break;
+    case FT_ENCODING_MS_BIG5:
+        encoding_list = (EncodingList) {{"CP950", "BIG5", NULL}};
+        break;
+    case FT_ENCODING_MS_WANSUNG:
+        encoding_list = (EncodingList) {{"CP949", "EUC-KR", NULL}};
+        break;
+    case FT_ENCODING_MS_JOHAB:
+        encoding_list = (EncodingList) {{"CP1361", "JOHAB", NULL}};
+        break;
+    default:
+        return codepoint;
+    }
+
+    // open iconv context
+    const char **encoding_str = encoding_list.names;
+    iconv_t cd = (iconv_t) -1;
+    while (*encoding_str) {
+        cd = iconv_open(*encoding_str, "UTF-32LE");
+        if (cd != (iconv_t) -1) break;
+        ++encoding_str;
+    }
+    if (cd == (iconv_t) -1)
+        return codepoint;
+
+    char input_buffer[4];
+    char output_buffer[2]; // MS-flavour encodings only need 2 bytes
+    uint32_t result = codepoint;
+
+    // convert input codepoint to little endian uint32_t bytearray
+    for (int i = 0; i < 4; ++i) {
+        input_buffer[i] = result & 0xFF;
+        result >>= 8;
+    }
+    result = codepoint;
+
+    // do actual convert, only reversible converts are valid, since we are converting unicode to something else
+    size_t inbuf_size = sizeof(input_buffer);
+    size_t outbuf_size = sizeof(output_buffer);
+    char *inbuf = input_buffer;
+    char *outbuf = output_buffer;
+    if (iconv(cd, &inbuf, &inbuf_size, &outbuf, &outbuf_size))
+        goto clean;
+
+    // now we have multibyte string in output_buffer
+    // assemble those bytes into uint32_t
+    size_t output_length = sizeof(output_buffer) - outbuf_size;
+    result = pack_mbcs_bytes(output_buffer, output_length);
+
+clean:
+    iconv_close(cd);
+    return result;
+}
+#else
+static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint) {
+    // just a stub
+    return codepoint;
+}
+#endif
 
 /**
  * Select a good charmap, prefer Microsoft Unicode charmaps.
@@ -79,7 +230,7 @@ void charmap_magic(ASS_Library *library, FT_Face face)
 
 /**
  * Adjust char index if the charmap is weird
- * (currently just MS Symbol)
+ * (currently all non-Unicode Microsoft cmap)
  */
 
 uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
@@ -87,12 +238,22 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
     if (!face->charmap)
         return symbol;
 
-    switch (face->charmap->encoding) {
-    case FT_ENCODING_MS_SYMBOL:
-        return 0xF000 | symbol;
-    default:
-        return symbol;
+    if (face->charmap->platform_id == TT_PLATFORM_MICROSOFT) {
+        switch (face->charmap->encoding) {
+        case FT_ENCODING_MS_SYMBOL:
+            return 0xF000 | symbol;
+        case FT_ENCODING_MS_SJIS:
+        case FT_ENCODING_MS_GB2312:
+        case FT_ENCODING_MS_BIG5:
+        case FT_ENCODING_MS_WANSUNG:
+        case FT_ENCODING_MS_JOHAB:
+            return convert_unicode_to_mb(face->charmap->encoding, symbol);
+        default:
+            return symbol;
+        }
     }
+
+    return symbol;
 }
 
 static void set_font_metrics(FT_Face face)
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -82,7 +82,7 @@ static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
         codepage = 1361;
         break;
     default:
-        return codepoint;
+        return 0;
     }
 
     WCHAR input_buffer[2];
@@ -105,7 +105,7 @@ static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
     int output_length = WideCharToMultiByte(codepage, WC_NO_BEST_FIT_CHARS, input_buffer, inbuf_size,
                                           output_buffer, sizeof(output_buffer), NULL, &conversion_fail);
     if (output_length == 0 || conversion_fail)
-        return codepoint;
+        return 0;
 
     return pack_mbcs_bytes(output_buffer, output_length);
 }
@@ -137,7 +137,7 @@ static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
         encoding_list = (EncodingList) {{"CP1361", "JOHAB", NULL}};
         break;
     default:
-        return codepoint;
+        return 0;
     }
 
     // open iconv context
@@ -149,18 +149,18 @@ static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
         ++encoding_str;
     }
     if (cd == (iconv_t) -1)
-        return codepoint;
+        return 0;
 
     char input_buffer[4];
     char output_buffer[2]; // MS-flavour encodings only need 2 bytes
     uint32_t result = codepoint;
 
-    // convert input codepoint to little endian uint32_t bytearray
+    // convert input codepoint to little endian uint32_t bytearray,
+    // result becomes 0 after the loop finishes
     for (int i = 0; i < 4; ++i) {
         input_buffer[i] = result & 0xFF;
         result >>= 8;
     }
-    result = codepoint;
 
     // do actual convert, only reversible converts are valid, since we are converting unicode to something else
     size_t inbuf_size = sizeof(input_buffer);
@@ -182,7 +182,8 @@ clean:
 #else
 static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint) {
     // just a stub
-    return codepoint;
+    // we can't handle this cmap, fallback
+    return 0;
 }
 #endif
 
@@ -578,7 +579,9 @@ int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
 
     for (i = 0; i < font->n_faces && index == 0; ++i) {
         face = font->faces[i];
-        index = FT_Get_Char_Index(face, ass_font_index_magic(face, symbol));
+        index = ass_font_index_magic(face, symbol);
+        if (index)
+            index = FT_Get_Char_Index(face, index);
         if (index)
             *face_index = i;
     }
@@ -592,14 +595,19 @@ int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
         face_idx = *face_index = add_face(fontsel, font, symbol);
         if (face_idx >= 0) {
             face = font->faces[face_idx];
-            index = FT_Get_Char_Index(face, ass_font_index_magic(face, symbol));
+            index = ass_font_index_magic(face, symbol);
+            if (index)
+                index = FT_Get_Char_Index(face, index);
             if (index == 0 && face->num_charmaps > 0) {
                 int i;
                 ass_msg(font->library, MSGL_WARN,
                     "Glyph 0x%X not found, broken font? Trying all charmaps", symbol);
                 for (i = 0; i < face->num_charmaps; i++) {
                     FT_Set_Charmap(face, face->charmaps[i]);
-                    if ((index = FT_Get_Char_Index(face, ass_font_index_magic(face, symbol))) != 0) break;
+                    index = ass_font_index_magic(face, symbol);
+                    if (index)
+                        index = FT_Get_Char_Index(face, index);
+                    if (index) break;
                 }
             }
             if (index == 0) {
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -283,7 +283,9 @@ get_glyph_nominal(hb_font_t *font, void *font_data, hb_codepoint_t unicode,
     FT_Face face = font_data;
     struct ass_shaper_metrics_data *metrics_priv = user_data;
 
-    *glyph = FT_Get_Char_Index(face, ass_font_index_magic(face, unicode));
+    *glyph = ass_font_index_magic(face, unicode);
+    if (*glyph)
+        *glyph = FT_Get_Char_Index(face, *glyph);
     if (!*glyph)
         return false;
 
@@ -300,7 +302,9 @@ get_glyph_variation(hb_font_t *font, void *font_data, hb_codepoint_t unicode,
     FT_Face face = font_data;
     struct ass_shaper_metrics_data *metrics_priv = user_data;
 
-    *glyph = FT_Face_GetCharVariantIndex(face, ass_font_index_magic(face, unicode), variation);
+    *glyph = ass_font_index_magic(face, unicode);
+    if (*glyph)
+        *glyph = FT_Face_GetCharVariantIndex(face, *glyph, variation);
     if (!*glyph)
         return false;
 
@@ -794,7 +798,9 @@ static void shape_fribidi(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
         GlyphInfo *info = glyphs + i;
         FT_Face face = info->font->faces[info->face_index];
         info->symbol = shaper->event_text[i];
-        info->glyph_index = FT_Get_Char_Index(face, ass_font_index_magic(face, shaper->event_text[i]));
+        info->glyph_index = ass_font_index_magic(face, shaper->event_text[i]);
+        if (info->glyph_index)
+            info->glyph_index = FT_Get_Char_Index(face, info->glyph_index);
     }
 
     free(joins);


astiob:
Thanks!

I think we want to avoid a dependency on iconv, plus its code pages tend to be slightly different from Microsoft’s. But this is definitely very good progress, as we finally have an idea of what conversions actually need to be done.

astiob:
Oh, and each platform’s iconv has its own set of code pages and names for them.

Apache553:
> I think we want to avoid a dependency on iconv

After I read `configure.ac` later, it turns out that it seems we already has iconv as a dependency... Ignore my lines maybe.

> its code pages tend to be slightly different from Microsoft’s

I did some search and read the source code of libiconv(from GNU and Apple). I changed those encoding names to "CPxxxx" names, which means the microsoft variation charsets according to several lines of comments in the source code. Except the "Johab" encoding(CP1361), the source code did't mention. But it seems that's the only option for that encoding.

> each platform’s iconv has its own set of code pages and names for them.

I borrowed a MacBook from my roommate :-D, then I found those "CPxxxx" names exists on both macos and linux. And the windows version is built with mingw(maybe?), I expect it should be the same as linux.

astiob:
> After I read `configure.ac` later, it turns out that it seems we already have iconv as a dependency...

Right; I’d forgotten about that. But it’s an optional dependency used only if the consumer explicitly asks us to transcode subtitles.

But I suppose this could be sufficient as a first step. In that case, please add `#ifdef CONFIG_ICONV` guards though.

> I changed those encoding names to "CPxxxx" names, which means the microsoft variation charsets according to several lines of comments in the source code

That’s a good change. But those comments and intentions are not always fully reliable :-) Maybe the situation is better with Microsoft encodings, but I know the Macintosh encodings were definitely out of date when I last checked.

> I found those "CPxxxx" names exists on both macos and linux

I think macOS and most Linux distributions all use GNU iconv. But other implementations exist e. g. on FreeBSD and NetBSD, and they support different sets of encodings and name them differently. Unfortunately, there is no single standard name list.

Luckily, it looks like `CP936` `CP949` `CP950` exist on [FreeBSD](https://github.com/freebsd/freebsd-src/tree/main/share/i18n/esdb/CP) and [NetBSD](https://github.com/NetBSD/src/tree/trunk/share/i18n/esdb/CP), but `CP1361` doesn’t; and there may be (likely are) yet other implementations of iconv out there. Even the encodings that do exist here haven’t been updated in [19 years](https://github.com/NetBSD/src/tree/trunk/share/i18n/csmapper/CP), so there’s a decent chance Microsoft has changed them since then.

astiob:
Probably not useful, but I’ll just leave this here: there’s a [debug-fonts](https://github.com/astiob/libass/commits/debug-fonts) branch in my personal working repository that uses iconv, Windows and macOS native mechanisms to decode font _names._ (For some reason I skipped Japanese and Johab… Johab probably because I wasn’t sure what encoding that was, but Japanese I don’t remember why.)

Apache553:
> In that case, please add `#ifdef CONFIG_ICONV` guards though.

OK, I will do this. If iconv is present then enable the iconv part, else not. 

> Luckily, it looks like `CP936` `CP949` `CP950` exist on [FreeBSD](https://github.com/freebsd/freebsd-src/tree/main/share/i18n/esdb/CP) and [NetBSD](https://github.com/NetBSD/src/tree/trunk/share/i18n/esdb/CP), but `CP1361` doesn’t;

Seems I should change `CP1361` to `JOHAB`,which may have better compatibility across those iconv implementations. At least I can find the name in directory trees of [FreeBSD](https://github.com/freebsd/freebsd-src/tree/main/share/i18n/csmapper/KS) and [NetBSD](https://github.com/NetBSD/src/tree/trunk/share/i18n/csmapper/KS). And in case of GNU iconv, it turns out that  `CP1361` is an alias of `JOHAB`.

> Even the encodings that do exist here haven’t been updated in [19 years](https://github.com/NetBSD/src/tree/trunk/share/i18n/csmapper/CP), so there’s a decent chance Microsoft has changed them since then.

Yes I think this is very possible to happen. But as for Microsoft's habit of backward-compatibility, and those encodings are very old, and iconv is a very widely used library, I think we don't need to worry about that.

TheOneric:
>  there may be (likely are) yet other implementations of iconv out there.

Haiku seems to use GNU's libiconv by default.


musl has a smaller set of encoding, but appears to support all but `JOHAB`/`CP1361` of the used sets ([ref](https://git.musl-libc.org/cgit/musl/tree/src/locale/iconv.c#n38)).


illumos has its own implementation. In `iconv -l`'s output only `BIG5` and `GB1830` appear from the ones currently used in the patches. However, the manual page listing which combinations are supported for conversions doesn't list the mat all. In the best case the docs are just outdated and at least two encoding work. ([ref](https://github.com/illumos/illumos-gate/blob/9ecd05bdc59e4a1091c51ce68cce2028d5ba6fd1/usr/src/man/man5/iconv.5)).
```sh
# from iconv -l
zh_TW-big5 (BIG5, BIG-5, BIG_5, zh_TW.big5),
zh_CN.gbk (GB18030, GB-18030, GB_18030, BGK, zh_CN.gb18030),
```

---

If this is the correct place to look at uclibc doesn't support any of the used encoding. ([looked here](https://git.uclibc.org/uClibc/tree/extra/locale/charmaps?id=d46dc8bc88e38251bfa3712efe7abf62933f5419)).

bionic-libc (used in Android) only supports ASCII and UTF encodings ([ref](https://android.googlesource.com/platform/bionic/+/e2c784f15912115f778e427f0ce9a21189a3b494/libc/include/iconv.h#n63)).

TheOneric:
> [convert Unicode to] multibyte charset bytes, then assemble those bytes into a uint32_t, assuming those bytes present a prefix-zero-bytes-omitted big-endian integer.

Are you sure that this — using the *multi-byte text encoding* for lookups — is actually how it is supposed to work? Without any prior knowledge, I'd expect that the *codepoints* of a character set would be used for lookups, just as with Unicode. Eg for `ア` the JIS codepoint is represented by the two bytes `0x25 0x22` but the *Shift-JIS-encoded text* form is `0x83 0x41`.

Apache553:
> Are you sure that this — using the _multi-byte text encoding_ for lookups — is actually how it is supposed to work?

Actually, the _multi-byte text encoding_ is the codepoint, at least for GBK, BIG5 and SHIFT-JIS (I don't know much about UHC and Johab, but I can speculate this is way how those charsets work.). The multibyte seqence is usually the prefix-zero-bytes-omitted big-endian integer which represents the codepoint directly. JIS and SHIFT-JIS are two charsets. SHIFT-JIS don't encode JIS codepoints like UTF-8 does to Unicode(UCS-4). The codepoint of `ア` in Shift-JIS is 0x8341. So I think this is how it works.

TheOneric:
> Actually, the multi-byte text encoding is the codepoint (numerically equal), at least for GBK, BIG5 and SHIFT-JIS.
> […] SHIFT-JIS doesn't encode JIS codepoints

I don't know about GBK, BIG5 and the others, but SHIFT-JIS definitely is a *text encoding* for JIS-codepoints, which can be  different from the encoded bytes.
SHIFT-JIS encodes the union of the *charsets* [JIS X 0213](https://www.asahi-net.or.jp/~AX2S-KMTN/ref/jisx0213/index.html) (originally: [JIS X 0208](https://web.archive.org/web/20150317193520/http://kikaku.itscj.ipsj.or.jp/ISO-IR/042.pdf)), in which all characters use two bytes with the first byte not zero, and the single byte [JIS X 0201](https://en.wikipedia.org/wiki/JIS_X_0201#Codepage_layout). SHIFT-JIS maps JIS X 0213 *codepoints* into two bytes where the first byte lies in the unused space of JIS X 0201; so only for the single byte JIS X 0201 characters the encoding is identical to the codepoint; see eg [Wikipedia](https://en.wikipedia.org/wiki/Shift_JIS#Shift_JISx0213_and_Shift_JIS-2004) for the mapping function.
EUC-JP would be another *text encoding* for JIS X 0213 *and* JIS X 0201 in which only the 7bit codepoints (=ISO646-JP) are indetical to their encoding.
*(For reference: ISO646-JP has latin characters and is except for two characters identical to ASCII. JIS X 0201 extends this to 8bits with (half-width) Katakana. JIS X 0213 has fullwidth characters, including Kanji.)*

Of course, it is possible, that fonts use SHIFT-JIS-text-encoding and not the JIS-codepoints to identify characters. But it would be good to have a confirmation either way. If it does use Shift-JIS-encoded, I wonder if there are also fonts using EUC-JP-encoded text for characters or if EUC-JP had to be reencoded to Shift-JIS for rendering.

astiob:
> But it would be good to have a confirmation either way.

That’s precisely what the screenshots in this PR provide. Of course, it wouldn’t hurt _still_ to cross-check this against FontForge source code or some such, as I was planning to do from the beginning, but if this works well enough for the screenshots, then it seems quite right.

> I wonder if there are also fonts using EUC-JP-encoded text for characters

Precisely the following [Microsoft font encodings](https://docs.microsoft.com/en-us/typography/opentype/spec/cmap#windows-platform-platform-id--3) exist:

* Symbol
* Unicode
* “ShiftJIS”
* “PRC”
* “Big5”
* “Wansung”
* “Johab”

TheOneric:
> > But it would be good to have a confirmation either way.
>
> That’s precisely what the screenshots in this PR provide

Ah, #575 is a Shift-JIS sample. And also confirmed to render the same in a VSFilter, so it's unlikely the particular font is just borked in an odd way. Good.

TheOneric:
> Luckily, it looks like CP936 CP949 CP950 exist on FreeBSD and NetBSD, but CP1361 doesn’t

On NetBSD 9.2:
```
$ iconv -l | grep -Ei 'johab|1361'
cp1361
johab
```
```

**Comment by Apache553** [comment]
> Still, musl and illumos do not recognise the `CPXXX` names of some supported encodings. Perhaps we'll need to try multiple names _(or try to find names recognised by all implementations we looked at and hope those name will also be recognised in all other iconv implementations supporting the relevant encoding but we didn't consider)_.

For musl, its [documentation](https://wiki.musl-libc.org/functional-differences-from-glibc.html#iconv) says that it can only convert **from** those East-Asian charsets, but not convert **to**.

For illumos, I have just checked its output of `iconv -l` on OmniOS 11. It looks like that it don't support Microsoft code pages. And there's no support for SHIFT-JIS, Wansung and Johab. 

I think it might be better to let the user to handle those conversions by providing a better iconv or writting their own code to use something capable. It seems there's nothing we can do on such platforms.

**Comment by TheOneric** [comment]
> For musl, its [documentation](https://wiki.musl-libc.org/functional-differences-from-glibc.html#iconv) says that it can only convert from those East-Asian charsets, but not convert to.

As the linked wiki page writes, musl *can* convert **to** Shift-JIS, which is one of the encoding used in this patch but not recognised by the `CPXXX` name. Future versions may expand the support.

> For illumos, I have just checked its output of iconv -l on OmniOS 11. It looks like that it doesn't support Microsoft code pages.

Not by the Microsoft codepage ids, exactly; but two of the used encoding are supported.  That's why the name for them should fall back to one recognised by illumos, see [my previous post](https://github.com/libass/libass/pull/583#issuecomment-1009279438) (which was updated). I can also confirm, that of [illumos-966c588f82](https://github.com/illumos/illumos-gate/commit/966c588f828c4693b5a80ec27b24ff16afa11ec9) the encodings are supported as targets from an UTF-8 source.

> It seems there's nothing we can do on such platforms.

For musl and illumos there is: (also) use the names understood by the native iconv implementation. Some encodings working is better than none. There really isn't anything we can do on uclibc and bionic-libc though.

> providing a better iconv

When building libass (or in some cases via preloading at startup) it is possible to use GNU-libiconv instead of the native iconv on many platforms. But we shouldn't *require* this. Rather strive to utilise the various native iconvs as much as possible if the effort is reasonable; adding an alternative name is imho reasonable.

**Comment by Apache553** [comment]
> adding an alternative name is imho reasonable.

OK now it will try a list of names to open iconv context. Since it looks like musl supports only UTF8 and there's no alias in mingw/msys iconv for UTF8 (On my computer, mingw/msys supports only "UTF-8" but not "UTF8"). I also implemented a UCS4 -> UTF8 encoding function and a list of "UTF8 names" to try out.

Did I miss something?

**Comment by TheOneric** [comment]
Thanks! Regarding the input encoding, I think all iconv implementations we looked at supported `UCS-4LE` – except bionic where `UTF-32LE` would work but none of the other relevant encodings. With regard to naming I think a single name is fine for the *input*.  

There's another issue though: ... the input encoding isn't always little endian, but rather the platforms native endianess. GNU and musl have an `WCHAR_T` encoding which for platforms with 32bit `wchar_t` will be equal to “native endian UTF32”. It may not be a good idea to rely on a specific `wchar_t` size though and illumos doesn't know about `WCHAR_T` *(haven't checked NetBSD and the rest)*.
We'll want to check the native endianess somehow *(how portable are predefined macros here?)* and choose the input encoding name based on that.

**Comment by TheOneric** [comment]
I guess we can use the same check for big-endian as wyhash:
```C
#if defined(__BIG_ENDIAN__) || (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)
```
and otherwise assume little-endian.

Apache553:
> There's another issue though: ... the input encoding isn't always little endian

There's a little loop:
```
for (int i = 0; i < 4; ++i) {
    input_buffer[i] = result & 0xFF;
    result >>= 8;
}
```
`result` is the input UCS-4 codepoint at the beginning.
This should make `input_buffer`'s content always little endian bytes no matter what the host is.

TheOneric:
Sorry, forgot about the loop. Unless this turns out to be performance relevant, it is a suitable and simple solution to the endianess.

TheOneric:
Apparently some software uses `CP…` names to refer to variants with IBM's instead of Microsoft's vendor extensions; eg ICU supposedly uses `CP949` for `IBM949`. There exist `MSCP…`, `MS…` and `WINDOWS-…` names to unambigously refer to Microsoft variants, but which one of those names exists differs by codepage and implementation.
Below is an overview of which names are supported where. Could someone check what Apple supports? And if it isn't using GNU libiconv also what's available on MS Windows with MinGW?
If we're concerned about vendor extensions, it appears like the names preferred by IANA are supported by all implementations *(Johab doesn't have a IANA designation)*. Otherwise trying a (Microsoft) codepage name first, then IANA name, then subsets seems like the best strategy.

```C
/*
 * y: supported;  s: supported only as source;  n: unsupported;
 * .: untested/unknown;  ?: known name, unknown if MS, IBM or none
 *
 *                                |   GNU   | FreeBSD | NetBSD | illumos |  musl  |  Apple  |  MinGW
 * -------------------------------+---------+---------+--------+---------+--------+---------+---------
 * MS932/Shift_JIS/CP932          |  y/y/?  |  y/y/y  | y/y/y  |  n/n/n  | n/y/n  |         |
 * CP936/GB18030/GB2312/GBK       | y/y/y/y | y/y/y/y | y/y/y/y| n/y/n/y |n/s/s/s |         |
 * CP950/Big5                     |   y/y   |  y/y    |  y/y   |   n/y   |  s/s   |         |
 * MSCP949/CP949/EUC-KR           |  y/y/y  |  y/y/y  | y/y/y  |  n/n/n  | n/s?/s |         |
 * CP1361/Johab                   |   y/y   |   y/y   |  y/y   |   n/n   |  n/n   |         |
 * UTF-32LE/UCS-4LE               |   y/y   |   y/y   |  y/y   |   y/y   |  y/y   |         |
 *
 *  GNU is used by glibc-systems, Haiku and whenever GNU libiconv overrides the native iconv.
 *  uclibc only supports conversion to and from UTF-8 and none of the non-Unicode encodings of interest
 *  bionic only supports Unicode, ASCII and WCHAR_T and not the UCS names.
 * 
 * Alternatives: CP936:   WINDOWS-936 supported only by GNU and FreeBSD (not NetBSD)
 *                        MS936 only by GNU
 *                        MSCP936 supported by NetBSD and FreeBSD
 *               CP1361:  MSCP1361 supported by GNU and NetBSD (not FreeBSD)
 *               CP950:   MSCP950 supported only by NetBSD and FreeBSD
 */

```

libass/ass_font.c
```diff
@@ -35,6 +42,197 @@
...
+        break;
+    case FT_ENCODING_MS_WANSUNG:
+        encoding_str = WANSUNG_ENCODING;
+        break;
+    case FT_ENCODING_MS_JOHAB:
+        encoding_str = JOHAB_ENCODING;
+        break;
+    default:
+        return codepoint;
+    }
```
TheOneric, line 187: Imho it would be more readable to inline the name list here in the switch-case inside the function. E.g. like this *(names not final)* and then use a `for`-loop later:
```C
    typedef struct {size_t n; const char* names[5];} Enclist;
    Enclist tarenc;

    // map freetype encoding to iconv encoding
    switch (encoding) {
    case FT_ENCODING_MS_SJIS:
        tarenc = (Enclist){2, {"MS932", "SHIFTJIS"}};
        break;
    case FT_ENCODING_MS_GB2312:
        // IBM's codepage 936 is a whole different encoding, but rarely even supported
        tarenc = (Enclist){4, {"CP936", "GB18030", "GB2312", "GBK"}};
        break;
    case FT_ENCODING_MS_BIG5:
        tarenc = (Enclist){2, {"CP950", "BIG5"}};
        break;
    case FT_ENCODING_MS_WANSUNG:
        tarenc = (Enclist){3, {"MSCP949", "CP949", "EUC-KR"}};
        break;
    case FT_ENCODING_MS_JOHAB:
        tarenc = (Enclist){2, {"CP1361", "JOHAB"}};
        break;
    default:
        return codepoint;
    }
```

libass/ass_font.c
```diff
@@ -35,6 +42,197 @@
...
+
+static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
+{
+    uint32_t result = codepoint;
+    char input_buffer[4];
+    char output_buffer[4];
+    char* inbuf = (char*)&input_buffer;
+    char* outbuf = (char*)&output_buffer;
+    size_t inbuf_size = sizeof(input_buffer);
+    size_t outbuf_size = sizeof(output_buffer);
```
TheOneric, line 164: Some of the assignments could be moved closer to their first usage.

libass/ass_font.c
```diff
@@ -35,6 +42,197 @@
...
+
+    // open iconv context
+    iconv_t cd = (iconv_t)-1;
+    while (*ucs4_str) {
+        while (*encoding_str) {
+            cd = iconv_open(*encoding_str, *ucs4_str);
+            if (cd != (iconv_t)-1) break;
+            ++encoding_str;
+        }
+        if (cd != (iconv_t)-1) break;
```
TheOneric, line 204: Style Nit: iirc we usually a space after a cast, so `(icont_t) -1`. Applies to all casts in this patch.

libass/ass_font.c
```diff
@@ -35,6 +42,197 @@
...
+        ++ucs4_str;
+    }
+    if (cd == (iconv_t)-1) {
+        return codepoint;
+    }
+
+    // do actual convert, only reversible converts are valid, since we are converting unicode to something else
+    if (iconv(cd, &inbuf, &inbuf_size, &outbuf, &outbuf_size)) {
+        goto clean;
+    }
```
TheOneric, line 214: Style Nit: superfluous `{}`

libass/ass_font.c
```diff
@@ -35,6 +42,197 @@
 #include "ass_utils.h"
 #include "ass_shaper.h"
 
+
+/**
+ * Convert a UCS-4 code unit to a packed uint32_t in given multibyte encoding
+ */
+#if defined(_WIN32)
```
TheOneric, line 49: We usually use `#if defined(_WIN32) && !defined(__CYGWIN__)`.
In this specific case using Microsoft's API might be simpler since we specifically want encodings with Microsoft-specific extensions, but [Cygwin's FAQ](https://cygwin.com/faq/faq.html#faq.programming.win32-api) alludes to some restrictions regarding Win32-API calls so not excluding Cygwin might be bad still.

libass/ass_font.c
```diff
@@ -35,6 +42,197 @@
...
+    // do convert
+    int output_size = WideCharToMultiByte(
+            codepage,
+            0,
+            input_buffer,
+            inbuf_size,
+            output_buffer,
+            4,
+            NULL,
+            &conversion_fail);
```
TheOneric, line 101: Please don't use a magic `4` but `sizeof(output_buffer)`. Also setting the `WC_NO_BEST_FIT_CHARS`-flag might be good.
Finally, I'd prefer the arguments to be placed on one or two lines, not one line per argument.

libass/ass_font.c
```diff
@@ -35,6 +42,197 @@
...
 
+
+/**
+ * Convert a UCS-4 code unit to a packed uint32_t in given multibyte encoding
+ */
+#if defined(_WIN32)
+static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
+{
+    uint32_t result = codepoint;
+    uint16_t input_buffer[2];
```
TheOneric, line 53: I believe `wchar_t` is the canonical type for `input_buffer` for use with Win32-APIs.

libass/ass_font.c
Apache553: The only used win32 api is `WideCharToMultiByte`, which is in `kernel32.dll`. According to the FAQ, the restrictions are applied to environment variables and filesystem path formats. As far as I know, this function neither get/sets environment variables nor calls any filesystem api (implicitly read from some internal databases maybe, but that should be safe). So I think not excluding Cygwin is OK.

astiob:
When you say “GNU”, do you mean glibc or libiconv? Because they are distinct, and each has its own iconv implementation. It seems your table shows glibc data. At the same time, I assume when you spoke of non-Linux systems using “GNU”, you actually meant libiconv.

On my outdated macOS, I have GNU libiconv 1.11, which supports:
* `CP932` but not `MS932`, as well as a distinct encoding with the names `MS_KANJI` `SHIFT-JIS` `SHIFT_JIS` `SJIS` `CSSHIFTJIS`—this one really confuses me because I can’t tell which of these two encodings is really the Microsoft one;
* `CP936` with the aliases `MS936` and `WINDOWS-936`, as well as the encodings `GBK`, `GB2312` and `GB18030`, but they are all distinct from each other;
* `CP950`, as well as `BIG5` with some aliases, but it is a distinct encoding;
* `CP949` with the alias `UHC` but not `MSCP949`, as well as `EUC-KR` with some alises, but that is a distinct encoding;
* `CP1361` with the alias `JOHAB`;
* `UTF-32LE` and `UCS-4LE` (but they are displayed as distinct encodings; not sure if this is true).

On my outdated Linux, `iconv -V` prints `iconv (GNU libc) 2.26` and lists the names in your table, but I can’t tell from `iconv -l` alone which of them are aliases. When I was working on the `debug-fonts` branch, I think I looked at the source code, but I’m writing this comment without doing that.

Microsoft encodings are not the same as “standard” Shift-JIS, GBK, BIG5 or EUC, if there even is a true standard. Name choice is significant. Of course, if an iconv implementation doesn’t support Microsoft’s variant, then we may as well fall back to another variant that’s slightly wrong but better than nothing; but the order in which we try names matters. Unless, of course, GNU libiconv’s `iconv -l` is lying about what are aliases and what are distinct encodings.

astiob:
Ultimately, I still think the proper thing to do would be to integrate the right encodings directly into libass. Using iconv for this is too messy, too unreliable and too slow, and requires an otherwise-optional dependency to correctly display some fonts. But having it with iconv is still much better than nothing, so I’d be happy to have it until that integration takes place some day.

TheOneric:
> When you say “GNU”, do you mean glibc or libiconv?

I tested glibc but assumed glibc and GNU libiconv share their character mappings; I will test a recent GNU libiconv later.
Do newer Apple systems use recent GNU libiconv, continue to use an old version or replaced it with Apple's own implementation?

> Microsoft encodings are not the same as “standard” Shift-JIS, GBK, BIG5 or EUC, if there even is a true standard

Exactly, that's why we can't just use the IANA names but have to deal with inconsistent Codepage names. There is a “true standard” for all but Johab; the previously linked IANA page links to or describes where to find the definitions.

> Name choice is significant. Of course, if an iconv implementation doesn’t support Microsoft’s variant, then we may as well fall back to another variant that’s slightly wrong but better than nothing; but the order in which we try names matters.

Yes, as I wrote [before](https://github.com/libass/libass/pull/583#issuecomment-1012621683) *“Otherwise trying a (Microsoft) codepage name first, then IANA name, then subsets seems like the best strategy.”* (the sentence preceeding the quoted one was missing a negation I now added).
To expand a bit on that, we could either:
 - try (multiple) unambigously Microsoft'ish names first (preferring the names most widely supported), then `CP…`, then IANA, then subsets.
 - try just `CP...`, speculating that it will be the Microsoft variant if any vendor extensions are supported, then IANA, then subsets.
 - not care about subsets

> `UTF-32LE` and `UCS-4LE` (but they are displayed as distinct encodings; not sure if this is true).

From glancing over the first couple search results, it appears to me like `UCS-4LE` does not care if a value is actually a defined valid Unicode code point and just encodes it with the same scheme used for `UTF-32`. But the latter only accepts valid Unicode codepoints. `UTF-32LE` is recognised by IANA, `UCS-4LE` is not, only `ISO-10646-UCS-4`(alias: `csUCS4`) which is Big-Endian.
For our purposes `UTF-32LE` should be good as text is supposed to only contain valid Unicode code points and as a IANA recognised name, `UTF-32LE` should also be supported by all iconv implementation who support Unicode encodings.

libass/ass_font.c
TheOneric: Can you add a comment to the code regarding why Cygwin is not included then?

TheOneric:
> I have GNU libiconv 1.11, which supports:

I now checked with GNU libivonc 1.16 (the currently newest release) and got the same results, including the same names missing compared to glibc. For reference, I tested with glibc 2.31.
I updated the table from my previous post to distinguish between glibc and GNU libiconv.

libass/ass_font.c
Apache553: I finally feel those names are totally a mess. So I decide to try `MSCP*`, `MS*`, `CP*` and any possible names repectively. I think is will be the simplest solution.

libass/ass_font.c
Apache553: Adding a space after a cast looks weird. And existing code doesn't have a space after cast, too.

TheOneric: Which variant looks “weird” is subjective. The existing code is admittedly not 100%-consistent, but in `ass_font.c` alone there are 7 instances of casts with a space but only 3 *(all inside one function) without a space. Over the whole codebase the lead of the space-variant is only increasing and e.g. in `ass_rasterizer.c` all casts use a space.
When new code is added we should increase consistency, not make it worse.

Apache553: OK I will do it

libass/ass_font.c
```diff
@@ -90,77 +95,80 @@ static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
...
-    for (size_t i = 0; i < output_size; ++i) {
+    for (size_t i = 0; i < outbuf_size; ++i) {
         result <<= 8;
         result |= (uint8_t)output_buffer[i];
     }
 
     return result;
 }
 #elif defined(CONFIG_ICONV)
+typedef struct { const char* names[8]; } EncodingList;
```
TheOneric, line 112: There's no reason to expose this typedef to the whole file's scope. Please move it inside the `convert_unicode_to_mb` function.

Apache553:
Also, I found another problem when handling conversion failure. Currently, I returns the input codepoint in case of failure. If the input character's unicode codepoint is valid in target charset, the render result may not be correct. For example, in script file, there's a line has character `U+82A0 芠`. Then the script requires the line to be rendered using a SHIFT-JIS cmap font. The result will be `U+3042 あ`, because `0x82A0` in SHIFT-JIS maps to `あ`. 

I think we need a mechanism to report conversion failures. Then we can find a fallback font to render or just let it become a tofu. I don't know how libass use `ass_font_get_index` function. It seems it is used in many places. Maybe returning `(uint32_t)-1` is a good choice. But I don't know if this will cause problem. Could anyone tell me please?

libass/ass_font.c
TheOneric: Sorry, I meant to say ”declarations“ not “assignments”. You did already move the assignments; can you also move the declarations?

libass/ass_font.c
Apache553: Done.

libass/ass_font.c
Apache553: Done.

astiob:
> I don't know how libass use `ass_font_index_magic` function. It seems it is used in many places. Maybe returning `(uint32_t)-1` is a good choice. But I don't know if this will cause problem. Could anyone tell me please?

It’s used in several places in `ass_render.c` and `ass_shaper.c`, always as part of the same pattern:

```c
glyph_index = FT_Get_Char_Index(face, ass_font_index_magic(face, codepoint));
```

The right thing would be to return 0 (probably better than −1, as we should never encounter NUL characters in source text and I assume all the encodings under consideration encode it as 0) from `ass_font_index_magic` and modify each use of it like this:

```c
glyph_index = ass_font_index_magic(face, codepoint);
if (glyph_index)
    glyph_index = FT_Get_Char_Index(face, glyph_index);
```

This way it will get mapped to “glyph not found” and skipped or tofu’d as appropriate.

Apache553:
> This way it will get mapped to “glyph not found” and skipped or tofu’d as appropriate.

No it doesn't work. I have just tested, `FT_Get_Char_Index` returned `1` when second parameter is `0` for font `麗流隷書`, which is a SHIFT-JIS cmap font. The character just renders to nothing.

TheOneric:
Some comments regarding the iconv encoding names:
 - `UHC` is not a name registered by IANA; it seems highly unlikely to me any software supporting the Windows mapping will recognise `UHC` but not one of the `…949` names.
 - English Wikipedia suggest the order of the `GB…` name should be `"GB18030", "GBK", "GB2312"` instead of the current order.
 - Regarding trying "any possible names", if this is the goal the `WINDOWS-…` names are missing.
    However trying all the names of which implementations often only support one or two will almost certainly lead to repeated failing attempts to create iconv handles for unknown names. As @astiob brought up performance as a drawback of the iconv appraoch with less names already, it might be better to cut down the list to only supported names.
    Going further, in the interest of performance, the `MSCP…`, `MS…` and `WINDOWS-…` names could all be dropped with the justification that the tested implementations were we could determine aliases, appear to all use the Microsoft mapping for the `CP…` names used in this patchset.
    @astiob: What's your preference?
    - Specifically there is no implementation we know about recognising `MS1361` or `WINDOWS-1361`.
 
On another topic, I think the first 4 commits should be squashed into a single one.
Potentially, it'd also be good to log which encoding name was used for conversion for debugging, but if performance is of concern maybe this is undesirable as it happens per codepoint rather than per font.

astiob:
Jumping ahead:

None of the font encodings under consideration allow code points longer than 2 bytes. (They’re not even directly supported in TrueType/OpenType cmaps, which is why UCS-4 cmaps need subtable hacks and dedicated encoding IDs.)

We may as well limit the `iconv` output buffer to 2 bytes and reject any input that finds this too short.

---

> Going further, in the interest of performance, the [extra “all possible”] names could all be dropped with the justification that tested implementations […] appear to [not benefit from them].

If my interpretation (see quote) is correct, this sounds desirable indeed. Then again, maybe we really just need to cache the iconv handles. Then again, maybe we just need to measure this before optimizing!

---

> tested implementations […] appear to all use the Microsoft mapping for the `CP…` names

I’ve just tested libiconv’s `CP932` vs `SHIFT_JIS`/`MS_KANJI`/etc. using VSFilter as a reference. I can confirm that `CP932` is the (only) one we want there.

(`CP932` accepts Unicode `\` and `~` and rejects Unicode `¥` and `‾`; this matches VSFilter. `SHIFT_JIS` does the opposite.)

---

> English Wikipedia suggest the order of the `GB…` name should be `"GB18030", "GBK", "GB2312"` instead of the current order. Microsoft's codepage 936 actually encodes GBK+vendor extension (since Win95), “GB2312” is supposedly a subset of “GBK” while “GB18030” is supposedly a superset.

My understanding is that `GBK` should be first but the differences are limited to:

* whether the euro sign is supported;
* whether the euro sign is encoded as a single byte;
* whether the encoder rejects Unicode code points that aren’t mapped at all in Microsoft’s CP936.

Then, it comes down to whether the font itself defines glyphs for the GB code points that are invalid in Microsoft’s CP936.

If we want to be really rigorous, we could add a post-processing step after `iconv` to convert the euro sign and (if desired) reject the invalid code points.

If we don’t, my logic is (without looking at the actual behaviour of various iconvs):

1. `GBK` has a decent chance to refer to Microsoft’s CP936 exactly.

2. If it doesn’t, it probably refers to GBK 1.0, which is a small strict superset of Microsoft’s CP936 minus the euro sign. That is, euro signs will be rejected and fail to render, and a small amount of Unicode Private Use Area code points will fail to be rejected but won’t have glyphs in an honest CP936 font anyway.

3. Even if it primarily targets GBK 1.0, there’s a chance that the Unicode→GB encoder supports the single-byte euro sign for compatibility with Microsoft’s CP936. It may also support Unicode code points first mapped in GB18030 and thus actually target GB18030 rather than GBK 1.0; see next point.

4. GB18030 is a larger superset that changes some GB→Unicode mappings (but we’re going in the other direction) and still lacks the single-byte euro sign.

    Explicitly requesting `GB18030` seems like it has a higher chance to use the standard double-byte encoding for the euro sign rather than Microsoft’s single-byte encoding.

    It also supports Unicode code points first mapped in GB18030, which should be safe for valid CP936-font text (because it won’t contain them) but will fail to reject newer Unicode code points if they happen to be used, thus making it more likely that a subtitle author who tests with libass on a platform where we pick `GB18030` accidentally uses such a nonportable code point. However, that would require the font to actually map a glyph to this GB18030 code point, which… seems unlikely, but I have no real way to judge this risk.

    It may or may not reject Unicode Private Use Area code points that were used in GBK 1.0 but are no longer used in GB18030. But they likewise don’t exist in Microsoft’s CP936, so it’s all the better if they’re rejected. But in any case, this benefit will be offset by the amount of newer non-PUA code points that fail to be rejected (see above), or the font won’t have these code points mapped in the first place.

    Finally, the GB18030 encoder can produce 4-byte output, which cannot be used in fonts.

5. GB2312 is indeed a subset of Microsoft’s CP936, so it will reject valid inputs and should be tried last.

TheOneric:
> >  Going further, in the interest of performance, the [extra “all possible”] names could all be dropped with the justification that tested implementations […] appear to [not benefit from them].
>
> If my interpretation (see quote) is correct, this sounds desirable indeed.

For clarification: the sentences was referring to dropping the extra, unambgiously Microsoft names, but keep the `CP…` name as the first choice, followed by the corresponding IANA name (lacking vendor changes) and then optionally subsets of the IANA name. This way we could at least mostly get it right where iconv supports the vendor-change-free variants like illumos and musl.

Now with a failed conversion leading to fontfallback instead of retrieving the wrong glyphs, there's also an argument to *only* try the `CP…` name. But then again there's no guarantee the fallback uses Unicode mapping, that the different metrics of the fallback don't mess things up, that a suitable fallback even exists or that there is a system font provider. So omitting the IANA names also has potential to make things worse.

> making it more likely that a subtitle author who tests with libass on a platform where we pick GB18030 accidentally uses such a nonportable code point [which is only present in GB18030 but not Microsoft's CP936 codepage]. However, that would require the font to actually map a glyph to this GB18030 code point, which… seems unlikely, but I have no real way to judge this risk.

I'd expect newly created fonts to all use Unicode. Even if the font-author insists on using a legacy encoding, the software used to create the font will map the characters to the correct Microsoft-flavour positions *(if it supports creating legacy-encoded fonts at all)*. So I'm not too concerned about this. But sure we can try `GBK` before `GB18030` if the euro sign doesn't work in either one anyway.

---

In conclusion, does this look like a good order to try and set of names? *(With caching iconv handles, translating the GB18030 euro sign and if caching is in place trying unambgious Microsoft-flavour names first as potential future improvements.)*
```C
 // CP... names are in theory ambigous and may also refer to IBM's variants rather than Microsoft's variants.
 // In practice tested implementations seem to use MS's variant at least for the encodings used here.
 switch (encoding) {
    case FT_ENCODING_MS_SJIS:
        encoding_list = (EncodingList){{"CP932", "SHIFT_JIS", NULL}};
        break;
    case FT_ENCODING_MS_GB2312:
        encoding_list = (EncodingList){{"CP936", "GBK", "GB18030", "GB2312", NULL}};
        break;
    case FT_ENCODING_MS_BIG5:
        encoding_list = (EncodingList){{"CP950", "BIG5", NULL}};
        break;
    case FT_ENCODING_MS_WANSUNG:
        encoding_list = (EncodingList){{"CP949", "EUC-KR", NULL}};
        break;
    case FT_ENCODING_MS_JOHAB:
        encoding_list = (EncodingList){{"CP1361", "JOHAB", NULL}};
        break;
    default:
        return codepoint;
 }
 
 // + changing the output buffer's size to 2
```

TheOneric:
> When I was testing Shift-JIS earlier, VSFilter displayed the font’s .undef glyph (?) for unencodable code points instead of trying font fallback.

Can confirm, that this happens on MS Windows 7 with the xy-VSF bundled with Aegisub-r8250, even when there is an applicable fallback. The Shift-JIS font is the one from #575, the other characters are provided by “DejaVu Sans” and “DejaVu Math TeX Gyre”.
U+00DF is part of JIS X 0213 but not JIS X 0208, not sure if it's part of Microsoft's codepage. The other two codepoints are not in JIS X 0213. Notice how for the character, whose codepoint is larger than 16bit, **two**(!) default glyphs appear instead.  
It would be interesting to know if later versions of MS Windows behave the same, especially with regards to codepoints outside the BMP.

xyVSF-aegiR8250_shiftjis

```
[Script Info]
; Script generated by Aegisub r8250
ScriptType: v4.00+
PlayResX: 640
PlayResY: 360
Video File: ?dummy:25.000000:40000:1280:720:47:191:225:

[V4+ Styles]
Style: kasi,麗流隷書,20,&H00FFFFFF,&HF0000000,&H00000000,&H8C000000,-1,0,0,0,100,100,0,0,1,0,3,8,30,15,6,1

[Events]
Dialogue: 0,0:01:47.68,0:01:54.41,kasi,NTP,0,0,0,,{\an8}君の隣でずっと変わらずに護るだろう\N(U+0001D49C Mathematical Script Capital A)\N(U+00DF Latin Small Letter Sharp S)\N(U+1E9E Latin Capital Letter Sharp S)
Dialogue: 0,0:01:47.68,0:01:54.41,kasi,NTP,0,0,0,,{\an2}君の隣でずっと変わらずに護るだろう\N{\fs80}𝒜 \N{\fs}ß | ẞ
Dialogue: 0,0:01:47.68,0:01:54.41,kasi,NTP,0,0,0,,{\an4\fn(DejaVu Math TeX Gyre)\fs80}𝒜
Dialogue: 0,0:01:47.68,0:01:54.41,kasi,NTP,0,0,0,,{\an6\fn(DejaVu Sans)}ß | ẞ
```

astiob:
I get this on up-to-date Windows 10 that lacks DejaVu fonts, with your script muxed into an MKV:

The lines set in the Shift-JIS font look the same as in Oneric’s screenshot: two default glyphs for the supplementary-plane U+1D49C and one each for the BMP code points. The DejaVu Sans line with the German BMP characters displays fine, but not in DejaVu Sans as it isn’t installed. The DejaVu Math TeX Gyre line with the supplementary-plane character displays two tofu boxes.

Note how the supplementary-plane character displays _two_ tofu boxes even when not using the Shift-JIS font. In fact, I can’t get the character to display even by explicitly naming a font that has a glyph for it.

This seems to be a bug in VSFilter specifically rather than GDI, although I don’t immediately understand how it is possible. It depends on where the subtitles come from:

Subtitle source | Renderer | Effect
-|-|-
MKV | XySubFilter | Two tofu for 𝒜 regardless of named font
MKV | MPC-HC\* | Two tofu for 𝒜 regardless of named font
External UTF-16 | XySubFilter | Correct glyph for 𝒜 even with font fallback; two dots in SJIS line
External UTF-16 | MPC-HC\* | Correct glyph for 𝒜 even with font fallback; two dots in SJIS line
External UTF-8 | XySubFilter | Correct glyph for 𝒜 even with font fallback; two dots in SJIS line
External UTF-8 with BOM | MPC-HC\* | Question marks and missing line, see screenshot below
External UTF-8 w/o BOM | MPC-HC\* | Mojibake

\* The MPC-HC I used for this test isn’t the most up-to-date version. It’s CCCP’s 1.7.8.162 (ff72737) (blitzkrieg). If desired, I can retest on clsid’s newest MPC-HC on another computer. (The XySubFilter is pinterf’s 3.2.0.804: not quite the latest one but almost up-to-date, and the change log since then has nothing about encodings.)

MPC-HC with UTF-8 external subtitles:

Both lines containing the supplementary-plane character show question marks, each in the font named on that line. The Shift-JIS line with the German BMP characters is entirely missing, as if it didn’t exist in the subtitle file.

I seem to remember seeing a commit in XySubFilter that precisely mentioned surrogate-plane characters in external UTF-8 subtitles. FWIW I vaguely remember it not being in xy-VSFilter. At any rate, that suggests there was indeed a bug that got fixed.

Best-case rendering:

As described above, the supplementary-character is visible in the line that doesn’t use a Shift-JIS font but there are still two dots in its place on the line that does.

> Can confirm that this happens […] even when there is an applicable fallback.

Unfortunately it’s not as easy as ensuring you have a font with that glyph. For example, I don’t have any DejaVu fonts on this computer, but I see both eszetts, so font fallback finds them. But if I change the declared font on that line to `MS PGothic`, the capital version turns into a tofu while the lowercase version merely changes font:

Screenshot as described.

I suspect GDI’s font fallback works on a codepoint-block basis or something. I can’t exclude the possibility that the Shift-JIS font we’re looking at declares that it supports—or that GDI thinks all Shift-JIS fonts support—_some_ code points from the Unicode block that contains the eszett and therefore doesn’t attempt fallback. This seems less likely for the capital eszett though, which lives in a completely different, far less common block.

Apache553:
> I suspect GDI’s font fallback works on a codepoint-block basis or something.

This [msdn](https://docs.microsoft.com/en-us/globalization/input/font-technology#font-linking) page says GDI uses _Font Linking_. I have done some testing with my MPC-HC(1.9.15.79-759de8ceb) with XySubFilter(3.2.0.806) for the case of declaring font to `MS PGothic`.

There's a registry in `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\FontLink\SystemLink` named `MS PGothic`. It has several filenames and fontface names pairs in it. On my computer, the content is:
```
MINGLIU.TTC,PMingLiU
SIMSUN.TTC,SimSun
GULIM.TTC,Gulim  (Not found on my computer)
YUGOTHM.TTC,Yu Gothic UI (Not found on my computer)
MSJH.TTC,Microsoft JhengHei UI
MSYH.TTC,Microsoft YaHei UI
MALGUN.TTF,Malgun Gothic
SEGUISYM.TTF,Segoe UI Symbol
```
I have checked all fonts listed here. It turns out that none of them contains `U+1E9E`. I think this is reason why it becomes a tofu in this case.

> I don’t have any DejaVu fonts on this computer, but I see both eszetts, so font fallback finds them.

I think this is the result of font linking, too. On my computer, GDI will choose SimSun(宋体) for non-existent fonts. In the font linking list of `SimSun`:
```
MICROSS.TTF,Microsoft Sans Serif,108,122
MICROSS.TTF,Microsoft Sans Serif
MINGLIU.TTC,PMingLiU
MSMINCHO.TTC,MS PMincho
BATANG.TTC,Batang
MSYH.TTC,Microsoft YaHei UI
MSJH.TTC,Microsoft JhengHei UI
YUGOTHM.TTC,Yu Gothic UI
MALGUN.TTF,Malgun Gothic
```
`Microsoft Sans Serif` is present, which is a font that contains `U+1E9E`. I have a screenshot for you to distinguish the source of the glyphs.

fontlink1

As you can see, `U+00DF`'s glyph is from `SimSun`, `U+1E9E`'s glyph is from `Microsoft Sans Serif`.

---

By the way, in my opinion I think it's good for libass to have a good font fallback like the one we already have. There's no need to imitate how XySubFilter (or something similar) does font fallback. I think few scripts will rely on font fallback to achieve their purposes. And something readable is always better than tofu or mojibake.

astiob:
Font linking may be involved in the process in some way, but it isn’t limited to that. On most computers, mine included, VSFilter/GDI defaults to Arial, so much so that fansubs actually [do depend on this behaviour](https://github.com/libass/libass/issues/42). (Are you sure yours is using SimSun? That’s actually worrying. How did you determine this?) On the same computer I did the above test on, Arial has no font linking defined. Yet GDI finds fallbacks for VSFilter for all sorts of glyphs that Arial lacks. FWIW I even tried `\fnArial` before posting the last comment and saw no difference from `\fnDejaVu...`

astiob:
> Are you sure yours is using SimSun? […] How did you determine this?

Disregard this. I’ve just remembered that I’ve seen it with my own eyes in https://github.com/libass/libass/issues/550#issuecomment-924289844.

Apache553:
> On most computers, mine included, VSFilter/GDI defaults to Arial, so much so that fansubs actually [do depend on this behaviour](https://github.com/libass/libass/issues/42).

I don't know that there's subtitle depends on this behavior before this... As for the default font, I think this is windows language related, mine is Windows 10,  Simplified Chinese. I wrote a tiny program that creates font with non-existent font name, creates dc, selects gdi object into dc then call `GetTextFace` to get what font it actually picked.



> FWIW I even tried `\fnArial` before posting the last comment and saw no difference from `\fnDejaVu...`

If I didn't make a mistake, Arial does have `U+00DF` and `U+1E9E`.

TheOneric:
>  [astiob:] I seem to remember seeing a commit in XySubFilter that precisely mentioned surrogate-plane characters in external UTF-8 subtitles. FWIW I vaguely remember it not being in xy-VSFilter.

Yep: https://github.com/Cyberbeing/xy-VSFilter/commit/fc4cc19b30a26fbf3c5983db111a24dd6ef7bc75

astiob:
And of course, in-string font fallback is also performed (seemingly via Arial, hence the ticket linked above, but this may well depend on the locale too) for the myriad of fonts fansubbers use, none of which have font linking defined.

> Arial does have U+00DF and U+1E9E.

Welp, that invalidates this test. But do note that `{\fnArial}CJK ideographs here` does display just fine, too.

I’ve done a new test to exclude font linking: I took MS PGothic, renamed it in FontForge, attached it to my MKV and referenced it in my subtitles. To demonstrate that some font fallback is happening, I changed the `\fn` of the mathematical script A to my custom font, too. I still get a plain eszett and a tofu, although the tofu looks different now:

Screenshot as described.

FWIW the eszett comes from MS PGothic itself. And while it’s still only a guess, my guess about Unicode blocks is corroborated by the fact the font does contain glyphs for U+1E80 to U+1E85. The different tofu may perhaps be a result of font linking somehow causing a different font to be used for the tofu glyph itself.

---

At any rate, the Shift-JIS line consistently displays two dots for the supplementary-plane character.

Maybe we need, for testing, a custom font that uses (only) a non-Unicode cmap and limits its glyph repertoire to the bare minimum. But I’ve just tried to create one in FontForge and failed: it seems to unconditionally write Unicode cmaps.

On the other hand, I’ve managed to convert the Shift-JIS font from #575 to a Unicode cmap by reexporting it from FontForge. It apparently messed up the font in several regards, but hiragana/kanji still look the same, though with weird metrics. And lo and behold, the mathematical script A started showing up in VSFilter.

TheOneric:
I've split the font-fallback on conversion failures commit from the others again and while at it reduced the output buffer size to 2 chars; otherwise the same as before: [old diff to unsquashed version](https://github.com/libass/libass/compare/ecdd7b18a1c33c443ba187edd3cb5199355ee9fb..a542c01717887e2106aa1d659431e12c64e7234d) which also squashed the font fallback commit, [new diff to unsquashed version](https://github.com/libass/libass/compare/ecdd7b18a1c33c443ba187edd3cb5199355ee9fb..77a4cdf1ac1e8051919ff118e932b223ee8c0d7e) only squashing the first 4 commits.

libass/ass_font.c
astiob: FWIW the _actual_ canonical type for Win32 is `WCHAR`, but we use `wchar_t` in libass. This distinction doesn’t matter on native Win32, but on midipix, these types are different. Of course, midipix still has the chance to never take off.

libass/ass_font.c
```diff
@@ -26,15 +26,165 @@
...
+    wchar_t input_buffer[2];
+    size_t inbuf_size;
+
+    // encode surrogate pair, assuming codepoint > 0 && codepoint <= 10FFFF
+    if (codepoint >= 0x10000) {
+        // surrogate pair required
+        inbuf_size = 2;
+        uint32_t code = (codepoint - 0x10000);
+        input_buffer[0] = 0xD800 | (code >> 10);
+        input_buffer[1] = 0xDC00 | (code & 0x3FF);
```
astiob, line 88: This isn’t wrong, but you can do shorter:
```c
// no intermediate uint32_t code
input_buffer[0] = 0xD7C0 + (codepoint >> 10);
input_buffer[1] = 0xDC00 + (codepoint & 0x3FF);
```

libass/ass_font.c
```diff
@@ -26,15 +26,165 @@
...
+    } else {
+        inbuf_size = 1;
+        input_buffer[0] = codepoint;
+    }
+
+    // do convert
+    char output_buffer[4];
```
astiob, line 95: This can be reduced to 2 bytes, too.

libass/ass_font.c
```diff
@@ -26,15 +26,165 @@
...
+        result <<= 8;
+        result |= (uint8_t) output_buffer[i];
+    }
+
+    return result;
+}
+#elif defined(CONFIG_ICONV)
+static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint)
+{
+    typedef struct { const char* names[8]; } EncodingList;
```
astiob, line 113: Another style nit, here and below: libass prefers to put the space before the asterisks, like so:

```c
const char *names[8]
const char **encoding_str
char *inbuf
char *outbuf
```

astiob, line 113: This array’s size can be reduced if we’re no longer using that many names. But maybe it doesn’t matter.

libass/ass_font.c
```diff
@@ -87,12 +237,22 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
     if (!face->charmap)
         return symbol;
 
-    switch (face->charmap->encoding) {
-    case FT_ENCODING_MS_SYMBOL:
-        return 0xF000 | symbol;
-    default:
-        return symbol;
+    if (face->charmap->platform_id == TT_PLATFORM_MICROSOFT) {
```
astiob, line 240: This is a good and innocent change (bug fix), but a nonobvious change nevertheless. Might be worth mentioning in the commit message, at least?

astiob:
FWIW:

> I'd expect newly created fonts to all use Unicode.

Yes, but (a) GB18030 has been around for quite a while; (b) font authors never do only what you expect. :-) Much like subtitle authors!

> Even if the font-author insists on using a legacy encoding, the software used to create the font will map the characters to the correct Microsoft-flavour positions

I’m confused and worried by FontForge’s list of options in the Encodings menu, which has multiple standards for each language and then separately “cp932”. On the one hand, (modern) FontForge always writes a Unicode cmap, so this doesn’t matter. On the other, this doesn’t instil confidence that other software, or even older FontForge, stick to Microsoft’s encodings and enforce them precisely.

Still don’t expect this to matter _much_ though.

astiob (approved):
The current/new code looks good to me. I’ve finally looked through the whole patch, too. Found a few more nits, but nothing major.

libass/ass_font.c
```diff
@@ -26,15 +26,165 @@
...
+    if (iconv(cd, &inbuf, &inbuf_size, &outbuf, &outbuf_size))
+        goto clean;
+
+    // now we have multibyte string in output_buffer
+    // assemble those bytes into uint32_t
+    size_t output_bytes = sizeof(output_buffer) - outbuf_size;
+    for (size_t i = 0; i < output_bytes; ++i) {
+        result <<= 8;
+        result |= (uint8_t) output_buffer[i];
+    }
```
astiob, line 174: Would it be worth it to extract this little loop into its own function and reuse it in both implementations of `convert_unicode_to_mb`? Probably not significantly less code, but maybe a little less opportunity for bugs?

libass/ass_font.c
TheOneric: Prior to the patches at hand `wchar_t` is only used in `ass_directwrite.c`, which should probably all use `WCHAR` instead.  `dwrite_c.h` already uses `WCHAR` and #573 would currently add more uses of `wchar_t`. Perhaps we should use `WCHAR` where applicable in all new code and also update `ass_directwrite.c`?

astiob: Yeah, because we don’t currently use Win32 APIs anywhere but in `ass_directwrite.c`. `dwrite_c.h` copies Microsoft’s declarations fairly verbatim, so that explains that.

> Perhaps we should use `WCHAR` where applicable in all new code and also update `ass_directwrite.c`?

Yes, that’s an option. Indeed, that’s the “right” thing to do, probably.

Apache553: > these types are different

If `WCHAR` and `wchar_t` are different, the whole module is likely to break. Since the library functions like `wcsncmp` expect `wchar_t` and Windows APIs expect `WCHAR`. I think we need to place a warning sign here to reminder anyone who want to compile this project on midipix. They will need to use utf-16 variant of the library functions provided by midipix. Or we will need some ifdef-endif to select functions.

astiob: You mean `ass_directwrite`, right? You’re not using any C library `wchar_t` functions in this PR, are you?

But yes. Good catch. `ass_directwrite` does use `wcsncmp` once. But it can be replaced by a call to `CompareStringW` or `CompareStringEx`, and I don’t immediately see any other `wcs...` calls. But to be safe, a more careful audit of the code would be in order if we were to actually do this.

Apache553: However, `CompareStringOrdinal` is available since Vista. Maybe providing our own version of `wcsncmp` would be better, if we don't want to drop Windows XP.

astiob: Right, `CompareStringOrdinal` is even better; I overlooked it.

On the one hand, DirectWrite itself is only available on Vista+, so we can safely call that function. On the other hand, we can’t reference the symbol directly lest we fail to launch on XP; we’d have to go via `GetProcAddress` like we do for `DWriteCreateFactory`. That’s kind of annoying for such a simple function used only once.

Another alternative is [`StrCmpW`](https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-strcmpw) from `Shlwapi.dll`, but that means linking to a whole new DLL.

I thought there was a Win32 function for this that had always existed… oh well.

Implementing our own `wcsncmp` is a bulletproof solution, but also feels kinda dumb.

---

Actually, there’s another issue here besides `wcsncmp`: wide string literals. They use `wchar_t`, of course, but we need specifically UTF-16.

At the end of the day, midipix still isn’t exactly a common reality, so we _might_ want to settle for just using `wcsncmp` and wide string literals with `WCHAR` until it materializes better. Or not touch anything at all just yet and stick to `wchar_t` for the time being.

libass/ass_font.c
```diff
@@ -26,15 +26,167 @@
...
 #include "ass_font.h"
 #include "ass_fontselect.h"
 #include "ass_utils.h"
 #include "ass_shaper.h"
 
+/**
+ *  Get the mbcs codepoint from the output bytes of iconv/WideCharToMultiByte,
+ *  by treating the bytes as a prefix-zero-byte-omitted big-endian integer.
+ */
+static uint32_t pack_mbcs_bytes(const char *bytes, size_t length)
```
astiob, line 54: Thanks! I wonder whether this is worth marking `static inline` or it doesn’t make any difference.

libass/ass_font.c
```diff
@@ -79,20 +231,30 @@ void charmap_magic(ASS_Library *library, FT_Face face)
 
 /**
  * Adjust char index if the charmap is weird
- * (currently just MS Symbol)
+ * (currently just MS Symbol and non-Unicode Microsoft cmap)
```
astiob, line 235: At this point, this simply covers all [non-Unicode] Microsoft encodings.

libass/ass_font.c
Apache553: So let's just use `wcsncmp` for string comparison and `WCHAR` for where `wchar_t` is needed. This is done by latest commit.

TheOneric: Regarding string literals, ISO-C11 has the `u` and `U` prefixes for `char16_t` and `char32_t` respectively. MinGW does not complain about them even with our current setting of `-std=gnu99`.
As of current master the `wcsncmp`-call in `ass_directwrite` appears to be the only standard `wchar_t`-API used. *(Checked by grepping for all functions and macros declared in ISO-C99's `wchar.h`: `-E '(fgetwc|fgetws|fputwc|fputws|fwide|fwprintf|fwscanf|getwc|getwchar|putwc|putwchar|swprintf|swscanf|ungetwc|vfwprintf|vwprintf|wprintf|wscanf|wcstod|wcstof|wcstol|wcstold|wcstoll|wcstoull|wcstoul|btowc|mbrlen|mbrtowc|mbsinit|mbsrtowcs|wcrtomb|wctomb|wcsrtombs|wcscat|wcschr|wcscmp|wcscoll|wcscpy|wcscspn|wcslen|wcncat|wcsncmp|wcsncpy|wcspbrk|wcsrchr|wcsspn|wcsstr|wcstok|wcsxfrm|wmemchr|wmemcmp|wmemcpy|wmemmove|wmemset|wcsftime)\(|WCHAR_'`)*

The non-Unicode-font patches use `wchar_t`/`WCHAR` only with Win32-APIs and no "wide" string- or character-literals or standard "wide" functions; so at least here we can safely use `WCHAR` even if we delay adjusting the other occurrences of `wchar_t` with Win32-APIs.

libass/ass_directwrite.c
```diff
@@ -471,18 +471,18 @@ static int encode_utf16(wchar_t *chars, uint32_t codepoint)
...
     HRESULT hr = IDWriteLocalizedStrings_GetStringLength(names, k, &length);
     if (FAILED(hr))
         goto cleanup;
 
-    if (length >= (UINT32) -1 || length + (size_t) 1 > SIZE_MAX / sizeof(wchar_t))
+    if (length >= (UINT32) -1 || length + (size_t) 1 > SIZE_MAX / sizeof(WCHAR))
         goto cleanup;
 
-    temp_name = (wchar_t *) malloc((length + 1) * sizeof(wchar_t));
+    temp_name = (WCHAR *) malloc((length + 1) * sizeof(WCHAR));
```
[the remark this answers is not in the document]
astiob: I think I have an explanation for this: `ass_directwrite` was originally written in C++ and then translated to C. These casts may have been overlooked during that translation. I don’t think there’s any benefit in keeping them in for consistency; but neither is it urgent to take them out.

TheOneric (approved):
I'd slightly prefer if the patch used `WCHAR` from the beginning and the `WCHAR`-transition for `ass_directwrite` were left to a more throughout change also updating the string literals and standard API-calls. But I'm not feeling strongly about this. Otherwise looking good!

MrSmile:
I certainly dislike polluting `ass_font.c` with system-specific code, especially as intended functionality is system-agnostic. Maybe it's better to introduce a dedicated file for this, with the aim to gradually replace `iconv`/`WideCharToMultiByte` with custom conversion functions later?

TheOneric:
> Maybe it's better to introduce a dedicated file for this

I'm not strictly against this, but it seems a bit overkill to me for only ~150 lines. When/if there's a custom conversion-implementation splitting it out makes perfect sense, but that's not (yet) the case. Would you be content if the `windows.h` and `iconv.h` headers were included in the same `#if`-block that defines the functions, so there's only one `#ifdef`-construct?

>  aim to gradually replace `iconv`/`WideCharToMultiByte` with custom conversion functions

As custom conversion functions will likely require embedding large tables into the library, and `WideCharToMultiByte` is “the source of truth” so to say for MS-flavour encodings as we want to emulate, it might be that we'll continue to use this on Windows instead of our custom implementation to keep the binary size down. Unless, there's a relevant performance advantage or the tables end up smaller than I'm expecting.

TheOneric:
After short discussion on IRC, not *yet* creating a new file for this seems ok iiuc.
@Apache553, can you move the `iconv.h`/`windows.h` inclusions into the respective `#if`-block defining the using functions?

libass/ass_font.c
```diff
@@ -35,6 +36,159 @@
...
+        ret |= (uint8_t) bytes[i];
+    }
+    return ret;
+}
+
+/**
+ * Convert a UCS-4 code unit to a packed uint32_t in given multibyte encoding
+ */
+static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint);
+
```
TheOneric, line 57: I'm not too fond of the declaration outside the `#if`-blocks added by the latest change.
At one hand it is unnecessary, because we define the static function in all cases directly below the declaration,
on the other hand, if this weren't the case, adding a unconditional declaration and conditional definitions would delay errors due to a missing definition to the linker stage *(we use `-Werror-implicit-function-declaration`)*.

If you want to preserve the doc comment for `convert_unicode_to_mb` maybe merge it in the comment preceding the `#if defined(_WIN32)`, eg similar to this:
```C
/**
 * Convert a UCS-4 code unit to a packed uint32_t in given MS-flavour multibyte encoding
 *
 * We don't exclude Cygwin from _WIN32, since we only use WideCharToMultiByte which 
 * shouldn't violate any Cygwin restrictions and unlike iconv is guaranteed to be MS-flavour
 */
```

MrSmile:
I think we should also create a performance test, with full page of small letters, to check repeated `iconv_open()` impact.

PR #585, opened 2022-01-19
WCHAR - wchar_t distinction

TheOneric:
Concept for a more throughout `WCHAR`-`wchar_t` distinction likely required for midipix as discussed in #583.
Uses C11's `u"..."` prefix for `WCHAR *`-literals, which MinGW (currently) accepts even with `-std=gnu99`, and replaces the `wchar_t`-API `wcsncmp` with Microsoft's `CompareStringOrdinal`.
Not actually tested beyond compilation, feel free to force-push this branch if you prefer.

### DIFF
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -482,7 +482,7 @@ static char *get_utf8_name(IDWriteLocalizedStrings *names, int k)
     if (length >= (UINT32) -1 || length + (size_t) 1 > SIZE_MAX / sizeof(wchar_t))
         goto cleanup;
 
-    temp_name = (wchar_t *) malloc((length + 1) * sizeof(wchar_t));
+    temp_name = malloc((length + 1) * sizeof(wchar_t));
     if (!temp_name)
         goto cleanup;
 
@@ -495,7 +495,7 @@ static char *get_utf8_name(IDWriteLocalizedStrings *names, int k)
     if (!size_needed)
         goto cleanup;
 
-    mbName = (char *) malloc(size_needed);
+    mbName = malloc(size_needed);
     if (!mbName)
         goto cleanup;
 
@@ -616,7 +616,7 @@ static void add_font_face(IDWriteFontFace *face, ASS_FontProvider *provider,
             goto cleanup;
     }
 
-    FontPrivate *font_priv = (FontPrivate *) calloc(1, sizeof(*font_priv));
+    FontPrivate *font_priv = calloc(1, sizeof(*font_priv));
     if (!font_priv)
         goto cleanup;
 
@@ -815,7 +815,7 @@ static void add_font(IDWriteFont *font, IDWriteFontFamily *fontFamily,
     if (!get_font_info_IDWriteFont(font, fontFamily, &meta))
         goto cleanup;
 
-    FontPrivate *font_priv = (FontPrivate *) calloc(1, sizeof(*font_priv));
+    FontPrivate *font_priv = calloc(1, sizeof(*font_priv));
     if (!font_priv)
         goto cleanup;
     font_priv->font = font;
@@ -1086,7 +1086,7 @@ ASS_FontProvider *ass_directwrite_add_provider(ASS_Library *lib,
         goto cleanup;
     }
 
-    priv = (ProviderPrivate *)calloc(sizeof(*priv), 1);
+    priv = calloc(sizeof(*priv), 1);
     if (!priv)
         goto cleanup;
 
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -21,7 +21,6 @@
 #include "ass_compat.h"
 
 #include <initguid.h>
-#include <wchar.h>
 
 #include "dwrite_c.h"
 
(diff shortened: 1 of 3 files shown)


astiob:
Dislike, because this breaks compilation on older compilers for (IMO) no good reason.

astiob:
In particular, midipix itself [used to have](https://github.com/mintty/mintty/issues/559#issuecomment-255352813), in 2016, an old GCC that wouldn’t understand these string literals. They were going to upgrade it “immediately after midipix alpha release”, which has not happened to date, and I can’t immediately find any mention of GCC in [their repositories](https://dev.midipix.org/) at all, so I don’t know whether their GCC ever got upgraded. In the linked thread, they were saying Unicode literals sounded like the right solution that was up to them to make work by upgrading their GCC, but if we don’t gain any immediate benefit from it now, I don’t think we can justify the cost of dropping old compilers.

astiob:
It’s also not clear whether midipix defines `_WIN32` in the first place. Someone would have to join their IRC channel (which is the only way to get installation instructions) and get a working setup to test.

Even if it does, our `configure` probably (just guessing) disables `ass_directwrite` on it. So that would need fixing as well.

TheOneric:
> this breaks compilation on older compilers for (IMO) no good reason.

As all instances of `u"..."` are tied to `WCHAR`, they will be hidden behind `_WIN32` or `CONFIG_DIRECTWRITE` guards, both of which can only be true when compiling for Windows, meaning the only relevant compilers will be MinGW's gcc and clang, Microsoft's compiler and midipix's compiler. MinGW's gcc does accept it and Microsoft docs also mention `u"..."`. Now, if midipix's gcc-builds still does not support `u"..."` the change of course doesn't make any sense.
When not compiling for Windows a wide variety of compilers can be used, but none of them will see any `u"..."`-literals.

I don't know how to solve this without using `u"..."` or inserting midipix-specific conversion functions for literals everywhere; do you have a better suggestion (or prefer to wait with any changes until a full midipix-toolchain is publicly available)?

> Our assembler mode selection might be broken, though.

Midipix's triplet is supposedly `x86-64-midipix-nt64`. This will end up in the `x86-64`-but-unknown-platform case, printing an error message suggesting to contact upstream and refuse to configure without `--disable-asm` in the meantime *(but yes, previously configure and perhaps the build would have succeeded but resulting in an unusable binary)*.

astiob:
> I don't know how to solve this without using `u"..."` or inserting midipix-specific conversion functions for literals everywhere

The best alternative I have is a platform- or compiler-dependent macro:

```c
#ifdef whatever_midipix_defines
#define U16(s) u ## s
#else
#define U16(s) L ## s
#endif
U16("Arial")
```

Microsoft itself has a similar macro (`_T` iinm) for marking strings narrow or wide depending on the `_UNICODE` macro. I don’t know what that macro does on midipix. Maybe we can actually define `_UNICODE` and use that same macro without making our own. But I don’t see it mentioned in that mintty thread I linked earlier, so I expect not.

> MinGW's gcc does accept it and Microsoft docs also mention `u"..."`.

Right, but that’s _current_ versions. Before this change, people can use much older MinGW GCCs (on Windows and cross-compiling alike) to build libass, which would stop working. I’m not sure which MSVC versions we require and whether this applies there too. There’s also Intel’s compiler, which is switching to LLVM now but I get the impression still has its own frontend even now, and of course was completely separate until now. I think it would also be disingenuous to continue to say we need “C99” when using C11 string literals, but maybe we could live with that.

> or prefer to wait with any changes until a full midipix-toolchain is publicly available

Personally, I’m inclined towards this; but it’s merely an inclination. I think we could use `WCHAR` for consistency with Microsoft’s API declarations and other types but not bother changing anything else right now.

It seems that adding midipix support requires some litter in our code (even if not for the string literals, then at least for the `StringCompareOrdinal`), and it’s unclear whether there is (ever going to be) any actual benefit to this. Maybe I’m too ignorant and midipix is actually, like, really used? I’d be happy to do whatever we need to support it if it’s actually a thing. Or maybe I’m overthinking the negative impact from the changes on our code clarity and compilability. It doesn’t exactly add motivation that you can’t just download a test setup from their website, either, but maybe I’m being too petty.

TheOneric:
Putting the macro into `ass_compat.h` with `u"..."` only for `ifdef __midipix__` will work yeah. AndI agree that it would be better if there was a publicly available way to test midipix.
Poking around in the aforementioned mintty-port, I [found it uses](https://github.com/midipix-ports/mintty/blob/midipix/src/host/midipix/std.h#L57) `u16_wcsncmp` function, likely from the included `u16ports/u16ports.h`-header which GitHub's search doesn't find in the repository. Perhaps it's provided with midipix and we could also use macros to override the standard `wchar_t` APIs on midipix. As we can't test to be sure, let's close this PR and take another look when midipix's toolchain is fully public.

TheOneric:
For reference: #573 adds another three uses of the standard `wchar_t` `wcslen`-API, assuming 16bit-width.

astiob:
Thanks for the effort and for keeping track of the `wcslen`s. It is unfortunate, but I agree it’s best to wait until midipix becomes more public.

> I [found it uses](https://github.com/midipix-ports/mintty/blob/midipix/src/host/midipix/std.h#L57) `u16_wcsncmp` function, likely from the included `u16ports/u16ports.h`-header which GitHub's search doesn't find in the repository.

Ah, that’s another interesting option. For reference, I can see that file here: https://dev.midipix.org/cross/u16ports/blob/main/f/include/u16ports/u16ports.h

PR #587, opened 2022-02-01
Fix wrong "size"/"priv" arguments name in ass_add_font

CastagnaIT:
The argument `size` and `priv` not match the argument as defined in the header
with the result of compiler warnings
Header: https://github.com/libass/libass/blob/master/libass/ass.h#L696-L697

### DIFF
--- a/libass/ass_library.c
+++ b/libass/ass_library.c
@@ -95,24 +95,24 @@ void ass_set_style_overrides(ASS_Library *priv, char **list)
         *q = strdup(*p);
 }
 
-void ass_add_font(ASS_Library *priv, const char *name, const char *data, int size)
+void ass_add_font(ASS_Library *library, const char *name, const char *data, int data_size)
 {
     size_t idx = priv->num_fontdata;
-    if (!name || !data || !size)
+    if (!name || !data || !data_size)
         return;
     if (!(idx & (idx - 32)) && // power of two >= 32, or zero --> time for realloc
             !ASS_REALLOC_ARRAY(priv->fontdata, FFMAX(2 * idx, 32)))
         return;
 
     priv->fontdata[idx].name = strdup(name);
-    priv->fontdata[idx].data = malloc(size);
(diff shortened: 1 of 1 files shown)


astiob:
Thanks!

Out of curiosity, what compiler does this?

The same function’s first parameter also has a different name in the header and in the file. Is that not producing the same warning?

rcombs:
Wouldn't just changing the parameter name in the header be a smaller change? Also, yeah, what compiler does this? The parameter names don't affect the function signature, so afaik it's perfectly spec-compliant for the declaration and definition not to match.

CastagnaIT:
my exhaustion i have completely forgotten it...
on Kodi we have patched libass for Windows to compile it with CMake

TheOneric:
> > Out of curiosity, what compiler does this?
>
> on Kodi we have patched libass for Windows to compile it with CMake

CMake is not a compiler; it's a build-system which (indirectly) invokes a compiler among others.

I'm not adverse to making the names more consistent, but as far as the standard is concerned the names do not matter beyond each individual declaration or definition and are allowed to be different (in fact they can be left out in declarations entirely). So this is more a cosmetic change rather than a “fix” for something wrong.

Furthermore, there is another instance of a `priv`/`library`-mismatch between public declaration and (internal) definition with
`ass_clear_fonts` and `ass_renderer_done` uses `priv`/`render_priv`. Were there no compiler warnings about those?
If the names are to be changed at all, I think it'd be better to make the header (and definitions) consistently use `priv` for `ASS_Library *` parameters. Not sure about `ASS_Renderer *` though, most declarations and definitions currently also use `priv` like for `ASS_Library *`.

CastagnaIT:
i was using visual studio 2019 MSVC
but recently i have switched to 2022
until now i hadn't recompiled anymore I just brought in a commit I had done a long time ago
now that i have tried recompile i notice that the warning is no longer given
feel free to merge or close

PR #593, opened 2022-02-14
Clearly document the need for ass_set_storage_size

TheOneric:
The `NOTE:` at the end of `ass_set_storage_size`'s documenting comment matches how it is already done for `ass_set_fonts`. For better visibility this is again pointed out for both in `ass_renderer_init`'s comment.

I'm not sure if the phrasing for the anamorphic part is correct (or good).

### DIFF
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -330,6 +330,11 @@ void ass_set_message_cb(ASS_Library *priv, void (*msg_cb)
  * \brief Initialize the renderer.
  * \param priv library handle
  * \return renderer handle or NULL if failed
+ *
+ * NOTE: before rendering starts the renderer should also be
+ *       configured with at least ass_set_storage_size(),
+ *       ass_set_frame_size() and ass_set_fonts();
+ *       see respective docs.
  */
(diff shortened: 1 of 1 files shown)


libass/ass.h
```diff
@@ -358,11 +362,14 @@ void ass_set_frame_size(ASS_Renderer *priv, int w, int h);
  * This is used to calculate the source aspect ratio and the blur scale.
  * The source image size can be reset to default by setting w and h to 0.
  * The value set with this function can influence the pixel aspect ratio used
- * for rendering.
+ * for rendering. The values must be from before anamorphic stretching.
```
rcombs, line 365: The meaning here isn't entirely clear; I'd say something like "The values must be the actual storage size of the video stream, without any anamorphic de-squeeze applied."

rcombs (approved):

libass/ass.h
TheOneric: Changed it to the suggested, thanks!

TheOneric:
Will merge in ~2 days if there are no objections.

Issue #591, opened 2022-02-13
Many API-users ignore ass_set_storage_size leading to incorrect rendering


TheOneric:
As we noticed recently on IRC while discussing commit https://github.com/libass/libass/commit/d60f2dce1dcafaed0e21406213e66072f3036752 (October 2020), which changed the behaviour when `ass_set_storage_size` was never called, many API-users ignore `ass_set_storage_size`.
We briefly thought, that before the above commit API-users who're rendering at storage resolution would still have gotten the correct behaviour by accident — but that was never the case for transforms.

This leads to incorrect results and is particular bad for Aegisub as it means sub authors will work with wrong results and moreover if the final result is only checked in a similarly bugged player (e.g. VLC) the error will not show itself even outside of Aegisub.

We should probably more prominently point out the need for `ass_set_storage_size` in our docs *(EDIT_ now done)* and send issues/patches to ensure at least the most prominent libass users do call `ass_set_storage_size`. We should also suggest to backport the fix to LTS/stable releases where applicable.

**mpv** and **gstreamer** are **not affected** due to calling `ass_set_storage_size`; below is an incomplete list of affected programs:

 - [x] **Aegisub**:
       There are also AviSynth- and cli-interfaces to its csri; I don't know if they always render at video resolution.
       At least the most prominent, active'ish forks should be fixed:
   - [x] [wangqr](https://github.com/wangqr/Aegisub)
   - [x] [TypesettingTools](https://github.com/TypesettingTools/Aegisub)
   - [x] [DC](https://github.com/Ristellise/AegisubDC), is derived from wangqr afaik ~~so a fix there should eventually end up in DC anyway, but pinging DC is probably a good idea~~
      Mysteriously the last AegisubDC binaries v9215 already show the correct rendering with its libass backend, eventhough the sourcecode in git (https://github.com/Ristellise/AegisubDC/commit/35db2ad6c0b64599249175ed6904010f200e0be2) does never call `ass_set_storage_size`. Also AegisubDC is now basically unmaintained.
      **Update:** As it turns out, the 9215B binary build apparently already includes the patch, eventhough it was neveer pushed to the public git repo, see: https://github.com/Ristellise/AegisubDC/issues/54
 - [x] **ffmpeg**: always renders at (what will become after filters?) video resolution, already has `original_size` parameter. Fix included in master after (not including) 5.0; got backported to 5.0.1 and several 4.x releases.
[...]

PR #601, opened 2022-03-01
Documentation Corrections 

TheOneric:
Possibly there's another issue with our description of `YCbCr Matrix: None` in `ass_types.h`.
Our docs claim and mpv interprets “None” to mean: “use sRGB values unchanged” and that this would match all but "some buggy" xy-VSF versions. However Cyberbeing’s and pinterf’s xy-VSFilter’s rc5 branches seem to effectively treat "None" the same as TV.601 afaict?

As mentioned on IRC we might want to deprecate setting PAR explicitly in favour of `ass_Set_storage_size` in the future *(and for `LayoutRes` it currently seems like ignoring the explicitly set PAR is required anyways)*. Does someone have a usecase for explicilty setting PAR with non-isotropic scaling?

### DIFF
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -353,8 +353,10 @@ void ass_renderer_done(ASS_Renderer *priv);
  * \brief Set the frame size in pixels, including margins.
  * The renderer will never return images that are outside of the frame area.
  * The value set with this function can influence the pixel aspect ratio used
- * for rendering. If the frame size doesn't equal to the video size, you may
- * have to use ass_set_pixel_aspect().
+ * for rendering.
+ * If after compensating for configured margins the frame size
+ * is not an isotropicly scaled version of the video display size,
+ * you may have to use ass_set_pixel_aspect().
  * @see ass_set_pixel_aspect()
  * @see ass_set_margins()
  * \param priv renderer handle
@@ -409,10 +411,6 @@ void ass_set_shaper(ASS_Renderer *priv, ASS_ShapingLevel level);
  * the bottom "black bar" between video bottom screen border when playing 16:9
  * video on a 4:3 screen.)
  *
- * When using this function, it is recommended to calculate and set your own
- * aspect ratio with ass_set_pixel_aspect(), as the defaults won't make any
- * sense.
- * @see ass_set_pixel_aspect()
  * \param priv renderer handle
  * \param t top margin
  * \param b bottom margin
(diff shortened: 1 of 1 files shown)


TheOneric:
Is “avs” AviSynth? What is “vs”?
If it's AviSynth, it currently doesn't call `ass_set_storage_size` but should do so. After looking at its readme and glancing over the code, it seems to me like PAR is only set if explicitly requested by an user, did I get this right? Furthermore, I’m assuming assrender probably renders subtitles at the videos storage size *(i.e. before anamorphic de-squeezing)* so without user override the automatically derived PAR for libass would normally always be 1.0, is this correct?

Digressing a bit from the core topic of this PR:
If that’s the case and assrender wants to continue to have such an override, there will be a problem with potential future files using `LayoutRes{X,Y}` as those will likely ignore the API-supplied PAR. For the override to continue to always work we’d need to add a new “force PAR-API”. However, if doing so with `LayoutRes{X,Y}` is actually desirable is not clear. *If* assrender doesn't need to provide a stable argument interface, replacing the DAR and SAR arguments with “original_size” arguments which would be used for PAR calculation and be passed to `ass_set_storage_size` might be favourable *(that’s what [ffmpeg did in 2012](https://github.com/FFmpeg/FFmpeg/commit/247fbf071bb9d30f5341844a3d4ebbd53a986968) – except for not setting storage size yet)*.

As for this PR, I asked about the usage to determine if the PAR wording should perhaps be changed to favour `ass_set_storage_size` more over setting PAR. But with such an override existing in assrender, let’s keep the current wording which keeps PAR relevant for non-isotropic scaling until we actually discuss whether to deprecate `ass_set_pixel_aspect`.

Masaiki:
vs means vapoursynth, also a kind of scripts for video manipulation like avisynth.
if no dar and sar passed, par will be default to 1.0 in `ass_start_frame`. in theory, whether or not to call `ass_set_storage_size` (with the same parameters as `ass_set_frame_size`) should render the same result at this time, but # 591 really puzzles me.
besides, clip (the video stream structure) of avisynth and vapoursynth only has "storage_size", and the "frame_size" is more of container side.

TheOneric:
> vs means vapoursynth, also a kind of scripts for video manipulation like avisynth.

Thanks. Does the VapourSynth assrender call `ass_set_storage_size`? If not we'll need to file an issue there too.

> besides, clip (the video stream structure) of avisynth and vapoursynth only has "storage_size"

Thanks for the info; so it seems for those two frame size equals storage size and in absence of additional filters PAR would be 1.

> if no dar and sar passed, par will be default to 1.0 in `ass_start_frame`. in theory, whether or not to call `ass_set_storage_size` (with the same parameters as `ass_set_frame_size`) should render the same result at this time, but #591 really puzzles me.

I'm not sure I understood what part of #591 puzzles you.
PAR explicitly defaulting to `1.0` matches what libass does when neither `ass_set_storage_size` nor one of the PAR setters are called. Regarding only PAR this also matches what libass does when frame and storage size are set to the same value and PAR is not set explicitly. However, `ass_set_storage_size` affects more then just PAR *(as the docs already say and these patches expand upon)*. So while setting PAR to `1` and not setting storage size leads to the same PAR as if storage size was set to equal frame size, other aspects of rendering will differ between those cases. Setting storage size provides more info than just setting PAR.

If the goal of assrender’s DAR and SAR options was to preserve rendering between anamorphic <-> non-anamorphic conversions, it is insufficient since it’s missing the additional information of the original storage size. It also can't preserve rendering after isotropic scaling, where PAR doesn't change at all, but the storage size does. ffmpeg replaced its DAR option for ASS with an `original_size` option in the past *(though they didn't know about the non-PAR affects yet, but it’ll come in handy to fix just that)*. Such an `original_size` option can correctly deal with those use cases, PAR alone cannot.
The only case where *maayybe* in some very specific cases forcing a PAR different from what storage + frame size would yield might be desirable, is if a previous filter scaled the video display size non-isotropicly — but I don't think this would even give sensible results in the general case.

PR #603, opened 2022-03-11
Dont check avx2 feature bit unconditionally

Zapeth:
Proper cpuid usage is to check the highest function parameter (eax=0) before querying extended features

### DIFF
--- a/libass/ass_utils.c
+++ b/libass/ass_utils.c
@@ -57,7 +57,11 @@ int has_avx(void)
 
 int has_avx2(void)
 {
-    uint32_t eax = 7, ebx, ecx, edx;
+    uint32_t eax = 0, ebx, ecx, edx;
+    ass_get_cpuid(&eax, &ebx, &ecx, &edx);
+    if (eax < 7) // check highest function parameter
+        return 0;
+    eax = 7;
     ass_get_cpuid(&eax, &ebx, &ecx, &edx);
     return (ebx >> 5) & has_avx();
 }


TheOneric:
Thanks. Do you know if `eax=1` is always safe or should we check with `eax=0` first also in `has_avx` and `has_sse2`?

Out of interest, does it `SIGILL` if it’s called with `eax=7` on a non-supporting CPU or does something else happen?

Zapeth:
From the Intel instruction set manual:
> If a value entered for CPUID.EAX is higher than the maximum input value for basic or extended function for that processor then the data for the highest basic information leaf is returned.

This is indeed the case on my CPU, but I wouldn't expect this to be consistent across different vendors (or even models).

It also says that even the cpuid instruction itself cannot be assumed to be available (`The ID flag (bit 21) in the EFLAGS register indicates support for the CPUID instruction`), so technically that should be checked as well.

Though realistically I don't expect this to be a problem for any post-2000 x86 CPUs, so I guess its up to you how thorough you want to check it. Usually its safer/easier to just use the compiler-provided functions than manually calling the instruction (`__get_cpuid` in gcc)

MrSmile:
I think, I should finally merge my PR https://github.com/libass/libass/pull/559 instead. It already has that check coupled with the better CPU flag architecture.

Zapeth:
Ok, that implementation should fix the issue as well, so closing this in favor of your PR

TheOneric:
Both #400 and #559 check for the highest supported CPUID leaf, but neither checks if CPUID is supported via EFLAGS meaning a i386 binary compiled with `--enable-asm` *(the default)* will compile but not run on pre Pentium+ CPUs iiuc.
And afaik rcombs and MrSmile did not reach a consensus on which of the two approaches/PRs to take.

TheOneric:
[AMD’s](https://www.amd.com/system/files/TechDocs/24594.pdf) and [Intel’s](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf) both state software “at any privilege level” *(p.165, p.3-190)* can execute the CPUID instruction.
AMD’s manual also provides the following example to test for CPUID support on 32bit (and 64bit) processors:
```asm
pushfd        ; save EFLAGS
pop eax       ; store EFLAGS in EAX
mov ebx, eax  ; save in EBX for later testing
xor eax, 00200000h ; toggle bit 21
push eax      ; push to stack
popfd         ; save changed EAX to EFLAGS
pushfd        ; push EFLAGS to TOS
pop eax       ; store EFLAGS in EAX
cmp eax, ebx  ; see if bit 21 has changed
jz NO_CPUID   ; if no change, no CPUID
```

But for `pushf` there are privilege restrictions documented:
> **AMD, *p. 290***
> In virtual-8086 mode, if system software has set the IOPL field to a value less than 3, a general-protection exception occurs if application software attempts to execute PUSHFx or POPFx while VME is not enabled or the operand size is not 16-bit

> **Intel, *p. 4-516***
> When in virtual-8086 mode and the I/O privilege level (IOPL) is less than 3, the PUSHF/PUSHFD instruction causes a general protection exception (#GP).

Are those restrictions something we could trip up on?

Zapeth:
I don't think this is an issue for applications that run at user level, not to mention that its also implemented that way [in gcc](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/cpuid.h#L243), so I would use that as reference (or just directly use it).

TheOneric:
Requiring compiler-specific functions is not an option because they are non-portable and libass supports a wide range of compilers.

The GCC code you linked assumes CPUID is always available on x86_64. I didn't spot any such guarantee in AMD’s or Intel’s manual, but they are quite large and I’m no expert on x86. Do you happen to know where this assumption comes from?

I haven't been able to test this with a CPU which doesn't support CPUID, but without assuming CPUID is mandatory in x86_64 this probably works:
```asm
;-------------------------------------------------------------------
; uint32_t has_cpuid( void );
;-------------------------------------------------------------------
%if ARCH_X86_64
    %define PUSHFN pushfq
    %define POPFN  popfq
%else
    %define PUSHFN pushfd
    %define POPFN  popfd
%endif

INIT_XMM
cglobal has_cpuid, 0, 2, 0
    PUSHFN
    pop r1
    mov r0, r1
    xor r0, 0x00200000
    push r0
    POPFN
    PUSHFN
    pop r0
    xor r0, r1
    and r0, 0x00200000 ; non-zero if bit is writable
    push r1            ; Restore original EFLAGS
    POPFN
    mov eax, r0d
    RET

```

rcombs:
There is no reason for anyone to ever try to run libass on virtual-8086 mode (and if they did, they'd have to build for 8086, not for i386, so our i386 assembly wouldn't be built anyway). 

I don't see any reason to support x86 CPUs without CPUID. Nothing else in open-source multimedia supports it, and essentially everything using libass uses at least one other such package. I'm not aware of anyone ever having complained about the requirement across the existence of libass, ffmpeg, x264, or any other related project. If someone really needs to run on pre-Pentium x86 CPUs, they can file a separate issue, but I would be very surprised if that ever happened.

TheOneric:
Iiuc CPUID is not part of the base-i386 or even -i486 set, so a build with "i386 assembly" enabled crashing there seems odd to me. If preventing these crashes is as simple as checking writability of one bit and we don't need to worry about wirtual-8086, I see no reason not to do so.

rcombs:
Oh huh, I just double-checked myself and it turns out ffmpeg, x264, and LAME actually _do_ check eflags writability to verify CPUID availability (though opus, dav1d, libvpx, and libaom do not). So, alright, fair enough, that's enough precedent for me.

Zapeth:
> The GCC code you linked assumes CPUID is always available on x86_64. I didn't spot any such guarantee in AMD’s or Intel’s manual, but they are quite large and I’m no expert on x86. Do you happen to know where this assumption comes from?
> 
I'm also just going by assumption and haven't found a definitive statement that clarifies this, but as far as I can tell its availability is implied due to x86_64 CPUs being required to support certain features and extensions, like SSE and SSE2 instructions. 

And those generally need to be queryable by applications, using the CPUID instruction.

Also I'd think that if that assumption was incorrect the affected code would have been updated since then (looking at the git log, the comment was written in 2007).

rcombs:
I can confirm that none of the other multimedia projects I checked do a CPUID check on x86_64; they all assume it'll always be available there.

PR #604, opened 2022-03-17
Fix x86 cpuid

TheOneric:
As discussed in #603 we need to check for cpuid support and if it is supported for the highest leaf available to avoid crashes and bogus results respectively.

The content of `ass_cpu_capabilities` is mostly copied from #400 and #559 but instead of using bitflags and moving it into a new / into the bitmap header it for now remains in `ass_utils`. It can be moved and switched to bitflags with the checkasm- and aarch64-related refactoring.

I'm still not 100% sure why x86_64 supposedly always supports cpuid, but gcc and [ffmpeg](https://github.com/FFmpeg/FFmpeg/commit/d05f808dc90a55f2a6eabe6ff62eb3a056e428d3) assume this *(x264 appears to have copied the check from ffmpeg; LAME only has 32bit assembly afaict)* and Zapeth’s guess about an implicit dependency SSE or so seems plausible.

### DIFF
--- a/libass/ass_utils.c
+++ b/libass/ass_utils.c
@@ -40,6 +40,9 @@ void ass_cpu_capabilities(bool *sse2, bool *avx2)
     *sse2 = false;
     *avx2 = false;
 
+    if (!ass_has_cpuid())
+        return;
+
     uint32_t eax = 0, ebx, ecx, edx;
     ass_get_cpuid(&eax, &ebx, &ecx, &edx);
     uint32_t max_leaf = eax;
--- a/libass/ass_utils.c
+++ b/libass/ass_utils.c
@@ -35,33 +35,39 @@
 
 #include "x86/cpuid.h"
 
-int has_sse2(void)
+void ass_cpu_capabilities(bool *sse2, bool *avx2)
 {
-    uint32_t eax = 1, ebx, ecx, edx;
-    ass_get_cpuid(&eax, &ebx, &ecx, &edx);
-    return (edx >> 26) & 0x1;
-}
+    *sse2 = false;
+    *avx2 = false;
 
-int has_avx(void)
-{
-    uint32_t eax = 1, ebx, ecx, edx;
+    uint32_t eax = 0, ebx, ecx, edx;
     ass_get_cpuid(&eax, &ebx, &ecx, &edx);
-    if (!(ecx & (1 << 27))) // not OSXSAVE
-        return 0;
-    uint32_t misc = ecx;
-    ass_get_xgetbv(0, &eax, &edx);
-    if ((eax & 0x6) != 0x6)
-        return 0;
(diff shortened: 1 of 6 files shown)


libass/ass_utils.c
```diff
@@ -35,33 +35,39 @@
...
-    return (misc >> 28) & 0x1;
-}
+    uint32_t max_leaf = eax;
+    bool avx = false;
+
+    if (max_leaf >= 1) {
+        eax = 1;
+        ass_get_cpuid(&eax, &ebx, &ecx, &edx);
+        if (edx & (1 << 26))  // SSE2
+            *sse2 = true;;
```
rcombs, line 55: ;;

rcombs (commented):
One nit; otherwise looks fine.

libass/ass_utils.c
TheOneric: done

libass/x86/cpuid.asm
```diff
@@ -22,6 +22,31 @@
...
+    xor r0, 0x00200000
+    push r0
+    popfd
+    pushfd
+    pop r0
+    xor r0, r1
+    and r0, 0x00200000 ; non-zero if bit is writable
+    push r1            ; Restore original EFLAGS
+    popfd
+    mov eax, r0d
```
MrSmile, line 46: This is `mov eax, eax`. I think it's better to use `eax/ecx` directly instead of `r0/r1` to not trigger unwanted side effects like this. Also maybe rewrite it to use only one register? Something like this (not tested);
```asm
pushfd
mov eax, [esp]
xor [esp], dword 0x00200000
popfd
pushfd
xor eax, [esp]
xor [esp], eax
popfd
and eax, 0x00200000
ret
```

TheOneric: What side-effects does `mov eax, eax` have? Afaik this is just a no-op.

Regarding `r0/r1`, sure I can change it to use `eax/ecx`, if altering `ecx` is allowed in all supported calling conventions and you think avoiding one potential no-op is worth it.

As for using only one register, I don't see what benefit this is supposed to bring. Even 32bit x86 has 3 caller saved registers in the calling convention I know about. Are compilers (with LTO?) capable of omitting saving some normally caller-saved registers when they know the called function doesn't use them? And if so, does this really matter here; `has_cpuid` is not performance-critical and iinm only called once per `ASS_Renderer`.

astiob: > Are compilers (with LTO?) capable of omitting saving some normally caller-saved registers when they know the called function doesn't use them?

They are, but here the C compiler won't know that the nasm-compiled function doesn't use some registers.

TheOneric: Now using `eax`, `ecx` and `cglobal 0,0,0` instead of `r0/r1`.

PR #606, opened 2022-04-04
Non-recursive Make

TheOneric:
This is a demo for a non-recursive build containing some comments explaining or pointing out noteworthy matters and is intentionally not fully consistent to show off different styles/approaches. A final version would drop the comments and be made fully consistent using the chosen style. gitignore paths are also left unchanged for now.

The main advantages of using a non-recursive build are that it provides better automatic dependency resolution across targets and enables higher parallelism for shorter build times. Furthermore it can reduce some duplication *(like eg the NASM target rules in #400) and reduces clutter in `make`’s output. *(If everything is already built there still 22 lines of output with our current recursive build if all utilities are enabled, with this patch there are only 3 lines.)*

Regarding the dependency resolution, take a look at the `bench` rule in #400’s 3906012d9cf4afc19e522a7de749847657d0fbeb: If the library hasn't been built yet it will fail as the Makefile in the checkasm subdir does not know how to create the library. To fix this we’d need to add `bench: libass.la` to the Makefile.am in the libass dir – an unrelated Makefile.  
The mockup `sptarget` in this demo commit on the other hand will work just fine without any additional changes in other fragments as the Makefile it ends up in also knows how to build the library.

The speedup comes from subdirs being required to be processed sequentially, which reduces how much can be compiled in parallel and even worse means all libraries and utilities are *linked* one by one instead of simultaneously.
For the default build of only the library, i.e. only one subdir, any possible speed change will be negligible but for development builds with utilities enabled there’s a noticeable difference and we’ll add at least two more utilities (checkasm, fuzzer) in the future further increasing the difference.

To quantify the speedup claim, here are some rounded timings (in seconds) for `profile`, `test` and `compare` enabled, using nasm and 12 threads. There clearly is a benefit in relative terms, though in absolute terms the build time is rather short — at least on the tested hardware — either way *(but this also isn't the only benefit)*.

 |                                  | recursive | non-recursive | Relative time reduction |
 |----------------------------------|-----------|---------------|-------------------------|
 | Clean Build, default flags       |  4.3±0.1  |   3.5±0.05    |  19%                    |
 | Changed ass.c, default flags     |  1.3±0.05 |   1.0±0.05    |  18%                    |
 | Clean Build, `-flto=jobserver`   | 10.1±0.1  |   6.3±0.05    |  37%                    |
 | Changed ass.c, `-flto=jobserver` |  7.6±0.1  |   4.3±0.1     |  43%                    |

.

Downsides of using a non-recursive Make are that the paths of the source files end up a bit longer, but longterm not much else as far as I can see.
Midterm, if the first commit is kept and the binaries are placed directly into the toplevel: when one changes between an older and newer commit eg in a bisect, running `make clean` *(for new→old, if using an in-tree build)* or manually removing the `Makefile.in` from the utility directories *(old→new)* is required to avoid conflicts due to name-collisions.

A user-visible change, with “user” being someone who builds libass, with this is that the binaries can potentially be placed in a different directory (see comments in commit) and that the uninstalled libraries move from `libass/.libs/` to `.libs/` — once installed there’s no difference.

---

The actual change is rather simple once a style is decided. This RFC state commit is intended to make sure everyone is aware of alternatives and that we can find an approach everyone likes. Thus it’d be great if I could get some feedback on the options even if it’s just “idc”:
 - Are there any reservations about using a non-recursive build in general?
 - If not, should we:
   - use `%reldir%`, `%D%` or hardcode paths?
   - place binaries directly into the toplevel, into `bin/` or hardcode the paths to keep similar behaviour to now?
   - gather all utilities into one directory? *(answering “no” here conflicts with binaries at the toplevel)*
   - use a single `Makefile.am` fragment for all utilities or one fragment per utility?

Personally I’m inclined to gather all utilities in a common directory with a single `Makefile.am` and using `%reldir%` everywhere for ease of understanding even if it’s a bit longer. Regarding the placement of the binaries I don't like hardcoding the path, but otherwise do not have a particular preference.
In the end though I'm fine with any of the options if they are commonly accepted.


TheOneric:
Turns out support for `%reldir%` and `%D%` was "only" added to Automake in 2013 with [this commit](https://git.savannah.gnu.org/cgit/automake.git/commit/?id=ba25a9f1c295d7799575afa39c4c9854a85ad8e3) contained in releases since 1.14 *(December 2013)*.
Does anyone care about dropping support for pre 1.14 Automake? Note this only affects builds from git and tarball generation. Building from an already generated tarball does not require Automake at all.

rcombs (commented):
Generally looks good to me. A few thoughts on the nitty-gritty:
- It'd be nice if we could use put just the filenames in the .am file, then set the `_SOURCES` variable by automatically prepending `%reldir/%` to each item in the list. In Make, that'd be something like `libass_la_SOURCES := $(libass_sourcelist:%=$(SUBDIR)%)`. Does that kind of thing work in Automake?
- If not, I think I'd go with the first approach (hardcoded paths); it looks cleaner.
- I have no issue with dropping support for very old automake versions.

astiob:
To be very honest: do we even need separate Automake files? Our code/object base is pretty small. Maybe it’s easier to just manually put it all into a single top-level Makefile.am?

As for the location of the target files:
* I don’t have a particular preference, but I think any change will want to be prominently noted in the commit message/change log/release notes. I imagine the exact path may be used in custom install scripts or similar. (Myself, I’m using a symlink to `.../libass/.libs/libass.9.dylib` as my system-wide libass.)
* The poor Meson PR that we keep neglecting will want to match. It may be worth considering whether any one option is more convenient for Meson than the others.
* The bisect issue is unfortunate, especially because if you forget to `clean` and keep the wrong binaries around, you may end up continuing to run them while thinking you’re testing completely different commits. All things equal, I’d prefer to avoid this, but if that avoidance brings significant disadvantages of its own, I’m sure we can live with it.

TheOneric:
Repushed to fix a mistake: It ***is*** possible to use `%reldir%` for executable/library locations, the associated varaible names should then use %canon_reldir%, or %C% for short, to get the sanitised form.

---

> It'd be nice if we could use put just the filenames in the .am file, then set the _SOURCES variable by automatically prepending %reldir/% to each item in the list.  In Make, that'd be something like `libass_la_SOURCES := $(libass_sourcelist:%=$(SUBDIR)%)`.

Automake 1.16’s release news note that it is since then possible to use Make variables in the source list, which might mean the macro expansion might work as well. However, the pattern macro you used, is a GNU Make extension. POSIX Make only has `$(var:subst1=subst2)` which adds suffixes, *(though the future direction notes pattern macros might at some undetermined point be added to POSIX)*. So this won't work in any case, without suddenly requiring GNU Make.

TheOneric:
> Maybe it’s easier to just manually put it all into a single top-level Makefile.am?

I like the separation between the library and auxiliary programs, but it would be possible.

> The poor Meson PR that we keep neglecting will want to match.

Non-installed Meson generated files are already in a different place anyway. The only relevant change for the Meson patch here, is the move of the profile sources *(the Meson patch does not support building any auxiliary programs but profile)*. Note though, there are other build changes already in master, which the Meson patch didn't yet adjust to.

> The bisect issue is unfortunate [...] All things equal, I’d prefer to avoid this

Regarding the executables, we could not move the sources and keep the binary locations. This would avoid the bisect issue, but means we would soon have at least 5 different folders for test programs at the toplevel. The non-recursive change could be a good opportunity to reorganise the sources. If we keep the locations, I'm not sure if it’d be better to have a separate fragment for the auxiliary programs at the toplevel (`Makefile-util.am`?) or add their rules to the root `Makefile.am`.
If we really want to we could also preserve the uninstalled library location.

eli-schwartz:
One possible difference between recursive and non-recursive autotools make, is that you can do per-subdirectory targets of all kinds -- not just building, but installing too. This is similar to what the Meson PR could do using 0.60.0 `install_tag` support.

It's probably not actually important. Avoiding recursive make is a tremendous benefit and worth paying many prices, and this is a really small one.

TheOneric:
> One possible difference between recursive and non-recursive autotools make, is that you can do per-subdirectory targets of all kinds -- not just building, but installing too.

Regarding installing, this should be completely irrelevant for libass, as the library and headers from the `libass` subdirectory and the pkg-config-file are the only things which get installed anyway. Regarding building, there still are distinct targets for each library or program, so the current `make -C compare` will be equivalent to the non-recursive `make bin/compare` (or adjusted to wherever we decide to put the executables). Cleaning rules would no longer be able to be run per current subdirectory, but to me this doesn't seem all that useful to begin with.

MrSmile (commented):
I'm certainly prefer to not move testing sources one level deeper.
As for paths, I think all variants look ugly, so I'm with **rcombs** here: automation is better, after that is hardcoded approach.
Toplevel binary location is good, but that conflicts with not moving sources, so I think we should keep the current location.

TheOneric:
As the most conservative approach appears to also be the most popular one, here it is: no source-tree changes, all uninstalled targets remain in the same location, our minimal Automake x<=1.11 does not need to be bumped *(and the paths are all hardcoded, but tbf we’re unlikely to ever rename or move the `libass/` directory)*.

Note that I increased the separator between source files to 2 spaces as I otherwise found the list harder to read now that it also contains the directory.

Issue #404, opened 2020-06-16
Opentype lang tag


adipose:
libass currently supports a two character lang tag.  Harfbuzz (a dependency of libass) uses up to 3 characters.  It seems recommended to support the same limit in libass.  

see https://github.com/libass/libass/blob/master/libass/ass.c line 615

and https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-ot-tag-table.hh

adipose:
It also seems harfbuzz may support even longer strings with subtypes.

TheOneric:
Currently libass only considers [ISO 639-1 codes valid](https://github.com/libass/libass/commit/c979365946b2dc2499ede862b6f7da15f9bc0ed1). As it is supported by Harfbuzz, allowing users to specify languages not covered by ISO 639-1 seems sensible to me. Eg ~~Ainu `ain`~~ *(doesn't seem to be recognized by Harfbuzz; Lower Sorbian `dsb` is an actually supported example)*
Afaik the `Language` tag is a libass extension – so no compatibility concerns.

Some things to consider:
 [Harfbuzz uses IETF BCP 47 codes](https://github.com/harfbuzz/harfbuzz/commit/2f1f961cc0f0f907916b4b00342d8a2dda8b4ee1) to denote languages and their variants. Not all libraries can understand this. It's probably a good idea to advise users to use two-letter ISO 639-1 codes when possible, as those are more widely understood.

E.g. pr #372 needs ISO 639-1 to work with libunicode and probably should also chop off anything starting with the first dash `-` (variants) before passing it to libunicode.

adipose:
The directwrite api also uses the "locale" parameter, which may have limitations.  I believe libass will pass the Language tag to this field, but I'm not entirely sure they are BCP 47 codes.

TheOneric:
> The directwrite api also uses the "locale" parameter, which may have limitations. I believe libass will pass the Language tag to this field, but I'm not entirely sure they are BCP 47 codes.

LIbass currently does **not** pass any [language to directwrite](https://github.com/libass/libass/blob/master/libass/ass_directwrite.c#L440).  
I'm not familiar with directwrite, but I don't think there's any benefit in passing language info to it, as afaik – by libass – it is only used to get the correct font/fallback and not glyph rendering.

adipose:
> 
> 
[...]

Issue #319, opened 2018-11-02
Fonts not working on greek characters 


peterthunder:
I have a .mkv anime episode that contains .ass subtitles and the fonts do not work on the greek characters when using mpv.

How it should look like: https://cdn.discordapp.com/attachments/507680995311616000/507681693956571147/unknown.png
The font only works on the english characters.

How it actually looks like:
https://cdn.discordapp.com/attachments/507680995311616000/507681810734514176/unknown.png

A sample of the video can be found here: https://mega.nz/#!tFYRSYKC!ennXFJ7osIvzisl4RGtNfS2syw0NQPwzV6b0R9KvPZo

astiob:
Here’s what I _think_ is happening here:

The font in question has a Microsoft symbol character map. The characters in such fonts are encoded in the Unicode private use area with codes starting at F000. What we currently do for symbol charmaps in `ass_font_index_magic` is just bitwise-or the desired code with 0xF000. This works for base ASCII, but, as it turns out, not for other stuff such as the Greek in this example. What is actually supposed to be done, as far as I can tell, is that we convert the original Unicode codepoint to the appropriate Windows 8-bit encoding and then add 0xF000 to _that_ codepoint.

For example, I’m getting this error on the provided sample file:

    [ass] Glyph 0x3A3 not found, selecting one more font for (AG_Avero_C, 400, 0)

U+03A3 is the Greek capital sigma. FontForge shows me a capital sigma at 0xF0D3 in these fonts. https://en.wikipedia.org/wiki/Windows-1253 shows that the capital sigma is 0xD3, which matches this index.

This seems to match the behaviour described at https://blogs.msdn.microsoft.com/murrays/2016/01/04/wingdings-with-cyrillic-greek-etc-characters/.

**Ideal fix in libass:** do what Word, Excel and RichEdit do and figure out the Windows encoding for the character based on its Unicode script property. I’m not sure which encodings are supposed to be used in case of ambiguities, but I’m guessing 1250–1258 and 874. (There are probably ambiguities even within this set though…)

**Fix for subtitle author:** avoid symbol fonts. In this case, the font being a symbol font makes no sense. Find real Unicode fonts that look the same or somehow make your own out of these symbol fonts.

[...]

PR #607, opened 2022-04-05
fontselect: use language from track for fallback selection

### LINKED ISSUE #550: Take track language into account when performing font fallback

rcombs:
Closes #550.

We'll want to encourage players to set `track->Language` to any value they might have from file metadata. Demo mpv patch: https://gist.github.com/1619f3a49ee72b08aa4428b2a7cfa70c.

This doesn't remove the existing code that removes the default FC_LANG value on Fontconfig, as I'm still not entirely sure what cases that was referencing having issues with.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -884,7 +884,12 @@ static int process_info_line(ASS_Track *track, char *str)
         char *p = str + 9;
         while (*p && ass_isspace(*p)) p++;
         free(track->Language);
-        track->Language = strndup(p, 2);
+        track->Language = strdup(p);
+        if (track->Language) {
+            char *end = track->Language + strlen(track->Language);
+            rskip_spaces(&end, track->Language);
+            *end = 0;
+        }
     } else if (!strncmp(str, "ScriptType:", 11)) {
         check_duplicate_info_line(track, SINFO_SCRIPTTYPE, "ScriptType");
         parse_script_type(track, str + 11);
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -29,6 +29,10 @@
 
 #include "ass_coretext.h"
 
+#ifndef __has_builtin
+#define __has_builtin 0
+#endif
+
 #define SAFE_CFRelease(x) do { if (x) CFRelease(x); } while (0)
 
 static const ASS_FontMapping font_substitutions[] = {
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -33,6 +33,12 @@
 #define __has_builtin 0
 #endif
 
+#if __has_builtin(__builtin_available)
+#define CHECK_AVAILABLE(sym, ...) __builtin_available(__VA_ARGS__)
+#else
+#define CHECK_AVAILABLE(sym, ...) (!!&sym)
+#endif
+
 #define SAFE_CFRelease(x) do { if (x) CFRelease(x); } while (0)
 
 static const ASS_FontMapping font_substitutions[] = {
@@ -252,7 +258,8 @@ cleanup:
 }
 
 static char *get_fallback(void *priv, ASS_Library *lib,
-                          const char *family, uint32_t codepoint)
+                          const char *family, uint32_t codepoint,
+                          ASS_StringView locale)
 {
     CFStringRef name = CFStringCreateWithCString(NULL, family, kCFStringEncodingUTF8);
     if (!name)
@@ -273,7 +280,25 @@ static char *get_fallback(void *priv, ASS_Library *lib,
         return NULL;
     }
 
-    CTFontRef fb = CTFontCreateForString(font, r, CFRangeMake(0, 1));
+    CTFontRef fb = NULL;
+#if defined(__MAC_10_9)
+#ifndef __MAC_10_15
+    // This function was available as early as 10.9, but wasn't exposed in the headers. We'll create our own compatible declaration.
+    CTFontRef CTFontCreateForStringWithLanguage(
+        CTFontRef               currentFont,
+        CFStringRef             string,
+        CFRange                 range,
+        CFStringRef _Nullable   language) __OSX_AVAILABLE_STARTING(__MAC_10_9, __IPHONE_7_0);
+#endif
+    if (CHECK_AVAILABLE(CTFontCreateForStringWithLanguage, macOS 10.9, iOS 7.0, watchOS 2.0, tvOS 9.0, *)) {
+        CFStringRef locStr = locale.len ? CFStringCreateWithBytes(NULL, (const UInt8*)locale.str, locale.len, kCFStringEncodingUTF8, false) : NULL;
+        fb = CTFontCreateForStringWithLanguage(font, r, CFRangeMake(0, 1), locStr);
+        SAFE_CFRelease(locStr);
+    } else
+#endif
+    {
+        fb = CTFontCreateForString(font, r, CFRangeMake(0, 1));
+    }
     CFRelease(font);
     CFRelease(r);
     if (!fb)
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -250,8 +250,7 @@ cleanup:
 static char *get_fallback(void *priv, ASS_Library *lib,
                           const char *family, uint32_t codepoint)
 {
-    CFStringRef name = CFStringCreateWithBytes(
-        0, (UInt8 *)family, strlen(family), kCFStringEncodingUTF8, false);
+    CFStringRef name = CFStringCreateWithCString(NULL, family, kCFStringEncodingUTF8);
     if (!name)
         return NULL;
 
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -32,16 +32,8 @@
 
 #define MAX_NAME 100
 
-typedef struct fc_locale_private {
-    ASS_StringView locale;
-    FcFontSet *fallbacks;
-    FcCharSet *fallback_chars;
-} ProviderLocale;
-
 typedef struct fc_private {
     FcConfig *config;
-    size_t nb_locales;
-    ProviderLocale **locales;
 } ProviderPrivate;
 
 static bool check_postscript(void *priv)
@@ -81,16 +73,6 @@ static void destroy(void *priv)
 {
     ProviderPrivate *fc = (ProviderPrivate *)priv;
 
-    for (size_t i = 0; i < fc->nb_locales; i++) {
-        if (fc->locales[i]->fallback_chars)
-            FcCharSetDestroy(fc->locales[i]->fallback_chars);
-        if (fc->locales[i]->fallbacks)
-            FcFontSetDestroy(fc->locales[i]->fallbacks);
-        if (fc->locales[i]->locale.str)
-            free((char*)fc->locales[i]->locale.str);
-        free(fc->locales[i]);
-    }
-    free(fc->locales);
     FcConfigDestroy(fc->config);
     free(fc);
 }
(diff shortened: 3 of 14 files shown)


TheOneric:
Currently the `Language` Script Info header only accepts two-letter ISO 3166-1 codes. Any extension of the supported notation should imo also be made for the in-file header (which is a libass-specific extension) not just for manual struct assignments.

Something to consider is that different libraries we interact with support different language notations. Your commit mentions UTS#35, HarfBuzz appears to use IETF BCP 47 codes as noted in [#404](https://github.com/libass/libass/issues/404#issuecomment-648306753) and libunibreak can only handle ISO 3166-1 iirc.
I don't know how compatible UTS#35 and IETF BCP 47 are, at first glance they appear to have at least some similarities.
If this PR lands before #486 it doesn't need to consider the only-ISO-3166-1 case much, but we should recommend to prefer two-letters codes over three-letter codes or numeric notations when available. If it lands after #486 the language tag must be sanitised before being passed to libunibreak.

libass/ass_fontselect.h
```diff
@@ -148,13 +148,15 @@ typedef void    (*SubstituteFontFunc)(void *priv, const char *name,
  * \param lib ASS_Library instance
  * \param family original font family name (try matching a similar font) (never NULL)
  * \param codepoint Unicode codepoint (UTF-32)
+ * \param locale UTS#35 locale to use for the lookup, or NULL for system default
```
astiob, line 151: Minor: the mention of UTS#35 seems awfully specific. Do we know that our font provider backends specifically expect _Unicode locale identifiers?_ Does HarfBuzz (which is what the tracks’s `Language` is currently used for) expect the same? And at any rate, we only currently read two bytes of the `Language` header (but technically a consumer may put arbitrary data in through the API), although IIRC we’ve had requests to fix—er, expand this.

astiob (commented):
One clear bug aside, looks okayish.

I’m worried by the linearly searched, quadratically reallocated array in `ass_fontconfig`, but then one would hope this should be irrelevant. But as an idea, how about making the `ProviderLocale`s form a singly-linked list? That would avoid the quadratic reallocation and possibly look cleaner (as `nb_locales` and `ASS_REALLOC_ARRAY` would be unnecessary), while of course maintaining the linear search.

Do I understand correctly that `locale` is needed in `ass_font_desc` because a single font cache may be used with multiple `ASS_Track`s?

libass/ass_fontconfig.c
```diff
@@ -190,11 +201,11 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
     }
 }
 
-static void cache_fallbacks(ProviderPrivate *fc)
+static void cache_fallbacks(ProviderPrivate *fc, ProviderLocale *pLoc)
 {
     FcResult result;
 
-    if (fc->fallbacks)
+    if (!pLoc->fallbacks)
```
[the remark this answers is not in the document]
rcombs: Whoops! Fixed.

libass/ass_fontselect.h
rcombs: UTS#35 is what CoreText expects; Fontconfig takes RFC3066 in a way that makes both forms effectively equivalent. We could call it "RFC 3066" here if we wanted since that's technically slightly stricter.

rcombs:
Before landing this, we should probably tweak the `Language` header parsing and add sanitization on the field regardless. I've been experimenting with this file:
```
#include <fontconfig/fontconfig.h>
#include <stdio.h>

int main(int argc, char** argv)
{
    if (argc < 2)
        return 1;

    printf("%s -> %s\n", argv[1], FcLangNormalize(argv[1]));

    return 0;
}
```
Notably, most 3-letter codes don't get normalized to their 2-letter forms, so I'd expect 3-letter language codes wouldn't match the 2-letter values in fonts. This means that we'll need to either tell callers to pre-normalize the language codes they pass, or do it ourselves internally; the latter is probably preferable.

We should also consider what to do in the no-language case. Currently, this leaves that as status quo, which means that the results are system-locale-specific in CoreText and DirectWrite, but not in fontconfig. I think we should probably figure out a way to make this locale-specific on fontconfig as well, tbh; the user's locale is a better guess for an appropriate font language than whatever coincidental behavior we currently have.

Additionally, what should we do for cases where the caller wants to set the default language to something other than whatever the providers autodetect? Should we just tell consumers to set the track language in that case, or provide a separate API?

One potential issue with fontconfig: on a lot of systems, the default font for all CJK languages will be something like `Noto Sans CJK JP`. For instance, `fc-match :lang=ko` on my Linux system gives `NotoSansCJK-Regular.ttc: "Noto Sans CJK JP" "Regular"`. Sometimes this is due to fontconfig weirdness (I have absolutely no idea why this isn't picking the perfectly good `Noto Sans CJK KR` install that the config files point to), but some systems might actually only have one of the fonts installed, either because they're slightly smaller or because they want that language to be the default. I _think_ Harfbuzz should handle this, but we should verify that. It also means that if we decide to disable locale-variant glyph selection by default in the future (for vsfilter compat), we'll want to make sure the change only applies to attached fonts and authoring mode.

libass/ass_fontselect.h
rcombs: ("BCP 47" is another common name for these)

TheOneric: Fontconfig according to its documentation accepts “the form Ll-Tt where Ll is a two or three letter language from ISO 639 and Tt is a territory from ISO 3166”.

Notably, this differs from BCP47 in that the latter can have a “script” identifier before the region identifier which Fontconfig doesn't expect or understand and in that BCP47 requires the shortest ISO 369 language code to be used while Fontconfig doesn't.

If we take BCP47 then without maintaining our own database of identifiers, we can iinm pass it as is to HarfBuzz, check if the first part consists of two letters and only pass it to libunibreak if that’s the case. It appears like script tags are always 4letter while region tags can have 2 or 3 letters. We could use this to sanitise the identifier to a form accepted by Fontconfig. Or can `FcLangNormalize` already handle this?

Nit: IETF BCP 47 actually contains RFC5646, which obsoleted the RFC obsoleting RFC3066.

rcombs:
Force-pushed with:
- Language field extended to support longer strings
- Docs changed to indicate that the value should be a BCP47 tag containing only language and region

I've also tested behavior when using fonts that support multilingual glyph variants (e.g. `Noto Sans CJK JP`): Harfbuzz _does_ ensure this works correctly. If you select `Noto Sans CJK JP`, but pass language values of `zh-CN`, `zh-TW`, `zh-HK`, `ja`, or `ko`, you get the correct variants for e.g. U+8FD4:

| Lang | Glyph |
|---|---|
| zh-CN | <span lang="zh-CN">返</span> |
| zh-TW | <span lang="zh-TW">返</span> |
| zh-HK | <span lang="zh-HK">返</span> |
| ja | <span lang="ja">返</span> |
| ko | <span lang="ko">返</span> |

libass/ass.c
```diff
@@ -884,7 +884,17 @@ static int process_info_line(ASS_Track *track, char *str)
...
+        track->Language = strdup(p);
+        if (track->Language) {
+            size_t len = strlen(track->Language);
+            for (size_t i = len; i-- > 0;) {
+                if (track->Language[i] == ' ' ||
+                    track->Language[i] == '\t')
+                    track->Language[i] = 0;
+                else
+                    break;
+            }
```
TheOneric, line 896: Can't this just use `rskip_spaces` and insert a single `\0` as needed?

libass/ass_types.h
```diff
@@ -274,7 +274,7 @@ typedef struct ass_track {
     int WrapStyle;
     int ScaledBorderAndShadow; // 0 or 1 (boolean)
     int Kerning; // 0 or 1 (boolean)
-    char *Language; // zero-terminated ISO-639-1 code
+    char *Language; // if non-null, zero-terminated BCP47 code including language and optional region
```
TheOneric, line 277: I feel like this can be misread as requiring not only the the – usually optional – region part to be included in `Language`’s BCP47 code.
Due to its importance for interoperability with ISO-639-1-only interfaces, it may also be worth to reiterate BCP47’s requirement to use 2letter codes when available.

TheOneric (commented):
It may be cleaner to split extension to BCP47 codes and `Language` usage for font fallback into separate commits.

Since `WRAP_UNICODE` was merged half a year ago, this part from my first comment here is now relevant:
> Something to consider is that different libraries we interact with support different language notations. Your commit mentions UTS#35, HarfBuzz appears to use IETF BCP 47 codes as noted in https://github.com/libass/libass/issues/404#issuecomment-648306753 and libunibreak can only handle ISO 3166-1 iirc.
> [...]
> If it lands after https://github.com/libass/libass/pull/486 the language tag must be sanitised before being passed to libunibreak.

rcombs:
Rebased with comments applied, and:
- Use ASS_StringView instead of a bare `const char*` in new internal API
- Have the cache items own a copy of the language tag instead of relying on the track's value never changing
- Keep runtime support for macOS versions prior to 10.9 (since we apparently care about that)
- Split into as many distinct commits as is reasonable

Anybody wanna test this on Windows?

rcombs:
I did some testing on fontconfig and found that the earlier version of this _sorta_ worked, but got tripped up by the `get_fallbacks` function, which gave us a long list of fallbacks to try, which would often include one for a different language before the one we were looking for. I don't think it's possible to make that list prefer locale-matching languages (or, if it _is_ possible, I think it'd be slower than what I ended up landing on without being better in any meaningful way).

The new version of this PR has a commit that vastly simplifies our fallback behavior on fontconfig (relative to the existing code, and _especially_ relative to the previous version of the PR). It now simply asks fontconfig for a match with the passed-in family, locale, and code point. This gives reasonable results for me on macOS for all tested inputs except Hong Kong Chinese (due to [this upstream configuration issue](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues/353)), and gives reasonable results for all cases on Linux.

The main downside is that it'll be slightly slower in the case where we'd previously have simply looked up a font in the substitutions list, and the slow lookup runs once per font name that we need to fallback from. It should be much faster than the `FcFontSort` call we used to make in the `get_fallback` case, but that's offset to some extent by the results not being cached.

Also in this version:
- Added a compile-time SDK version check for the macOS language check code
- Added an `ASS_TEST_FORCE_FONTCONFIG` env var check in the `test` program to simplify testing that path on systems with other font providers

libass/ass_coretext.c
```diff
@@ -270,7 +274,18 @@ static char *get_fallback(void *priv, ASS_Library *lib,
         return NULL;
     }
 
-    CTFontRef fb = CTFontCreateForString(font, r, CFRangeMake(0, 1));
+    CTFontRef fb = NULL;
+#if __has_builtin(__builtin_available) && defined(MAC_OS_X_VERSION_10_9)
+    if (__builtin_available(macOS 10.9, iOS 7.0, *)) {
```
astiob, line 279: It turns out this symbol isn’t actually declared in the C headers until the 10.15 SDK. As discussed on IRC, let’s declare it ourselves and hope for the best.

Further, as discussed in https://github.com/libass/libass/pull/664#discussion_r1119732310, in https://github.com/libass/libass/pull/664/commits/77e4bfb977ebd8bcde56f21da88760046a025f39 and on IRC, the following will restore compatibility with GCC and older Clang, which are viable compilers on 10.9, and make the `__builtin_available` polyfill redundant:

```suggestion
#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 1090 || __IPHONE_OS_VERSION_MAX_ALLOWED >= 70000
    CTFontRef CTFontCreateForStringWithLanguage(CTFontRef currentFont, CFStringRef string, CFRange range, CFStringRef language) __OSX_AVAILABLE_STARTING(__MAC_10_9, __IPHONE_7_0);
    if ((intptr_t) &CTFontCreateForStringWithLanguage) {
```

plus `#include <Availability.h>` (or we can rely on the fact that it already happens to be included transitively, but I’d slightly prefer an explicit include),

assuming I can trust the claim in Availability.h that `__MAC_OS_X_VERSION_MAX_ALLOWED` is undefined when targeting iOS.

rcombs: This introduces a `-Wpartial-availability` warning when targeting pre-10.9, which is often configured as an error.

astiob: Current docs say `-Wpartial-availability` is an alias for `-Wunguarded-availability` which is the exact flag that warns about `availability`, and as you and I have both mentioned previously, it is only enabled by default in its `-Wunguarded-availability-new` which affects 10.13+ APIs, not 10.9 APIs. Our Makefile doesn’t enable it, either. This is no different from #664.

Unless perhaps you actually mean some other warning that pertains to the declaration rather than the invocation?

If you want us to support build setups that explicitly add `-Werror=unguarded-availability` to extra cflags, we need completely separate compilation paths (or a macro) for new Clang and all other compilers.

libass/ass_fontconfig.c
```diff
@@ -202,165 +184,78 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
...
+        FcPatternDestroy(pat);
+
+    return ret;
 }
 
 static ASS_FontProviderFuncs fontconfig_callbacks = {
     .check_postscript   = check_postscript,
     .check_glyph        = check_glyph,
     .destroy_provider   = destroy,
-    .get_substitutions  = get_substitutions,
```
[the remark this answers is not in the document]
rcombs: Ah, I've been testing this using `test/test.c`, which sets `default_family` to `"sans-serif"`. With that configuration and this patchset, the usual `serif`, `sans-serif`, and `monospace` aliases work fine for me. I get `DejaVuSerif`, `DejaVuSans`, and `DejaVuMono` for `en`, with CJK fallback on `MS-Mincho` or `MS-Gothic`; for `zh-CN`, I get `SimSun`, `ArialUnicodeMS`, and `NSimSun`; for `ja`, I get `MS-Mincho`, `MS-Gothic`, and `MS-Gothic`. This is all as expected.

With `Arial`, though, I get the same behavior you're describing here. I'm not sure what the best solution for this is; I can't see any way to implement the `get_substitutions` route in a locale-aware manner.

libass/ass_coretext.c
```diff
@@ -248,10 +252,10 @@ static void match_fonts(void *priv, ASS_Library *lib, ASS_FontProvider *provider
 }
 
 static char *get_fallback(void *priv, ASS_Library *lib,
-                          const char *family, uint32_t codepoint)
+                          const char *family, uint32_t codepoint,
+                          ASS_StringView locale)
 {
-    CFStringRef name = CFStringCreateWithBytes(
-        0, (UInt8 *)family, strlen(family), kCFStringEncodingUTF8, false);
+    CFStringRef name = CFStringCreateWithCString(0, family, kCFStringEncodingUTF8);
```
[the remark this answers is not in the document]
rcombs: I ended up removing that usage after checking and finding that I don't get reliably consistent retain counts on these objects; best not to risk it.

libass/ass_coretext.c
rcombs: Re-pushed with a new `CHECK_AVAILABLE` macro, which should work fine for both new and old compilers.

libass/ass_fontconfig.c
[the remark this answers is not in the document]
FlyGoat: Sorry to chime in here, was bothered by mpv's font selection process for a while, so I decided to do my investigation.
`FcFontMatch` is supposed to work with alias, because it performs partial match on `FC_FAMILY`, so `ass_fontconfig` won't fail at first instance for `{\fnserif}`.
For `Arial` I guess that's because @rcombs does not have `Arial Unicode MS` installed. That's the only `Arial` family font with CJK support out there.
IMO `get_substitutions` is not really useful with `fontconfig`, given that it already handles family alias well, but perhaps performing a search without locale if first search fails can handle this situation better.

moi15moi:
Is there a reason why this PR is stalling?
It would help resolving: https://github.com/libass/libass/issues/319

libass/ass_fontselect.h
TheOneric: > It appears like script tags are always 4letter while region tags can have 2 or 3 letters. We could use this to sanitise the identifier to a form accepted by Fontconfig. Or can FcLangNormalize already handle this?

With the documented restrictions on which BCP-47 subtag we take this _should_ not be a problem anymore and just work if every API users follows docs

TheOneric:
> Is there a reason why this PR is stalling?

Yes, see the open review threads: it breaks fontconfig substitutions

_(It technically also is an API change (`Language` is no longer either `NULL` or exactly three bytes long) for readers of the struct, but this should not be an issue for any remotely sane code imho.)_

PR #612, opened 2022-04-15
Unsafe leftshifts and float cast overflows

TheOneric:
This needs some discussion wrt VSFilter-compatibility and could probably be expanded to cover more casts not found in the samples yet; see commit messages. Reproducing samples here: [612_more-ub-fixes.zip](https://github.com/libass/libass/files/8498916/612_more-ub-fixes.zip)
Also see here for the tests probing what happens in practice on x86 WIndows for the relevant UB: https://github.com/TheOneric/libass/runs/6033094405?check_suite_focus=true#step%3A7%3A115=

@MrSmile: I haven't encountered such a sample yet, but from a quick glance it seems like the leftshift in `quantize_blur` can potentially end up with assigning -1 to the `shadow_mask` if the radius is so large that `ord > 31`. Would that be a problem or did I miss some guards protecting from this scenario?

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -662,7 +662,8 @@ static void blend_vector_clip(ASS_Renderer *render_priv, ASS_Image *head)
     ol_key.u.drawing.text = render_priv->state.clip_drawing_text;
 
     double m[3][3] = {{0}};
-    double w = render_priv->font_scale / (1 << (render_priv->state.clip_drawing_scale - 1));
+    int32_t scale_base = lshiftwrapi(1, render_priv->state.clip_drawing_scale - 1);
+    double w = scale_base > 0 ? (render_priv->font_scale / scale_base) : 0;
     m[0][0] = render_priv->font_scale_x * w;
     m[1][1] = w;
     m[2][2] = 1;
@@ -1105,7 +1106,8 @@ get_outline_glyph(ASS_Renderer *priv, GlyphInfo *info)
             return;
         }
 
-        double w = priv->font_scale / (1 << (info->drawing_scale - 1));
+        int32_t scale_base = lshiftwrapi(1, info->drawing_scale - 1);
+        double w = scale_base > 0 ? (priv->font_scale / scale_base) : 0;
         scale.x = info->scale_x * w;
         scale.y = info->scale_y * w;
         desc = 64 * info->drawing_pbo;
--- a/libass/ass_utils.h
+++ b/libass/ass_utils.h
@@ -182,6 +182,12 @@ static inline int double_to_d22(double x)
     return lrint(x * 0x400000);
 }
 
+static inline int32_t lshiftwrapi(int32_t i, int32_t shift)
+{
+    // '0u +' to avoid automatic integer promotion causing UB
+    return (0u + (uint32_t)i) << (shift & 31);
+}
+
 static inline int mystrtod(char **p, double *res)
 {
     char *start = *p;
--- a/libass/ass_utils.h
+++ b/libass/ass_utils.h
@@ -182,14 +182,6 @@ static inline int double_to_d22(double x)
     return lrint(x * 0x400000);
 }
 
-static inline int mystrtoi(char **p, int *res)
-{
-    char *start = *p;
-    double temp_res = ass_strtod(*p, p);
-    *res = (int) (temp_res + (temp_res > 0 ? 0.5 : -0.5));
-    return *p != start;
-}
-
 static inline int mystrtod(char **p, double *res)
 {
     char *start = *p;
--- a/libass/ass_utils.h
+++ b/libass/ass_utils.h
@@ -190,14 +190,6 @@ static inline int mystrtoi(char **p, int *res)
     return *p != start;
 }
 
-static inline int mystrtoll(char **p, long long *res)
-{
-    char *start = *p;
-    double temp_res = ass_strtod(*p, p);
-    *res = (long long) (temp_res + (temp_res > 0 ? 0.5 : -0.5));
-    return *p != start;
-}
-
 static inline int mystrtod(char **p, double *res)
 {
     char *start = *p;
--- a/libass/ass_drawing.c
+++ b/libass/ass_drawing.c
@@ -19,8 +19,6 @@
 #include "config.h"
 #include "ass_compat.h"
 
-#include <ft2build.h>
-#include FT_OUTLINE_H
 #include <math.h>
 #include <stdbool.h>
 #include <limits.h>
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -768,10 +768,10 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
         } else if (tag("be")) {
             double dval;
             if (nargs) {
-                int val;
+                int32_t val;
                 dval = argtod(*args);
                 // VSFilter always adds +0.5, even if the value is negative
-                val = (int) (render_priv->state.be * (1 - pwr) + dval * pwr + 0.5);
+                val = dtoi32(render_priv->state.be * (1 - pwr) + dval * pwr + 0.5);
                 // Clamp to a safe upper limit, since high values need excessive CPU
                 val = (val < 0) ? 0 : val;
                 val = (val > MAX_BE) ? MAX_BE : val;
@@ -816,7 +816,7 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
             if (render_priv->state.effect_timing)
                 render_priv->state.effect_skip_timing +=
                     render_priv->state.effect_timing;
-            render_priv->state.effect_timing = val * 10;
+            render_priv->state.effect_timing = dtoi32(val * 10);
         } else if (tag("shad")) {
             double val, xval, yval;
             if (nargs) {
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -36,13 +36,6 @@ struct arg {
     char *start, *end;
 };
 
-static inline int argtoi(struct arg arg)
-{
-    int value;
-    mystrtoi(&arg.start, &value);
-    return value;
-}
-
 static inline int32_t argtoi32(struct arg arg)
 {
     int32_t value;
@@ -230,7 +223,7 @@ static bool parse_vector_clip(ASS_Renderer *render_priv,
 
     int scale = 1;
     if (nargs == 2)
-        scale = argtoi(args[0]);
+        scale = argtoi32(args[0]);
 
     struct arg text = args[nargs - 1];
     render_priv->state.clip_drawing_text.str = text.start;
@@ -367,11 +360,11 @@ char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
                 render_priv->state.fay = 0.;
         } else if (complex_tag("iclip")) {
             if (nargs == 4) {
-                int x0, y0, x1, y1;
-                x0 = argtoi(args[0]);
-                y0 = argtoi(args[1]);
-                x1 = argtoi(args[2]);
-                y1 = argtoi(args[3]);
+                int32_t x0, y0, x1, y1;
+                x0 = argtoi32(args[0]);
+                y0 = argtoi32(args[1]);
+                x1 = argtoi32(args[2]);
+                y1 = argtoi32(args[3]);
                 render_priv->state.clip_x0 =
                     render_priv->state.clip_x0 * (1 - pwr) + x0 * pwr;
                 render_priv->state.clip_x1 =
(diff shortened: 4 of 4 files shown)


libass/ass_utils.h
```diff
@@ -182,19 +183,37 @@ static inline int double_to_d22(double x)
     return lrint(x * 0x400000);
 }
 
+static inline int double_to_int(double d)
+{
+    // For the usual 64bit double, 32bit int this
+    // differs for exactly INT_MAX.
+    // For other sizes (int) ((double) INT_MAX) may be UB
+    // if INT_MAX is not accurately representable as a double
```
rcombs, line 206: This comment is inaccurate: if `((double)INT_MAX)` is not accurately representable, it's rounded in an implementation-defined manner. It's only UB if `INT_MAX` is out of range for `double` _and_ IEEE arithmetic isn't supported. We only really support 64-bit-double anyway, though, so it's all kinda moot.

rcombs: Also, couldn't we use `[l]lrint` or `[l]lround` here, and then return `INT_MIN`/`LLONG_MIN` on `fetestexcept(FE_INVALID)`? This would only work on IEEE platforms, but… do we care about non-IEEE platforms?

I looked through every platform on godbolt and the only ones without `math_errhandling & MATH_ERREXCEPT` I found were: 
- AVR
- KVX
- MRISC32
- MSP430
- RISC-V rv32gc
- ESP32 (inc. S2/S3)

Out of these platforms, only rv32gc seems to be a plausible target for libass, and as far as I can tell, it's only missing this because of lack of existing implementation support (the hardware supports it).

We could also use these checks and return `INT_MIN` in cases where we're sure the result will be incorrect, and otherwise use `lrint`/`lround` to handle the straggler cases in a manner that's portable for conformant IEEE platforms, and otherwise may return some edge-case value (but is never UB).

TheOneric: The `(int)` is not meant to be a comment about what type can have different sizes, but part of the expression, so the full expression is `(int) ((double) INT_MAX)` with two casts.

The first one from an integer to a floating-point type is to my understanding¹ indeed not UB as you pointed out if absolutely larger regular numbers or ±Infinity exists *(e.g. IEEE)*.
The comment was about the second cast, from a floating-point to integer type, though which will be UB if `*_MAX` is not accurately representable and the implementation-defined rounding to the available precision adjusts the value away from zero. Then the value of `(double) *_MAX` is larger than the original `INT_MAX` value and ISO C99, section 6.1.3.4 makes it very clear that casting a float type whose integral part cannot be represented in the targeted integer representation is undefined behaviour *(yes, even `(unsigned) -2.4`)*. That's why the check has to be `>=` and not `>` and in fact this does happen on my machine for the `long long` version of this function.

----

¹: The exact wording in section 6.1.3.4 is: 
> “If the value being converted is in the range of values that can be represented but cannot be represented exactly, the result is either the nearest higher or nearest lower representable value, chosen in an implementation-defined manner. If the value being converted is outside the range of values that can be represented, the behavior is undefined.”

I interpret this to mean that if an integer value is larger than the largest regular number the floating type can represent it can still be "represented" as `INFINITY` if it exists, though I haven't checked the rest of the spec to closely about possible restrictions/exceptions about `±INFINITY`. If that’s correct, then only something like ARM's alternative 16bit float mode could have UB in the first cast; assuming IEEE float seems reasonable to me though.

astiob: For `int32_t`, this already exists in `ass_parse.c` as `dtoi32` since #261. And it handles `INT(32)_MAX` _without_ deviating from x86 by using `INT32_MAX + 1LL`. Other than that, it’s identical to your code (notably with equally-strict comparison signs), which is reassuring.

libass/ass_utils.h
```diff
@@ -182,19 +183,52 @@ static inline int double_to_d22(double x)
...
+    bool neg_i = i < 0;
+    if (neg_i) i = -i;
+
+    // Some compilers behave differently for negative compile-time constants
+    // (e.g. on Linux gcc, clang and tcc - but not MSYS2's gcc on Windows), see:
+    // if (shift < 0)
+    //     u = ((uint32_t)i) >> -shift;
+    // else
+    uint32_t u = ((uint32_t)i) << (((uint32_t)shift) & 31);
+    i = u > INT32_MAX ? -(int32_t)(~u) - 1 : u;
```
astiob, line 197: Elsewhere, we just cast `uint32_t` to `int32_t`, and we seem to have had the consensus opinion that it is good enough given the definition of `int32_t` as specifically two’s complement (even in C99). If we want to be extra pedantic and avoid this implementation-defined behaviour, then those other places should be sought out and adjusted as well. Some instances are in `interpolate_alpha` in `ass_parser.c` since #261, but there may be more.

libass/ass_render.c
```diff
@@ -662,7 +662,8 @@ static void blend_vector_clip(ASS_Renderer *render_priv, ASS_Image *head)
     ol_key.u.drawing.text = render_priv->state.clip_drawing_text;
 
     double m[3][3] = {{0}};
-    double w = render_priv->font_scale / (1 << (render_priv->state.clip_drawing_scale - 1));
+    int32_t scale_base = lshiftwrapi(1, render_priv->state.clip_drawing_scale - 1);
+    double w = render_priv->font_scale / (scale_base ? scale_base : 1);
```
astiob, line 666: Here and below, it may make more sense to set `w = 0` or something if `!scale_base`, because I imagine VSFilter wouldn’t display anything for such an item, as it would end up with every single coordinate inside the drawing equal to `INT32_MIN`, reducing the drawing to an infinitesimal, off-screen point.

astiob (commented):
We don’t actually use `mystrtoll` any more, so that entire function can be dropped and `double_to_ll` need not be added.

And where we use `mystrtoi`—via `argtoi`, which is its only direct use—perhaps we should use `argtoi32` instead, allowing us to similarly drop `mystrtoi` and `double_to_int`.

libass/ass_utils.h
TheOneric: I forgot casts from integer type to signed integer type are never UB *(unlike non-cast overflows and double→int casts)*, but "only" have an implementation defined result or raise an implementation-defined signal(?). Given fixed width types are required to be two-complements, jsut casting them seems pretty reasonable then, or if we wanted to make really sure  `*(int32_t *) &uint32_variable` would work too and be probably optimised away by the compiler.

libass/ass_utils.h
TheOneric: Switched all `argtoi` usage to argtoi32; I'm assuming this should also improve VSFilter-compatibility in some edge cases unless VSF sometimes uses a function which first parses an int as a double like `argtoi` did.

libass/ass_render.c
TheOneric: I haven't checked the following calculations too closely, but using `0` seems sensible so it does now.

MrSmile: Do we even need to emulate never used VSFilter bugs? It's ugly cumbersome code without any tangible benefit. Also, it's compatibility not with some existing VSFilter source (as it's UB that should be fixed there too), but with compiled binary with some particular behavior for this UB. I think that
```c
double w = ldexp(priv->font_scale, 1 - info->drawing_scale);
```
is much cleaner and more intuitively correct in this case.

MrSmile:
> if the radius is so large that `ord > 31`.

Maximal blur radius is 100, so overflows there can be only due to the large `blur_scale`. I think it's better to sanitize that than deal with its consequences in `quantize_blur`. But if you think otherwise, you can replace `ord` with `FFMIN(ord, 31)` or something.

TheOneric:
Simplified `lshiftwrapi`, the conditional negation was not necessary for the MSYS2 and general non-compile-constnat behaviour. However, if I'm not mistaken the `(uint32_t)` casts do not protect from UB in (the theoretical) case of `UINT32_MAX < INT_MAX < 2^63 - 1` due to automatic promotion to `int`. To deal with that, there's now a preprocessor check and an alternative version.

TheOneric:
> Maximal blur radius is 100, so overflows there can be only due to the large blur_scale.

Correct me if I got anything wrong. Looking at the relevant macros, it appears to me like any reasonable choices would not let `ord` grow beyond 32. But ord itself can become negative if `POSITION_PRECISION` is defined to something smaller than 2. Maybe we should add a comment to its defintion warning that it must be at least `2.0`?

MrSmile:
Even with `POSITION_PRECISION = 1.0` `ord` would be nonnegative for nonnegative `radius`, so there's no problem here.

TheOneric:
Right, the significant from `frexp` is in `[0.5, 1)` not `[1, 2)` so only `POSITION_PRECISION < 1` can lead to negative `ord`. Then at least in `ass_render.c` all but the two shifts already identified by AFL++ should be safe, great!

libass/ass_render.c
TheOneric: How did you conclude that this particular quirk is “never used”? I'm inclined to default to sticking closely to VSFilter, since there are many weird VSFilter quirks relied upon in real-world scripts and emulating this is sufficiently easy *(and I expect it to probably not negatively effect sane and otherwise VSFilter-compatible scripts)*.

MrSmile: This quirk doesn't introduce any new behavior that can be useful, hard to trigger accidentally, and no authoring/automation tool output it (actually anything except `\p1` is outdated by floating-point coordinates extension). So there's basically no way it can creep into real subs outside of fuzzing. There's already a ton of widely used VSFilter bugs, so I think it's better to focus on that than trying to emulate every possible UB from its source code.

libass/ass_utils.h
```diff
@@ -182,6 +182,20 @@ static inline int double_to_d22(double x)
...
+#if UINT32_MAX > INT_MAX
+    // If INT_MAX >= UINT32_MAX, but also INT_MAX < 2^63 - 1
+    // this may still be UB due to auto int promotion
+    return ((uint32_t)i) << (((uint32_t)shift) & 31);
+#else
+    // but this on the other hand is wrong if UINT_MAX < UINT32_MAX.
+    // Extra variable to prevent the compiler from dropping the uint32_t cast.
+    uint32_t tmp = ((unsigned)i) << (((uint32_t)shift) & 31);
+    return tmp;
+#endif
```
astiob, line 196: Could you elaborate on the extra-variable hack? I don’t immediately understand how it can matter.

Keeping in mind that `int32_t` is always two’s complement, I don’t think the cast in `(uint32_t)shift` is necessary, either, is it?

Anyway, how about simplifying all of this to just:
```c
    return 0u + (uint32_t) i << (shift & 31);
```

libass/ass_utils.h
astiob: > unless VSF sometimes uses a function which first parses an int as a double like `argtoi` did.

No, FWIW I don’t _think_ it ever does this, and I distinctly remember this being one of our longstanding quirks. But I’m not quite sure why I didn’t replace those calls immediately when I introduced `argtoi32`. Maybe it’s worth giving these instances another quick cross-check with VSFilter source code just in case.

libass/ass_utils.h
TheOneric: > Could you elaborate on the extra-variable hack? I don’t immediately understand how it can matter.

I'm pretty sure that as far as the spec is concerned, it should be identical to using no extra variable but just a cast to `uintN_t)` *(and before the last push, it did use the latter)*. In practice, testing with `uint8_t`, the compiler apparently to drop both the `unsigned` and `uint8_t` cast and then the sanitizer complains about UB because `left shift of 128 by 25 places cannot be represented in type 'int'` unless the temporary variable is used. Arguably this seems to be more of a compiler/sanitizer bug.

Using your proposed `0u + (uintN_t)i)` instead also makes the compiler and sanitizer happy _and_ allows to drop the preprocessor check, so let’s go with this instead :)

libass/ass_utils.h
TheOneric: From quick glance a guliverkli2, all paramters replaced here and the `fade` paramters already replaced in #261 are using `wcstol` and afaik in all current Microsoft platforms `long` is 32bit.

libass/ass_render.c
TheOneric: Sufficiently high enough drawscale values can easily occur if an additional key was hit by accident while typing. How common this is or not and the remains unnoticed with VSFilter- and “current-libass on x86”-behaviour, I do not know.  After some rounds of simplification, the emulation code is now just one line. And since the effects of drawscale values higher than 31 can also be achieved by decimal coordinates as you pointed out, sub authors don't gain anything useful by us not emulating VSFilter here, but everyone will end up with a new compatibility concern/issue. Thus I believe explicitly emulating VSF on leftshifts is the better choice.

TheOneric:
After discussion on IRC, the division by zero check was replaced by a positivity check since the value will never be zero but can be `INT32_MIN`, we'd like to avoid negative `w` and this shouldn't matter too much for drawings portable between guliverkli2- and xy-VSFilter as far as I can tell. See reworded commit message for details.

If there are no objections to treating `\p31` as an effective `\fscx0`, mistakes in my interpretation of *VSF code or delays requested, I'm intending to merge this soon.

astiob:
> treating `\p31` as an effective `\fscx0`

\*`\p32`

TheOneric:
> *`\p32` (the shift amount is the value minus one)

oups; now fixed, thx.

Issue #610, opened 2022-04-12
Negative Fontsize: assert fail in outline_stroke


TheOneric:
[Once](https://github.com/libass/libass/issues/397)  [again](https://github.com/libass/libass/issues/431) this assert fails on some inputs.

```
libass/ass_outline.c:1498: outline_stroke: Assertion `rad >= eps && rad <= OUTLINE_MAX' failed.
```
Tested on top of cbeea94fc70a6322ecce4c4ff30bef2a9028c18a, see attached archive for three sample files triggering the failure and info on the selected fonts: [assert-radeps-fail.zip](https://github.com/libass/libass/files/8477053/assert-radeps-fail.zip)

If asserts are disabled, a stack-overflow occurs in `process_quadratic` instead *(old versions of libass just produced slightly odd renderings and did not crash)*.

EDIT: The `fuzzer_w3:id:000159,sig:04,src:000968+000719,time:452375,execs:57769,op:splice,rep:16` sample already hits UB before failing the assert which might cause the issue for this particular case. The other two samples do not hit any UB afaict.
For this one sample, the preceeding UB is:
```
libass/ass_render.c:1108:42: runtime error: shift exponent 9999999 is too large for 32-bit type 'int'
```

MrSmile:
Cases `id:000000` and `id:000248` are unsanitized (negative) `Fontsize` style field. I think we should do the same as [for other fields](https://github.com/libass/libass/blob/master/libass/ass.c#L557-L561). Case `id:000159` is indeed [an UB with excessive shift](https://github.com/libass/libass/blob/master/libass/ass_render.c#L665). I think we should forget about emulating VSFilter UB in this case and fix it by replacing with corresponding `ldexp()` call.

TheOneric:
> Case id:000159 is indeed [an UB with excessive shift](https://github.com/libass/libass/blob/master/libass/ass_render.c#L665).

For reference, I do not observe an excessive shift at the linked location but at `libass/ass_render.c:1108` for the `id:000159` sample with `profile` or my fuzzing consumer — for other samples excessive and negative shifts do occur at `libass/ass_render.c:665` though, see #612.

TheOneric:
> (negative) Fontsize style field. I think we should do the same as [for other fields](https://github.com/libass/libass/blob/master/libass/ass.c#L557-L561) [i.e. `FFMAX(val, 0)`].

[...]

PR #615, opened 2022-04-24
Don't crash on negative Fontsize

TheOneric:
See #610 for samples and VSFilter screenshots *(the assert fail in the `id:000159` samples is actuallly only fixed by #612 not this, which is why it is not listed in the commit)*.
Before this changes libass crashes if a Style with a negative `Fontsize` field is rendered either by failing an assert or if asserts are disabled by overflowing the stack. This brings the rendering closer in line with VSFilter and prevents crashes.

The change isn't fully correct though. Upon closer inspection the positive and negative fontsizes aren't quite identical in VSFilter (but would be with this change); the negative version is slightly **larger** than the positive one. I have no idea where this difference is coming from. At first glance the only bit affected by the sign of the fontsize in VSFilter is the `(LONG)(x+0.5)` rounding at the very beginning of `RTS.cpp` — but this should at most *reduce* the absolute value for negative sizes; here the opposite is observed. Any ideas what might be causing this?

For comparison, here are libass screenshots with v1 of the patch applied:
libass-patched_01
libass-patched_02

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -1991,7 +1991,7 @@ static bool parse_events(ASS_Renderer *render_priv, ASS_Event *event)
         info->effect_timing = render_priv->state.effect_timing;
         info->effect_skip_timing = render_priv->state.effect_skip_timing;
         info->font_size =
-            render_priv->state.font_size * render_priv->font_scale;
+            fabs(render_priv->state.font_size * render_priv->font_scale);
         info->be = render_priv->state.be;
         info->blur = render_priv->state.blur;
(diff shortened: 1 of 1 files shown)


TheOneric:
I still have no clue where the difference to VSFilter/GDI comes from. On IRC astiob mentioned font metrics also changing in VSF for large positive size values beyond some threshold, which might or might not have the same source.
I'm considering to merge this as a "closer than before and doesn't crash anymore" interim solution for the release and open a new issue to track the different font metrics for negative and large positive fontsizes.

PR #622, opened 2022-05-14
Add fuzzing utility

TheOneric:
Last commit will be dropped before a merge and included in a follow up patch series. It is included as a preview to illustrate how `ASS_FUZZMODE` or its alternative `FUZZ_CPPFLAGS` (better matching `FUZZ_LDFLAGS`) will affect it.

This adds the fuzzer used to find the recently fixed UB, e.g. negative fontsize. Having it in the repo simplifies reproducing the issues for others, allows us to utilise it in ART which will be more efficient than profile and no longer requires the start and end times of samples to be known. It can also be used in OSS-Fuzz, which currently has its own fuzzer source producing suspected bogus memory leaks we couldn't reproduce locally *(the version here always fully reinitialises and cleans up the library handles, so this shouldn't happen anymore)*.

As for the OSS-Fuzz-specific build target, after some experimenting I believe this manual linking is the least ugly way to allow OSS-Fuzz to use the new fuzzer.  
Including the link rules here will also enable us to push fixes (like for the memory leaks) or updates if anything about out build changed without having to go through a CLAed PR in OSS-Fuzz’s repo. It certainly isn't ideal to include OSS-Fuzz-specifics into our build system though, so if you don't like it don't hesitate to complain about it. We could leave the final link command in OSS-Fuzz’s own build rules.

For reference, this is how OSS-Fuzz could be updated to make use of these changes (removal of old fuzzer and dict omitted): 
[ossfuzz-diff.txt](https://github.com/libass/libass/files/8693984/ossfuzz-diff.txt)

----


fuzz/fuzz.c
```diff
@@ -0,0 +1,240 @@
...
+
+        ASS_Track *track = ass_read_memory(ass_library, (char *)buf, len, NULL);
+        if (!track)
+            continue;
+        consume_track(ass_renderer, track);
+
+        ass_free_track(track);
+        ass_clear_fonts(ass_library);
+        // Renderer caches cannot be cleared (cache_limit just updates the bound values)
+        // Re-calling set_fonts destroys and re-initialises the fontselector at least
```
astiob, line 208: It also clears the font and shaper caches and calls `ass_reconfigure`, which in turn clears the composite, bitmap and outline caches… which, I think, is all of our caches.

And isn’t it unsafe to call `ass_clear_fonts` before clearing the font-related caches? Should that be moved below the `ass_set_fonts` call?

fuzz/ass.dict
```diff
@@ -0,0 +1,192 @@
...
+"ScriptType"
+"PlayResX"
+"PlayResY"
+"Timer"
+"WrapStyle"
+"ScaledBorderAndShadow"
+"YCbCr Matrix"
+"Kerning"
+"Language"
+#"; Script generated by FFmpeg/Lavc"
```
astiob, line 48: Is this commented out intentionally?

fuzz/fuzz.c
```diff
@@ -0,0 +1,240 @@
...
+#ifndef ASS_FUZZMODE
+    #define ASS_FUZZMODE FUZZMODE_STANDALONE
+#endif
+
+ASS_Library *ass_library;
+ASS_Renderer *ass_renderer;
+
+void msg_callback(int level, const char *fmt, va_list va, void *data)
+{
+#if ASS_FUZZMODE == FUZZMODE_STANDALONE
```
MrSmile, line 40: I think we should always generate the full error string here to check error reporting. Also calculate simple hash (xor?) from the result string into some global variable to force uninitialized memory check and keep optimizers from removing dead code.

fuzz/fuzz.c
```diff
@@ -0,0 +1,240 @@
...
+    if (!ass_renderer) {
+        ass_library_done(ass_library);
+        ass_library = NULL;
+        printf("ass_renderer_init failed!\n");
+        return false;
+    }
+
+    ass_set_fonts(ass_renderer, NULL, "sans-serif",
+                  ASS_FONTPROVIDER_AUTODETECT, NULL, 1);
+    ass_set_frame_size(ass_renderer, 854, 480);
```
MrSmile, line 68: I think it's better to make some variations here to check bitmap alignment issues and such. Maybe we can use some seed value from the start of the input buffer (or command line parameter for standalone mode) to generate random sizes in some range.

fuzz/fuzz.c
```diff
@@ -0,0 +1,240 @@
...
+
+    return true;
+}
+
+static void consume_track(ASS_Renderer *renderer, ASS_Track *track)
+{
+    for (int n = 0; n < track->n_events; ++n) {
+        int change;
+        ASS_Event event = track->events[n];
+        ass_render_frame(ass_renderer, track, event.Start, &change);
```
MrSmile, line 79: I think that instead of 3 fixed time moments it's better to render only once, but at random time using the aforementioned seed. Also we should really "consume" results by calculating bitmap hash into global variable.

fuzz/fuzz.c
```diff
@@ -0,0 +1,240 @@
...
+    if (!init()) {
+        printf("library init failed!\n");
+        return 1;
+    }
+
+    __AFL_INIT();
+    buf = __AFL_FUZZ_TESTCASE_BUF;
+    while (__AFL_LOOP(100000)) {
+        len = __AFL_FUZZ_TESTCASE_LEN;
+        if (len < 50) // Can't be a useful file
```
MrSmile, line 195: I think that such arbitrary magic values aren't needed (maybe except `len < sizeof(seed)` check). Also there could be potential problems that emerge only for small inputs.

fuzz/ass.dict
TheOneric: Kind of, AFL++ was warning that such a large single dictionary entry may negatively affect fuzzing performance. Considering one sample with the comment will suffice for coverage and the effects can be acheived otherwise the fuzzer probably doesn't even need to be told about this being an active sequence. It would have been better to just remove it, but I didn't because I missed it before opening the PR — now I removed it properly.

fuzz/fuzz.c
TheOneric: Now that you mention it, yes, [FT_New_Memory_Face](https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_new_memory_face)'s notes clearly spell out: “You must not deallocate the memory before calling FT_Done_Face”. Maybe I just got lucky not to run into any problems with it. Switched it around now.

fuzz/fuzz.c
```diff
@@ -0,0 +1,235 @@
...
+     * -fsanitize-undefined-trap-on-error which will SIGILL without an
+     * error report being printed), see https://reviews.llvm.org/D35085
+     */
+    enum {
+        FUZZ_OK = 0,
+        //SANITISER_FAIL = 1,
+        // Invalid parameters passed etc
+        FUZZ_BAD_USAGE = 2,
+        // Error before rendering starts
+        FUZZ_INIT_ERR = 0
```
TheOneric, line 135: I set this to zero after some fuzzer-generated files were rejected by our parser as `TRACK_TYPE_UNKNOWN`, in the cases at hand because the fuzzer inserted a null-byte before the styles sections but iinm it could also happen if there was no Styles section at all. There might be some worth in keeping this enum for better communicating intent?

On a side note, iirc this differs from VSFilters which use the ScriptInfo `ScriptType` header and if at all reject based on that. Do we actually want to reject Styles-less files?

fuzz/ass.dict
```diff
@@ -0,0 +1,191 @@
...
+#   and non-printable characters must be escaped as \x<hex>.
+
+# Some common font names
+"Arial"
+"DejaVu Sans"
+
+# Various common terminals
+"yes"
+"no"
+"0x"
```
TheOneric, line 20: I’m actually not sure if `0x` is also accepted as a hexadecimal prefix anywhere.

astiob: It is (`mystrtoi32`). In colour headers, even `0x0x` is accepted.

fuzz/fuzz.c
TheOneric: Other than possibly very small and very large frame/storage_sizes triggering something about the (re)alloc strategy, it’s not clear too me what code paths could be unlocked by varying the size not reachable with this fixed value. Also varying the sizes and codepaths based on some random(ly selected) bytes might confuse the fuzzing engines code coverage based mutator, doing more harm than good.

fuzz/fuzz.c
TheOneric: Regarding “random time” the same comment as above applies.

As for processing the bitmaps to trigger sanitisers on it, this sounds like probably a good idea. If hashing them turns out to be costly still, maybe this could also be another `ifdef`, in which case `ASS_FUZZMODE` would be replaced with `FUZZ_CPPFLAGS`.

fuzz/fuzz.c
MrSmile: No, I mean instead of real printing it's better to use `vsnprintf` and hash its result.

fuzz/fuzz.c
MrSmile: I mean not real random, but `deterministic_noise_function(seed)`, so it should be perfectly ordinary for fuzzing engine to mutate `seed`.

fuzz/fuzz.c
MrSmile: We don't need a good hash, the only goal is to introduce dependency, so something like xor of all bytes should be enough and easily vectorizable.

fuzz/fuzz.c
TheOneric: I was already assuming deterministic behaviour for identical seed. AFL and libFuzzer aren't just doing naïve mutations, though they try to maximize coverage. If now alteration of say the last few bits changes the initial setup and thus possibly a lot of branches, this might interfere with their ability to explore the new paths deeper.
Furthermore, rendering for larger frame sizes is more expensive and as noted before, I don't see much benefit in it other than for extremly small or extremly large frame_sizes. All other effects of different frame sizes should also be able to be triggered by altering the input.

*(Side note: as for your other suggestions, I want to measure the performance impact first to decide if this will be always be enabled or conditionally via another macro define. It might take a few days until I get to that.)*

TheOneric:
Any more comments on this, or is it good to merge as is?

TheOneric:
As noted on IRC, the memory font cleanup for the AFL++ variant wasn't conforming to FreeType and libass documentation and has now been changed to be by fully recreating the renderer handle for each sample.
In testing, at least with the version of FreeType used, the old code "worked" without issues, but [`FT_New_memory_Face`'s documentation](https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_new_memory_face) states: *“You must not deallocate the memory before calling [FT_Done_Face](https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_done_face).”*. Ignoring this probably isn't a good idea.

If there are any question, concerns left or a planned review speak up now. In absence of such I intended to merge this soon.

TheOneric:
Here's patch making OSS-Fuzz use the newly added consumer instead of a custom one, tested on the final version: [0001-libass-use-upstream-s-fuzzing-consumer-and-dict.patch.gz](https://github.com/libass/libass/files/9049039/0001-libass-use-upstream-s-fuzzing-consumer-and-dict.patch.gz). If someone, who actually can submit patches to OSS-Fuzz, wants to pick it up that’d be great.

PR #637, opened 2022-08-14
ass_parse: fix premature truncation on \t(\Xa)

rcombs:
This caused animations to out-of-range alpha values to be clipped to the uint8_t range _before_ performing the lerp, rather than only after. VSFilter only truncates after. Using e.g. `\1a0\t(\1a1FF)` should run through the entire 0-FF range _twice_.


astiob:
Looks like an accidental regression in 9b04e56ff6e1ed82bc48ac7307e627bdbbaf082d (0.15.0)? Thanks for catching this. We should add a test for this, now that we have tests.

PR #645, opened 2022-09-06
PlayResX scaling and Effect’s delay precison and (non-)scaling

TheOneric:
First commit is just a rebased from astiob’s [draft PlayResX fix](https://github.com/astiob/libass/commit/a550fc5115bd36cb9d0510976d7357b609410645). Afaict this is functionality-wise correct and was only missing the Effect `delay` unscaling. @astiob: Was this the missing thing you mentioned or is there something I overlooked?

Rebasing was mostly just moving things around, but also dealing with the newly emulated „shadow origin“ quirks, and dropping one change as the affected code was removed when the SBAS=no defaults were adjusted before 0.15.0.

One possible issue: anonymous unions used in this patch to alias `font_scale` and `screen_scale_y` are not part of ISO C99 our current baseline *(with some not documented machine assumptions)*. However, those are part of ISO C11 and prior to that were an extension in GNU and some other compilers. So far all compilers I checked accepted it without complaints with default build options:
 - GCC supports it since [at least version 3.1.1](https://gcc.gnu.org/onlinedocs/gcc-3.1.1/gcc/Unnamed-Fields.html#Unnamed-Fields) *(2002-07)*
 - since GCC supported them before clang even existed, it’s likely every release of clang supports them though I haven't tested that
 - tcc *(tested: 0.9.27)*
 - chibicc *(tested: git main)*
 - cproc *(tested: master)*
 - and apparently MSVC [supports them as well](https://docs.microsoft.com/en-US/cpp/cpp/anonymous-class-types?view=msvc-170)

So perhaps we can just rely on them being supported and use it anyway.

---

Some artifical sample files and xy-VSF renderings: [645_samples.tar.gz](https://github.com/libass/libass/files/9499516/645_samples.tar.gz)

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -211,27 +211,27 @@ static ASS_Image *my_draw_bitmap(unsigned char *bitmap, int bitmap_w,
  */
 static double x2scr_pos(ASS_Renderer *render_priv, double x)
 {
-    return x * render_priv->orig_width / render_priv->font_scale_x / render_priv->track->PlayResX +
+    return x * render_priv->frame_content_width / render_priv->par_scale_x / render_priv->track->PlayResX +
         render_priv->settings.left_margin;
 }
 static double x2scr_left(ASS_Renderer *render_priv, double x)
 {
     if (render_priv->state.explicit || !render_priv->settings.use_margins)
         return x2scr_pos(render_priv, x);
-    return x * render_priv->fit_width / render_priv->font_scale_x /
+    return x * render_priv->fit_width / render_priv->par_scale_x /
         render_priv->track->PlayResX;
 }
 static double x2scr_right(ASS_Renderer *render_priv, double x)
 {
     if (render_priv->state.explicit || !render_priv->settings.use_margins)
         return x2scr_pos(render_priv, x);
-    return x * render_priv->fit_width / render_priv->font_scale_x /
+    return x * render_priv->fit_width / render_priv->par_scale_x /
         render_priv->track->PlayResX +
         (render_priv->width - render_priv->fit_width);
 }
 static double x2scr_pos_scaled(ASS_Renderer *render_priv, double x)
 {
-    return x * render_priv->orig_width / render_priv->track->PlayResX +
+    return x * render_priv->frame_content_width / render_priv->track->PlayResX +
         render_priv->settings.left_margin;
 }
 /**
@@ -239,7 +239,7 @@ static double x2scr_pos_scaled(ASS_Renderer *render_priv, double x)
  */
 static double y2scr_pos(ASS_Renderer *render_priv, double y)
 {
-    return y * render_priv->orig_height / render_priv->track->PlayResY +
+    return y * render_priv->frame_content_height / render_priv->track->PlayResY +
         render_priv->settings.top_margin;
 }
 static double y2scr(ASS_Renderer *render_priv, double y)
@@ -979,18 +979,18 @@ ASS_Vector ass_layout_res(ASS_Renderer *render_priv)
 
     ASS_Track *track = render_priv->track;
     if (settings->par <= 0 || settings->par == 1 ||
-            !render_priv->orig_width || !render_priv->orig_height)
+            !render_priv->frame_content_width || !render_priv->frame_content_height)
         return (ASS_Vector) { track->PlayResX, track->PlayResY };
     if (settings->par > 1)
         return (ASS_Vector) {
-            lround(track->PlayResY * render_priv->orig_width / render_priv->orig_height
+            lround(track->PlayResY * render_priv->frame_content_width / render_priv->frame_content_height
                     / settings->par),
             track->PlayResY
         };
     else
         return (ASS_Vector) {
             track->PlayResX,
-            lround(track->PlayResX * render_priv->orig_height / render_priv->orig_width
+            lround(track->PlayResX * render_priv->frame_content_height / render_priv->frame_content_width
                     * settings->par)
         };
 }
@@ -999,8 +999,8 @@ static void init_font_scale(ASS_Renderer *render_priv)
 {
     ASS_Settings *settings_priv = &render_priv->settings;
 
-    double font_scr_w = render_priv->orig_width;
-    double font_scr_h = render_priv->orig_height;
+    double font_scr_w = render_priv->frame_content_width;
+    double font_scr_h = render_priv->frame_content_height;
     if (!render_priv->state.explicit && render_priv->settings.use_margins) {
         font_scr_w = render_priv->fit_width;
         font_scr_h = render_priv->fit_height;
@@ -1148,7 +1148,7 @@ get_outline_glyph(ASS_Renderer *priv, GlyphInfo *info)
 
         int32_t scale_base = lshiftwrapi(1, info->drawing_scale - 1);
         double w = scale_base > 0 ? (1.0 / scale_base) : 0;
-        scale.x = info->scale_x * w * priv->screen_scale_x / priv->font_scale_x;
+        scale.x = info->scale_x * w * priv->screen_scale_x / priv->par_scale_x;
         scale.y = info->scale_y * w * priv->screen_scale_y;
         desc = 64 * info->drawing_pbo;
         asc = val->asc - desc;
@@ -1323,8 +1323,8 @@ static void calc_transform_matrix(ASS_Renderer *render_priv,
     double dist = 20000 * render_priv->blur_scale;
     z4[2] += dist;
 
-    double scale_x = dist * render_priv->font_scale_x;
-    double offs_x = info->pos.x - info->shift.x * render_priv->font_scale_x;
+    double scale_x = dist * render_priv->par_scale_x;
+    double offs_x = info->pos.x - info->shift.x * render_priv->par_scale_x;
     double offs_y = info->pos.y - info->shift.y;
     for (int i = 0; i < 3; i++) {
         m[0][i] = z4[i] * offs_x + x4[i] * scale_x;
@@ -1387,7 +1387,7 @@ get_bitmap_glyph(ASS_Renderer *render_priv, GlyphInfo *info,
 
         ASS_DVector bord = {
             64 * info->border_x * render_priv->border_scale_x /
-                render_priv->font_scale_x,
+                render_priv->par_scale_x,
             64 * info->border_y * render_priv->border_scale_y,
         };
         double width = info->hspacing_scaled + info->advance.x;
@@ -1428,7 +1428,7 @@ get_bitmap_glyph(ASS_Renderer *render_priv, GlyphInfo *info,
 
         double bord_x =
             64 * render_priv->border_scale_x * info->border_x / tr->scale.x /
-                render_priv->font_scale_x;
+                render_priv->par_scale_x;
         double bord_y =
             64 * render_priv->border_scale_y * info->border_y / tr->scale.y;
 
@@ -2143,7 +2143,7 @@ static bool parse_events(ASS_Renderer *render_priv, ASS_Event *event)
 
         if (!drawing_text.str) {
             info->hspacing_scaled = double_to_d6(info->hspacing *
-                    render_priv->screen_scale_x / render_priv->font_scale_x *
+                    render_priv->screen_scale_x / render_priv->par_scale_x *
                     info->scale_x);
             fix_glyph_scaling(render_priv, info);
         }
@@ -2367,7 +2367,7 @@ static void calculate_rotation_params(ASS_Renderer *render_priv, ASS_DRect *bbox
         while (info) {
             info->shift.x = info->pos.x + double_to_d6(device_x - center.x +
                     info->shadow_x * render_priv->border_scale_x /
-                    render_priv->font_scale_x);
+                    render_priv->par_scale_x);
             info->shift.y = info->pos.y + double_to_d6(device_y - center.y +
                     info->shadow_y * render_priv->border_scale_y);
             info = info->next;
@@ -2430,7 +2430,7 @@ static void render_and_combine_glyphs(ASS_Renderer *render_priv,
 {
     TextInfo *text_info = &render_priv->text_info;
     int left = render_priv->settings.left_margin;
-    device_x = (device_x - left) * render_priv->font_scale_x + left;
+    device_x = (device_x - left) * render_priv->par_scale_x + left;
     unsigned nb_bitmaps = 0;
     bool new_run = true;
     CombinedBitmapInfo *combined_info = text_info->combined_bitmaps;
@@ -2518,7 +2518,7 @@ static void render_and_combine_glyphs(ASS_Renderer *render_priv,
             assert(current_info);
 
             ASS_Vector pos, pos_o;
-            info->pos.x = double_to_d6(device_x + d6_to_double(info->pos.x) * render_priv->font_scale_x);
+            info->pos.x = double_to_d6(device_x + d6_to_double(info->pos.x) * render_priv->par_scale_x);
             info->pos.y = double_to_d6(device_y) + info->pos.y;
             get_bitmap_glyph(render_priv, info, &current_info->leftmost_x, &pos, &pos_o,
                              &offset, !current_info->bitmap_count, flags);
@@ -2558,7 +2558,7 @@ static void render_and_combine_glyphs(ASS_Renderer *render_priv,
 
         if (info->effect_type == EF_KARAOKE_KF)
             info->effect_timing = lround(d6_to_double(info->leftmost_x) +
-                d6_to_double(info->effect_timing) * render_priv->font_scale_x);
+                d6_to_double(info->effect_timing) * render_priv->par_scale_x);
 
         for (int j = 0; j < info->bitmap_count; j++) {
             info->bitmaps[j].pos.x -= info->x;
@@ -2946,9 +2946,9 @@ ass_render_event(ASS_Renderer *render_priv, ASS_Event *event,
     event_images->height =
         text_info->height + text_info->border_bottom + text_info->border_top;
     event_images->left =
-        (device_x + bbox.x_min) * render_priv->font_scale_x - text_info->border_x + 0.5;
+        (device_x + bbox.x_min) * render_priv->par_scale_x - text_info->border_x + 0.5;
     event_images->width =
-        (bbox.x_max - bbox.x_min) * render_priv->font_scale_x
+        (bbox.x_max - bbox.x_min) * render_priv->par_scale_x
         + 2 * text_info->border_x + 0.5;
     event_images->detect_collisions = render_priv->state.detect_collisions;
     event_images->shift_direction = (valign == VALIGN_SUB) ? -1 : 1;
@@ -3018,16 +3018,16 @@ ass_start_frame(ASS_Renderer *render_priv, ASS_Track *track,
     // PAR correction
     double par = render_priv->settings.par;
     if (par == 0.) {
-        if (render_priv->orig_width && render_priv->orig_height) {
-            double dar = ((double) render_priv->orig_width) /
-                         render_priv->orig_height;
+        if (render_priv->frame_content_width && render_priv->frame_content_height) {
+            double dar = ((double) render_priv->frame_content_width) /
+                         render_priv->frame_content_height;
             ASS_Vector layout_res = ass_layout_res(render_priv);
             double sar = ((double) layout_res.x) / layout_res.y;
             par = dar / sar;
         } else
             par = 1.0;
     }
-    render_priv->font_scale_x = par;
+    render_priv->par_scale_x = par;
 
     render_priv->prev_images_root = render_priv->images_root;
     render_priv->images_root = NULL;
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -672,9 +672,9 @@ static void blend_vector_clip(ASS_Renderer *render_priv, ASS_Image *head)
 
     double m[3][3] = {{0}};
     int32_t scale_base = lshiftwrapi(1, render_priv->state.clip_drawing_scale - 1);
-    double w = scale_base > 0 ? (render_priv->font_scale / scale_base) : 0;
-    m[0][0] = render_priv->font_scale_x * w;
-    m[1][1] = w;
+    double w = scale_base > 0 ? (1.0 / scale_base) : 0;
+    m[0][0] = render_priv->screen_scale_x * w;
+    m[1][1] = render_priv->screen_scale_y * w;
     m[2][2] = 1;
 
     m[0][2] = int_to_d6(render_priv->settings.left_margin);
@@ -999,21 +999,31 @@ static void init_font_scale(ASS_Renderer *render_priv)
 {
     ASS_Settings *settings_priv = &render_priv->settings;
 
+    double font_scr_w = render_priv->orig_width;
     double font_scr_h = render_priv->orig_height;
-    if (!render_priv->state.explicit && render_priv->settings.use_margins)
+    if (!render_priv->state.explicit && render_priv->settings.use_margins) {
+        font_scr_w = render_priv->fit_width;
         font_scr_h = render_priv->fit_height;
+    }
 
-    render_priv->font_scale = font_scr_h / render_priv->track->PlayResY;
-    render_priv->blur_scale = font_scr_h / ass_layout_res(render_priv).y;
-    if (render_priv->track->ScaledBorderAndShadow)
-        render_priv->border_scale =
-            font_scr_h / render_priv->track->PlayResY;
-    else
-        render_priv->border_scale = render_priv->blur_scale;
+    render_priv->screen_scale_x = font_scr_w / render_priv->track->PlayResX;
+    render_priv->screen_scale_y = font_scr_h / render_priv->track->PlayResY;
+
+    ASS_Vector layout_res = ass_layout_res(render_priv);
+    render_priv->blur_scale = font_scr_h / layout_res.y;
+    if (render_priv->track->ScaledBorderAndShadow) {
+        render_priv->border_scale_x = render_priv->screen_scale_x;
+        render_priv->border_scale_y = render_priv->screen_scale_y;
+    } else {
+        render_priv->border_scale_x = font_scr_w / layout_res.x;
+        render_priv->border_scale_y = render_priv->blur_scale;
+    }
 
     if (render_priv->state.apply_font_scale) {
-        render_priv->font_scale *= settings_priv->font_size_coeff;
-        render_priv->border_scale *= settings_priv->font_size_coeff;
+        render_priv->screen_scale_x *= settings_priv->font_size_coeff;
+        render_priv->screen_scale_y *= settings_priv->font_size_coeff;
+        render_priv->border_scale_x *= settings_priv->font_size_coeff;
+        render_priv->border_scale_y *= settings_priv->font_size_coeff;
         render_priv->blur_scale *= settings_priv->font_size_coeff;
     }
 }
@@ -1137,9 +1147,9 @@ get_outline_glyph(ASS_Renderer *priv, GlyphInfo *info)
         }
 
         int32_t scale_base = lshiftwrapi(1, info->drawing_scale - 1);
-        double w = scale_base > 0 ? (priv->font_scale / scale_base) : 0;
-        scale.x = info->scale_x * w;
-        scale.y = info->scale_y * w;
+        double w = scale_base > 0 ? (1.0 / scale_base) : 0;
+        scale.x = info->scale_x * w * priv->screen_scale_x / priv->font_scale_x;
+        scale.y = info->scale_y * w * priv->screen_scale_y;
         desc = 64 * info->drawing_pbo;
         asc = val->asc - desc;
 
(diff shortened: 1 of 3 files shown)


MrSmile:
I'm definitely against aliasing seemingly different variables with each other. Also we have `ASS_DVector` for x/y-pairs, better to use it. Also, I'm not sure, but isn't [anamorphic blur branch](https://github.com/libass/libass/commits/anamorphic-blur) related to the same problem too?

TheOneric:
> I'm definitely against aliasing seemingly different variables with each other.

I agree that non-obvious aliasing can be problematic/confusing when read *and writes* are involved. However, here the values are only written to in `init_font_scale` and otherwise only read, so there shouldn't be any issues. Imho, being able to refer to the scale factor which applies to fonts in all directions as just `font_scale` is a good thing and makes the code easier to understand than if one had to wonder why the x-direction is scaled by `screen_scale_y`.

An alternative would be to actually create `font_scale` and `screen_scale_y` as actually distinct memory locations, which are assigned the same value in `init_font_scale` — but then we'd have a redunandant value which could also result in bugs later on if the calcuation of one is updated but the other one forgotten. Apart from possible compiler/ISO-C99 concerns, aliasing seems like the superior approach to me.

> Also we have ASS_DVector for x/y-pairs, better to use it.

Sure. As the pairs is never passed around as parameters I don't think it’s necessary or easier to use the Vector Type, but it's also not more complicated so I guess why not. *(the only difference will be typing a `.` instead of the last `_`)*.

> Also, I'm not sure, but isn't anamorphic blur branch related to the same problem too?

Nope, the first commit fixes rendering for `PlayResX / StorageX  ≠  PlayResY / StorageY`. The second one fixes a superset thereof: `PlayRes  ≠  StorageRes`. But anamorphic blur is about `StorageRes ≠ VideoDisplayRes` or if you prefer `PAR ≠ 1`.

VSFilter scales `\blur` isotropical using the heights to determine the factor, but as it renders at storage resolution the finished blur gets de-queezed along with the video, creating a non-uniform blur appearance in display space. The required scaling here can already be expressed with pre-existing scale information iinm: `blur_scale` for y and `blur_scale * font_scale_x` for x. *(`font_scr_w / PlayResX` is not recorded in any scale value before the change at hand)*

astiob:
> Do not scale delay parameter of legacy effects
> VSFilter explicitly unscales those parameters, to make it relative to storage rather than Play* resolution

This wording doesn’t make sense to me. The delay is in milliseconds per pixel, and that’s a PlayRes pixel, not any other. The division by `m_scale` precisely ensures that it _is_ scaled from PlayRes to storage resolution.

libass is the one that isn’t scaling `delay` and is treating it as milliseconds per display pixel.

> Note: VSFilteralso [sic] stores the delay parameter as an integer after it has been scaled and also applies the max(., 1) after scaling. However, casting to int or applying FFMAX(., 1) to the denominator in libass actually results in different rendering than xy-VSFilter.

Keep in mind that VSFilter’s quotient is multiplied by 8 subpixels-per-pixel. Our `scroll_shift` is itself `int`, so we’re losing precision compared to VSFilter. Is this enough to explain the difference you’re seeing?

> Was this the missing thing you mentioned or is there something I overlooked?

No idea 😓 Do I understand correctly that you’ve gone and checked all instances of `m_scale` in VSFilter? If you have, that should be all.

But here’s what concerns me after a fresh new look at this code: there’s a bunch of `* priv->screen_scale_x / priv->font_scale_x`. Why is that a thing? It seems like an oxymoron. I trust it’s probably correct for our current pipeline, although I’d be delighted if this was double-checked with some quick tests; I do now recall that we, uh, work in anamorphic coordinates until the very end and then stretch them out? But do you think we should & could perhaps change it so that we instead stretch font glyphs early and work in display coordinates, such that the `font_scale_x` disappears here?

> I'm definitely against aliasing seemingly different variables with each other.

They aren’t different; that’s precisely the point. It is a single coefficient that affects two things and is therefore known by two names.

The alternatives are to replace `font_scale` by `screen_scale_x` everywhere and possibly have to add comments explaining why specifically `_x` is used, or to rename `screen_scale_x` to something like `screen_scale_x_and_font_scale`. (Or indeed duplicate the field and the value, just to avoid the word `union`.)

Then again, how many instances of `font_scale` do actually remain after this? Maybe there’s only one and it’s fine to use `screen_scale_x` and move the explanatory comment there?

> The required scaling [for blur] can already be expressed with pre-existing scale information iinm: `blur_scale` for y and `blur_scale * font_scale_x` for x.

Good point.

TheOneric:
> Do I understand correctly that you’ve gone and checked all instances of `m_scale` in VSFilter? If you have, that should be all.

I checked all instances of `m_scalex` in xy-VSFilter (not `m_scaley`).

> Keep in mind that VSFilter’s quotient is multiplied by 8 subpixels-per-pixel. Our scroll_shift is itself int, so we’re losing precision compared to VSFilter. Is this enough to explain the difference you’re seeing?

Possibly, I don't think I tested any scaling values not expressable as `n/8`. Though, unlike `y0` and `y1` in `Scroll *`, there's no `MAX_SUB_PIXEL` multipliers in VSF’s `delay` calculation prior to the `int` cast.
If the supersampling is the cause, do you think it’s worthwhile to simulate the rounding error for scaled `delay` here? Otoh I don't recall seeing any `((int) (8 * val)) / 8.0` constructs in libass.

> > Do not scale delay parameter of legacy effects
> > VSFilter explicitly unscales those parameters, to make it relative to storage rather than Play* resolution
>
> This wording doesn’t make sense to me. The delay is in milliseconds per pixel, and that’s a PlayRes pixel, not any other. The division by m_scale precisely ensures that it is scaled from PlayRes to storage resolution.

I agree that the title could be clearer and am open to suggestions.
I got to the “unscaled”/“non-scaled” naming, because for ASS, storage resolution is basically the native coordinate space and being set relative to any other grid, like the PlayRes canvas, requires scaling. See also how *non-scaled* borders and shadows (`ScaledBorderAndShadow: no`) are expressing storage res pixels, while *scaled* values refer to PlayRes canvas units.

> Then again, how many instances of `font_scale` do actually remain after this? Maybe there’s only one and it’s fine to use `screen_scale_x` and move the explanatory comment there?

Huh.. a simple grep actually only shows one remaining use:
https://github.com/libass/libass/blob/8c43c08e0da780af542f7f1829ad7db217f0b73b/libass/ass_render.c#L2098-L2099

> But here’s what concerns me after a fresh new look at this code: there’s a bunch of * priv->screen_scale_x / priv->font_scale_x. Why is that a thing? It seems like an oxymoron.

It scales to the storage frame fit into the margin-less rendering area iinm, so the storage aspect ratio is preserved. As for why, I haven't thought too hard about it tbh and just inherited it from your earlier commit after checking that my artificial samples now match VSFilter. *(`font_scale_x` is basically PAR, if `> 1` font glyphs will be stretched, if `< 1` they will be squeezed to simulate rendering at storage res)*.

astiob:
> I agree that the title could be clearer and am open to suggestions.
> I got to the “unscaled”/“non-scaled” naming, because for ASS, storage resolution is basically the native coordinate space and being set relative to any other grid, like the PlayRes canvas, requires scaling. See also how _non-scaled_ borders and shadows (`ScaledBorderAndShadow: no`) are expressing storage res pixels, while _scaled_ values refer to PlayRes canvas units.

Well, quite. But again, `delay` is in milliseconds ***per PlayRes pixel.*** It _should_ be scaled, but we currently don't. Your commit title and message suggest the opposite.

TheOneric:
> But again, `delay` is in milliseconds _**per PlayRes pixel.**_ It _should_ be scaled, but we currently don't. Your commit title and message suggest the opposite.

If I take the smae video file and the same ASS file except for a changed `PlayResX`, then the Banner effect still moves the same amount of *storage pixels* per second; see samples. So it appears to me like it is in *storage pixels*, not *PlayRes canvas units*.
*Before* this PR, reducing `PlayResX`, did accelarate the movement in *storage pixels* in libass, now it doesn't.

astiob:
> If I take the smae video file and the same ASS file except for a changed `PlayResX`, then the Banner effect still moves the same amount of _storage pixels_ per second

It doesn't for me.

astiob:
Don't forget that VSFilter rounds the delay to an integer as you pointed out yourself. Try setting the delay to 33 or 40, whatever your video's FPS is, to get 1 frame per pixel and hence 1 pixel per frame. Then halve the PlayRes. Then double the delay because the pixels have doubled in size.

astiob:
And speaking of this:

> Your samples seem to all use delay 0, that is 1 due to the `max(., 1)`, and that 1 is indeed 1 ms per storage pixel. But it is 1 ms per storage pixel completely regardless of whether the division by `m_scale` is present.

it could be that the difference you were seeing here:

> However, casting to int or applying FFMAX(., 1) to the denominator in libass actually results in different rendering than xy-VSFilter.

is caused by the same fact. libass with your PR in its current state substitutes 0 by 1 _first_ and _then_ divides by the scaling factor. VSFilter divides by the scaling factor first and then applies max, and if the original value is 0, it becomes 1 precisely due to the max and nothing else—and remains unscaled, as it was the zero that got scaled, not the one.

libass/ass_render.c
```diff
@@ -975,24 +975,37 @@ static void init_font_scale(ASS_Renderer *render_priv)
...
-            font_scr_h / render_priv->track->PlayResY;
-    else
-        render_priv->border_scale = render_priv->blur_scale;
+    if (render_priv->track->ScaledBorderAndShadow) {
+        render_priv->border_scale_x = render_priv->screen_scale_x;
+        render_priv->border_scale_y = render_priv->screen_scale_y;
+    } else {
+        if (settings_priv->storage_width)
+            render_priv->border_scale_x =
+                font_scr_w / settings_priv->storage_width;
```
astiob, line 998: I suspect this additionally needs `/ font_scale_x`. Otherwise it seems to include the anamorphic factor in itself, which the other branches don’t. Imagine a simple situation where PlayRes = display res = storage height but storage width is anamorphic.

To test this: anamorphic video, `ScaledBorderAndShadow: no`, storage size set, and
 the ASS contains… uh… borders or shadows?

astiob:
> I checked all instances of `m_scalex` in xy-VSFilter (not `m_scaley`).

Sounds good enough.

> Huh.. a simple grep actually only shows one remaining use

Great! Sounds like we have a solution then. Drop `font_scale` and just move the VSFilter-compatibility comment to the one place it remained in.

> If the supersampling is the cause, do you think it’s worthwhile to simulate the rounding error for scaled `delay` here? Otoh I don't recall seeing any `((int) (8 * val)) / 8.0` constructs in libass.

Right now it's sounding like the issue you were seeing was separate, but regardless, the rounding remains a fact. If I understand correctly, any rounding error/difference in the delay adds up very quickly to an obvious difference in the event's position, so yes, I think it'd be worthwhile to reproduce the rounding exactly. That's for the delay itself though, which doesn't involve any 8-factors. Sorry if I confused you.

The other part, which involves the 8-factor, is subpixel positioning of the event. Our `scroll_shift` is an integer, so we can move the event only by whole pixels. This is probably just a visual-quality issue, but it should be noticeable if the delay is large (and the scroll slow). Can we just change `scroll_shift` to a `double`?

---

Regarding `font_scale_x`: shouldn’t we perhaps completely disregard any user-supplied PAR if we have the storage size? Is it ever useful to override/disable `sub-ass-vsfilter-aspect-compat` while `sub-ass-vsfilter-blur-compat` is on? How much sense does the arithmetic even make in such a situation? The code, at least, is already starting to confuse me 😅 It’s always been confusing, but until now it was at least kinda warranted; but now if we have all the right sizes and scaling factors, it would be more intuitive to just scale horizontal things by `border_scale_x` and `screen_scale_x` and scale vertical things based on `border_scale_y`, `screen_scale_y` and `blur_scale`.

libass/ass_render.c
TheOneric: I don't see any difference between anamorphic shadow/border scaling between xy-VSFilter and libass with this patch, comparison screenshots: [645_compare_anamorphic_border-shadow.tar.gz](https://github.com/libass/libass/files/9507579/645_compare_anamorphic_border-shadow.tar.gz)
Since with `SBAS=no` borders and shadows are storage pixels, and those factors are all meant to scale to rendering frame width it does make sense to me to scale borders/shadows also by the anamorphic (de-)squeeze factor.

astiob: Right, that makes sense. Thanks for the tests!

I think I got lost in my logical labyrinth last night. It started when I got suspicious of `* border_scale_x / font_scale_x`, which I thought was doubly applying the anamorphism, but I failed to notice that this operation is the inverse and the change I ended up proposing in this thread would precisely make it worse!

libass/ass_render.c
```diff
@@ -975,24 +975,37 @@ static void init_font_scale(ASS_Renderer *render_priv)
...
+        else
+            render_priv->border_scale_x = 1.;
```
astiob, line 1000: Shouldn’t this be `screen_scale_x` to match the Y-axis scaling in the line below since d60f2dce1dcafaed0e21406213e66072f3036752?

astiob (changes requested):
Thanks for the update!

While I was away, it did occur to me that libass might’ve been handling `scroll_shift` as PlayRes-pixels and scaling it later, which explains another layer of misunderstanding. I was gonna check later, but you figured this out and pushed the new commits faster :-) But now looking at ass_render.c, I do indeed see this.

---

I’ve figured out what was _really_ confusing me about `font_scale_x`: it’s the inconsistency with which the first commit (and my original patch) changes `foobar_scale` into either `foobar_scale_x / font_scale_x` or just plain `foobar_scale_x`. Again, it may very well be _correct_ in the sense that the new code matches VSFilter behaviour, but in that case it seems that the commit message needs some extra elaboration: currently it sounds as if the commit only changes our behaviour in cases where the X scale (for some definition of X scale) differs from the Y scale, but the code seems to change part of our behaviour on _every_ anamorphic video, no matter how carefully you select/construct the sample’s `PlayResX`.

astiob:
And if we lack storage size but have `font_scale_x`, shouldn’t we calculate border/delay scales as if by estimating `storage_width` as `scr_w / PlayResX / font_scale_x` rather than (as we currently do) `scr_w / PlayResX`?

astiob:
Ah, maybe I meant to say something like `PlayResY * scr_w / scr_h / font_scale_x`? To scale consistently with the Y axis plus the _known_ aspect ratio regardless of whether `PlayResX` was originally the storage or display width?

libass/ass_render.c
TheOneric: Indeed, good catch, thanks.

libass/ass_parse.c
```diff
@@ -898,8 +898,22 @@ void apply_transition_effects(ASS_Renderer *render_priv, ASS_Event *event)
             render_priv->state.scroll_direction = SCROLL_RL;
 
         delay = v[0];
+        // VSF works in storage coordinates, but scales delay to PlayRes canvas
+        // before applying max(delay, 1). This means, if delay=0 we end up with
+        // 1 _storage pixel_ per second without any PlayRes scaling.
```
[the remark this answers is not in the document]
TheOneric: Done.

libass/ass_parse.c
```diff
@@ -898,8 +898,22 @@ void apply_transition_effects(ASS_Renderer *render_priv, ASS_Event *event)
...
+        // The way libass deals with delay, it is automatically relative to the
+        // PlayRes canvas, so we only need to "unscale" the delay=0 fallback.
+        // VSF also casts the scaled delay to int, but we cannot do this as this
+        // would ruin the scale factor we require for delay=0 and lead to noticeable
+        // deviations from VSFilter. But we still want to emulate the rounding error
+        // from VSF as this can cause noticeable differences over time.
+        double scale_x = (1.0 * render_priv->settings.storage_width) / render_priv->track->PlayResX;
```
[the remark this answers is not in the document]
TheOneric: Done.

libass/ass_parse.c
```diff
@@ -898,8 +898,22 @@ void apply_transition_effects(ASS_Renderer *render_priv, ASS_Event *event)
...
         if (delay == 0)
-            delay = 1;          // ?
+            delay = scale_x;
+        if (delay == 0) // PlayRes not set
```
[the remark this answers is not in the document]
TheOneric: Yep, now fixed, thanks.

TheOneric:
Summary of changes since v1:
 - no longer using an anonymous union, dropped `font_scale`
 - added commit to change scroll_shift to double
 - completely changed the delay scaling commit, as it turns out only `scaled delay < 1` is unscaled in VSFilter and added rounding error emulation *(correct emulation for delay non-zero but scaled less than 1 still pending)*
 
Here are the newly added samples and xy-VSF renderings: [645_samples_02.tar.gz](https://github.com/libass/libass/files/9521963/645_samples_02.tar.gz)

Will reply to remaining comments and suggestion later.

libass/ass_parse.c
```diff
@@ -898,8 +898,22 @@ void apply_transition_effects(ASS_Renderer *render_priv, ASS_Event *event)
...
+            delay = 1;
+        // Emulate VSFilter's rounding error
+        delay = ((int) (delay / scale_x)) * scale_x;
```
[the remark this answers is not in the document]
TheOneric: VSF’s `max(scaled_delay, 1)` should now be fully emulated and no div by zero should be possible. `storage_*` checks should also be more consistent now. Thanks!

TheOneric:
Changes since v2:
 - Effect delay: Emulate VSF’s `max` for negative and scaled to `< 1` values (untested)
 - x-scaling: tweak commit message and mention anamorphic scaling
 - New: Make storage res checks consistently check both dims for zero by already introducing `ass_layout_res` from the `LayoutRes*` patch
 - New: Improve storage res fallback when PAR info was supplied
 - New: sanitise some user-supplied values early in render_api
 - New RFC: prefer `set_storage_size` over `set_pixel_aspect`
 - New, unrelated:
     - minor spelling fix in docs
     - readme update and tweak

Commits were also reordered. Barring newly discovered defects, everything up to and including `render_api: sanitise user-supplied values early` should probably be okay to merge early if the rest takes longer.


Potential future changes?:
 - Rename `orig_{width,height}` to `render_*` or `frame_*`?
 - `screen_scale_*` aren't the best names, but the only alternative I can come up with is `default_scale_*` which also isn't great.
 - It would probably make sense to rename `font_scale_x` as the anamorphic correction actually applies to more than just fonts. I couldn't come up with an easily understood *and short* substitute for “anamorphic” though. Then again, as it’s basically PAR, just calling it `par`, or in line with other names `par_scale_x` might be better?
 - Possibly, use `ASS_DVector` for xy scaling pairs *(leaving only the anamorphic factor unpaired)*

TheOneric:
> Regarding `font_scale_x`: shouldn’t we perhaps completely disregard any user-supplied PAR if we have the storage size? Is it ever useful to override/disable sub-ass-vsfilter-aspect-compat while sub-ass-vsfilter-blur-compat is on?

I’m not really aware of any usecase for it. I even [once](https://github.com/libass/libass/pull/601) suggested maybe deprecating `ass_set_pixel_aspect` wholly in favour of `ass_set_storage_size`, but maybeee — but that’s unverified — setting PAR but not storage size, makes the result closer to MPC-HC in case a script targeted that? Even so, that wouldn't be a usecase for setting both storage and PAR but with incompatible values.
New version has a RFC commit for preferring storage size.

> Ah, maybe I meant to say something like `PlayResY * scr_w / scr_h / font_scale_x`? To scale consistently with the Y axis plus the known aspect ratio regardless of whether PlayResX was originally the storage or display width?

Probably doesn't affect many files but seems reasonable. Implemented in current version, but taking into account the [y-axis may be the one](https://github.com/mpv-player/mpv/issues/3948) being stretched for display.

> I’ve figured out what was really confusing me about `font_scale_x`: it’s the inconsistency with which the first commit (and my original patch) changes foobar_scale into either `foobar_scale_x / font_scale_x` or just plain `foobar_scale_x`. Again, it may very well be correct in the sense that the new code matches VSFilter behaviour, but in that case it seems that the commit message needs some extra elaboration

There already is some semmingly inconsistencies regarding `font_scale_x`’s use before the x-scaling patch (e.g. [1](https://github.com/libass/libass/blob/a48c98c274df25cc8d60cac1402948387d49840c/libass/ass_render.c#L2677) and [2](https://github.com/libass/libass/blob/a48c98c274df25cc8d60cac1402948387d49840c/libass/ass_render.c#L2330-L2332)), so for the current pipeline it’s probably correct the patch also is. Added some notes about anamorphic scaling to the commit message.
Streamlining the coordinate conversions at some point in the future to be more straightforward and consistent would be ideal.

astiob:
> e.g. [1](https://github.com/libass/libass/blob/a48c98c274df25cc8d60cac1402948387d49840c/libass/ass_render.c#L2677) and [2](https://github.com/libass/libass/blob/a48c98c274df25cc8d60cac1402948387d49840c/libass/ass_render.c#L2330-L2332)

FWIW, this difference may be one of those that are relatively easy to explain. I _think_ (based on my memory and on the code I’ve seen glimpses of within this PR, but without a comprehensive review of ass_render.c) that what’s happening in the existing code is that we compute _sizes_ in display pixels but _positions_ in anamorphic-compressed pixels, to ultimately transform them to display coordinates only at the very end ([here?](https://github.com/libass/libass/blob/a48c98c274df25cc8d60cac1402948387d49840c/libass/ass_render.c#L2910-L2914)).

astiob:
Oh, by the way, [1] is in `add_background`, which is the libass-only BorderStyle 4. I bet nobody has ever tested it with anamorphic video, neither before nor after the patch!

TheOneric:
Merged the first part with the Effect changes.

Any comments on the remaining and RFC commit(s) or suggested additional changes?

TheOneric:
Tentatively reworded the PAR preference commit message and added renames for `font_scale_x` to `par_scale_x` and `orig_*` to `frame_content_*`.

I still didn't find a better alternative for `screen_scale_*` and `ASS_DVector` changes are omitted as they'd just create a bunch of rebase conflicts for any potential future changes after feedback.

TheOneric:
If anyone plans to do more testing or take another look at this speak up now, otherwise I'll merge this soon.

libass/ass_render.c
```diff
@@ -978,7 +978,20 @@ ASS_Vector ass_layout_res(ASS_Renderer *render_priv)
         return (ASS_Vector) { settings->storage_width, settings->storage_height };
 
     ASS_Track *track = render_priv->track;
-    return (ASS_Vector) { track->PlayResX, track->PlayResY };
+    if (settings->par <= 0 || settings->par == 1)
+        return (ASS_Vector) { track->PlayResX, track->PlayResY };
+    if (settings->par > 1)
+        return (ASS_Vector) {
+            lround(track->PlayResY * render_priv->orig_width / render_priv->orig_height
```
astiob: This could benefit from an `orig_height != 0` guard (and similarly below). We do have such a guard in the default `par` calculation in `ass_start_frame`.

astiob (commented):
> Both omissions combined actually happen to cancel each other out for `PlayRes`-relative features if subs for an anamorphic video use isotropically scaled storage resolution as `PlayRes`, meaning libass was already rendering `PlayRes`-relative features of such subs correctly before by chance.
>
> However, any other choice of `PlayRes` exhibited at least one of those issues.

I believe this is not accurate.

In general, _no_ scenario with complex-enough typesetting on anamorphic video was rendered correctly in libass before this PR. _Neither_ PlayResX=anamorphic storage width _nor_ PlayResX=unsqueezed display width would always match the rendering of libass-after-this-PR. (Hence my insistence that this should be mentioned in the commit message.)

This is why this commit changes _some_ instances of `border_scale` to `border_scale` but some others to `border_scale_x / font_scale_x`. (Which is what kept confusing me, as I tried to point out previously.)

In particular, my test shows that \xshad was wrong (too close to the base glyphs); collisions were wrong (not detected when they happen, presumably because the computed border was too small); and BorderStyle 4 (which applies BorderStyle 1 _and_ a background box) looked wrong (the box was smaller than the glyph outline).

Between this and what your test files already contained, I believe and hope that all instances of `border_style_x` have been covered. And I _think_ the new code is, indeed, correct in all cases, but could you retest [this](https://github.com/libass/libass/files/9639169/test-x-scaling.zip) (from 0:14:00 on) with a patched build? My only remaining concern is that we seem to compute shadow in two places and I’m not sure how to test each, but let’s hope that it’s fine.

libass/ass_render.c
TheOneric: Good catch, thanks, now fixed!
I guess zero area `frame_content_*` is allowed, to avoid margin-using dialogue lines disappearing if the video is zoomed to zero?

TheOneric:
> I believe this is not accurate.

You are correct, only some things like vector drawings rendered correctly in the mentioned case. IN retesting this now, `\xshad` seems to have been rendered correctly only for PlayRes equal to scaled display res.
I thought I tested more than just vector drawings, but I also remember being confused at one point if storage or display res yielded correct rendering so I probably mixed up the results for some features.
I reworded the commit message, does the new version look better?

> And I think the new code is, indeed, correct in all cases, but could you retest [this](https://github.com/libass/libass/files/9639169/test-x-scaling.zip) (from 0:14:00 on) with a patched build?

Here you go, results for 3d2f66d49eeb4ed033260c14408c668a6a7ac52b: [645_astiob-extendedTests_results.tar.gz](https://github.com/libass/libass/files/9639407/645_astiob-extendedTests_results.tar.gz)
Notably, for `stor=8x` the “near collision” case actually *is* a collision.
All but `x_anamorphic` were tested on non-anamorphic video and because the `8x_sbas=no` sub file also had unique events mentioning anamorphic behaviour I additionally tested this also with anamorphic video.

libass/ass_render.c
astiob: TBVH I can’t say why or even _whether_ it is “allowed”, and I’d rather it wasn’t, given that it completely breaks this and the par calculation. But I guess it ultimately boils down to: what other option do we have, if someone does give us a zero? Just refuse to render anything at all?

astiob (approved):
Thanks!

The new commit message looks great. Everything LGTM, with just one caveat.

The caveat: I know I’m the one who proposed it, but here’s an argument (for the sake of it) against preferring storage size over an explicitly set PAR: it turns mpv’s `sub-ass-vsfilter-aspect-compat` toggle (which apparently was historically so important that it has its own hot key `V`!) into a no-op unless `sub-ass-vsfilter-blur-compat` is also disabled. My weak counter-argument to this argument is that I suspect that subtitles that did want this toggle to be disabled had `PlayResX` equal to a desqueezed width and should therefore start working correctly now (regardless, indeed, of the PAR setting, even if the API keeps working)… except the actual text will get stretched. I have never encountered a file like this myself, so I can’t tell how relevant this really is; but it kinda feels wrong to break this mpv feature given the emphasis mpv puts on it. Maybe we should ask mpv people and maybe they’ll just say they’d actually love to merge it into `blur-compat`?

> Notably, for `stor=8x` the “near collision” case actually _is_ a collision.

My bad; I see I put a value there that gives a total width of _exactly_ PlayResX, which does indeed produce a collision in other situations in the same test set. I assume a rounding error is involved and a collision is correct enough, too. If you change the `\bord` to `68`, _then_ it should stop colliding.

But for what it’s worth, this is what I got in CCCP’s xy-VSFilter:

No collision on anamorphic video.
No collision on square-pixel video.

TheOneric:
> If you change the \bord to 68, then it should stop colliding.

Yup, even just changing one to `68.9` avoids a collision. Curious, that VSFilter sometimes collides and sometimes not with combined border equal to PlayRes.

> The new commit message looks great. Everything LGTM, with just one caveat.

I opened a mpv issue here: https://github.com/mpv-player/mpv/issues/10680 and removed the commit from the patch series. The commit is preserved as a format-patch [0001-api-prefer-storage-size-over-pixel-aspect.patch.gz](https://github.com/libass/libass/files/9639618/0001-api-prefer-storage-size-over-pixel-aspect.patch.gz) and depending on mpv feedback can be applied later.
As the rest appears good to merge I'll do so shortly.

astiob:
> Curious that VSFilter sometimes collides and sometimes not with combined border equal to PlayRes.

OK, correction: it doesn’t collide (both events are visible) in _any_ of the tests. So this sounds like a libass bug. (As much as in the _general_ case, this can be affected by floating-point error, like I suspected initially—but not relevant here as the scaling factors are all exactly representable.)

TheOneric:
Right, this wasn't checked in the test files. Now added the following to `test-x-scaling_anamorphic_sbas-yes_playx-is-{stor,display}x.ass` from the previous samples:
```
Dialogue: 1,0:00:11.00,0:00:12.00,Label,,0000,0000,0000,,Shadow + Origin + Rotations (SBAS=yes)
Dialogue: 1,0:00:11.00,0:00:12.00,Probe,,0000,0000,0000,,{\an5\bord4\xshad100\frz40}Testing
Dialogue: 1,0:00:11.00,0:00:12.00,Probe,,0000,0000,0000,,{\an5\bord4\xshad100\fry40}Testing
Dialogue: 1,0:00:11.00,0:00:12.00,Probe,,0000,0000,0000,,{\an5\bord4\xshad100\frx40}Testing
```

Other than the unrelated difference in border creation it appears the same to me:

**xy-VSF:**
xy-VSF_desqueezed_shadow+org+rot_displayx
xy-VSF_desqueezed_shadow+org+rot_storx

**libass master:**
libass-master_shadow+org+rot_displayx
libass-master_shadow+org+rot_storx

PR #651, opened 2022-10-04
[ABI BREAK] v4++ MarginB/T

TheOneric:
**Not for immediate merge.**
This breaks ABI but is by itself not too useful, so it is postponed until it can be confalted with another ABI or API break or perhaps until at least the more useful `RelativeTo` is also implemented.

The first actual commit of this is “Recognise v4++ files”, prior ones are from #649 and once this is merged I'll rebase this to get rid of them.

In its current state this implements:

 - Recognise v4++ files
 - Parse v4++ files, while disabling custom format lines for v4++
 - Implement `MarginB/T` with added complexity to ensure its only an *ABI* and not also an *API* break.
   If this gets conflated with an API break, this complexity can be removed and `MarginV` struct members renamed to `MarginT`.

Missing for full v4++ support is `RelativeTo`, we currently assume this field allows to basically enable `use_margins` per Style. However, xy\* dropped support for it completly, and renderers using the external VSFilter interface cannot meaningfully act on it. Thus tests to confirm our understanding must use guliverkli(2) MPC with its internal subtitle renderer and may additionally also test MPC-HC'S ISR which did not just retain but expand `RelativeTo`-related logic.

As a side note: if we end up breaking API anyway, we can also drop `treat_fontname_as_pattern`, which is currently still used in MPlayer *(and in MPlayer's internal libass copy the field actually still has an effect)*. Ideally MPlayer would receive patches before the field is removed from libass.


libass/ass_utils.c
```diff
@@ -256,9 +256,9 @@ uint32_t parse_color_tag(char *str)
     return ass_bswap32((uint32_t) color);
 }
 
-uint32_t parse_color_header(char *str)
+int32_t parse_int_header(char *str)
```
Masaiki, line 259: uint32_t here?

TheOneric: No, signed `int32_t` is correct; headers can have negative values. E.g. in the commit you’re commenting on, the VSF-style clamping for negative alpha values `FFMAX(FFMIN(front_alpha, 0xFF), 0)`  wouldn't work *(as straight forward)* using only unsigned integers.
For more info see the corresponding commit which was already merged into master: https://github.com/libass/libass/commit/b7b816a8ea09d26574ba4090b6e1362a59065b66

Issue #654, opened 2022-10-19
build: no longer builds with ffmpeg past 18d6c07267994398f99b2241f577f8e7118af099


1480c1:
### libass version

https://github.com/libass/libass/commit/75a3dbac9bd41842a4d00b0d42c9513e2c8aec67

### Is it a regression?

recent change since yesterday (see ffmpeg's commit)

### How did you invoke configure?

../configure --prefix=/local32 --enable-static --disable-shared

### System Information

Windows compiling using msys2 and mingw64 gcc

### Toolchain

gcc 12.2.0 from msys2

### Log of the error

copied from https://github.com/m-ab-s/media-autobuild_suite/issues/2305

```c
I:/MABS/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/12.2.0/../../../../i686-w64-mingw32/bin/ld.exe: I:/MABS/local32/lib/libass.a(ass.o):ass.c:(.text+0x2ea0): multiple definition of `read_file'; fftools/ffmpeg_opt.o:ffmpeg_opt.c:(.text+0x2fb0): first defined here
```

### Additional info

read_file() added to ffmpeg at https://github.com/FFmpeg/FFmpeg/commit/18d6c07267994398f99b2241f577f8e7118af099

read_file() added to libass (the resulting library symbol) at https://github.com/libass/libass/commit/571185341f3e58bd57a823070dde30c66491a4ee

possible solution is to add a prefix to read_file from libass so it matches the rest of the function

astiob:
For reference, this happens to be a duplicate of #222, although the FFmpeg development is new.

TheOneric:
While it is indeed a good idea to rename `read_file` and other non-static generic internal functions in libass (as observed in #222), this alone doesn't really fix this problem. There are many existing libass releases with `read_file` which still won't be able to be linked statically to `ffmpeg` even if we rename the function in master. So you should definitely also contact ffmpeg to ensure the (now no longer static) `read_file` there is also renamed. I don't yet see a report about this in [their issue tracker](https://trac.ffmpeg.org/query?status=closed&status=new&status=open&status=reopened&col=id&col=summary&col=status&col=type&col=component&col=version&col=time&col=changetime&report=1&desc=1&order=time).

*(Also strictly speaking this is a ffmpeg regression not a libass one as libass had the function non-static for 7 years and ffmpeg only since yesterday)*.

rcombs:
[...]

Issue #222, opened 2016-05-07
Namespace nonstatic symbols to support static linking


mulabro:
Hello,

I get some issue when I did the compilation of ffmbc.
I contacted them to tell them the message below  and they told me to see it with you.
I hope this will help improve the libass.

Message from FFMBC :

> Actually, libass should be contacted, exporting a function "read_file" from a library could cause clashes in many other programs, it seems strange.
> None the less I'll fix it

My message :

> Hello,
> 
> I try to compile ffmbc 0.7.4 with the libass-0.13.1 to be able to add subtitle to video but I get an error.
> I succeed to to the following compilation :
> 
> ./configure --prefix=/usr/local/ffmbc/ffmbc-0.7.4-x264-10bit-faac-zlib-freetype-lame-openjpeg-ass-static/ --enable-gpl --enable-libx264 --enable-libfaac --enable-nonfree --enable-doc --enable-zlib --enable-libfreetype --enable-libmp3lame --enable-libopenjpeg --enable-static --disable-shared --extra-libs=-static --extra-cflags='-I/usr/local/include -static' --extra-ldflags='-L/usr/local/lib -static'
> 
> but when I add the libass it's not working
> 
> ./configure --prefix=/usr/local/ffmbc/ffmbc-0.7.4-x264-10bit-faac-zlib-freetype-lame-openjpeg-ass-static/ --enable-gpl --enable-libx264 --enable-libfaac --enable-nonfree --enable-doc --enable-zlib --enable-libfreetype --enable-libmp3lame --enable-libopenjpeg --enable-libass --enable-static --disable-shared --extra-libs=-static --extra-cflags='-I/usr/local/include -static' --extra-ldflags='-L/usr/local/lib -static'
> 
> My libass is compiled with the parameters below :
> ./configure --prefix=/usr/local --enable-shared=no --enable-static=yes --disable-fontconfig --disable-require-system-font-provider
> I can't make it work with fontconfig because I have an issue with the xml library. I use libass without fontconfig for my compilation of ffmpeg and it work. I don't know if this information help.
> 
> I got the message below :
> LD    ffmbc_g
> /usr/local/lib/libass.a(ass.o): dans la fonction « read_file »:
> 
> /root/source/libass-0.13.2/libass/ass.c:1078: définitions multiples de « read_file »
> 
> cmdutils.o:/root/source/FFmbc-0.7.4/cmdutils.c:708: défini pour la première fois ici
> 
[...]

PR #655, opened 2022-10-21
Namespace internal API

TheOneric:
See #222 and #654.
One unsued function is dropped, ~~11~~ 12 are static'fied and possibly moved, 3 are moved into headers as `static inline` and all others renamed. It also adds a check to our CI to avoid adding new non-namespaced internal functions.

There are a few more functions which technically could be moved and static'fied like in the thrid commit, but I either wasn't sure if or don't think it's a good idea. They (might) logically fit better into their current files. See below:
```
--- Not sure if moving would be good ---

apply_transition_effects (parse -> render)
event_has_hard_overrides (parse -> render)

fix_outline (bitmap -> render)
be_padding (bitmap -> render)


--- Probably a bad idea to move ---

resolve_base_direction (shaper -> render)

outline_add_rect (outline -> font)
outline_convert (outline -> font)
outline_rotate_90 (outline -> font)

outline_scale_pow2 (outline -> render)
outline_stroke (outline -> render)
outline_to_bitmap (outline -> render)
outline_transform_2d (outline -> render)
outline_transform_3d (outline -> render)
outline_update_cbox (outline -> render)
outline_update_min_transformed_x (outline -> render)

rasterizer_done (rasterizer -> render)
rasterizer_init (rasterizer -> render)
rasterizer_set_outline (rasterizer -> bitmap)

copy_bitmap (bitmap -> render)
realloc_bitmap (bitmap -> blur)
shift_bitmap (bitmap -> render)
```


MrSmile:
There's also a third method: make function `static inline` and put it into appropriate header. I think `skip_spaces`, `rskip_spaces`, `change_alpha`, `be_padding` and maybe `numpad2align` should be handled in that way.

Also, `be_blur_pre` and `be_blur_post` are good candidates for assembly implementation, so maybe it's better to use assembly naming scheme for them instead, similarly to `ass_be_blur`.

astiob (approved):

TheOneric:
Good idea re `static inline` in headers! I converted `skip_spaces`, `rskip_spaces` and  `numpad2align` to those. `be_padding` is only used in `ass_render.c`, so if its moved out of `ass_bitmap.c` anyway, I just moved it directly to `ass_render.c` instead of a header.

`change_alpha` though calls another function `static` to `ass_parse.c` which also calls another `static` function. To convert `change_alpha` to header `static inline` those would need to be moved as well, eventhough they're not used anywhere else which imho doesn't seem worth it.
I always get confused about non-static inline with its differences in GNU89 and ISO-C99/GNU99, but it seems like with `change_alpha` being explicitly `extern` in the header and the definition in `ass_parse.c` marked as `inline`, the intnent and likely effect is that hthe compiler gets a hint to and can inline usage in `ass_parse.c`, but an external version is also compiled which is then used in other compilation units. Given `change_alpha` is only called once outside of `ass_parse.c` there's probably also not much perf to be gained from turning it into `static inline`.

> Also, `be_blur_pre` and `be_blur_post` are good candidates for assembly implementation, so maybe it's better to use assembly naming scheme for them instead, similarly to `ass_be_blur`.

Do you mean the `ass_` prefix, the `_c` suffix or both? Imho it's fine as is for now. The suffix only makes sense if there's already an alternative version. For them to be used in `BitmapEngine`, I believe the functions will need to be un`static`fied at which point the prefix can be added — and the CI check will enforce this.

MrSmile:
`ass_change_alpha` looks ugly as it's no longer complimentary to `change_color`, also `inline` without `static` is questionable. Maybe it's better to make
```c
void ass_apply_fade(uint32_t *clr, int fade)
{
    // VSFilter compatibility: apply fade only when it's positive
    if (fade > 0)
        change_alpha(clr, mult_alpha(_a(*clr), fade), 1);
}
```
and export it instead? There's no need to move `mult_alpha` in that case too.

Also there's a lot of misalignment, where indent of second and later lines in prototypes and calls is no longer corresponds to first line after inserting `ass_` prefix.

libass/ass_bitmap.c
```diff
@@ -122,7 +122,7 @@ void ass_synth_blur(const BitmapEngine *engine, Bitmap *bm,
     ass_aligned_free(tmp);
 }
 
-bool alloc_bitmap(const BitmapEngine *engine, Bitmap *bm,
+bool ass_alloc_bitmap(const BitmapEngine *engine, Bitmap *bm,
```
MrSmile, line 125: Missing indent.

libass/ass_bitmap.c
```diff
@@ -154,29 +154,29 @@ void ass_free_bitmap(Bitmap *bm)
...
+    if (!ass_alloc_bitmap(engine, dst, src->w, src->h, false))
         return false;
     dst->left = src->left;
     dst->top  = src->top;
     memcpy(dst->buffer, src->buffer, src->stride * src->h);
     return true;
 }
 
-bool outline_to_bitmap(ASS_Renderer *render_priv, Bitmap *bm,
+bool ass_outline_to_bitmap(ASS_Renderer *render_priv, Bitmap *bm,
```
MrSmile, line 171: Indent.

libass/ass_bitmap.c
```diff
@@ -202,12 +202,12 @@ bool outline_to_bitmap(ASS_Renderer *render_priv, Bitmap *bm,
...
     int32_t tile_w = (w + mask) & ~mask;
     int32_t tile_h = (h + mask) & ~mask;
-    if (!alloc_bitmap(render_priv->engine, bm, tile_w, tile_h, false))
+    if (!ass_alloc_bitmap(render_priv->engine, bm, tile_w, tile_h, false))
         return false;
     bm->left = x_min;
     bm->top  = y_min;
 
-    if (!rasterizer_fill(render_priv->engine, rst, bm->buffer,
+    if (!ass_rasterizer_fill(render_priv->engine, rst, bm->buffer,
```
MrSmile, line 210: Indent.

libass/ass_outline.c
```diff
@@ -245,7 +245,7 @@ bool outline_convert(ASS_Outline *outline, const FT_Outline *source)
  * Outline should be preallocated to a sufficient size
  * and coordinates should be in the allowable range.
  */
-void outline_add_rect(ASS_Outline *outline,
+void ass_outline_add_rect(ASS_Outline *outline,
```
MrSmile, line 248: Indent.

libass/ass_outline.c
```diff
@@ -345,11 +345,11 @@ bool outline_rotate_90(ASS_Outline *outline, ASS_Vector offs)
  * Result outline should be uninitialized or empty.
  * Source outline can be NULL.
  */
-bool outline_scale_pow2(ASS_Outline *outline, const ASS_Outline *source,
+bool ass_outline_scale_pow2(ASS_Outline *outline, const ASS_Outline *source,
```
MrSmile, line 348: Indent.

libass/ass_outline.c
```diff
@@ -397,15 +397,15 @@ bool outline_scale_pow2(ASS_Outline *outline, const ASS_Outline *source,
  * Result outline should be uninitialized or empty.
  * Source outline can be NULL.
  */
-bool outline_transform_2d(ASS_Outline *outline, const ASS_Outline *source,
+bool ass_outline_transform_2d(ASS_Outline *outline, const ASS_Outline *source,
```
MrSmile, line 400: Indent.

libass/ass_outline.c
```diff
@@ -432,15 +432,15 @@ bool outline_transform_2d(ASS_Outline *outline, const ASS_Outline *source,
  * Result outline should be uninitialized or empty.
  * Source outline can be NULL.
  */
-bool outline_transform_3d(ASS_Outline *outline, const ASS_Outline *source,
+bool ass_outline_transform_3d(ASS_Outline *outline, const ASS_Outline *source,
```
MrSmile, line 435: Indent.

libass/ass_outline.c
```diff
@@ -469,7 +469,7 @@ bool outline_transform_3d(ASS_Outline *outline, const ASS_Outline *source,
 /*
  * \brief Find minimal X-coordinate of control points after perspective transform
  */
-void outline_update_min_transformed_x(const ASS_Outline *outline,
+void ass_outline_update_min_transformed_x(const ASS_Outline *outline,
```
MrSmile, line 472: Indent.

libass/ass_outline.c
```diff
@@ -1485,11 +1485,11 @@ static bool close_contour(StrokerState *str, int dir)
  * \param eps approximate allowable error
  * \return false on allocation failure
  */
-bool outline_stroke(ASS_Outline *result, ASS_Outline *result1,
+bool ass_outline_stroke(ASS_Outline *result, ASS_Outline *result1,
```
MrSmile, line 1488: Indent.

libass/ass_outline.h
```diff
@@ -87,40 +87,40 @@ typedef struct {
...
+                              const double m[2][3]);
+bool ass_outline_transform_3d(ASS_Outline *outline, const ASS_Outline *source,
+                              const double m[3][3]);
 
 // info queries
-void outline_update_min_transformed_x(const ASS_Outline *outline,
-                                      const double m[3][3],
-                                      int32_t *min_x);
-void outline_update_cbox(const ASS_Outline *outline, ASS_Rect *cbox);
+void ass_outline_update_min_transformed_x(const ASS_Outline *outline,
```
MrSmile, line 116: One space is missing.

libass/ass_parse.c
```diff
@@ -280,7 +280,7 @@ static ASS_Style *lookup_style_strict(ASS_Track *track, char *name, size_t len)
  *            of a number of spaces immediately preceding '}' or ')'
  * \param pwr multiplier for some tag effects (comes from \t tags)
  */
-char *parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
+char *ass_parse_tags(ASS_Renderer *render_priv, char *p, char *end, double pwr,
```
MrSmile, line 283: Indent.

libass/ass_rasterizer.c
```diff
@@ -262,7 +262,7 @@ static bool add_cubic(RasterizerData *rst, const ASS_Vector *pt)
 }
 
 
-bool rasterizer_set_outline(RasterizerData *rst,
+bool ass_rasterizer_set_outline(RasterizerData *rst,
```
MrSmile, line 265: Indent.

libass/ass_rasterizer.c
```diff
@@ -719,7 +719,7 @@ static bool rasterizer_fill_level(const BitmapEngine *engine, RasterizerData *rs
     return true;
 }
 
-bool rasterizer_fill(const BitmapEngine *engine, RasterizerData *rst,
+bool ass_rasterizer_fill(const BitmapEngine *engine, RasterizerData *rst,
```
MrSmile, line 722: Indent.

libass/ass_render.c
```diff
@@ -1239,26 +1239,26 @@ size_t ass_outline_construct(void *key, void *value, void *priv)
                 break;
 
             ASS_Outline src;
-            if (!outline_scale_pow2(&src, &k->outline->outline[0],
+            if (!ass_outline_scale_pow2(&src, &k->outline->outline[0],
```
MrSmile, line 1242: Indent.

libass/ass_render.c
```diff
@@ -1239,26 +1239,26 @@ size_t ass_outline_construct(void *key, void *value, void *priv)
...
                                     k->scale_ord_x, k->scale_ord_y))
                 return 1;
-            if (!outline_stroke(&v->outline[0], &v->outline[1], &src,
+            if (!ass_outline_stroke(&v->outline[0], &v->outline[1], &src,
```
MrSmile, line 1245: Indent.

libass/ass_render.c
```diff
@@ -2650,12 +2650,12 @@ size_t ass_composite_construct(void *key, void *value, void *priv)
...
 
-    int bord = be_padding(k->filter.be);
+    int bord = ass_be_padding(k->filter.be);
     if (!bord && n_bm == 1) {
-        copy_bitmap(render_priv->engine, &v->bm, last->bm);
+        ass_copy_bitmap(render_priv->engine, &v->bm, last->bm);
         v->bm.left += last->pos.x;
         v->bm.top  += last->pos.y;
-    } else if (n_bm && alloc_bitmap(render_priv->engine, &v->bm,
+    } else if (n_bm && ass_alloc_bitmap(render_priv->engine, &v->bm,
```
MrSmile, line 2658: Indent.

libass/ass_render.c
```diff
@@ -2677,10 +2677,10 @@ size_t ass_composite_construct(void *key, void *value, void *priv)
         }
     }
     if (!bord && n_bm_o == 1) {
-        copy_bitmap(render_priv->engine, &v->bm_o, last_o->bm_o);
+        ass_copy_bitmap(render_priv->engine, &v->bm_o, last_o->bm_o);
         v->bm_o.left += last_o->pos_o.x;
         v->bm_o.top  += last_o->pos_o.y;
-    } else if (n_bm_o && alloc_bitmap(render_priv->engine, &v->bm_o,
+    } else if (n_bm_o && ass_alloc_bitmap(render_priv->engine, &v->bm_o,
```
MrSmile, line 2683: Indent.

MrSmile (approved):
I've marked places with incorrect alignment, otherwise LGTM.

TheOneric:
All indents should now be adjusted.

Issue #410, opened 2020-07-06
Anamorphic blur


astiob:
In anamorphic video in a classical VSFilter setup (`render_priv->font_scale_x != 1`), blur is applied before the image is stretched horizontally. To faithfully reproduce this, we’d need to stretch our own blur horizontally by `font_scale_x`, making the x-radius and the y-radius different.

@MrSmile I may have asked you before, but how hard would this be?

Of course, this also applies to \be, but the solution I offered in #407 can handle that. Let’s stick to discussing Gaussian blur here.

MrSmile:
Should be mostly straightforward as 2D Gaussian blur is essentially a pair of 1D blurs.

PR #660, opened 2022-11-09
Use compatible blur parameters on anamorphic video

astiob:
@MrSmile I’ve taken your [commit](https://github.com/libass/libass/commit/5f7349b54bb7718291db499389bb3277fad35cd1) as it was and edited the commit message. Does the new message look OK to you?

Test: https://github.com/libass/libass-tests/pull/1

---

As a reminder/for reference, XySubFilter also has correct anamorphic blur, whereas MPC-HC’s ISR doesn’t have anamorphic blur and indeed doesn’t scale blur _at all_ (and also seems to have worse blur quality; perhaps it’s still using the old guliverkli2 blur code, whereas xy-VSFilter switched to libass’s old, then-current implementation at some point).


compare/compare.c
```diff
@@ -569,7 +571,7 @@ static int *parse_cmdline(int argc, char *argv[])
 fail:
     free(pos);
     const char *fmt =
-        "Usage: %s ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:1-8>] [-p <pass-level:0-3>]\n";
+        "Usage: %s ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:1-8>] [-a <anamorphic-scale:1-8>] [-p <pass-level:0-3>]\n";
```
astiob, line 574: Or maybe `-s` should just accept a pair of integers? Like, `[-s <scale:1-8>[x<scaleY:1-8>]]`?

libass/ass_render.c
```diff
@@ -1010,21 +1010,23 @@ static void init_font_scale(ASS_Renderer *render_priv)
     render_priv->screen_scale_y = font_scr_h / render_priv->track->PlayResY;
 
     ASS_Vector layout_res = ass_layout_res(render_priv);
-    render_priv->blur_scale = font_scr_h / layout_res.y;
+    render_priv->blur_scale_x = font_scr_w / layout_res.x;
+    render_priv->blur_scale_y = font_scr_h / layout_res.y;
```
astiob, line 1014: ### `LayoutRes` vs `ass_set_pixel_aspect` for blur and 3D transforms:

Normally, `(font_scr_w / layout_res.x) == par_scale_x * (font_scr_y / layout_res.y)`. Indeed, when exactly one of `par_scale_x` and `layout_res` is known, the other is defaulted to a value that ensures this equation is satisfied.

But our API allows the consumer to explicitly set _both_ `storage_size` and `pixel_aspect` to such values that this equation doesn’t hold. In practice, this happens (only) in mpv with `--sub-ass-vsfilter-blur-compat=yes --sub-ass-vsfilter-aspect-compat=no`.

In such a situation, it’s anyone’s guess as to what “should” happen in the user’s opinion. The only scenarios I’ve been able to come up with when a user might want to do this involve mixed-and-match video and subtitles:

* Maybe they’re playing an anamorphic video file with custom style overrides or converted subtitles and seeing that without `aspect-compat=no` their overridden style is getting stretched ugly.

    In this case, it’s obvious we shouldn’t be stretching any text (so we don’t want to disregard the PAR completely as we [tried](https://github.com/libass/libass/pull/645#pullrequestreview-1119327502)), but if there’s any blur or transforms left from the original ASS, it may make more sense still to apply them at the anamorphically-compressed LayoutRes, as originally intended, despite the `aspect-compat=no`.

     This PR currently does this for blur, but transforms are applied (in master, and this PR doesn’t currently change this) to an anamorphically-stretched video-height rectangle. Maybe we ought to change the transform code, too.

    Then again, user style overrides can include `Blur` and `Angle`, too, and in that case it’d make more sense to apply them at the display aspect ratio. Maybe we ought to adjust user overrides for `par_scale_x` in `handle_selective_style_overrides` like we adjust them for `PlayResY`?

* Maybe they’re playing an anamorphic video file with an external subtitle file that was authored for 
an unrelated non-anamorphic video file, as in https://github.com/mpv-player/mpv/issues/10680#issuecomment-1257127239.

    In this case, well, what the user really wants us to do is to pretend `layout_res` itself has the anamorphically-stretched aspect ratio. Indeed, this is what we assume with `blur-compat=no aspect-compat=no` (and before b6db99a3930b609ca1ba5a0b9930b5c6dc5701dc assumed with `blur-compat=no` regardless of `aspect-compat`). But at the end of the day, `blur-compat=no` is the _only_ way to handle this situation: if the external subs don’t declare `LayoutRes`, they may have been authored for _any_ video, and the current video’s resolution is completely irrelevant.

compare/compare.c
```diff
@@ -679,7 +691,8 @@ int main(int argc, char *argv[])
         if (!track)
             continue;
         Result res = process_image(renderer, track, list.items[i].dir, output,
-                                   name, list.items[i].time, scale);
+                                   name, list.items[i].time,
+                                   scale * anamorphic_scale, scale);
```
MrSmile, line 695: This is up to &times;512 original bitmap size (previously only &times;64), I think it's overkill.

MrSmile (commented):
> Does the new message look OK to you?

Yes, looks good enough.

Also `compare` docs should be modified accordingly too.

compare/compare.c
MrSmile: Yes, that variant would be more obvious.

compare/compare.c
astiob: TBH I don’t really see why we need to limit the scales at all. This is a utility that takes the scale explicitly. If the user wants a bigger one, who are we to stop them?

But sure, with `<scaleX:1-8>:<scaleY:1-8>` the limit will be smaller.

compare/compare.c
astiob: Amended to take `-s <scale:1-8>[x<scale:1-8>]` with an expanded usage message.

astiob: And again to update `compare/README.md`.

compare/compare.c
MrSmile: `uint16_t` for intermediate during averaging can overflow. Also division approximation can break down too (actually I should check that).

compare/compare.c
```diff
@@ -171,11 +171,11 @@ static int compare1(const Image16 *target, const uint16_t *grad,
...
 static int compare(const Image16 *target, const uint16_t *grad,
                    const ASS_Image *img, const char *path,
-                   double *result, int scale)
+                   double *result, int scale_x, int scale_y)
 {
-    if (scale == 1)
+    if (scale_x == 1 && scale_y == 1)
         return compare1(target, grad, img, path, result);
-    int scale2 = scale * scale;
+    int scale2 = scale_x * scale_y;
```
MrSmile, line 178: It's no longer a correct name, maybe something like `area` would be better?

compare/compare.c
```diff
@@ -593,14 +596,21 @@ int main(int argc, char *argv[])
...
-        if (arg[0] < '1' || arg[0] > '8' || arg[1]) {
+        if (arg[0] < '1' || arg[0] > '8' || (arg[1] && arg[1] != 'x')) {
             printf("Invalid scale value, should be 1-8!\n");
             goto end;
         }
-        scale = arg[0] - '0';
+        scale_x = scale_y = arg[0] - '0';
+        if (arg[1]) {
+            if (arg[2] < '1' || arg[2] > '8' || arg[3]) {
+                printf("Invalid scale value, should be 1-8!\n");
```
MrSmile, line 609: Maybe it's better to move scale parsing into separate function to eliminate copy-paste? I.e.
```c
        if (!parse_scale(argv[pos[SCALE]], &scale_x, &scale_y)) {
            printf("Invalid scale value, should be 1-8!\n");
            goto end;
        }
```

compare/compare.c
astiob: Hmm, does that mean it was broken for `-s 6`, too? Or is it something specific to the divisor of 35?

MrSmile: No, for scales up to 8 the only problematic pair is 5&times;7. Why it's so specific I don't know, I've simply tested for all possible inputs.

astiob: Committed in your name. If you prefer, I can squash it into the main `compare` commit.

compare/compare.c
astiob: Renamed to `scale_area`.

compare/compare.c
astiob: Done.

By the way, does the control flow within the block (now the function) seem fine to you, or would you prefer it to be tweaked in some way? I’m finding myself not quite liking any of the ways I can write it.

MrSmile: Hmm... maybe something like
```c
static bool parse_scale(const char *arg,  int *scale_x, int *scale_y)
{
    if (arg[0] < '1' || arg[0] > '8')
        return false;
    *scale_x = *scale_y = arg[0] - '0';
    if (!arg[1])
        return true;
    if (arg[1] != 'x' || arg[2] < '1' || arg[2] > '8' || arg[3])
        return false;
    *scale_y = arg[2] - '0';
    return true;
}
```

compare/compare.c
MrSmile: I think it's better to not keep regression in history.

compare/compare.c
```diff
@@ -593,14 +609,12 @@ int main(int argc, char *argv[])
     ItemList list = {0};
     int result = R_ERROR;
 
-    int scale = 1;
+    int scale_x = 1, scale_y = 1;
     if (pos[SCALE]) {
-        const char *arg = argv[pos[SCALE]];
-        if (arg[0] < '1' || arg[0] > '8' || arg[1]) {
+        if (!parse_scale(argv[pos[SCALE]], &scale_x, &scale_y)) {
```
MrSmile, line 614: Looks like it's possible to merge `if`s too.

compare/compare.c
astiob: Squashed.

compare/compare.c
astiob: Applied, thanks.

compare/compare.c
astiob: Merged.

astiob:
In the last push, I’ve also tweaked the README a tiny bit:

```diff
 * `<scale>` sets an oversampling factor or a pair of independent factors for horizontal and vertical oversampling
-  (in any case each factor is a positive integer up to 8, default 1);
+  (in any case each factor is a positive integer up to 8, default 1x1);
```

in hopes of making it extra clear that if `-s` is omitted entirely, the default is `-s 1x1`, but if a single number `-s N` is given, it doesn’t default to `-s Nx1`.

compare/compare.c
```diff
@@ -593,14 +609,10 @@ int main(int argc, char *argv[])
...
-    if (pos[SCALE]) {
-        const char *arg = argv[pos[SCALE]];
-        if (arg[0] < '1' || arg[0] > '8' || arg[1]) {
-            printf("Invalid scale value, should be 1-8!\n");
-            goto end;
-        }
-        scale = arg[0] - '0';
+    int scale_x = 1, scale_y = 1;
+    if (pos[SCALE] && !parse_scale(argv[pos[SCALE]], &scale_x, &scale_y)) {
+        printf("Invalid scale value, should be 1-8!\n");
```
MrSmile, line 614: This message isn't completely correct too, should be something like `1-8[x1-8]`.

MrSmile (approved):

compare/compare.c
astiob: Suggestion applied.

libass/ass_render.c
TheOneric: > [...] or converted subtitles and seeing that without aspect-compat=no their overridden style is getting stretched ugly.

For converted subs storage resolution is never set and pixel aspect is at most set to the *display’s* pixel aspect.

> Then again, user style overrides can include `Blur` and `Angle`, too, and in that case it’d make more sense to apply them at the display aspect ratio. Maybe we ought to adjust user overrides for par_scale_x in handle_selective_style_overrides like we adjust them for PlayResY?

I can see how this could work for `Angle`, but I’m not sure how you want to adjust the single `Blur` and in fact also `Shadow` and `Outline` values of a Style to avoid stretching in one direction. Do you want to involve the renderer state with its split `(x|y)(shad|bord)` values, override the scaling factors for the current event (also affecting any original ASS snippets) or something else? *(+ with SBAS=yes Shadow and Outline would need to be treated differently depending on whether PlayRes* matches storage, display res or none of the two)*

compare/README.md
```diff
@@ -5,11 +5,12 @@ To build a test utility configure libass with the `--enable-compare` flag.
...
 
 Test program command line:  
-`compare ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:1-8>] [-p <pass-level:0-3>]`
+`compare ([-i] <input-dir>)+ [-o <output-dir>] [-s <scale:1-8>[x<scale:1-8>]] [-p <pass-level:0-3>]`
 
 * `<input-dir>` is a test input directory, can be several of them;
 * `<output-dir>` if present sets directory to store the rendering results;
-* `<scale>` sets an oversampling factor (positive integer up to 8, default 1);
+* `<scale>` sets an oversampling factor or a pair of independent factors for horizontal and vertical oversampling
+  (in any case each factor is a positive integer up to 8, default 1x1);
```
TheOneric, line 13: If you’re worried about `-s N` seeming like `-s Nx1`, perhaps it would be best to explictly state that a single value `NN` acts like `NxN`, or give an example like further down in the “Usage” text?

TheOneric (approved):

libass/ass_render.c
astiob: Right; I didn’t actually think about it. But it looks simple enough: currently, we take user overrides in `handle_selective_style_overrides` and put them in a `ASS_Style`, which lacks the split x/y values. But the only place that calls `handle_selective_style_overrides` immediately converts that style to the `RenderContext`, which does have them. So perhaps we could make `handle_selective_style_overrides` just modify the context directly. It looks like the only part of the `state.style` that isn’t just copied to another field is the `Margin{L,R,V}`: perhaps we should just give them their own fields in the `RenderContext`.

Ironically, I’m _less_ clear on `Angle` now that I think about this. Normally, `Angle` means `\frz`. For one, if the user overrides `\frz`, do they also want to override `\frx0\fry0`? Even if so, `\frz` followed by horizontal stretching of the anamorphic video is no longer representable as a single rotation: we’d need to decompose it into multiple values. Or just create an extra flag and set it to disable `par_scale_x` for this one glyph run… ew.

Besides, we can never be quite sure what the right way to “disable” anamorphic compensation is. mpv apparently supports rendering to non-square-pixel displays and sets `par` appropriately. But we can’t distinguish the display PAR from the video PAR, so we can’t compensate overrides for just one of the two PARs. Ugh.

On another note, though, even if not for PAR, we probably totally should compensate horizontal-measure overrides for `PlayResX` now that we scale them. In particular, yes, `\xbord` and `\xshad`, so we need to apply overrides directly to `RenderContext` anyway.

> override the scaling factors for the current event

wouldn’t affect borders and shadows.

> with SBAS=yes Shadow and Outline would need to be treated differently depending on whether PlayRes matches storage, display res or none of the two

Not sure I understand.

TheOneric: > > override the scaling factors for the current event
> wouldn’t affect borders and shadows.

I was referring to `border_scale_(x|y)`, `blur_scale_(x|y)`, etc

> > with SBAS=yes Shadow and Outline [...]
> Not sure I understand.

Nvm, I got something wrong.

TheOneric:
This PR or one of the commits should also close https://github.com/libass/libass/pull/410

astiob:
Oh. Not sure how I missed that. I thought I’d searched, but guess not! Will add in the morning.

compare/README.md
astiob: Thanks for the feedback. I’ve tweaked the text further, and I think it’s significantly clearer now:

```
* `<scale>` sets factors for horizontal and vertical oversampling (positive integers up to 8)
  or a single factor to be used in both directions (default 1);
```

astiob:
Added the issue reference to the appropriate commit.

PR #664, opened 2022-11-15
Support Core Text on Mac OS X 10.5 and prerelease 10.6

### LINKED ISSUE #595: Compilation errors on prerelease PowerPC Mac OS X Snow Leopard
Compilation fails on this:
```
In file included from ass_utils.c:30:
ass.h:421: error: wrong number of arguments specified for ‘deprecated’ attribute
ass.h:520: error: wrong number of arguments specified for ‘deprecated’ attribute
In file included from ass.c:35:
ass.h:421: error: wrong number of arguments specified for ‘deprecated’ attribute
ass.h:520: error: wrong number of arguments specified for ‘deprecated’ attribute
libtool: compile:  /usr/bin/gcc-4.2 -std=gnu99 -DHAVE_CONFIG_H -I. -I.. -I/opt/local/include -std=gnu99 -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter -Werror-implicit-function-declaration -Wstrict-prototypes -Wpointer-arith -Wredundant-decls -Wno-missing-field-initializers -D_GNU_SOURCE -pipe -Os -arch ppc -I/opt/local/include/freetype2 -I/opt/local/include/libpng16 -I/opt/local/include/fribidi -I/opt/local/include/harfbuzz -I/opt/local/include/glib-2.0 -I/opt/local/lib/glib-2.0/include -I/opt/local/include/freetype2 -I/opt/local/include/libpng16 -I/opt/local/include/ossp -I/opt/local/include/freetype2 -I/opt/local/include/libpng16 -MT ass_string.lo -MD -MP -MF .deps/ass_string.Tpo -c ass_string.c -o ass_string.o >/dev/null 2>&1
make[2]: *** [ass_utils.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
libtool: compile:  /usr/bin/gcc-4.2 -std=gnu99 -DHAVE_CONFIG_H -I. -I.. -I/opt/local/include -std=gnu99 -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter -Werror-implicit-function-declaration -Wstrict-prototypes -Wpointer-arith -Wredundant-decls -Wno-missing-field-initializers -D_GNU_SOURCE -pipe -Os -arch ppc -I/opt/local/include/freetype2 -I/opt/local/include/libpng16 -I/opt/local/include/fribidi -I/opt/local/include/harfbuzz -I/opt/local/include/glib-2.0 -I/opt/local/lib/glib-2.0/include -I/opt/local/include/freetype2 -I/opt/local/include/libpng16 -I/opt/local/include/ossp -I/opt/local/include/freetype2 -I/opt/local/include/libpng16 -MT ass_strtod.lo -MD -MP -MF .deps/ass_strtod.Tpo -c ass_strtod.c -o ass_strtod.o >/dev/null 2>&1
mv -f .deps/ass_string.Tpo .deps/ass_string.Plo
make[2]: *** [ass.lo] Error 1
```

The older version builds:
```
36-47% port -v installed libass
The following ports are currently installed:
  libass @0.13.7_0 (active) requested_variants='' platform='darwin 10' archs='ppc' date='2022-02-14T20:14:23+0800'
```

astiob:
This should fix #595 and further make the Core Text system font provider usable on Mac OS X 10.5 “Leopard”—because why not?

@TheOneric Do we have any tests on PostScript-name vs full-name font selection? Looks like not? I don’t think it’s in any way urgent or a blocker for this, but at some point we probably should add some. And it would be great to have some tests targeting specifically system font providers, although I’m not quite sure how we could do that.

### DIFF
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -61,29 +61,6 @@ static void destroy_font(void *priv)
     CFRelease(fontd);
 }
 
-static bool is_postscript_font_format(CFNumberRef cfformat)
-{
-    bool ret = false;
-    int format;
-    if (CFNumberGetValue(cfformat, kCFNumberIntType, &format)) {
-        ret = format == kCTFontFormatOpenTypePostScript ||
-              format == kCTFontFormatPostScript;
-    }
-    return ret;
-}
-
-static bool check_postscript(void *priv)
-{
-    CTFontDescriptorRef fontd = priv;
-    CFNumberRef cfformat =
-        CTFontDescriptorCopyAttribute(fontd, kCTFontFormatAttribute);
-    if (!cfformat)
-        return false;
-    bool ret = is_postscript_font_format(cfformat);
-    CFRelease(cfformat);
-    return ret;
-}
-
 static bool check_glyph(void *priv, uint32_t code)
 {
     if (code == 0)
@@ -295,7 +272,6 @@ static void get_substitutions(void *priv, const char *name,
 }
 
 static ASS_FontProviderFuncs coretext_callbacks = {
-    .check_postscript   = check_postscript,
     .check_glyph        = check_glyph,
     .destroy_font       = destroy_font,
     .match_fonts        = match_fonts,
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -57,8 +57,8 @@ static char *cfstr2buf(CFStringRef string)
 
 static void destroy_font(void *priv)
 {
-    CTFontDescriptorRef fontd = priv;
-    CFRelease(fontd);
+    CFCharacterSetRef set = priv;
+    SAFE_CFRelease(set);
 }
 
 static bool check_glyph(void *priv, uint32_t code)
@@ -66,16 +66,12 @@ static bool check_glyph(void *priv, uint32_t code)
     if (code == 0)
         return true;
 
-    CTFontDescriptorRef fontd = priv;
-    CFCharacterSetRef set =
-        CTFontDescriptorCopyAttribute(fontd, kCTFontCharacterSetAttribute);
+    CFCharacterSetRef set = priv;
 
     if (!set)
         return true;
 
-    bool result = CFCharacterSetIsLongCharacterMember(set, code);
-    CFRelease(set);
-    return result;
+    return CFCharacterSetIsLongCharacterMember(set, code);
 }
 
 static char *get_font_file(CTFontDescriptorRef fontd)
@@ -139,8 +135,9 @@ static void process_descriptors(ASS_FontProvider *provider, CFArrayRef fontsd)
 
         char *path = NULL;
         if (get_font_info_ct(fontd, &path, &meta)) {
-            CFRetain(fontd);
-            ass_font_provider_add_font(provider, &meta, path, index, (void*)fontd);
+            CFCharacterSetRef set =
+                CTFontDescriptorCopyAttribute(fontd, kCTFontCharacterSetAttribute);
+            ass_font_provider_add_font(provider, &meta, path, index, (void*)set);
         }
 
         free(meta.postscript_name);
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -29,6 +29,18 @@
 
 #include "ass_coretext.h"
 
+#ifndef __has_builtin
+#define __has_builtin 0
+#endif
+
+#if __has_builtin(__builtin_available)
+#define CHECK_AVAILABLE(sym, ...) __builtin_available(__VA_ARGS__)
+#else
+// Cast to suppress "comparison never succeeds" warnings in some compilers
+// when the build target is new enough that sym isn't a weak symbol
+#define CHECK_AVAILABLE(sym, ...) (!!(intptr_t) &sym)
+#endif
+
 #define SAFE_CFRelease(x) do { if (x) CFRelease(x); } while (0)
 
 static const ASS_FontMapping font_substitutions[] = {
--- a/libass/ass_coretext.c
+++ b/libass/ass_coretext.c
@@ -19,6 +19,7 @@
 #include "config.h"
 #include "ass_compat.h"
 
+#include <Availability.h>
 #include <CoreFoundation/CoreFoundation.h>
 #include <TargetConditionals.h>
 #if TARGET_OS_IPHONE
@@ -88,7 +89,28 @@ static bool check_glyph(void *priv, uint32_t code)
 
 static char *get_font_file(CTFontDescriptorRef fontd)
 {
-    CFURLRef url = CTFontDescriptorCopyAttribute(fontd, kCTFontURLAttribute);
+    CFURLRef url = NULL;
+    if (false) {}
+#ifdef __MAC_10_6
+    // Declared in SDKs since 10.6, including iOS SDKs
+    else if (CHECK_AVAILABLE(kCTFontURLAttribute, macOS 10.6, *)) {
+        url = CTFontDescriptorCopyAttribute(fontd, kCTFontURLAttribute);
+    }
+#endif
+#if !TARGET_OS_IPHONE && (!defined(__MAC_10_6) || __MAC_OS_X_VERSION_MIN_REQUIRED < __MAC_10_6)
+    // ATS is declared deprecated in newer macOS SDKs
+    // and not declared at all in iOS SDKs
+    else {
+        CTFontRef font = CTFontCreateWithFontDescriptor(fontd, 0, NULL);
+        if (!font)
+            return NULL;
+        ATSFontRef ats_font = CTFontGetPlatformFont(font, NULL);
+        FSRef fs_ref;
+        if (ATSFontGetFileReference(ats_font, &fs_ref) == noErr)
+            url = CFURLCreateFromFSRef(NULL, &fs_ref);
+        CFRelease(font);
+    }
+#endif
     if (!url)
         return NULL;
     CFStringRef path = CFURLCopyFileSystemPath(url, kCFURLPOSIXPathStyle);
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -78,6 +78,9 @@ struct font_info {
 
     // private data for callbacks
     void *priv;
+
+    // unused if the provider has a check_postscript function
+    bool is_postscript;
 };
 
 struct font_selector {
@@ -113,13 +116,6 @@ struct font_data_ft {
     FT_Face face;
     int idx;
 };
-
-static bool check_postscript_ft(void *data)
-{
-    FontDataFT *fd = (FontDataFT *)data;
-    PS_FontInfoRec postscript_info;
-    return !FT_Get_PS_Font_Info(fd->face, &postscript_info);
-}
 
 static bool check_glyph_ft(void *data, uint32_t codepoint)
 {
@@ -161,7 +157,6 @@ get_data_embedded(void *data, unsigned char *buf, size_t offset, size_t len)
 
 static ASS_FontProviderFuncs ft_funcs = {
     .get_data          = get_data_embedded,
-    .check_postscript  = check_postscript_ft,
     .check_glyph       = check_glyph_ft,
     .destroy_font      = destroy_font_ft,
 };
@@ -265,6 +260,7 @@ get_font_info(FT_Library lib, FT_Face face, const char *fallback_family_name,
     int slant, weight;
     char *fullnames[MAX_FULLNAME];
     char *families[MAX_FULLNAME];
+    PS_FontInfoRec postscript_info;
 
     // we're only interested in outlines
     if (!(face->face_flags & FT_FACE_FLAG_SCALABLE))
@@ -324,6 +320,7 @@ get_font_info(FT_Library lib, FT_Face face, const char *fallback_family_name,
     info->width  = 100;     // FIXME, should probably query the OS/2 table
 
     info->postscript_name = (char *)FT_Get_Postscript_Name(face);
+    info->is_postscript = !FT_Get_PS_Font_Info(face, &postscript_info);
 
     if (num_family) {
         info->families = calloc(sizeof(char *), num_family);
@@ -488,6 +485,7 @@ ass_font_provider_add_font(ASS_FontProvider *provider,
     info->width         = width;
     info->n_fullname    = meta->n_fullname;
     info->n_family      = meta->n_family;
+    info->is_postscript = meta->is_postscript;
 
     info->families = calloc(meta->n_family, sizeof(char *));
     if (info->families == NULL)
@@ -612,9 +610,12 @@ void ass_font_provider_free(ASS_FontProvider *provider)
 static bool check_postscript(ASS_FontInfo *fi)
 {
     ASS_FontProvider *provider = fi->provider;
-    assert(provider && provider->funcs.check_postscript);
+    assert(provider);
 
-    return provider->funcs.check_postscript(fi->priv);
+    if (provider->funcs.check_postscript)
+        return provider->funcs.check_postscript(fi->priv);
+    else
+        return fi->is_postscript;
 }
 
 /**
--- a/configure.ac
+++ b/configure.ac
@@ -146,7 +146,7 @@ AS_IF([test "x$enable_coretext" != xno], [
     AC_COMPILE_IFELSE([
         AC_LANG_PROGRAM( dnl# First test for legacy include
             [[#include <ApplicationServices/ApplicationServices.h>]],
-            [[CTFontDescriptorCopyAttribute(NULL, kCTFontURLAttribute);]]
+            [[CTFontDescriptorCopyAttribute(NULL, kCTFontNameAttribute);]]
         )
     ], [
         pkg_libs="$pkg_libs -framework ApplicationServices -framework CoreFoundation"
@@ -158,7 +158,7 @@ AS_IF([test "x$enable_coretext" != xno], [
         AC_COMPILE_IFELSE([
             AC_LANG_PROGRAM( dnl# Otherwise check newer include style
                 [[#include <CoreText/CoreText.h>]],
-                [[CTFontDescriptorCopyAttribute(NULL, kCTFontURLAttribute);]]
+                [[CTFontDescriptorCopyAttribute(NULL, kCTFontNameAttribute);]]
             )
         ], [
             pkg_libs="$pkg_libs -framework CoreText -framework CoreFoundation"
--- a/libass/ass_fontselect.h
+++ b/libass/ass_fontselect.h
@@ -158,7 +158,7 @@ typedef char   *(*GetFallbackFunc)(void *priv,
 
 typedef struct font_provider_funcs {
     GetDataFunc         get_data;               /* optional/mandatory */
-    CheckPostscriptFunc check_postscript;       /* mandatory */
+    CheckPostscriptFunc check_postscript;       /* optional */
     CheckGlyphFunc      check_glyph;            /* mandatory */
     DestroyFontFunc     destroy_font;           /* optional */
     DestroyProviderFunc destroy_provider;       /* optional */
@@ -211,6 +211,12 @@ struct ass_font_provider_meta_data {
                         // See FONT_WEIGHT_*
     int width;          // Font weight in percent, normally 100
                         // See FONT_WIDTH_*
+
+    /**
+     * Whether the font contains PostScript outlines.
+     * Unused if the font provider has a check_postscript function.
+     */
+    bool is_postscript;
 };
 
 struct ass_font_stream {


TheOneric:
> Do we have any tests on PostScript-name vs full-name font selection?

nope

> it would be great to have some tests targeting specifically system font providers, although I’m not quite sure how we could do that.

You could either continune to check via the resulting bitmap and patch `compare` to enable a system font provider mode, or somehow parse the log messages to find out what fonts get selected.
In either case, for DirectWrite and CoreText such tests probably need to rely on specific fonts being available. For Fontconfig a custom config file and font dir could be used to control which "system" fonts are available.

rcombs (approved):

libass/ass_fontselect.c
```diff
@@ -608,9 +606,12 @@ void ass_font_provider_free(ASS_FontProvider *provider)
...
     ASS_FontProvider *provider = fi->provider;
-    assert(provider && provider->funcs.check_postscript);
+    assert(provider);
 
-    return provider->funcs.check_postscript(fi->priv);
+    if (provider->funcs.check_postscript)
+        return provider->funcs.check_postscript(fi->priv);
+    else
+        return fi->is_postscript;
 }
```
TheOneric, line 615: Since I didn't spot any `if (funcs.check_postscript)` guards elsewhere, do I understand correctly `fi->is_postscript` is already computed regardless of whether the font provider has `check_postscript?
If so, what’s the benefit of a provider having `check_postscript`? And do the performance considerations for DirectWrite’s  eager loading also apply to Fontconfig’s still existing eager loading on startup?

astiob: `fi->is_postscript` always exists, but it only has a meaningful value for font providers that invoke `get_font_info` one way or another during the addition of the font. Currently, that’s Core Text and the memory font provider. To declare this fact, these providers don’t define a `check_postscript` function (which would be useless after `is_postscript` is already computed).

Potentially, the value returned by `check_postscript` could be cached in `is_postscript` for the other providers. But this would require `is_postscript` to be something more complex than a simple `bool`: perhaps a tri-value `enum` or a pair of Boolean bit-fields. My first attempt actually was something like that (minus the caching), until I realized I could simplify it to this.

> what’s the benefit of a provider having `check_postscript`?

It lets the provider avoid expensive operations to figure out this value until the font actually matches some requested name.

For example, this presumably still applies (taken in isolation) to the DirectWrite provider just as it did back then: when enumerating fonts, DirectWrite seems to operate on a fast cache, but that cache doesn’t include the font format information. To get that, `check_postscript` must materialize the cache entry to the actual font and query the font for its format, but this is extra slow: per the original commit message, this step alone was apparently costing seconds for a font library that must’ve been small enough that it was otherwise loading reasonably quickly. Of course, when we actually want to select and to start using a font, we must materialize it anyway, but this lets us avoid materializing DirectWrite fonts that we are definitely not interested in.

Now, with the DirectWrite font provider now adding fonts lazily, this performance gain may have become negligible by Amdahl’s law. In fact, when we lazy-add fonts, we look them up by name, so in the ideal case, every lazily-added font will be a name match and cause `check_postscript` to run and/or the font to be selected and materialized for use. But technically, it’s possible for the font provider to return more fonts due to matching some name that we choose to ignore; although then again, I don’t remember if this is possible for the DirectWrite provider in particular.

Being unsure, I chose to keep this PR as small & safe as possible.

> And do the performance considerations for DirectWrite’s eager loading also apply to Fontconfig’s still existing eager loading on startup?

Nobody’s ever reported any performance issues with the Fontconfig system font provider.

TheOneric: Thanks for the detailed explanation!

TheOneric (approved):

TheOneric:
@barracuda156, could you test this on 10.5?

barracuda156:
@TheOneric @astiob Should I try this again now?

astiob:
No, sorry, I think I simply rebased this after a big change on master but hadn’t pushed until now.

astiob:
Although, I mean, if it doesn’t cause you too much trouble, rerunning the “Manrope Regular” test just to confirm that it hasn’t magically fixed itself by some coincidence would be good, too.

By the way, have you tested this on your prerelease Snow Leopard?

Generally speaking, if this is close enough to a working solution, I think we may/should merge it as is, given the lack of immediate ideas for what else might be wrong.

barracuda156:
@astiob I will try today. And yes, I also think it doesn’t hurt at all merging it.

astiob:
My current guess for Manrope Regular is that Leopard’s Core Text and current DirectWrite have special handling either for names ending literally in ` Regular` or for some kind of more flexible combination of the `REGULAR` bit in [`fsSelection`](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#fsselection) and arbitrary subfamily name. To find out more, I think we’ll need to make a bunch of slightly-tweaked versions of the font and see what happens to each of them.

libass/ass_coretext.c
```diff
@@ -76,7 +76,19 @@ static bool check_glyph(void *priv, uint32_t code)
...
+    CTFontRef font = CTFontCreateWithFontDescriptor(fontd, 0, NULL);
+    if (!font)
+        return NULL;
+    ATSFontRef ats_font = CTFontGetPlatformFont(font, NULL);
+    FSRef fs_ref;
+    if (ATSFontGetFileReference(ats_font, &fs_ref) == noErr)
+        url = CFURLCreateFromFSRef(NULL, &fs_ref);
+    CFRelease(font);
+#else
+    url = CTFontDescriptorCopyAttribute(fontd, kCTFontURLAttribute);
```
rcombs, line 90: I'd prefer to keep the new (and standard, and supported) behavior behind a `__builtin_available` or weak-symbol check.

astiob: Pushed an attempt with the weak-symbol check. It’s quite ugly, though.

On iOS, ATS doesn’t exist (I think), so the ATS branch must not be compiled.
On Leopard, `kCTFontURLAttribute` doesn’t exist, so the `kCTFontURLAttribute` branch must not be compiled.
Both branches must be conditionally-compiled, and if and only if both are present, we want to add an `else` between them.

Even if I reduce repetition by including the address check when the symbol isn’t weak, I’ll merely trade it for an extra preprocessor ifdef around the `else`. It also likely produces a compiler warning from `-Waddress`, included in `-Wall` (which may be able to be silenced with an ugly cast to `intptr_t`).

rcombs: Could do `__builtin_available` instead to deal with the `Waddress`?

astiob: That doesn’t help with the rest though. And breaks compilation with compilers that lack `__builtin_available`, like GCC or older Clang.

rcombs: I'm handling this on my language branch like this: rcombs@be8b501391d60f7d68171e44d1ad4f725fb4c242.

astiob: Not sure how that helps `__builtin_available`.

astiob: Basically, you’re forbidding the new-style code from being compiled on any compiler other than recentish Clang. How is that significantly better than compiling it _never,_ like this PR originally did?

astiob: It’s even worse for your branch, because `__builtin_available` first appeared in Xcode 9 which first ran on El Capitan (10.11), but you’re testing for an API that appeared in 10.9, so using `__builtin_available` means anyone building on 10.9 or 10.10 won’t get language support even though their system supports it.

In this PR, at least both code branches have equivalent functionality.

astiob: Going back to that `else`, here’s an odd way to minimize ifdeffery at the cost of less intuitive code:

```c
switch (true) {
#if TARGET_OS_IPHONE || MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
case !!&kCTFontURLAttribute:
    new code;
    break;
#endif
#if !TARGET_OS_IPHONE && MAC_OS_X_VERSION_MIN_REQUIRED < 1060
    old code;
#endif
}
```

astiob: Never mind; C requires constant expressions in `case`, doesn’t it. An equivalent `goto` would work, but that might just be even uglier.

astiob: > use an `if (0) {}`

Right, I suppose that could work.

---

You are literally writing this on a PR adding support for an even more ancient platform. And you do find it necessary to use a check at all, rather than unconditionally use the newer code.

> Because the original version lacked the new path on the vast majority of builds that will ever be made of this code

It did not: by your own logic, the majority of builds are made with `MAC_OS_X_VERSION_MIN_REQUIRED` way above 1060, which has always disabled the old path in this PR. The new symbol is strongly linked in this situation, so there’s no point doing anything else.

> the standard recommended mechanism for determining feature availability.

It isn’t standard (it’s an Apple extension), and it doesn’t even check features: it checks versions. This alone could be enough to avoid it, but we have to check the same versions in the preprocessor anyway. Which, by the way, the language branch is also missing.

> Nearly everyone building this is on "recentish" clang, and it's the far more important case to handle well.

We can handle modern Clang perfectly fine using weak-import checks. And this also handles old Clang, new GCC, old GCC and even LLVM-GCC. And it’s less code, too, as there is no need for `__has_builtin` at all: neither the polyfill definition nor the invocation.

> the vast majority of builds will be made with a minimum supported OS target newer than anything we're checking for here

And yet the vast majority of builds that will _run_ on platforms we’re bothering to check for at runtime will probably be made _on those same platforms._ Even when not exactly the same (if someone builds for redistribution), they’ll be somewhat close, because as you point out, Apple has been bumping the minimum allowed deployment target over time.

astiob: To be clear: in the scope of _this_ PR, the only scenario this check changes compared to the original state of this PR is if someone

1. builds libass on an old-enough Xcode that allows targeting both Leopard and Snow Leopard, setting Leopard as the deployment target (either building on Leopard itself or explicitly setting the deployment target if building on a newer one, IIUC that the default is to match the current machine),
2. and runs this build on Snow Leopard or newer,

which sounds like the exact opposite of what you consider a common scenario.

rcombs: > It isn’t standard (it’s an Apple extension)

It's the standard feature-check mechanism _for the platform_, and Apple's recommendation is that code be migrated from symbol null checks to `__builtin_available` checks; `__builtin_available` is also required to prevent `-Wunguarded-availability` and `-Wunguarded-availability-new` warnings, and the latter is on by default (not even just as part of `-Wall`). Android now uses the same mechanism in newer NDK versions, and specifically recommends setting `-Werror=unguarded-availability`. Using version checks allows us to declare entire functions as e.g. "available only on Android 29 or later", which permits unguarded use of newer symbols within that function, but require anything calling it to be behind an availability check (or to itself be an availability-annotated function).

barracuda156: > > recentish Clang
> 
> Yes, and also recentish libass. Nearly everyone building this is on "recentish" clang, and it's the far more important case to handle well.

It is important to handle all cases, as long as it doesn’t break anything. Not just personally relevant ones. That is how software improves and others have incentive to contribute and test.

> > anyone building on 10.9 or 10.10 won’t get language support even though their system supports it.
> 
> 
> I don't care about anyone building natively on a compiler so ancient that it doesn't even support the standard recommended mechanism for _determining feature availability_. The roughly 2 hobbyists in this situation can either get an OS from the past decade, get a _toolchain_ from the past decade, or stop being surprised when they don't new functionality.

You may not care, but other people do. Why break something for people who use the software?

astiob: Repushed with `if (false) {}` followed by a ladder of conditionally-compiled `else`s and with the commit message amended to add an updated test summary and a verbose explanation of my choices.

---

There’s a lot I wanted to say. (I took a break to cool down, but we can resume the debate elsewhere if you wish.)

But the bottom line is, using `__builtin_available` for a symbol older than `__builtin_available` itself simply nearly defeats the point of including a runtime check at all: in precisely the situations where it matters most, it will likely not be compiled in due to a lack of compiler support.

> the current SDK, for instance, doesn't support targeting anything older than 10.13

For my education, could you elaborate on what this means? In `AvailabilityMacros.h` (which, by the way, is careful to fall back to `weak_import` on compilers that lack `availability`), I don’t see as much as a warning even in MacOSX13.0.sdk, and all the old declarations are still there with appropriate availability attributes. In fact, I’ve been able to successfully build libass with `-mmacosx-version-min=10.5` [in GitHub CI](https://github.com/astiob/libass/actions/runs/4348435495/jobs/7596957526) with no warnings except warnings during the final linking that some components aren’t as backwards-compatible as I’m requesting. But the resulting binary apparently got heap overflows when running tests, which baffles me.

rcombs: > in precisely the situations where it matters most, it will likely not be compiled in due to a lack of compiler support.

I'm a lot more sympathetic about targeting old runtimes than old compilers. I'm fine with supporting older runtimes more-or-less indefinitely (though I wouldn't put in extra effort for it myself beyond disabling new features when informed they regress something badly), but it's essentially impossible to support all the functionality we want with old compilers without giving ourselves a whole lot of extra headaches (e.g. `stdatomic.h` is only supported in GCC 4.9 or clang 3.[?] or later, which is well past what Xcode 3 for macOS 10.5 included, and supporting threads without compiler atomics would be a whole lot of work).

People building software for (or even on) old runtimes can do so using modern compilers just fine; Apple doesn't ship a current clang that runs on macOS 10.5 (which isn't surprising; they don't ship _anything_ for that OS at this point), but there's no particular reason you couldn't use a third-party build, or cross-compile from a modern platform. The only serious reason to use a compiler that old is to compile old versions of software that no longer build with modern toolchains, and that's not really a concern for current software development. I don't see a lot of people seriously asking for software to support building on MSVC versions from 2009, much less having _new features_ support such builds; 13-year-old Xcode has similar "this should only be used by retro hobbyists for preservation of historical software" status, and that's not a scenario that affects us.

> For my education, could you elaborate on what this means? …In fact, I’ve been able to successfully build libass with `-mmacosx-version-min=10.5` [in GitHub CI](https://github.com/astiob/libass/actions/runs/4348435495/jobs/7596957526) with no warnings except warnings during the final linking that some components aren’t as backwards-compatible as I’m requesting.

Huh, current Xcode produces warnings when the deployment target is pre-10.13, but maybe that's just in the IDE? I suppose that implies that you could probably build a whole app targeting 10.5 with the current SDK as long as you didn't use Xcode itself, which if anything just makes building _on_ the Xcode 3 toolchain even less relevant.

barracuda156: > I suppose that implies that you could probably build a whole app targeting 10.5 with the current SDK as long as you didn't use Xcode itself

This won’t work, of course.

rcombs: > libass isn’t a Mac-specific software

Sure, but `ass_coretext` is (and `ass_afont` will be Android-specific).

> Users have no reason to expect that libass requires Clang specifically.

The modern SDK expects clang for a number of its features, and depending on clang for Apple-platform code is very widespread (see e.g. ffmpeg). It is essentially the only relevant C/C++ compiler for modern code on Apple platforms today.

> it would silently be compiled out, making the runtime check useless

The check would've done what it was supposed to: prevent code from attempting to run on a platform that may not support it, potentially leading to a crash.

>  both styles of runtime check work equally well.

This may be true in this specific case, but only because the API at issue here is so old that it's exempted from modern warnings. Newer APIs _require_ use of `__builtin_available` to avoid compiler warnings (unless the user disables those warnings, which is highly discouraged on both Apple and Android).

> The very goal of the affected code in this PR is to add support for Mac OS X 10.5, which runs quite officially on ppc32 and ppc64, so requiring Clang would soundly defeat the point.

I'm proposing that we require Clang in order to compile the _new-path_ code, not the old path.

Re: that Curl issue, note that they _didn't_ drop usage of `__builtin_available` in static builds, and continue to use it whenever they need to use such a check. The user didn't provide much information about the actual failure they were seeing, but my guess is that they were building `libcurl.a` on a modern clang, then trying to link against it using either an older clang or a gcc. The underlying symbol used by `__builtin_available` is `__isPlatformVersionAtLeast`, which is provided by the _static_ libcompiler_rt, not a dynamic lib.

To flesh out my priorities in a little more detail: runtime support for a reasonable set of modern platforms (going back at least as far back as the OS vendor still provides support) covering the vast majority of current usage > ability to make use of modern features at runtime > conformance with modern best practices > support for producing builds that _run_ on even older platform versions > ability to build the core feature set (but not necessarily all features) across a set of reasonably-modern toolchains > development simplicity > ability to build secondary features across moderately-older, proprietary, or otherwise-unpopular toolchains >>>>> support for compilers that are several years older than the software they're building.

So, for this specific check, since it's so old that the old path likely won't even be _compiled_ on the vast majority of builds, I suppose it doesn't really matter what we do here. My point here is that sooner or later, we're going to run into things that _do_ require newer compilers (C23! Apple and Android feature detection! Plenty of things in MSVC!), and I don't want us kneecapping our development process _forever_ by requiring that we maintain compatibility with ancient language versions just for the sake of a few weird retro-compiler hobbyists. There has to be _some point_ where we say a toolchain is no longer supported.

astiob: > Sure, but `ass_coretext` is

And a user has no way of knowing this, and no reason to.

> The modern SDK expects clang for a number of its features, and depending on clang for Apple-platform code is very widespread (see e.g. ffmpeg). It is essentially the only relevant C/C++ compiler for modern code on Apple platforms today.

A user on an old macOS has no reason to use a modern SDK, and indeed no _desire_ to use a modern SDK because it may be incompatible with their system.

> The check would've done what it was supposed to: prevent code from attempting to run on a platform that may not support it, potentially leading to a crash.

I don’t follow. Even if the platform is known to support it, even at the time of compilation, it’s still getting compiled out if the compiler doesn’t know what it means.

> This may be true in this specific case, but only because the API at issue here is so old that it's exempted from modern warnings.

This is a _big_ part of the issue. Not an afterthought; rather, a prime decision factor. This API, and the one you use in the language PR, are both old. They should therefore use the old check style.

> Newer APIs _require_ use of `__builtin_available` to avoid compiler warnings

And that’s fine, as fine as this can be: the APIs that do this come with compilers that support `__builtin_available`, and as I mention in the revised commit message, they (or rather, the things `ass_coretext.c` includes) fail to compile in anything other than Clang regardless of `__builtin_available` (due to block syntax unconditionally used in CGPath.h if nothing else), except perhaps old Apple GCC, but someone combining old Apple GCC with new SDKs is not a realistic scenario by any measure.

Those are not the APIs we’re discussing here.

> I'm proposing that we require Clang in order to compile the new-path code, not the old path.

But the new-path code contains nothing Clang-specific, either. Only the check does, and it brings no benefits.

> The underlying symbol used by `__builtin_available` is `__isPlatformVersionAtLeast`, which is provided by the _static_ libcompiler_rt, not a dynamic lib.

The point I was making is that, even if `libcompiler_rt` is static, its implementation may rely on 10.10+ or 10.13+ or whatever ABIs. _That_ may be the limiting factor when building with newer SDKs.

> requiring that we maintain compatibility with ancient language versions just for the sake of a few weird retro-compiler hobbyists.

I really don’t want this to become a personal argument, but I disagree with your attitude towards people on old platforms.

In any case, my belief is that using a new feature like `__builtin_available` really defeats the point of having code supporting old platforms in the first place. You haven’t fully explained what you want to achieve by adding it over just a compile-time check, but whatever it is, I don’t think it achieves that, and it regresses some scenarios.

barracuda156: To be clear, we do not necessarily need to support native Xcode compilers in order to support all systems from 10.4 onwards and all archs including ppc and ppc64. Supporting respective SDK is required, obviously, but recent GCC is available across the board (gcc7 on Tiger, possibly gcc10+, gcc13 on Leopard up).

TheOneric: > GCC upstream supports macOS at least from Leopard (10.5.8). There is zero problem with modern compilers on macOS PPC.

@barracuda156: For my education, what are the minimum x86 MacOS versions current upstream clang fully supports running on and building for? I know any PPC MacOS is not fully supported.

TheOneric:
@barracuda156, did you ever run this test? I can't immediately find the results.

> [2023-01-21]
> astiob: By the way, have you tested this on your prerelease Snow Leopard? [...]
> barracuda156: I will try today. [...]

libass/ass_coretext.c
astiob: Repushed with `Availability.h` instead of `AvailabilityMacros.h` (either of them works here) to match https://github.com/libass/libass/pull/607#discussion_r1133169911, which requires specifically that one.

barracuda156:
Could we merge this? Otherwise v. 0.17.1 still broken:
```
libass/ass_coretext.c: In function 'is_postscript_font_format':
libass/ass_coretext.c:69:25: error: 'kCTFontFormatOpenTypePostScript' undeclared (first use in this function); did you mean 'kCTFontFormatPostScript'?
   69 |         ret = format == kCTFontFormatOpenTypePostScript ||
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                         kCTFontFormatPostScript
libass/ass_coretext.c:69:25: note: each undeclared identifier is reported only once for each function it appears in
make[1]: *** [libass/ass_coretext.lo] Error 1
```

libass/ass_coretext.c
barracuda156: @TheOneric Sorry, somehow missed the question. I am not sure about Clang upstream, `clang-16` builds for x86_64 and i386 back to 10.6.8: https://ports.macports.org/port/clang-16/details (no idea about 10.5.8, no buildbot).

GCC upstream supports macOS at least back to 10.5.8. I have built GCC 13 on 10.6 PPC myself.

On 10.4.11 (Tiger) building newer GCC requires newer tools; otherwise the latest version may be restricted to GCC 10.4.0.

barracuda156:
> @barracuda156, did you ever run this test? I can't immediately find the results.
> 
> > [2023-01-21]
> > astiob: By the way, have you tested this on your prerelease Snow Leopard? [...]
> > barracuda156: I will try today. [...]

@TheOneric I believe so, but after 6 months I cannot recall for sure. (I definitely did run a test which I was asked to run here a few times.)

TheOneric:
@barracuda156 
> I think for discussed purpose SDK of 10.6 ppc and 10.5.8 match.

i vaguely remember something about pre-release 10.6 matching neither 10.5 nor 10.6 symbols and behaviour, but it has been a while.

> I believe so, but after 6 months I cannot recall for sure. (I definitely did run a test which I was asked to run here a few times.)

I can only find the [initial](https://github.com/libass/libass/issues/595#issuecomment-1312543802) probing test on pre-release 10.6 to get a grasp on how it deals with some fonts. This was *before* this patch series was created, since then I only see 10.5.8 test logs. Unless astiob, had some other concern I forgot about, I believe the lack of testing on actual pre-release 10.6 PPC is the only reason this hasn't been merged yet.
Also with the test reports being in the issue and not the PR comments its a bit hard to tell which version of the patches were tested.

@barracuda156, to remove any doubts can you build the current version of this PR and run [the test](https://github.com/libass/libass/issues/595#issuecomment-1043663556) on both 10.5.8 and pre-release 10.6 PPC and report the detailed logs for each?

TheOneric:
Sorry, actually hold on @barracuda156.

I thought the availability debate was settled, #607 declares a macro which also works for older compilers and I saw a “repushed with ...” comment here. But this PR currently does not use the new `CHECK_AVAILABLE` macro.

@astiob was there still something you disagree with about `CHECK_AVAILABLE` or was this patch just not yet updated?*(perhaps assuming #607 were to be merged imminently and waiting to be able to use the macro without introducing a conflict)*
For reference, this is the [current iteration](https://github.com/libass/libass/commit/b7e03ee03a6801aafb29597d4a339cec95ced9e4#diff-7166902c2f0e8e9dfbdda33ab79b30723451dfc27279983ed0b51402dffad650R36-R40) of `CHECK_AVAILABLE`: 

```C
#if __has_builtin(__builtin_available)
#define CHECK_AVAILABLE(sym, ...) __builtin_available(__VA_ARGS__)
#else
#define CHECK_AVAILABLE(sym, ...) (!!&sym)
#endif
```

barracuda156:
@TheOneric @astiob Please let me know when (and what specifically besides building) to test, I can do that on 10.5.8 and 10.6 ppc both.

astiob:
> was there still something you disagree with about `CHECK_AVAILABLE` or was this patch just not yet updated? _(perhaps assuming #607 were to be merged imminently and waiting to be able to use the macro without introducing a conflict)_

Off the top of my head, the current form of `CHECK_AVAILABLE` is good enough™ as we needed to reach a compromise. But as you point out, it’s in that other PR, and IIRC that PR ended up stuck due to Fontconfig font substitutes. Should I cherry-pick/copy-paste the macro here?

TheOneric:
yeah, copying the macro here seems like a good idea to me, given this will now probably be merged before #607

astiob:
@rcombs Please review https://github.com/libass/libass/pull/664/commits/4054edfdbee06264d1e916e336d6efc6bff52b8e, which I’ve put in your name as you wrote the code. (Tell me if you don’t want it or want it reworded or split.) Also, if a repeated test on my Mac (after I sleep) shows a warning in old Clang, I’d mildly prefer to readd a cast to `intptr_t` inside the `#else` to silence the warning, but I won’t if you say you dislike it.

@barracuda156 I think this is ready for a final test on your machines (hopefully I haven’t added any silly error during my rebase). More importantly on the Snow Leopard—in order to make sure the various changes haven’t accidentally broken it, given that Snow Leopard was the original reason you asked for any of this. But if it’s not too much trouble to re-verify both Leopard and Snow Leopard, that would be totally awesome too.

rcombs:
4054edfdbee06264d1e916e336d6efc6bff52b8e LGTM, and I have no issue with adding the `intptr_t` cast on the else path.

barracuda156:
@astiob Thank you very much! I will try and update you soon.

astiob:
Added the `intptr_t` cast with a comment. If @rcombs still doesn’t mind the edited commit in her name, then I believe this can be merged.

(For reference, as far as I can tell from Godbolt and documentation, Clang started to warn in version 3.5 and immediately supported this cast to suppress the warning, while GCC warns since version 4.2, including Apple’s, and supports this cast to suppress the warning since version 12.)

rcombs (approved):

PR #665, opened 2022-11-16
compare: fix 16-bit PNG read/write on big-endian platforms

astiob:
Before (bad): https://github.com/libass/libass-tests/actions/runs/3480858990/jobs/5821196824
~~After (good): https://github.com/libass/libass-tests/actions/runs/3482692960/jobs/5825311965~~
After (good): https://github.com/libass/libass-tests/actions/runs/3488593779/jobs/5837608707

### DIFF
--- a/compare/image.c
+++ b/compare/image.c
@@ -20,6 +20,11 @@
 #include <png.h>
 
 
+static inline bool is_little_endian(void)
+{
+    return *(char *) &(uint16_t) {1};
+}
+
 bool read_png(const char *path, Image16 *img)
 {
     FILE *fp = fopen(path, "rb");
@@ -79,7 +84,7 @@ bool read_png(const char *path, Image16 *img)
     ptrdiff_t half = 4 * w;
     if (depth == 8)
         ptr += half;
-    else
+    else if (is_little_endian())
         png_set_swap(png);
 
     for (uint32_t i = 0; i < h; i++) {
@@ -186,7 +191,7 @@ static bool write_png(const char *path, uint32_t width, uint32_t height,
                  PNG_COMPRESSION_TYPE_DEFAULT, PNG_FILTER_TYPE_DEFAULT);
     png_write_info(png, info);
 
-    if (depth > 8)
+    if (depth > 8 && is_little_endian())
         png_set_swap(png);
     png_write_image(png, rows);
     png_write_end(png, NULL);


compare/image.c
```diff
@@ -79,8 +92,6 @@ bool read_png(const char *path, Image16 *img)
     ptrdiff_t half = 4 * w;
     if (depth == 8)
         ptr += half;
-    else
-        png_set_swap(png);
```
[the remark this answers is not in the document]
MrSmile: I think that using an already existing infrastructure (`png_set_swap`) is better too. Also maybe autotools already have some standard tool for detecting endianness? It could be useful for big-endian ARM assembly too.

compare/image.c
```diff
@@ -20,6 +20,19 @@
...
+    const uint8_t *ptr8 = (const uint8_t *) ptr;
+    return (uint16_t) ptr8[0] << 8 | ptr8[1];
+}
+
+static inline void write_16be(uint16_t *ptr, uint16_t value)
+{
+    uint8_t *ptr8 = (uint8_t *) ptr;
+    ptr8[0] = value >> 8;
+    ptr8[1] = value;
+}
```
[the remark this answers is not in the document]
astiob: The POSIX headers that define them don’t exist on Windows. They seem to exist in a different, Windows-specific header, so we could get them with some ifdeffery. But why bother with inherently platform-specific stuff if we can make it work universally based on first principles?

compare/image.c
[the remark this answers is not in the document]
astiob: CPUs with runtime-modifiable endianness exist. Why limit ourselves when it’s so easy to make this work universally?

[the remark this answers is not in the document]
rcombs: Code still has to be compiled for a particular endianness on those systems; your `read_16be` will lower to either a straight load or a bswap depending on whether it's built for big- or little-endian, even on e.g. ARM (where endianness is indeed runtime-configurable).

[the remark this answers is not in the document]
MrSmile: In that case I think it's better to use `png_set_swap` anyway:
```c
static inline bool is_little_endian()
{
    uint16_t dst;
    uint8_t src[2] = { 1, 0 };
    memcpy(&dst, src, sizeof(dst));
    return dst == 1;

    // or use htons(), but it's additional dependency with different headers under *nix/windows
    // and looks like MSVC cannot inline it
}

...

if (is_little_endian())
    png_set_swap(png);
```

rcombs (approved):

astiob:
Pushed a variant with an explicit endianness check. Some things worth bikeshedding:
* Should it have more parentheses?
* Should it have fewer spaces?
* Is it too concise in any case and should it be expanded to a multi-liner?

It _is_ well-defined and well-behaved as far as I can tell. Casting pointers to `char` is allowed; reading them through `char` is allowed; `char` must be exactly 8 bits long because it’s the smallest type and `uint8_t` exists; neither `char` nor `uint16_t` can have padding bits (except this is mentioned in my draft copy of C11 but not in my draft copy of C99+TC3; but I haven’t been able to find any mention or discussion of this anywhere or any defect report).

MrSmile (approved):
I'm fine with it if you're sure it's not an UB.

compare/image.c
```diff
@@ -20,6 +20,11 @@
 #include <png.h>
 
 
+static inline bool is_little_endian(void)
+{
+    return *(char *) &(uint16_t) {1};
```
TheOneric, line 25: I was about to suggests changing `(char *)` to `(uint8_t *)` to avoid issues with C99-allowed implementations with padding bits *(consider e.g. `CHAR_BIT == 12 ` or `CHAR_BIT == 16` and to support `uint8_t` there are special read/write routines zeroing some upper bits)*, but then I found this in libpng’s `pngconf.h` header:
```C
#if CHAR_BIT == 8 && UCHAR_MAX == 255
   typedef unsigned char png_byte;
#else
#  error "libpng requires 8-bit bytes"
#endif
```

astiob: Nice find.

I chose `char` here to guarantee that the pointer punning is allowed. But, of course, `uint8_t` could be used with union punning instead.

But I’m not sure your scenario is possible. Are you thinking of the edge case I noted above where C99 seemingly fails to explicitly require that `uintN_t` have no padding bits? For what it’s worth, if we’re worried about that, we shouldn’t be using `uint16_t` for type punning, either—or `uint8_t` for that matter—but rather use signed types. But in any case, everything I’ve seen suggests everyone interprets this to be guaranteed in all versions of C. I must admit I can’t think of any better arguments though. It seems indeed that I already assumed `uint8_t` couldn’t have padding bits when I said this guaranteed `CHAR_BIT == 8` above.

But supposing it is possible, and supposing libpng didn’t have that check, our `image.c` would probably still fail spectacularly, because it assumes throughout that `uint16_t` is two `uint8_t`s.

(Now that I think about it, C doesn’t guarantee that the order of bits in an integer’s representation has _any_ particular properties, does it? They could legally be all jumbled up. So really, _any_ endianness check/assumption is inherently more platform-specific than the C standard allows, and only my original code is technically fully portable in this aspect? Which is indeed why it appealed to me, even if that portability is pretty much purely academic.)

astiob: Oh, here it is. At the start of the bigger section §7.18.1 <i><cite>Integer types</cite>:</i>

> When typedef names differing only in the absence or presence of the initial `u` are defined, they shall denote corresponding signed and unsigned types as described in 6.2.5; an implementation providing one of these corresponding types shall also provide the other.

So given that `int8_t` is defined to have no padding bits, and `uint8_t` is required to be present _iff_ `int8_t` is present and be its corresponding unsigned type, and corresponding types are defined to have the same storage size and the same representation for nonnegative values, there’s no room for `uint8_t` to have any padding bits.

(I must say I didn’t realize `uintN_t` required `intN_t` to be present. This implies `uintN_t` can’t be defined on platforms that lack an implementation of two’s complement. Oh well, file this under “useless trivia”.)

TheOneric: Ohh, with the presence of `uintN_t` and `intN_t` being required to be equivalent for each `N`, this is indeed not possible.
*If* it were allowed to only provide `uint8_t` and use padding bits for it, I think a machine could get away with using e.g. `CHAR_BIT == 12` and (because a type with 8bit arithmetic is recognised as important) provide special read/write instructions which always zero the upper 4 most significant bits. You’re still right though about other parts of `image.c` failing on such a hypothetical machine.

> Now that I think about it, C doesn’t guarantee that the order of bits in an integer’s representation has any particular properties, does it? They could legally be all jumbled up. So really, any endianness check/assumption is inherently more platform-specific than the C standard allows

I’m not aware of any requirements regarding bit ordering, but endianess checks should still be possible afaict.
If you obtain `char` and a type with `sizeof(...) == 2` you only need to ensure one of the `char`s will end up all zero and the other doesn't. E.g. by using `1` like here which sets the MSB to all zero and one bit (wherever it is actually located) of the LSB to one. Now checking the `char` located first in memory will either be zero (Big-Endian) or not zero (Little-Endian).

astiob: > I’m not aware of any requirements regarding bit ordering in a `char`, but endianess checks should still be possible afaict, since it acts on `char` rather than bit ordering.

No, I mean, you’re assuming that integers even _have_ an endianness to speak of—that is, this:

```c
/* given */ union { uint16_t u16; uint8_t u8[2]; };
assert((u16 >> 8) == u8[0] && (u16 & 0xFF) == u8[1]
    || (u16 >> 8) == u8[1] && (u16 & 0xFF) == u8[0]);
```

The 1 bit could be anywhere within the 16 bits of `uint16_t`. A `char` (or a `uint8_t`) doesn’t have to assign the same meaning to bits in the same positions as the lower half of a `uint16_t`, either. (In particular, it could be a trap representation in `signed char`, though `unsigned char` would avoid this.) In fact, the lower byte of a `uint16_t` may contain _both_ some of its most significant bits and some of its least significant bits, and the higher byte may contain the bits of middle significance (or vice versa).

TheOneric: I wanted to find out whether ISO C actually allows for this. Here are the relevant excerpts:

ISO C99 TC3, chapter 6.2.6.1 §2, §3 and the referenced footnote:
> Except for bit-fields, objects are composed of contiguous sequences of one or more bytes, the number, order, and encoding of which are either explicitly specified or implementation-defined.
>
> Values stored in unsigned bit-fields and objects of type unsigned char shall be represented using a pure binary notation. ⁴⁰⁾
>
> --------
>  40) A positional representation for integers that uses the binary digits 0 and 1, in which the values represented by successive bits are additive, begin with 1, and are multiplied by successive integral powers of 2, except perhaps the bit with the highest position. (Adapted from the American National Dictionary for Information Processing Systems.) A byte contains CHAR_BIT bits, and the values of CHAR_BIT type unsigned char range from 0 to 2^CHAR_BIT − 1 .

Further down in chapter 6.2.6.2 §1
> For unsigned integer types other than unsigned char, the bits of the object representation shall be divided into two groups: value bits and padding bits (there need not be any of the latter). If there are N value bits, each bit shall represent a different power of 2 between 1 and 2^N −1 , so that objects of that type shall be capable of representing values from 0 to 2^N − 1 using a pure binary representation; this shall be known as the value representation. The values of any padding bits are unspecified. ⁴⁴⁾

For C99 the following §2 starts with: 
> For signed integer types, the bits of the object representation shall be divided into three groups: value bits, padding bits, and the sign bit. There need not be any padding bits; there shall be exactly one sign bit. Each bit that is a value bit shall have the same value as the same bit in the object representation of the corresponding unsigned type [...]

Except for the last paragraph they didn't change with C11. For signed types C11 makes only a single signed bit being allowed more explicit and also requires `signed char` to have no padding bits.

The [current C23 draft N3047](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3047.pdf) adds this to 6.2.6.2 §1:
> The type bool shall have one value bit and (sizeof(bool)*CHAR_BIT)- 1 padding bits. Otherwise, there need not be any padding bits; unsigned char shall not have any padding bits.

And also changes the requirements for signed types, such that the sign bit must have been a value bit in the corresponding unsigned type and only a two’s complement representation is allowed. *(C99 and C11 also allowed for “sign and magnitude” and “one's complement”)*

---

The important takeaways are (assuming some other part in the standard doesn’t add further requirements):
 - Except for bitfields everything can be decomposed into full bytes (i.e. `(unsigned) char`). There are no gaps between bytes of an object.
 - `unsigned char` is required to use the naïve encoding and the value of each bit is monotone rising. So there’s at most the question whether `unsigned char`s bits are big- or little-endian; values/positions can't be jumbled.
 - In C11 `signed char` has no padding bits, meaning `unsigned char` has at most one padding bit. In C23 (which likely means already now no current vendor does this) `unsigned char` does not have any padding bits.
 - For other unsigned types each value bit must represent a distinct power of two
 - All unsigned integers *shall* be capable of representing 0-(2^N - 1) using the naïve position encoding. It doesn't read to me like they are actually mandated to use this encoding though.
 - I’m not sure if “value representation” refers to the pure binary notation they must be capable of or the representation actually used.
  - There's no requirements for padding bits to be contingent or to be placed at the start/end. (The latter also for the sign bit, but )

So tenatively yes, it seems like using a representation which jumbles the order/value for non-`unsigned char` types is allowed, albeit most likely not used in practice.

astiob: Continuing this (presumably) purely academic (but fun, as always) discussion:

> [For] `unsigned char` […] the value of each bit is monotone rising. So there’s at most the question whether `unsigned char`s bits are big- or little-endian; values/positions can't be jumbled.

I suppose it depends on how much you read into the words “positional” and “successive” in that footnote. Arguably, it says that a bit must exist for each power of two but not what order they’re physically stored in. Regardless, I don’t think it’s possible to _detect_ physical bit order within a byte, so specifying it for `unsigned char` doesn’t matter. For all intents and purposes, it does act _as if_ it was monotonically ordered, most significant bit on the “left” and least significant on the “right” in bit-shift operator terms. (I guess perhaps except through a union with bit fields, but IIRC bit fields themselves don’t have a specific defined order?)

What we _can_ detect is whether the same bits represent the same values in different types (by reinterpreting them via type punning). And what I was pointing out earlier is that this doesn’t seem to be required (other than for pairs of corresponding signed/unsigned integer types).

> For signed types C11 […] also requires `signed char` to have no padding bits.

> In C11 `signed char` has no padding bits, meaning `unsigned char` has at most one padding bit. In C23 (which likely means already now no current vendor does this) `unsigned char` does not have any padding bits.

C99 already forbids `unsigned char`—and arguably all character types—from having any padding bits, even if it may not make it quite as explicit. (This is precisely what allows us to read absolutely any type’s representation via an array of characters.)

  * For `unsigned char`, note in your quotes of §6.2.6.2 (emphasis mine):

    > For unsigned integer types _other than `unsigned char`,_ the bits of the object representation shall be divided into two groups: value bits and padding bits

    and of the footnote from §6.2.6.1:

    > A byte contains `CHAR_BIT` bits, and the values of type `unsigned char` range from $0$ to $2^{\mathtt{CHAR\\_BIT}} - 1$.

    So the concept of padding bits simply isn’t defined for `unsigned char`, and in any case `unsigned char` must necessarily be able to hold as many values as a byte’s bits could ever possibly form as value bits, which leaves no room for any of those bits to be padding.

  * For `signed char`, I was about to say the lack of padding bits follows from the previous fact combined with the requirements on corresponding signed/unsigned integer type pairs and on byte-based type representation. But as I was writing this out, I realized that if we’re being really technical, I can’t actually prove that a situation where e. g. `sizeof(signed char) == 2 && sizeof(unsigned char) == 2 && SCHAR_MAX == 127 && UCHAR_MAX == 65535` is truly forbidden. My best references so far are §6.2.5 ¶6:

    > For each of the signed integer types, there is a corresponding (but different) unsigned integer type (designated with the keyword `unsigned`) that uses the same amount of storage (including sign information) and has the same alignment requirements.

    ¶9:

    > The range of nonnegative values of a signed integer type is a subrange of the corresponding unsigned integer type, and the representation of the same value in each type is the same.

    §6.2.6.1 ¶4:

    > Values stored in non-bit-field objects of any other object type consist of $n \times \mathtt{CHAR\\_BIT}$ bits, where $n$ is the size of an object of that type, in bytes.

    and §5.2.4.2.1:

    > [The] implementation-defined values shall be equal or greater in magnitude (absolute value) to those shown, with the same sign.
    >
    >   *   minimum value for an object of type `signed char`  
            `SCHAR_MIN -127 //` $-(2^7-1)$
    >
    >   *   maximum value for an object of type `signed char`
            `SCHAR_MAX +127 //` $2^7-1$
    >
    > The value `UCHAR_MAX` shall equal $2^\mathtt{CHAR\\_BIT}-1$.  
    > <i>[But no such note is made about `SCHAR_...`]</i>

    Thus, `signed char` cannot have more bits than `unsigned char`, but I suppose nothing here really forbids some of `unsigned char`’s value bits from turning into padding in `signed char`. Valid nonnegative signed chars must be valid unsigned chars, but there’s no comparable requirement in the other direction.

    However, there’s also §6.2.6.1 ¶5, which describes trap representations (emphasis mine):

    > Certain object representations need not represent a value of the object type. If […] such a representation […] is read [or] produced […] by an […] expression _that does not have character type,_ the behavior is undefined. Such a representation is called a <i>trap representation.</i>

    which seems to single-handedly imply that all character type representations (including signed types) are always legal. But I guess it doesn’t define what they _mean?_

    (So actually, I _guess_ I really should’ve used `unsigned char`?)

  * For `char`, §6.2.5 ¶5:

    > An object declared as type `signed char` occupies the same amount of storage as a ‘‘plain’’ `char` object. 

    and ¶15:

    > The implementation shall define `char` to have the same range, representation, and behavior as either `signed char` or `unsigned char`.

    So if it’s unsigned, it’s definitely guaranteed padding-free, and if it’s signed, it hinges on your interpretation of the requirements for `signed char`.

PR #667, opened 2022-12-05
Checkasm v2

MrSmile:
Leftover from my checkasm branch. Changes from my previous version:
- updated to current dav1d source (`readtime` for macos is already there, so one less thing to worry);
- simplified rasterizer check (no copy-paste from actual rasterizer code).

As for `BENCH_RUNS` count, I feel that it's not enough: even with `(1 << 16)` and `rdtsc` from x86 there's quite large variation in results. I think the whole timing logic should be modified. Currently it's
```c
uint64_t tsum = 0;
int tcount = 0;
for (int ti = 0; ti < BENCH_RUNS; ti++) {
    uint64_t t = readtime();
    test_func(...);
    test_func(...);
    test_func(...);
    test_func(...);
    t = readtime() - t;
    if (some_black_magic) {
        tsum += t;
        tcount++;
    }
}
```
Basically, it's 4 inner timed iterations plus `BENCH_RUNS` outer iterations with "clever" averaging of the results (skipping of some partial timings). I think the following would be more stable, but I've not checked it yet:
```c
uint64_t tmin = -1;
for (int to = 0; to < OUTER_RUNS; to++) {
    uint64_t t = readtime();
    for (int ti = 0; ti < INNER_RUNS; ti++) {
        test_func(...);
        test_func(...);
        test_func(...);
        test_func(...);
    }
    t = readtime() - t;
    if (t < tmin)
        tmin = t;
}
```

Another point of concern is ABI check for `aarch64[_be]` (`checkasm_stack_clobber()`, `checked_call()`, etc.). It's explicitly disabled for macos and it's crashing under `aarch64_be` NetBSD from earlier. It's quite useful to disable completely, so I think I should look into it and try to understand what's going on there.

### DIFF
--- a/Makefile_util.am
+++ b/Makefile_util.am
@@ -50,6 +50,7 @@ endif
 	$(nasm_verbose)$(AS) $(ASFLAGS) -I$(top_srcdir)/libass/ -o $@ $<
 
 checkasm_checkasm_SOURCES = \
+    checkasm/rasterizer.c \
     checkasm/blend_bitmaps.c \
     checkasm/be_blur.c \
     checkasm/blur.c \
--- a/Makefile_util.am
+++ b/Makefile_util.am
@@ -52,6 +52,7 @@ endif
 checkasm_checkasm_SOURCES = \
     checkasm/blend_bitmaps.c \
     checkasm/be_blur.c \
+    checkasm/blur.c \
     checkasm/checkasm.h checkasm/checkasm.c
 
 checkasm_checkasm_CPPFLAGS = -I$(top_srcdir)/libass
--- a/Makefile_util.am
+++ b/Makefile_util.am
@@ -38,3 +38,35 @@ fuzz/fuzz_ossfuzz: fuzz/fuzz-fuzz.o libass/libass.la $(srcdir)/libass.pc
 		$$(pkg-config --static --libs $(srcdir)/libass.pc | sed -e 's/-lm //g' -e 's/-lass //g') \
 		-Wl,-Bdynamic
 endif
+
+
+if ENABLE_CHECKASM
+check_PROGRAMS += checkasm/checkasm
+TESTS += checkasm/checkasm$(EXEEXT)
+bench: run-checkasm-bench
+endif
+
+.asm.o:
+	$(nasm_verbose)$(AS) $(ASFLAGS) -I$(top_srcdir)/libass/ -o $@ $<
+
+checkasm_checkasm_SOURCES = \
+    checkasm/blend_bitmaps.c \
+    checkasm/be_blur.c \
+    checkasm/checkasm.h checkasm/checkasm.c
+
+checkasm_checkasm_CPPFLAGS = -I$(top_srcdir)/libass
+checkasm_checkasm_LDADD = libass/libass.la
+checkasm_checkasm_LDFLAGS = $(AM_LDFLAGS) -static
+
+if X86
+checkasm_checkasm_SOURCES += checkasm/x86/checkasm.asm
+endif
+if AARCH64
+checkasm_checkasm_SOURCES += checkasm/arm/checkasm_64.S
+endif
+
+run-checkasm: checkasm/checkasm$(EXEEXT)
+	checkasm/checkasm$(EXEEXT)
+
+run-checkasm-bench: checkasm/checkasm$(EXEEXT)
+	checkasm/checkasm$(EXEEXT) --bench
--- a/checkasm/checkasm.c
+++ b/checkasm/checkasm.c
@@ -1,0 +1,924 @@
+/*
+ * Copyright © 2018, VideoLAN and dav1d authors
+ * Copyright © 2018, Two Orioles, LLC
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#include "tests/checkasm/checkasm.h"
+
+#include <math.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <string.h>
+
+#include "src/cpu.h"
+
+#ifdef _WIN32
+#include <windows.h>
+#define COLOR_RED    FOREGROUND_RED
+#define COLOR_GREEN  FOREGROUND_GREEN
+#define COLOR_YELLOW (FOREGROUND_RED|FOREGROUND_GREEN)
+#else
+#include <unistd.h>
+#include <signal.h>
+#include <time.h>
+#ifdef __APPLE__
+#include <mach/mach_time.h>
+#endif
+#define COLOR_RED    1
+#define COLOR_GREEN  2
+#define COLOR_YELLOW 3
+#endif
+
+/* List of tests to invoke */
+static const struct {
+    const char *name;
+    void (*func)(void);
+} tests[] = {
+    { "msac", checkasm_check_msac },
+    { "refmvs", checkasm_check_refmvs },
+#if CONFIG_8BPC
+    { "cdef_8bpc", checkasm_check_cdef_8bpc },
+    { "filmgrain_8bpc", checkasm_check_filmgrain_8bpc },
+    { "ipred_8bpc", checkasm_check_ipred_8bpc },
+    { "itx_8bpc", checkasm_check_itx_8bpc },
+    { "loopfilter_8bpc", checkasm_check_loopfilter_8bpc },
+    { "looprestoration_8bpc", checkasm_check_looprestoration_8bpc },
+    { "mc_8bpc", checkasm_check_mc_8bpc },
+#endif
+#if CONFIG_16BPC
+    { "cdef_16bpc", checkasm_check_cdef_16bpc },
+    { "filmgrain_16bpc", checkasm_check_filmgrain_16bpc },
+    { "ipred_16bpc", checkasm_check_ipred_16bpc },
+    { "itx_16bpc", checkasm_check_itx_16bpc },
+    { "loopfilter_16bpc", checkasm_check_loopfilter_16bpc },
+    { "looprestoration_16bpc", checkasm_check_looprestoration_16bpc },
+    { "mc_16bpc", checkasm_check_mc_16bpc },
+#endif
+    { 0 }
+};
+
+/* List of cpu flags to check */
+static const struct {
+    const char *name;
+    const char *suffix;
+    unsigned flag;
+} cpus[] = {
+#if ARCH_X86
+    { "SSE2",               "sse2",      DAV1D_X86_CPU_FLAG_SSE2 },
+    { "SSSE3",              "ssse3",     DAV1D_X86_CPU_FLAG_SSSE3 },
+    { "SSE4.1",             "sse4",      DAV1D_X86_CPU_FLAG_SSE41 },
+    { "AVX2",               "avx2",      DAV1D_X86_CPU_FLAG_AVX2 },
+    { "AVX-512 (Ice Lake)", "avx512icl", DAV1D_X86_CPU_FLAG_AVX512ICL },
+#elif ARCH_AARCH64 || ARCH_ARM
+    { "NEON",               "neon",      DAV1D_ARM_CPU_FLAG_NEON },
+#elif ARCH_PPC64LE
+    { "VSX",                "vsx",       DAV1D_PPC_CPU_FLAG_VSX },
+#endif
+    { 0 }
+};
+
+typedef struct CheckasmFuncVersion {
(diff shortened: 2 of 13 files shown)


configure.ac
```diff
@@ -14,7 +14,7 @@ AM_PROG_AS
 AC_CHECK_HEADERS_ONCE([iconv.h])
 
 # Checks for library functions.
-AC_CHECK_FUNCS([strdup strndup])
+AC_CHECK_FUNCS([strdup strndup clock_gettime])
```
TheOneric, line 17: The check result for `clock_gettime` is never used *(`$ac_cv_func_clock_gettime` during configure or `HAVE_CLOCK_GETTIME` in the C preprocessor)*. If the result is not used, the check isn’t particularly useful.

checkasm/checkasm.c
```diff
@@ -0,0 +1,924 @@
...
+    LARGE_INTEGER i;
+    QueryPerformanceCounter(&i);
+    return i.LowPart;
+#elif defined(__APPLE__)
+    return (unsigned) mach_absolute_time();
+#else
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    return (unsigned) (1000000000ULL * ts.tv_sec + ts.tv_nsec);
+#endif
```
TheOneric, line 531: I know this is copied from dav1d, but do you know why `__APPLE__` isn't also using `clock_gettime` here? `clock_gettime`  is part of POSIX since 1993, so it should also exist on macOS.

TheOneric:
Regarding the `tcc`failure (only in checkasm) and `__attribute__(((aligned(x)))`  not working with `tcc` despite being documented brought up on IRC:

Attribute’s not working appears to actually be caused by a unfortunate (buggy?) macro definition in glibc which ends up removing all attributes during the preprocessor before the tcc compiler sees them: [tcc#57056](https://savannah.nongnu.org/bugs/?57056).
Employing either workaround of `#undef`ing the macro or using the alternative `__attribute((...))` variant helps.
However, it doesn't completly get rid of the failures, but the remaining ones are now limited to only AVX2 tests. Curiously the set of failing tests differs between the two workarounds. Tested with packaged tcc and current git HEAD. Maybe that’s just a side effect of the source being different like memory being placed differently though without deeper meaning.

If using musl or `#undef`ing `__attribute__`:
```
AVX2:
 - rasterizer.fill_solid_tile16     [OK]
 - rasterizer.fill_halfplane_tile16 [OK]
 - rasterizer.fill_generic_tile16   [OK]
 - rasterizer.merge_tile16          [OK]
 - rasterizer.fill_solid_tile32     [OK]
 - rasterizer.fill_halfplane_tile32 [OK]
 - rasterizer.fill_generic_tile32   [OK]
 - rasterizer.merge_tile32          [OK]
 - blend_bitmaps.add_bitmaps        [OK]
 - blend_bitmaps.imul_bitmaps       [OK]
 - blend_bitmaps.mul_bitmaps        [OK]
 - be_blur.be_blur                  [OK]
   stripe_unpack32_avx2 (segmentation fault)
 - blur.stripe_unpack32             [FAILED]
   stripe_pack32_avx2 (segmentation fault)
 - blur.stripe_pack32               [FAILED]
   shrink_horz32_avx2 (segmentation fault)
 - blur.shrink_horz32               [FAILED]
   shrink_vert32_avx2 (segmentation fault)
 - blur.shrink_vert32               [FAILED]
   expand_horz32_avx2 (segmentation fault)
 - blur.expand_horz32               [FAILED]
   expand_vert32_avx2 (segmentation fault)
 - blur.expand_vert32               [FAILED]
   blur4_horz32_avx2 (segmentation fault)
 - blur.blur4_horz32                [FAILED]
   blur4_vert32_avx2 (segmentation fault)
 - blur.blur4_vert32                [FAILED]
   blur5_horz32_avx2 (segmentation fault)
 - blur.blur5_horz32                [FAILED]
   blur5_vert32_avx2 (segmentation fault)
 - blur.blur5_vert32                [FAILED]
   blur6_horz32_avx2 (segmentation fault)
 - blur.blur6_horz32                [FAILED]
   blur6_vert32_avx2 (segmentation fault)
 - blur.blur6_vert32                [FAILED]
   blur7_horz32_avx2 (segmentation fault)
 - blur.blur7_horz32                [FAILED]
   blur7_vert32_avx2 (segmentation fault)
 - blur.blur7_vert32                [FAILED]
   blur8_horz32_avx2 (segmentation fault)
 - blur.blur8_horz32                [FAILED]
   blur8_vert32_avx2 (segmentation fault)
 - blur.blur8_vert32                [FAILED]
```

If using `__attribute((aligned(align)))` for `__TINYCC__`:
```
AVX2:
 - rasterizer.fill_solid_tile16     [OK]
 - rasterizer.fill_halfplane_tile16 [OK]
 - rasterizer.fill_generic_tile16   [OK]
 - rasterizer.merge_tile16          [OK]
   fill_solid_tile32_avx2 (segmentation fault)
 - rasterizer.fill_solid_tile32     [FAILED]
   fill_halfplane_tile32_avx2 (segmentation fault)
 - rasterizer.fill_halfplane_tile32 [FAILED]
   fill_generic_tile32_avx2 (segmentation fault)
 - rasterizer.fill_generic_tile32   [FAILED]
   merge_tile32_avx2 (segmentation fault)
 - rasterizer.merge_tile32          [FAILED]
 - blend_bitmaps.add_bitmaps        [OK]
 - blend_bitmaps.imul_bitmaps       [OK]
   mul_bitmaps_avx2 (segmentation fault)
 - blend_bitmaps.mul_bitmaps        [FAILED]
   be_blur_avx2 (segmentation fault)
 - be_blur.be_blur                  [FAILED]
 - blur.stripe_unpack32             [OK]
 - blur.stripe_pack32               [OK]
 - blur.shrink_horz32               [OK]
 - blur.shrink_vert32               [OK]
 - blur.expand_horz32               [OK]
 - blur.expand_vert32               [OK]
 - blur.blur4_horz32                [OK]
 - blur.blur4_vert32                [OK]
 - blur.blur5_horz32                [OK]
 - blur.blur5_vert32                [OK]
 - blur.blur6_horz32                [OK]
 - blur.blur6_vert32                [OK]
 - blur.blur7_horz32                [OK]
 - blur.blur7_vert32                [OK]
 - blur.blur8_horz32                [OK]
 - blur.blur8_vert32                [OK]
```

configure.ac
MrSmile: Removed.

MrSmile:
Disabled `checkasm` for `tcc` for the time being.

checkasm/checkasm.c
astiob: No, it doesn’t exit on macOS.

POSIX says:

> These functions are part of the Timers option and need not be available on all implementations.

TheOneric: Oh, I missed this being optional. Thanks!

Modern illumos [documents](https://illumos.org/man/3C/clock_gettime) `CLOCK_MONOTONIC` as being provided as an alias for `CLOCK_HIGHRES`. And sure enought checking the [current headers](https://github.com/illumos/illumos-gate/blob/9ecd05bdc59e4a1091c51ce68cce2028d5ba6fd1/usr/src/uts/common/sys/time_impl.h#L124-L126) both names are available (though the alias goes in the other direction).

rcombs: `clock_gettime` has been present on macOS since 10.12. Apple suggests replacing `mach_absolute_time` with `CLOCK_UPTIME_RAW`, which is also preferable over `CLOCK_MONOTONIC` for this use-case when available, since it isn't affected by NTP-driven rate changes (both are unaffected by NTP-driven time _jumps_). Specifically, Apple recommends using `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)`, which is slightly simpler to handle.

On macOS, `CLOCK_MONOTONIC`, `CLOCK_MONOTONIC_RAW`, and `CLOCK_UPTIME_RAW` are implemented in userspace, on top of `_mach_boottime_usec`, `mach_continuous_time`, and `mach_absolute_time` respectively. `mach_continuous_time` is just `mach_absolute_time`, except that it counts while the system is asleep, which is irrelevant to us. So, `mach_absolute_time` is probably preferable for our use-case, since we don't care about the actual time-base (optimally we'd be counting _cycles_, there's just no documented/unprivileged non-x86-specific way to do that on macOS), and `clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW)` and `clock_gettime_nsec_np (CLOCK_UPTIME_RAW)` have to call their underlying functions and then perform a time-base conversion.

Makefile_util.am
```diff
@@ -38,3 +38,39 @@ fuzz/fuzz_ossfuzz: fuzz/fuzz-fuzz.o libass/libass.la $(srcdir)/libass.pc
...
+	checkasm/checkasm$(EXEEXT)
+
+run-checkasm-bench: checkasm/checkasm$(EXEEXT)
+	checkasm/checkasm$(EXEEXT) --bench
+
+TESTS += checkasm/checkasm$(EXEEXT)
+
+bench: run-checkasm-bench
+
+endif
```
TheOneric, line 78: Since this is the only reference to `bench`, this means if libass was configured without checkasm support there’s no `bench` rule at all. I’m undecided on whether that’s good or bad.

A minor nit regarding the whole `if ENABLE_CHECKASM` block, but feel free to ignore: 
checkasm sources end up in the dist tarball either way, but it would be more consistent with our other conditional programs, to only enclose those lines into the conditional block which register the program. I.e. something like this:
```automake
if ENABLE_CHECKASM
check_PROGRAMS += checkasm/checkasm
TESTS += checkasm/checkasm$(EXEEXT)
# Not sure if the bench target should also be declared (without any operation) outside
bench: run-checkasm-bench
endif

checkasm_checkasm_SOURCES = ...
```

TheOneric (approved):
On IRC it was mentioned there might be a solution to running this on NetBSD aarch64_be, which would be great.
If it proves more stable in testing I have no objection to modifying the timing logic as suggested in the original post.

Makefile_util.am
MrSmile: Done.

MrSmile:
Implemented `aarch64_be` fix. Apple ARM has a different ABI, so it's impossible to enable register check for it.

As for timing, my earlier experiments show that there's no noticeable change to timing variation, both with enlarged `BENCH_RUNS` or two stage algorithm, so more experiments needed.

MrSmile:
I've not found a way to increase timing precision for my PC, so I've adopted a simple increase of `BENCH_RUNS` from the old **rcombs** branch for the time being. It runs around 40s on my x86 without any increase of precision, so if anybody has some good ideas I can try them instead.

TheOneric:
Unrelated to the following, I noticed dav1d’s checkasm now also prints out how the assembly routine’s throughput compares to the C implementation, like (`somefun_c: 120.0 (x1.0)` and `somefun_avx2: 34.5 (x3.5)`). If it’s not too troublesome, perhaps it’s a good idea to rebase on dav1d’s current `checkasm`? *(Nothing changed re timing afaict)*

---

> As for BENCH_RUNS count, I feel that it's not enough: even with (1 << 16) and rdtsc from x86 there's quite large variation in results.

I now tried to reproduce these variations to get a better idea of the problem and its gravity and it is quite bad indeed.
Visualising the results from 25 bench runs, the rectangle is centred around the arithmetic mean and extends in each direction by the standard deviation. The end of strokes extending from the rectangles upper/lower edge mark the maximum/minimum recorded time.¹ To deal with displaying results for routines whose time is on different magnitudes, everything is rescaled such that `avg == 100`.
runs_2pow16_excerpt

(Full plot: [runs_2pow12.pdf](https://github.com/libass/libass/files/10999394/runs_2pow12.pdf))

Foremost affected are `fill_generic_tile*` functions, then a couple functions of lesser severity follow and most functions do fine. Looking at the individual results of the tile-filling functions, their timing are actually all over the place and surprisingly evenly'ish spread unlike expected from proper benchmark results.
See for example times from `fill_generic_tile16_ssse3`; left column contains results ordered by time, right column shows the same results but in the order they were obtained in:
```
 110.1   133.8
 120.0   303.1
 121.7   544.4
 133.8   175.3
 174.9   544.3
 175.3   379.5
 178.9   121.7
 184.2   178.9
 184.4   446.4
 188.2   404.6
 191.0   303.2
 203.9   291.0
 220.0   256.3
 225.2   539.5
 254.7   638.8
 256.3   174.9
 291.0   617.7
 303.1   203.9
 303.2   120.0
 316.5   520.8
 319.3   220.0
 343.3   110.1
 379.5   319.3
 404.6   605.3
 417.0   441.3
 441.3   417.0
 446.4   254.7
 515.5   515.5
 520.8   343.3
 539.5   184.4
 544.3   225.2
 544.4   184.2
 605.3   316.5
 617.7   188.2
 638.8   191.0
```

A rerun gives consistently fishy results for tile-filling functions and a few other functions suffered from 1 or two outliers with greatly increased times. But apart from those they are consistent and outliers are plausibly explained by the environment having been less controlled during the rerun.
With `1<<12` the outlier problem got worse while tile-filling function times remain even'ish spread.

I suspect no matter how large `BENCH_RUNS` is chosen, the tile-filling functions won't get more consistent and something is flawed with the their test or timing procedure.

I haven't checked all, but it seems like tile-filling functions are the only ones which get such odd results. The other functions with high `stddev` I checked just had small absolute times and were therefore more affected by noise and resulting outliers.
For those high global `BENCH_RUNS` or specifically running those functions more often than long-running ones might make help.

Scripts to reproduce the setup and plots (requires GNU Awk and gnuplot) [attached](https://github.com/libass/libass/files/10999479/667_consistencyplot.zip). Process then looks like:
```
{ for i in {1..30} ; do ../checkasm/checkasm --bench; echo ""; sleep 5; done; } \
  | tee -a runs01.log
make runs01.pdf
```

@MrSmile, are the same functions affected for you? In the same way?
@rcombs, how are timing consistency and outliers without `rdtsc` on your aarch64 Mac?

.

If this doesn't lead to new ideas about the cause or a solution, imho it still makes sense to merge checkasm for its correctness tests and make a note about `bench` being iffy.

---

[1] While better than staring at a list of numbers,I’m not fully happy with the visualisation and open for better ideas. Symmetric extensions around the average doesn't make much sense, since we expect proper benchmark data to be inherently asymmetric. *(a hard lower bound and in a sufficiently controlled environment most results are close to the lower bound and (only) a few outlier of higher time consumption)*
On the upside the inwards facing minima strokes this often leads to, or rather their *lack* for tile-filling functions already serves as a hint of something being amiss.

MrSmile:
Yes, timing of  `fill_generic_tile` is [greatly dependent on its input](https://github.com/MrSmile/libass/blob/checkasm/checkasm/rasterizer.c#L161), which currently generated by random sampling. So for timing variation purposes it's necessary to use `--seed` parameter. I'll look into stabilizing `bench_new()` run (is it permitted to reset `checkasm` random generator for that function only?), but correctness tests before that call should be random. I've already hit a problem with not enough parameter variation (some blur functions have an even/odd height branch, but default tests had a fixed height), so I think that width/height should be randomized more compared to current deterministic approach.

MrSmile:
Stabilized timing of `fill_generic_tile`. As for upstream dav1d checkasm, there's the new `--affinity` parameter, which requires threading headers, so I think it's better to postpone update until threading PR.

TheOneric (approved):
Looks much better now! Shorter tests remain more susceptible to background noise like before, but that’s not as bad as the instability before and can in theory be avoided by better controlling the benchmarking environment.

Curiously I also observed the opposite of “background noise leading to a bogus spike in measured time” with `fill_solid_tile16_c` and `fill_halfplane_tile16_c` on two independent sets of 35 runs each. Here most runs congregated around a common time plus-minus a couple cycles, but one to three measurements per set ended up with a much lower time than the rest (e.g. usually around `60` with two measurements around `10`). I spot nothing causing such behaviour in the code and my best guess is, this is due to some lucky coincidence with speculative-execution and/or cache. We can probably gloss over it, though it’d be interesting to see if this also happens on different hardware.
Full plots (using the same visualisation as before): [set1.pdf](https://github.com/libass/libass/files/11182767/runs_pow16.pdf) and [set2.pdf](https://github.com/libass/libass/files/11182769/runs_pow16_repeat1.pdf) and [raw data](https://github.com/libass/libass/files/11182771/667_230403-32dea34_timings.tar.gz) if anyone’s interested.

PR #674, opened 2023-01-22
Simpler OS/2 table access

astiob:
@rcombs You introduced the code in `ass_get_font_weight` in https://github.com/libass/libass/pull/316/commits/6c4dca2c8ed20d02ac877d55fe25d7db13ace615, but it wasn’t consistent with our pre-existing uses of the OS/2 table, so this makes them consistent. I don’t think the `version != 0xffff` check does anything useful. I’ve tried to guess why you may have originally included it; see the commit message: does that sound plausible?

Cc @moi15moi: you’ve asked me about the version check.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -504,7 +504,7 @@ void ass_face_set_size(FT_Face face, double size)
 int ass_face_get_weight(FT_Face face)
 {
     TT_OS2 *os2 = FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
-    if (os2 && os2->version != 0xffff && os2->usWeightClass)
+    if (os2 && os2->usWeightClass)
         return os2->usWeightClass;
     else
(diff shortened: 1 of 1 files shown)


rcombs (approved):

moi15moi:
I think it is usefull when a font doesn't have an OS/2 table.
From [Freetype documentation](https://freetype.org/freetype2/docs/reference/ft2-truetype_tables.html#tt_os2)
> Note that we now support old Mac fonts that do not include an ‘OS/2’ table. In this case, the version field is always set to 0xFFFF.


You should this with this font: [Font without OS2 table.zip](https://github.com/libass/libass/files/10473574/Font.without.OS2.table.zip)

astiob:
Please re-read my commit message.

astiob:
But just to be sure, I’ve now run a basic libass test with the font you provided (thanks!): it does look like a plausible rendering of that font, at the very least.

PR #676, opened 2023-01-22
Improve weight and faux-italic compatibility with VSFilter

rcombs:
- Match GDI's special-case behavior for font weight values from 1 to 9.
- Match GDI's italicization transform matrix.

Both issues are demonstrated in the attached file:
[With_Jester.mkv.zip](https://github.com/libass/libass/files/10475605/With_Jester.mkv.zip)

This font used has weight 5.
[Jester Original.ttf.zip](https://github.com/libass/libass/files/10475612/Jester.Original.ttf.zip)

VSFilter rendering:
im6bZ3sR

Current libass rendering (faux-bolded unnecessarily; sheared roughly 12 degrees rather than vsfilter's 20):
8KbSGDdI

With this PR (perfect match):
italic-new7

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -504,10 +504,31 @@ void ass_face_set_size(FT_Face face, double size)
 int ass_face_get_weight(FT_Face face)
 {
     TT_OS2 *os2 = FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
-    if (os2 && os2->usWeightClass)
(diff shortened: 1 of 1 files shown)


astiob:
Ooh, nice!

I’ve been noticing lately that the amount of horizontal space we add after italics differs from VSFilter’s. Here’s hoping that this will now make it match or make it easier to figure out what amount VSFilter/GDI adds. If any.

Issue #613, opened 2022-04-18
Rendering: libass does not use the same font than VSFilter


moi15moi:
### Screenshots

It is useless in this case.

### Description of the issue

libass does not respect vsfilter font used.
Like you can see in the Additional info section, I give what I write in the \b tags and which font is been displayed.

I test it with AegisubDC 9214 and I use subtitle provider vsfiltermod_textsub and of course, libass.

Of course, this is only an example. VSFilter does maybe some other choice when x font weight is not installed.
In my example, all the possible font weight are available.

### libass version

AegisubDC 9214

### Is it a regression?

I don't know

### ASS Sample

See the MODDED_FONT.zip file

### Special Fonts

Here are the modded font I used and the .ass: [MODDED FONT.zip](https://github.com/libass/libass/files/8508672/MODDED.FONT.zip)
### System Information

OS: Windows 10
CPU: i5-7600

### Additional info

**VSFilter**

\bVALUE --> FONT_WEIGHT

b0 --> 400
b1 --> 700
b2-b99 --> 400
b100-b149 --> 100
b251-b258 --> 100
b150-b250 --> 200
b351-b358 --> 200
b259-b349 --> 300
b451-b461 --> 300
b350 --> 400
b359-b450 --> 400
b551-b561 --> 400
b462-b550 --> 500
b551-b661 --> 500
b562-b650 --> 600
b751-b758 --> 600
b662-b749 --> 700
b750 --> 800
b759-b850 --> 800
b951-b1050 --> 800
b851-b950 --> 900
b1051-infinite --> 900

**libass**

\bVALUE --> FONT_WEIGHT

b0 --> 400
b1 --> 700
b2 à b99 --> 400
b100-b149 --> 100
b150-b250 --> 200
b251-b349 --> 300
b350-b450 --> 400
b451-b550 --> 500
b551-b650 --> 600
b651-b749 --> 700
b750-b850 --> 800
b851-infinite --> 900

astiob:
> and I use subtitle provider vsfiltermod_textsub

It shouldn’t matter in this instance, but for future reference, please use “CSRI/vsfilter_textsub”. VSFilterMod is a separate beast that deviates from VSFilter in multiple ways and is explicitly meant only for hardsubs, where renderer compatibility is not a concern.

---

@rcombs IIRC you looked at GDI’s font selection including specifically faux bold. Does this happen to ring any bells?

[...]

Issue #305, opened 2018-03-11
Faux bold on Fansub Block does nothing in VSFilter


astiob:
While we’re on the topic of faux font styles, it seems that our faux bold is wrong too. However, all I know so far is that the popular “Fansub Block” font (which has a single glyph that is a rectangle) doesn’t seem to be emboldened in VSFilter _at all._

VSFilter:

![[Underwater-FFF] Saki Achiga-hen 04 [BD] rendered in VSFilter](https://user-images.githubusercontent.com/515193/37259104-e67039f8-258a-11e8-8664-e2442efd482a.png)

VSFilter with `\b0` looks the same.

libass:

![[Underwater-FFF] Saki Achiga-hen 04 [BD] rendered in libass](https://user-images.githubusercontent.com/515193/37259118-234a98dc-258b-11e8-9d83-0d9d4ec992c0.png)

Look at the right edge of the “Hariu Eri” box.

libass with `\b0`:

![[Underwater-FFF] Saki Achiga-hen 04 [BD], with an additional `\b0`, rendered in libass](https://user-images.githubusercontent.com/515193/37259236-5c84d15c-258c-11e8-8fb3-55d71ccd5f73.png)

```
Style: SIGN,Amira,45,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,-1,0,0,0,100,100,0,0,1,0,0,8,15,45,30,1
Dialogue: 0,0:10:54.17,0:10:57.13,SIGN,Ann,0,0,0,,{\fad(450,0)\fnFansub Block\an5\fscx792\fscy201.25\pos(980,432)\alpha&H90&\blur2\c&H000000&\iclip(m 768 312 l 1148 312 1149 466 768 466)}A
Dialogue: 1,0:10:54.17,0:10:57.13,SIGN,Ann,0,0,0,,{\fad(450,0)\fnFansub Block\an5\c&HEFEFEF&\fscx792\fscy201.25\pos(970,422)\blur0.5}A
Dialogue: 3,0:10:54.17,0:10:57.13,SIGN,Txt,0,0,0,,{\fnCalisto MT\fad(450,0)\fs32\blur0.5\c&H000000&\an5\pos(975,419)}Commentator\NHariu Eri
```

[...]

PR #678, opened 2023-01-23
fontselect: overhaul font scoring and faux-ing

### LINKED ISSUE #437: Improve score-based font selection and faux-bold behavior
- We should match the score system described here: https://github.com/libass/libass/pull/362#issuecomment-705885810
- An additional caveat: if fsSelection bit 5 is set, faux-bold will never occur, and the font will not receive the 120-weight-point-equivalent score bonus for it
- Whether or not a font is italic (for the purposes of both faux-italic and scoring) is determined entirely by fsSelection bit 0 (ITALIC), whereas freetype also uses bit 9 (OBLIQUE)

Today someone ran into a font (Bona Nova) which has its regular variant at weight=300 and its bold at weight=400, with regular having fSelection=REGULAR and bold having fSelection=BOLD. This means that at \b400 the bold variant is selected, but at \b700 the _regular_ variant is selected, as the 120-point simulated-bold bonus pushes it closer to the requested weight than 400…
image
except that it doesn't seem to actually get faux-bolded in the output???? I'm not sure why that is. It seems like the glyph metrics are affected, but not the actual shapes:
image
…which probably means our emboldener is handling at least some cases wrong, since afaik we never touch metrics. I'm not completely sure what's going on here. Seems like Windows's emboldener either adjusts metrics or adjusts glyph forms, but not both??? This part needs further investigation.

Simpler VSFilter demo of the matching behavior; \b600 matches the Regular variant with faux-bold (400+120=520, which is a closer match for 600 than 700 is).
image

For a font family that contains a weight=700 variant with fsSelection=BOLD, `\b900` gives the same result as `\b700` (vs in libass, where `\b900` results in faux-bolding the bold). This is likely a regression in ac75f67697845efc1b0a15693c96dcc9e0f71518; we need to check _both_ conditions.

rcombs:
Fixes #437.

This emulates GDI's behavior, both around scoring and faux decisions.

The fontconfig provider doesn't provide the full information we'd need to handle this 100% accurately, so it's best-guessed.

This branch is on top of #676, basically just to avoid rebase issues.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -548,6 +548,29 @@ int ass_face_get_weight(FT_Face face)
     }
 }
 
+static FT_Long fsSelection_to_style_flags(uint16_t fsSelection)
+{
+    FT_Long ret = 0;
+
+    if (fsSelection & 1)
+        ret |= FT_STYLE_FLAG_ITALIC;
+    if (fsSelection & (1 << 5))
+        ret |= FT_STYLE_FLAG_BOLD;
+
+    return ret;
+}
+
+FT_Long ass_face_get_style_flags(FT_Face face)
+{
+    // If we have an OS/2 table, compute this ourselves, since FreeType
+    // will mix in some flags that GDI ignores.
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
+    if (os2)
+        return fsSelection_to_style_flags(os2->fsSelection);
+
+    return face->style_flags;
+}
+
 /**
  * \brief Get maximal font ascender and descender.
  **/
@@ -693,9 +716,12 @@ bool ass_font_get_glyph(ASS_Font *font, int face_index, int index,
                 index);
         return false;
     }
-    if (!(face->style_flags & FT_STYLE_FLAG_ITALIC) && (font->desc.italic > 55))
+
+    FT_Long style_flags = ass_face_get_style_flags(face);
+    if (!(style_flags & FT_STYLE_FLAG_ITALIC) && (font->desc.italic > 55))
         ass_glyph_italicize(face->glyph);
-    if (font->desc.bold > ass_face_get_weight(face) + 150)
+    if (!(style_flags & FT_STYLE_FLAG_BOLD) &&
+        font->desc.bold > ass_face_get_weight(face) + 150)
         ass_glyph_embolden(face->glyph);
     return true;
 }
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -62,7 +62,6 @@ struct font_info {
 
     int slant;
     int weight;         // TrueType scale, 100-900
-    int width;
 
     // how to access this face
     char *path;            // absolute path
@@ -317,7 +316,6 @@ get_font_info(FT_Library lib, FT_Face face, const char *fallback_family_name,
     // fill our struct
     info->slant  = slant;
     info->weight = weight;
-    info->width  = 100;     // FIXME, should probably query the OS/2 table
 
     info->postscript_name = (char *)FT_Get_Postscript_Name(face);
     info->is_postscript = !FT_Get_PS_Font_Info(face, &postscript_info);
@@ -391,7 +389,7 @@ ass_font_provider_add_font(ASS_FontProvider *provider,
                            int index, void *data)
 {
     int i;
-    int weight, slant, width;
+    int weight, slant;
     ASS_FontSelector *selector = provider->parent;
     ASS_FontInfo *info = NULL;
     ASS_FontProviderMetaData implicit_meta = {0};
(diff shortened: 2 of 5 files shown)


TheOneric:
Does this help with #305?

rcombs:
No; I haven't yet determined why GDI sometimes fails to faux-bold when it nominally should (or does something weird with metrics instead, as seen in Bona Nova in #437).

astiob:
It feels a bit weird to convert `fsSelection` to a custom bitmask in fontselect but use it raw in `ass_font_get_glyph`. Maybe `ass_face_get_fsSelection` should just be `ass_face_get_style_flags` returning the custom bitmask instead?

rcombs:
Rebased with `ass_face_get_style_flags` returning the freetype mask constants. Might want to wait for #683?

astiob:
Nice, thanks!

> Might want to wait for #683?

I was thinking of merging this first, because #683 makes all of this PR’s DirectWrite changes redundant—but you’ve put in some good effort into them and we might find it useful to have them in Git history later.

moi15moi:
This should also close #613

libass/ass_directwrite_info_template.h
```diff
@@ -12,12 +12,22 @@ static bool NAME(FONT_TYPE)(FONT_TYPE *font,
...
-                 (style == DWRITE_FONT_STYLE_OBLIQUE)? FONT_SLANT_OBLIQUE :
-                 (style == DWRITE_FONT_STYLE_ITALIC) ? FONT_SLANT_ITALIC : FONT_SLANT_NONE;
+    meta->style_flags = style == DWRITE_FONT_STYLE_NORMAL ? 0 : FT_STYLE_FLAG_ITALIC;
+
+    const char* data;
+    UINT32 size;
+    void* context;
+    hr = font->lpVtbl->TryGetFontTable(font, DWRITE_MAKE_OPENTYPE_TAG('O', 'S', '/', '2'),
+                                       &data, &size, &context, &exists);
+    if (!FAILED(hr)) {
```
moi15moi, line 24: You should probably use [SUCCEEDED](https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-succeeded) instead of [!FAILED](https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-failed).

libass/ass_font.c
```diff
@@ -548,6 +548,29 @@ int ass_face_get_weight(FT_Face face)
...
+}
+
+FT_Long ass_face_get_style_flags(FT_Face face)
+{
+    // If we have an OS/2 table, compute this ourselves, since FreeType
+    // will mix in some flags that GDI ignores.
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
+    if (os2)
+        return ass_fsSelection_to_style_flags(os2->fsSelection);
+
```
moi15moi, line 570: GDI also has another edge case.
If there is a [head](https://learn.microsoft.com/en-us/typography/opentype/spec/head) table, it will use the bold/italic of the mac_style.
If will help for mac font like in #679

astiob (approved):

libass/ass_fontselect.c
```diff
@@ -445,21 +440,12 @@ ass_font_provider_add_font(ASS_FontProvider *provider,
     for (j = 0; j < meta->n_fullname; j++)
         printf("'%s' ", meta->fullnames[j]);
     printf("\n");
-    printf("  slant: %d\n", meta->slant);
+    printf("  style_flags: %x\n", (unsigned)meta->style_flags);
```
TheOneric, line 443: Considering it’s `FT_Long`, `%lx` and `(unsigned long)` or perhaps to make sure even `%llx` and `(unsigned long long)` should be a better fit here

libass/ass_fontselect.c
```diff
@@ -389,13 +385,12 @@ ass_font_provider_add_font(ASS_FontProvider *provider,
...
 {
     int i;
-    int weight, slant;
     ASS_FontSelector *selector = provider->parent;
     ASS_FontInfo *info = NULL;
     ASS_FontProviderMetaData implicit_meta = {0};
+    FT_Face face;
 
     if (!meta->n_family) {
-        FT_Face face;
```
TheOneric, line 400: afaict `face` isn’t actually used outside of this `if` block (anymore)

TheOneric (approved):
just two small nits

libass/ass_fontselect.c
astiob: `%lx` and no cast should be enough IMO. `FT_Long` is [documented](https://freetype.org/freetype2/docs/reference/ft2-basic_types.html#ft_long) to be `signed long` and doesn’t seem subject to change. For comparison, FreeType also has `FT_Int32` and `FT_Int64`, whose underlying types are unspecified.

libass/ass_fontselect.c
rcombs: Reverted.

libass/ass_fontselect.c
rcombs: Done, and also:
- Changed the type of `style_flags` to within `ASS_FontInfo` to `FT_Long`
- Made `fsSelection_to_style_flags` static (it's no longer used outside `ass_font.c`)

PR #681, opened 2023-01-31
vc: fix struct init error. supported since c23

wang-bin:
https://en.cppreference.com/w/c/language/struct_initialization


rcombs (approved):

astiob:
Thanks!

Not quite sure how this slipped in. Do GCC and Clang allow this nowadays? I didn’t know they do!

wang-bin:
> Thanks!
> 
> Not quite sure how this slipped in. Do GCC, Clang and even TCC allow this nowadays? I didn’t know they do!

only msvc fails. https://godbolt.org/z/GG6PYWeEe

Issue #675, opened 2023-01-22
ass_directwrite sees wrong full name in fonts with STAT table


astiob:
Manrope-Regular.ttf from https://fonts.google.com/specimen/Manrope is a named instance of a variable font exported as a non-variable font snapshot. It contains a [`STAT` table](https://learn.microsoft.com/en-us/typography/opentype/spec/stat) that explains how it relates to the other fonts in its family. This table says that this instance corresponds to the default position on each variation axis.

Hence, DirectWrite drops all axis names and reports the font’s “full name” as “Manrope”.  
But the (an?) actual name used by GDI is, as usual, the name with ID 4 in the `name` table: “Manrope Regular”.

As such, `ass_directwrite` fails to use this font for `\fnManrope Regular` (whereas it works with memory fonts, `ass_fontconfig`, and `ass_coretext` [except on old Mac OS X](https://github.com/libass/libass/issues/595#issuecomment-1324548120), as well as in VSFilter):

1. We request “Manrope Regular” from GDI.
2. GDI successfully finds it.
3. We convert it to a DirectWrite font.
4. We ask DirectWrite for the font’s full name.
5. DirectWrite says “Manrope”.
6. We add the font to our internal list under the name “Manrope”.
7. No fonts in our internal list have the name “Manrope Regular”.

[Removing](https://github.com/libass/libass/files/10474935/Manrope-Regular-woSTAT.zip) the font’s `STAT` table using fontTools TTX makes DirectWrite report “Manrope Regular”.

<i>Originally discovered in https://github.com/libass/libass/issues/595#issuecomment-1399540255</i>

This should be fixable by ignoring the names supplied by DirectWrite and reading them on our own, just as we do for Core Text. We can afford this, now that we query DirectWrite fonts on demand. (We have also long known that we need to do this for Fontconfig.)

Issue #459, opened 2020-12-07
Not loading font for word "SĄD"


MatiasMovie:
Not loading font for word "SĄD" (first second)
MPV
https://i.imgur.com/HrEUE0x.png
MPC (XySubFilter)
https://i.imgur.com/7m8LsJc.png

astiob:
We can’t help without seeing the MKV or at least the font file.

Your subtitles work for me with a copy of the same font from another file.

MatiasMovie:
Fonts
https://mega.nz/folder/1Qc0TA7A#s97DxtK0UP7b0J-iegNmMw

mkv name
[Die Hot 4 - 4 keepz] Great Pretender - 04 [475596EC].mkv

TheOneric:
On Tue, Dec 08, 2020 at 01:56:54 -0800, MatiasMovie wrote:
> Fonts
> https://mega.nz/folder/1Qc0TA7A#s97DxtK0UP7b0J-iegNmMw
> 
> mkv name
> [Die Hot 4 - 4 keepz] Great Pretender - 04 [475596EC].mkv

Going by the filename’s the linked font folder does _not_ have the required 
“FOT Manyo Sosho Std Stripped EB” font, only a similarly named 
“ManyoGyoshoStd” font (filename: FOTMANYOGYOSHOSTDSTRP-E.otf).

The name of the mkv is irrelevant, the important part is its contents, in this 
case especially its attached fonts. I suspect the mkv is missing the required 
font as well.
You can use eg MKVToolnix to either check the attached fonts and upload 
them if required, or if your not sure what to look for just make a cut (also 
MKVToolnix) of your mkv file and upload eg only the first four seconds of it. 

In case this is is not just a missing font afterall: What system are you using 
libass on (*Linux, MacOS, Windows, …) and if you know: which font provider is 
used (fontconfig, Coretext or DirectWrite)?

astiob:
Neither of those fonts is in the MKV:

```
$ mkvmerge -i "[Die Hot 4 - 4 keepz] Great Pretender - 04 [475596EC].mkv"
File '/home/common/unprotected/transmission/[Die Hot 4 - 4 keepz] Great Pretender - 04 [475596EC].mkv': container: Matroska
Track ID 0: video (MPEG-4p10/AVC/h.264)
Track ID 1: audio (AAC)
Track ID 2: subtitles (SubStationAlpha)
[...]

PR #683, opened 2023-02-08
directwrite: read metadata on our own, use DirectWrite-reported name as fallback

astiob:
DirectWrite sometimes returns names that differ from GDI’s. Instead of trusting it, use our own code that’s meant to be GDI-compatible, even if it costs us some extra resources.

This fixes https://github.com/libass/libass/issues/675 and facilitates future improvements to long font name matching (https://github.com/libass/libass/issues/459).

Plus a few minor improvements/precautions for exotic fonts/situations where we fail to find the same family name as DirectWrite does (or any family name at all).

Tested lightly: “Manrope Regular” works, and the added non-IDWriteFontFace3 code path works.

### DIFF
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -580,23 +580,6 @@ static char *get_fallback(void *priv, ASS_Library *lib,
     return family;
 }
 
-static int map_width(enum DWRITE_FONT_STRETCH stretch)
-{
-    switch (stretch) {
-    case DWRITE_FONT_STRETCH_ULTRA_CONDENSED: return 50;
-    case DWRITE_FONT_STRETCH_EXTRA_CONDENSED: return 63;
-    case DWRITE_FONT_STRETCH_CONDENSED:       return FONT_WIDTH_CONDENSED;
-    case DWRITE_FONT_STRETCH_SEMI_CONDENSED:  return 88;
-    case DWRITE_FONT_STRETCH_MEDIUM:          return FONT_WIDTH_NORMAL;
-    case DWRITE_FONT_STRETCH_SEMI_EXPANDED:   return 113;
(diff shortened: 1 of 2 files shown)


libass/ass_directwrite.c
```diff
@@ -592,8 +592,15 @@ static char *get_fallback(void *priv, ASS_Library *lib,
 #include "ass_directwrite_info_template.h"
 #undef FONT_TYPE
 
+#define FONT_TYPE IDWriteFont
+#define FAMILY_AS_ARG
+#include "ass_directwrite_info_template.h"
+#undef FONT_TYPE
+#undef FAMILY_AS_ARG
```
rcombs, line 599: should we just call `CreateFontFace` and use the `IDWriteFontFace` path instead of awkwardly operating on the `IDWriteFont` with a slightly different interface? Or am I misunderstanding how the API works?

astiob: `IDWriteFontFace` doesn’t provide any access to font names; only `IDWriteFontFace3` does, and it’s not available before Windows 10.

On the other hand, I forgot over time/didn’t know that it’s this easy to convert an `IDWriteFont` to an `IDWriteFontFace`; thanks for pointing it out. I suspect the original reason this wasn’t done was for performance, back when we loaded all fonts eagerly: I’m guessing the face initialization may read in the whole font file and take a lot of time, as opposed to a metadata cache–backed font object. Now that we load fonts lazily, I suppose this would no longer be a big concern if this was the only blocker.

And `FAMILY_AS_ARG` is a micro-optimization, which could, if desired, be simplified by including the argument either unconditionally (and passing `NULL`) or never (and calling `GetFontFamily` even when we already have a family handle: it’s used only in a presumably low-probability fallback branch, anyway).

PR #700, opened 2023-07-25
ci/gha: detect leaks on MacOS

TheOneric:
Depends on https://github.com/libass/libass-tests/pull/3

It fails the CI run with the `leaks` check as expected. However, the redirecting to silence `fuzz`’s *(originally `profile`’s)* detailed output also removes `leaks` report, so there are no immediately visible details on what went wrong and for some reason the fialing CI job gets stuck in the following cleanup steps. See e.g.: https://github.com/TheOneric/libass/actions/runs/5661487504/job/15339487365 or the preceding runs.  
**EDIT:** so the output of the `test` step isn't even available anymore thanks to getting stuck, but basically it just told you the crash tests did indeed crash (only for the run with `leaks`) but without any details on why, then `make check` exited and our manually defined steps were done with and it went to the automatically inserted cleanup stages where it gets stuck.

@rcombs, any ideas what `leaks` might do to cause the job to getstuck or how to improve reporting without also getting all the details from `fuzz`?

### DIFF
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -9,7 +9,7 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     name: build(${{ matrix.msystem || matrix.docker_image || matrix.os }},
-      ${{ matrix.cc }}${{ matrix.api && ', ' }}${{ matrix.api }})
+      ${{ matrix.cc }}${{ matrix.api && ', ' }}${{ matrix.api }}${{ matrix.jobname_suffix && ', ' }}${{ matrix.jobname_suffix }})
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


TheOneric:
Merged the first commit (and will merge the ART PR depending on it). Before we can actually merge the second commit and enable leak checks on MAcOS someone (preferably someone who can test MacOS locally) will need to look into what makes GHA timeout on cleanup when `leaks` is used.

TheOneric:
Apparently this fixed itself over a year later ^^

Note, while i don’t know what the `Idle exit: Untracked` output of `leaks` means, a simple web search shows the same line appearing in actual leaky reports so it doesn’t appear to refer to `leaks` not tracking memory

Issue #600, opened 2022-03-01
libass mangles positioning (and some rotation) of Japanese punctuation when rendering subtitles vertically


y3kcjd5:
### Screenshots

Example of (mostly) correct rendering by a media player using DirectVobSub:
![verticalsoftsubsgood](https://user-images.githubusercontent.com/100724690/156239287-8eec647a-d02c-43f6-8e5b-80b7e1387515.png)
Example of libass via ffmpeg "ass" filter hardcoded to video (also consistent with mpv):
![asscode](https://user-images.githubusercontent.com/100724690/156239541-21fda112-8be0-4e3e-ab0e-3b2eb8bee6e9.png)
Note: the above screenshots were trimmed on the lower and left sides only.

### Description of the issue

Discovered this issue while investigating a tangentially related problem in ffmpeg. Both DirectVobSub and multiple versions of Aegisub position, kern, and rotate the characters correctly for vertical display, while the latest libass version is literally all over the place.

### libass version

0.15.2-62-gba6bcb3a9c2f06272ca1ff1a65f52dc5bc4528b0 (as reported by ffmpeg)

### ASS Sample
```
[Script Info]
; Script generated by Aegisub 3.2.2
; http://www.aegisub.org/
Title: sub
ScriptType: v4.00+
WrapStyle: 0
PlayResX: 1920
PlayResY: 1080
ScaledBorderAndShadow: yes
Video Aspect Ratio: 0
Video Zoom: 4
YCbCr Matrix: TV.601

[Aegisub Project Garbage]
Last Style Storage: Default
Video File: 5-sec-blank.mp4
Video AR Mode: 4
Video AR Value: 1.777778
Video Zoom Percent: 0.500000
Video Position: 81

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: ROTEST,@MS UI Gothic,72,&H28000000,&H00000000,&H00FFFFFF,&HC8000000,0,0,0,0,100,100,0,-90,1,3,1,7,1910,20,20,1

[Events]
Format: Layer, Start, End, Style, Name, MarginL, MarginR, MarginV, Effect, Text
[...]

PR #702, opened 2023-08-06
GDI-compatible rotation handling

### LINKED ISSUE #701: Rendering: wrong direction of lines
### Screenshots

libass ------  vsfilter
 

### Description of the issue

It's re-report from https://github.com/Masaiki/xy-VSFilter/issues/18
Check the screenshots.

### libass version

master (mpv)

### Is it a regression?

_No response_

### ASS Sample

[test.zip](https://github.com/libass/libass/files/12265909/test.zip)
seek to 22:03

### Special Fonts

### System Information

OS: win10 22H2

### Log

_No response_

### Additional info

_No response_

rcombs:
I believe this should fix #701, as well as https://github.com/waddou/libass/issues/95 more broadly (or at least improve the situation there).

Caveats:
- As far as I can tell, GDI only applies VERT, not VKNA. I'm leaving this be for now until we can test to see what the impact of changing would be.
- GDI only applies VERT on characters that it deems to be fullwidth; we're applying it to all runs in @fonts. This is _probably_ fine, since it should be quite rare to have a VERT substitution for a halfwidth character as detected by this code.
- GDI prefers to use the `mort` table for vertical substitutions over `gsub`; it looks like harfbuzz has some support for that, but I'm not entirely clear on whether it would apply to our current usage. This seems to only matter for a few very old fonts.
- GDI makes use of the system code page in a couple places, which we don't want to do here; I attempt to guess the font's encoding instead.
- This definitely is going to need a decent amount of testing and comparison with vsfilter behavior.

### DIFF
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -76,7 +76,6 @@ struct ass_shaper {
 struct ass_shaper_metrics_data {
     Cache *metrics_cache;
     GlyphMetricsHashKey hash_key;
-    int vertical;
 };
 
 struct ass_shaper_font_data {
@@ -228,15 +227,8 @@ static FT_Glyph_Metrics *
 get_cached_metrics(struct ass_shaper_metrics_data *metrics,
                    hb_codepoint_t unicode, hb_codepoint_t glyph)
 {
-    bool rotate = false;
-    // if @font rendering is enabled and the glyph should be rotated,
-    // make cached_h_advance pick up the right advance later
-    if (metrics->vertical && unicode >= VERTICAL_LOWER_BOUND)
-        rotate = true;
-
     metrics->hash_key.glyph_index = glyph;
-    FT_Glyph_Metrics *val = ass_cache_get(metrics->metrics_cache, &metrics->hash_key,
-                                          rotate ? metrics : NULL);
+    FT_Glyph_Metrics *val = ass_cache_get(metrics->metrics_cache, &metrics->hash_key, NULL);
     if (!val)
         return NULL;
     if (val->width >= 0)
@@ -261,7 +253,9 @@ size_t ass_glyph_metrics_construct(void *key, void *value, void *priv)
 
     memcpy(v, &face->glyph->metrics, sizeof(FT_Glyph_Metrics));
 
-    if (priv)  // rotate
+    // if @font rendering is enabled and the glyph should be rotated,
+    // make cached_h_advance pick up the right advance later
+    if (k->vertical)
         v->horiAdvance = v->vertAdvance;
 
     return 1;
@@ -478,7 +472,6 @@ static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
         if (!metrics)
             return NULL;
         metrics->metrics_cache = shaper->metrics_cache;
-        metrics->vertical = info->font->desc.vertical;
 
         hb_font_funcs_t *funcs = hb_font_funcs_create();
         if (!funcs)
@@ -513,6 +506,7 @@ static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
     // update hash key for cached metrics
     struct ass_shaper_metrics_data *metrics =
         font->shaper_priv->metrics_data[info->face_index];
+    metrics->hash_key.vertical = !!(info->flags & DECO_ROTATE);
     metrics->hash_key.font = info->font;
     metrics->hash_key.face_index = info->face_index;
     metrics->hash_key.size = info->font_size;
@@ -894,6 +888,10 @@ void ass_shaper_find_runs(ASS_Shaper *shaper, ASS_Renderer *render_priv,
             // get font face and glyph index
             ass_font_get_index(render_priv->fontselect, info->font,
                     info->symbol, &info->face_index, &info->glyph_index);
+
+            if (info->font->desc.vertical &&
+                ass_codepoint_is_fullwidth(info->font->faces_cp[info->face_index], info->symbol))
+                info->flags |= DECO_ROTATE;
         }
         if (i > 0) {
             GlyphInfo *last = glyphs + i - 1;
--- a/libass/ass_cache_template.h
+++ b/libass/ass_cache_template.h
@@ -72,6 +72,7 @@ START(glyph_metrics, glyph_metrics_hash_key)
     GENERIC(double, size)
     GENERIC(int, face_index)
     GENERIC(int, glyph_index)
+    GENERIC(int, vertical)  // @font vertical layout
 END(GlyphMetricsHashKey)
 
 // describes an outline glyph
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -22,6 +22,7 @@
 #include <inttypes.h>
 #include <ft2build.h>
 #include FT_FREETYPE_H
+#include FT_SFNT_NAMES_H
 #include FT_SYNTHESIS_H
 #include FT_GLYPH_H
 #include FT_TRUETYPE_TABLES_H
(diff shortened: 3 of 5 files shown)


astiob:
Haven’t had a look at the code yet, but just a thought that popped in my mind after reading the description: GDI’s behaviour may also differ in runs where it detects that complex shaping is necessary and delegates to Uniscribe.

libass/ass_shaper.c
```diff
@@ -231,7 +231,7 @@ get_cached_metrics(struct ass_shaper_metrics_data *metrics,
     bool rotate = false;
     // if @font rendering is enabled and the glyph should be rotated,
     // make cached_h_advance pick up the right advance later
-    if (metrics->vertical && unicode >= VERTICAL_LOWER_BOUND)
+    if (metrics->vertical && ass_codepoint_is_fullwidth(metrics->hash_key.font->faces_cp[metrics->hash_key.face_index], unicode))
```
MrSmile, line 234: I fear this can be really slow. Maybe it's better to move this check into metrics cache?

rcombs:
Rebased with changes to match the behavior of WinXP instead of NT4; this mainly involves adding more Unicode ranges that are treated as wide. This now fully resolves all of the rotation mismatches in #600, though it still does nothing to address the positioning/metrics errors.

libass/ass_shaper.c
rcombs: Sure; done.

khaledhosny:
> GDI prefers to use the `mort` table for vertical substitutions over `gsub`; 

Are you sure about that? I’m surprised to learn that GDI has any support for `mort` table, or is this strictly for vertical rotation only?

rcombs:
It's exclusively for rotation, yeah.

rcombs:
…this no longer fixes #701, and now I'm not entirely sure how that works in GDI on XP+ at all (as far as I can tell, it treats — as fullwidth, and thus should rotate it?). It improves some metrics stuff from #600 further, though!

I do wonder if there are behavior variances in versions of windows that are newer than what I've analyzed, though.

astiob:
> as far as I can tell, it treats — as fullwidth, and thus should rotate it?

Er, I’m not sure I’m following you correctly, but — in vertical writing is supposed to be vertical. So, to counteract the expected 90° rotation of the whole text, GDI displays it horizontally in an @-font run. It’s the same thing that happens to other punctuation where it shifts position: the glyph itself is substituted. The font used in #701 is sans-serif, so you don’t see it, but in a serif font, the @-font’s — would actually have different serifs from the normal font’s, but both glyphs would be horizontal.

astiob:
Oh, my bad, the symbol in #701 is actually the em dash! I was the one thinking it was the chouon this time.

Still, probably substitutions? The font in question has this glyph in its `vert` table.

rcombs:
Okay, I think we're potentially running up against a harfbuzz limitation. The file in #701 has lines entirely composed of em dashes, which have HB_SCRIPT_COMMON. The font's gsub contains this:
```xml
    <ScriptList>
      <!-- ScriptCount=1 -->
      <ScriptRecord index="0">
        <ScriptTag value="hani"/>
        <Script>
          <DefaultLangSys>
            <ReqFeatureIndex value="65535"/>
            <!-- FeatureCount=1 -->
            <FeatureIndex index="0" value="0"/>
          </DefaultLangSys>
          <!-- LangSysCount=1 -->
          <LangSysRecord index="0">
            <LangSysTag value="CHN "/>
            <LangSys>
              <ReqFeatureIndex value="65535"/>
              <!-- FeatureCount=1 -->
              <FeatureIndex index="0" value="0"/>
            </LangSys>
          </LangSysRecord>
        </Script>
      </ScriptRecord>
    </ScriptList>
    <FeatureList>
      <!-- FeatureCount=1 -->
      <FeatureRecord index="0">
        <FeatureTag value="vert"/>
        <Feature>
          <!-- LookupCount=1 -->
          <LookupListIndex index="0" value="0"/>
        </Feature>
      </FeatureRecord>
    </FeatureList>
```

GDI sees this in an `@font` case, goes "yup, that's a `vert` in FeatureList", and applies it; the `ScriptList` is completely ignored. Harfbuzz, however, will only apply the `vert` when the buffer script is set to `hani` (which it isn't here, since the line is entirely composed of em dashes).

Telling harfbuzz that our text direction is TTB instead of LTR forces `vert` to always be on regardless of script, but also seems to break a lot of other stuff in ways that I don't see an obvious way to resolve. Maybe we _should_ be doing that, but maybe harfbuzz should provide a way to set this at the API level? Anyone have any thoughts?

astiob:
> Telling harfbuzz that our text direction is TTB […] seems to break a lot of other stuff in ways that I don't see an obvious way to resolve.

For reference, could you list some examples? HarfBuzz’s native TTB mode was what I thought was the way forward before this PR.

rcombs:
Harfbuzz's TTB mode may well be the answer here, I'm just definitely using it wrong. Here's my current WIP patch: https://gist.github.com/rcombs/69e9e7ec9cea11859855ef255dfe5fb2

This is definitely doing some stuff wrong; it falls apart badly on parts of the sample from #701: 
dmg

Here's output with this trivial test patch to use `HB_SCRIPT_HAN`:
```patch
diff --git a/libass/ass_shaper.c b/libass/ass_shaper.c
index f4e7367..74e8e31 100644
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -727,7 +727,7 @@ static bool shape_harfbuzz(ASS_Shaper *shaper, GlyphInfo *glyphs, size_t len)
 
         props.direction = FRIBIDI_LEVEL_IS_RTL(level) ?
             HB_DIRECTION_RTL : HB_DIRECTION_LTR;
-        props.script = glyphs[offset].script;
+        props.script = HB_SCRIPT_HAN;
         props.language  = hb_shaper_get_run_language(shaper, props.script);
         hb_buffer_set_segment_properties(buf, &props);
 

```
dmg

Notably, the em-dash lines show as contiguous lines instead of dashed as in vsfilter; not sure what's going on there. Some of the text might also be rendering wrong since only the font used for the em-dashes was provided. I also had to change the font name to `FZBaoSong-Z04`, as the Chinese name (`方正报宋_GBK`) doesn't match in CoreText's search (which is an issue unto itself that we should look into).

astiob:
The picture looks like you may just need to move the negation (within the macro) to a different place.

rcombs:
Well, here's some new and interesting screwiness: https://gist.github.com/rcombs/c428c575a1f82f6055e3c1e9e8f9398e

dmg

I'm mostly just throwing spaghetti at the wall here, and I'm not sure what to do about these offsets (maybe something about the bearing calculation?). Someone who understands this better than I do should probably take a stab at this.

MIRIMIRIM:
As tested, just as the documentation states, VSFilter / GDI checks for the ‘vrt2’ when processing OpenType fonts with CFF outlines for vertical writing. if it is not present, VSFilter will fall back to default font, but libass not check vrt2.

astiob:
Could you please post a link to the documentation you’re thinking of?

MIRIMIRIM:
https://learn.microsoft.com/zh-cn/typography/opentype/spec/features_uz#tag-vrt2
> Note that some implementations impose the following requirements for an OpenType font with CFF outlines to be used for vertical writing: the 'vrt2' feature must be present in the GSUB table, it must comprise a single type 1 (single substitution) lookup with lookupFlag set to 0, and the lookup must have a single subtable. The predecessor feature, ['vert'](https://learn.microsoft.com/zh-cn/typography/opentype/spec/features_uz#vert), is ignored.

https://learn.microsoft.com/zh-cn/typography/opentype/spec/recom#opentype-cjk-font-guidelines
> If an OpenType font with CFF outlines is to be used for vertical layout, some legacy platforms may require that a [Vertical Rotation ('vrt2') feature](https://learn.microsoft.com/zh-cn/typography/opentype/spec/features_uz#vrt2) be present in the Glyph Substitution (GSUB) table. See the Feature Tags section of the OpenType Layout Tag Registry for a description of and further requirements for this feature.

PR #705, opened 2023-08-27
Added meson build system

jambonmcyeah:


TheOneric (changes requested):
I’m not sure if you noticed, but there already is an on-going patch series for a secondary meson buildsystem in #330.

On a glance it seems the version here has some issues not found in #330:
 - has some unnecessary bits (e.g. setting `CONFIG` macros for *required* dependencies)
 - `--disable-xxx` appear broken as deps are always looked up and added to the deps of the libass library. The only distinction is in the `CONFIG` macros.
 - it pretends to support all options from autoconf, but some just aren't wired up to any logic at all like e.g. `--enable-fuzz`
 - reintroduces problems fixed and discussed in #330, like the duplication of symbol lists. No one wants to maintain the same list multiple times.
 
 On the upside, this version expands where symbol visibility control works in abscence of proper native Meson support by manually wiring up version scripts and clang export maps. That is quite nice for potential users!¹
 
 Overall it seems to me like #330 is closer to something useable *(but it too misses some stuff and has unaddressed review comments)*.  If you want to pursue this further I suggest you read through the unresolved review comments in #330, get in touch with @CoffeeFlux and start helping with #330 *(If that doesn't work out create a new PR based on #330, but with your additions and pending review comments from #330 addressed.)*
 
 ---

<details>
<summary>Footnote [1]</summary>
Ultimately though, this remains a temporary workaround and is something which should be addressed in meson itself. It is a basic feature for building C-ABI libs and it is asinine for every project to individually start checking and wiring up the necessary bits for all possible compiler+linker combinations. As a case in point, even this version misses out on e.g. <a href="https://www.illumos.org/man/1/ld)">illumos’ linker</a> (i think `-B -M <mapfile>` or similar should be able to do this). With meson only becoming a potential secondary buildsystem, I’m also not sure if we would want to take on the effort and maintenance burden resulting from manually getting symbol export control working everywhere.
</details>

PR #711, opened 2023-10-08
render_api: limit resolutions to be representable in a single ASS_Image

TheOneric:
The more general version of 358770835248e26241e4796f38f02073a3bae252.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -2812,7 +2812,7 @@ static void add_background(RenderContext *state, EventImages *event_images)
     bottom      = FFMINMAX(bottom, 0, render_priv->height);
     int w = right - left;
     int h = bottom - top;
-    if (w < 1 || h < 1 || w > FFMIN(INT_MAX, SIZE_MAX) / h)
+    if (w < 1 || h < 1)
         return;
     void *nbuffer = ass_aligned_alloc(1, w * h, false);
     if (!nbuffer)
--- a/libass/ass_render_api.c
+++ b/libass/ass_render_api.c
@@ -20,7 +20,11 @@
 #include "config.h"
 #include "ass_compat.h"
 
+#include <limits.h>
+#include <stdint.h>
+
 #include "ass_render.h"
+#include "ass_utils.h"
 
 static void ass_reconfigure(ASS_Renderer *priv)
 {
@@ -51,7 +55,7 @@ static void ass_reconfigure(ASS_Renderer *priv)
 
 void ass_set_frame_size(ASS_Renderer *priv, int w, int h)
 {
-    if (w < 0 || h < 0)
+    if (w <= 0 || h <= 0 || w > FFMIN(INT_MAX, SIZE_MAX) / h)
         w = h = 0;
     if (priv->settings.frame_width != w || priv->settings.frame_height != h) {
         priv->settings.frame_width = w;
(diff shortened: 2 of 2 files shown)


astiob:
Is it actually safe to remove that check?

Overall seems reasonable. Do we need to mention this in release notes?

TheOneric:
> Is it actually safe to remove that check?

I believe so, but please recheck: 

Copying the lines before the current check a bit beyond what the GitHub diff preview shows:
```C
    ASS_Renderer *render_priv = state->renderer;
    int size_x = state->shadow_x > 0 ?
        lround(state->shadow_x * state->border_scale_x) : 0;
    int size_y = state->shadow_y > 0 ?
        lround(state->shadow_y * state->border_scale_y) : 0;
    int left    = event_images->left - size_x;
    int top     = event_images->top  - size_y;
    int right   = event_images->left + event_images->width  + size_x;
    int bottom  = event_images->top  + event_images->height + size_y;
    left        = FFMINMAX(left,   0, render_priv->width);
    top         = FFMINMAX(top,    0, render_priv->height);
    right       = FFMINMAX(right,  0, render_priv->width);
    bottom      = FFMINMAX(bottom, 0, render_priv->height);
    int w = right - left;
    int h = bottom - top;
    if (w < 1 || h < 1 || w > FFMIN(INT_MAX, SIZE_MAX) / h)
```

Through `FFMINMAX` `w` and `h` are limited to `0 <= w <= render_priv->width` and analogous for `h`.
Since we now limit (afaict) `ASS_Renderer->width` and `-> height` such that their product is `width <= FFMIN(INT_MAX, SIZE_MAX) / height` the same will be automatically true for `w` and `h` here.

The limitations to `ASS_Renderer->width` and `-> height` are inherited from `ASS_Render->settings.width`/`height`.
Afaict the only place setting `ASS_Renderer->width` is `ass_reconfigure` in `ass_render_api.c` and it just copies `settings.width`. This also matches the documentation comment for `ASS-Renderer->width, height` which states `// screen dimensions (the whole frame from ass_set_frame_size)`

Confusingly, there is also `ASS_RenderPriv` which too has `width` and `height` members, but is different from `ASS_Renderer` whose associated variables we often name `render_priv` or just `priv`. For `ASS_RenderPriv` no limit is enforced *(but it just copies its values from `ASS_Image`s)*.  
A search with `grep -nR --include='*.c' --include='*.h' -E '(priv|render(er)?)(->|\.)height ='` only shows the one `ASS_Render->height` assignment and a couple ones for `ASS_RenderPriv`s.  


> Overall seems reasonable. Do we need to mention this in release notes?

I don't expect anyone to be affected by this new limit, but I guess it can’t hurt to mention it. And if someone is affected after all, this should draw their attention.

astiob:
> Through `FFMINMAX` `w` and `h` are limited to `0 <= w <= render_priv->width` and analogous for `h`.

Ah, right. Yeah, seems safe then.

TheOneric:
Or to fully spell out what’s happening: `left`, `right`, etc are limited to `0 <= left, right <= render_priv->width` via `FFMINMAX`. Thus `w` as their difference is limited to `-render_priv->width <= w <=render_priv->width` and combined with the preëxisting and kept check for `< 1`, the final limits are `1 <= w <= render_priv->width`.

PR #716, opened 2023-11-26
ass: explicitly read 32bit ints on timestring parts

TheOneric:
This matches VSFilter and avoids signed overflows for ABIs with 64bit `int`s, like ILP64.

In practice the latter is probably only a theoretical concern atm and this change more cosmetic.
Still, it amkes the intent clearer and matches how we’re moving more and more `int` to explicit `int32_t`.

My only concern comes from a vague memory, that `inttype.h` print/scan macros may sometimes under certain conditions *hurt* portability with some Windows compilation settings or so. @astiob, iinm this was something you brought up/discovered once, do you recall more and if this is applicable here (or am I just confused and mixing things up)?  
Though we already use `PRId64` further down in `ass.c`.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -228,9 +228,9 @@ static void set_default_style(ASS_Style *style)
 
 static long long string2timecode(ASS_Library *library, char *p)
 {
-    int h, m, s, ms;
+    int32_t h, m, s, ms;
     long long tm;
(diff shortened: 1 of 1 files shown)


astiob:
32-bit ones shouldn’t be able to hurt, as Windows codes for them are the same as everywhere else (just `"d"`).

PR #717, opened 2023-11-29
ass: don't fail to read files larger than INT_MAX

TheOneric:
Yet another `int` overuse in old code. Without this UB will occur on conversion to signed int and if the value wraps around to something negative, claim there was an IO error using whatever the last errno number was which is misleading.

Not sure if it’s best practice to check `ferror`, `feof` or both.


TheOneric:
> Not sure if it’s best practice to check ferror, feof or both.

In the previous revision checking just one should have been fine afaict.
However the patch omitted to also converted a successfully `sz` to `size_t` and there were two more preëxisting problems.
  - due to `ftell` returning a `long` file size was effectively still limited to 2GiB on Win32 LLP64
  - `ftell`+`fseek` is not guaranteed to work in ISO C (see message of second commit)

 A newly second commit now replaces the `fseek`+`ftell` approach

rcombs:
Why not just use `ftello` or `_ftelli64` when available? I believe that should cover all relevant platforms.

astiob:
FWIW I think it’s good to support files that are non-seekable; and even a seekable file’s size can change mid-read if someone concurrently writes to it. We could, of course, use ftello etc. as an optimization hint if it’s available.

On a separate note, why do we need to read the entire file into a buffer at all? Can’t we process it chunk by chunk? We probably need whole lines, but we should be able to read a chunk, process any whole lines, and buffer only the last, incomplete line as we proceed to read the next chunk.

astiob:
(Another non-ISO-C alternative for getting the file size is the `stat` family of functions.)

TheOneric:
> Why not just use ftello or _ftelli64 when available? I believe that should cover all relevant platforms.

I didn’t know Win32 has a `ftello` equivalent. Some care would need to be taken to correctly deal with the different types, possible `_FILE_OFFSET_BITS` values and `SIZE_MAX` but seems doable. However, apart from the easily avoidable `fonts_dir`, this would afaik be the first instance of us requiring a platform-specific API without an ISO fallback. Also, does POSIX and Win32 actually guarantee `fseek(..., SEEK_END)` to be working and meaningful for binary streams?

ISO C, here draft n3047 of C23 says:
> **§ 7.23.9.2** - Paragraph 3
>
> [...] A binary stream need not meaningfully support `fseek` calls with a *whence* value of `SEEK_END`.

[POSIX docs](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fseek.html) are silent about this particular case, but as usual defer to ISO C for everything not explicitly marked as an extension and [Win32 docs](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fseek-fseeki64?view=msvc-170) are also silent on the matter.

In practice it seems indeed unlikely anyone is using `ass_load_file` on a system without proper support for `SEEK_END` judging from how nobody complained about it so far. Non-POSIX and non-Win32 might perhaps also be rare.

There is a nice bonus to the read-loop though, apart from avoiding possibly purely theoretical problems: it works with streams which are never seekable and we can now do something like
```sh
zstd -c -d ./large.ass.zst | test/test out.png /dev/stdin 12
```

Side note: while I think attempts to seek on streams should fail cleanly with POSIX or ISO C, concerningly Microsoft docs contain the sentence *“On devices incapable of seeking, the return value is undefined.”*.

TheOneric:
> On a separate note, why do we need to read the entire file into a buffer at all? Can’t we process it chunk by chunk? We probably need whole lines, but we should be able to read a chunk, process any whole lines, and buffer only the last, incomplete line as we proceed to read the next chunk.

We could parse ASS files line by line, yeah. But `ass_load_file` is also used to open and load font files from some system font providers.

MrSmile:
I think even in the case of incremental read it's better to use some sensible approximation for initial buffer size (fstat?).
Actually, `ass_add_font` does malloc + memcpy again, so in the future it'd be good to eliminate that too.

TheOneric:
Now using `fstat` if available to allocate just the right size if it can be determine beforehand.

libass/ass.c
```diff
@@ -1455,18 +1455,21 @@ char *ass_load_file(ASS_Library *library, const char *fname, FileNameSource hint
         return NULL;
     }
 
-    const size_t CHUNK_SIZE = 16 * 1024;
-    size_t bufmax = 128 * CHUNK_SIZE;
+    size_t bufmax = 2 * 1024 * 1024;
+    ass_get_file_size(fp, &bufmax);
+    if (bufmax > SIZE_MAX - 1)
+        return NULL;
```
astiob, line 1461: This probably needs to be `goto fail`, because it leaks `fp`.

libass/ass_filesystem.c
```diff
@@ -50,6 +50,43 @@ static bool alloc_path(ASS_Dir *dir, size_t size)
 }
 
 
+#if HAVE_FSTAT || _WIN32
+// Microsoft docs require types.h to be included before stat.h
+// IN POSIX just stat.h suffices
+#if _WIN32
+#include <sys/types.h>
+#endif
+#include <sys/stat.h>
```
astiob, line 59: It is true that POSIX seems to require `sys/stat.h` alone to define everything necessary, but my Linux man page does list `sys/types.h` among the necessary includes and even adds `unistd.h`. I assume that there must be some older/less POSIX-compliant systems (besides Windows) out there that require `sys/types.h` before `sys/stat.h`. Why not just include both unconditionally?

libass/ass_filesystem.c
```diff
@@ -50,6 +50,43 @@ static bool alloc_path(ASS_Dir *dir, size_t size)
 }
 
 
+#if HAVE_FSTAT || _WIN32
+// Microsoft docs require types.h to be included before stat.h
+// IN POSIX just stat.h suffices
+#if _WIN32
```
astiob, line 56: Minor/optional, but I’d prefer `#ifdef` and `#if ... defined(_WIN32)`. This is what we use elsewhere and seems intuitively safer, although everything I can find does say it’s specifically defined with a value of `1` rather than an empty one.

libass/ass.c
TheOneric: yep, thx! Will also need to pre-initialise `buf` with `NULL` to avoid freeing an unitialised pointer

libass/ass_filesystem.c
TheOneric: seems to have been a (non)change in SUSv3/POSIX issue 6 (2004); from the change history:
> The requirement to include [<sys/types.h>](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/types.h.html) has been removed. Although [<sys/types.h>](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/types.h.html) was required for conforming implementations of previous POSIX specifications, it was not required for UNIX applications.

Now including both uncoditionally

libass/ass_filesystem.c
TheOneric: done

libass/ass.c
astiob, line 1461: You may want to short-circuit if `bufmax` is zero. Failing is probably fine, as there’s nothing useful in the file anyway.

astiob (commented):
We probably want to define `_FILE_OFFSET_BITS=64` as well, and whatever equivalent incantations other platforms may need. There’s probably (hopefully) a ready-to-use Autoconf macro for this.

libass/ass.c
TheOneric: good idea, done

TheOneric:
> We probably want to define _FILE_OFFSET_BITS=64 as well, and whatever equivalent incantations 

Now done via `AC_SYS_LARGEFILE`

astiob (approved):

libass/ass.c
```diff
@@ -1455,18 +1455,22 @@ char *ass_load_file(ASS_Library *library, const char *fname, FileNameSource hint
         return NULL;
     }
 
-    const size_t CHUNK_SIZE = 16 * 1024;
-    size_t bufmax = 128 * CHUNK_SIZE;
-    char *buf = malloc(bufmax + 1); // one extra for terminating zero
+    char *buf = NULL;
+    size_t bufmax = 2 * 1024 * 1024;
+    ass_get_file_size(fp, &bufmax);
+    if (bufmax > SIZE_MAX - 1 || bufmax == 0)
```
MrSmile, line 1461: Small `bufmax` can lead to suboptimal realloc scaling (i.e. 1, 2, 4,...), so I suggest to do `bufmax += CHUNK_SIZE` before allocation.

libass/ass.c
```diff
@@ -1455,18 +1455,22 @@ char *ass_load_file(ASS_Library *library, const char *fname, FileNameSource hint
...
+        goto fail;
+    buf = malloc(bufmax + 1); // one extra for terminating zero
     if (!buf)
         goto fail;
 
     size_t bytes_read = 0;
     while (true) {
-        size_t read = fread(buf + bytes_read, 1, CHUNK_SIZE, fp);
+        size_t maxread = bufmax - bytes_read;
```
MrSmile, line 1469: I suggest to ensure that `maxread` is always greater than `CHUNK_SIZE` to avoid suboptimal behavior at the file end.

libass/ass.c
```diff
@@ -1476,7 +1480,7 @@ char *ass_load_file(ASS_Library *library, const char *fname, FileNameSource hint
             }
         }
 
-        if (bufmax < bytes_read + CHUNK_SIZE) {
+        if (bufmax <= bytes_read) {
```
MrSmile, line 1483: This is always triggered for regular files: next to last read would be `bytes_read == bufmax` which triggers realloc, and then the final read return 0 and sets EOF. I think it's better to keen previous variant for this condition.

libass/ass_filesystem.c
```diff
@@ -50,6 +50,41 @@ static bool alloc_path(ASS_Dir *dir, size_t size)
...
+#endif
+
+bool ass_get_file_size(const FILE *fp, size_t *size)
+{
+    struct stat info;
+    int fd = fileno((FILE *) fp);
+    if (fstat(fd, &info))
+        return false;
+
+    if (!S_ISREG(info.st_mode) || info.st_size > SIZE_MAX)
```
MrSmile, line 75: For > 4G files at 32-bit system that triggers failure, so `ass_load_file` fallbacks to incremental read which eventually fail due to out of memory. I think `ass_get_file_size` should return `uint64_t` (or `int64_t` which can be combined with &minus;1 error) and let caller handle overflow.

libass/ass.c
MrSmile: I'm suggesting to read not only one additional byte after EOF, but `CHUNK_SIZE` additional bytes. It's to guarantee that any `fread` tries to read at least `CHUNK_SIZE` bytes for some sensible `CHUNK_SIZE`. Without that, there can be some miniscule reads immediately before reallocs in case of incremental processing. So yes, I'm suggesting to keep `CHUNK_SIZE` and old realloc condition (but new `fread` size is better, given that `maxread >= CHUNK_SIZE`).

libass/ass_filesystem.c
MrSmile: I think a function with the name `ass_get_file_size` should return a known file size even if that size is larger than `SIZE_MAX`. Currently we don't have a use for such sizes, but that can change in the future (for example, for a detailed error message).

TheOneric: While we are presumably unlikely to encounter files larger than >2^64 bytes soon, POSIX’s `stat` struct records the size in `off_t` (Windows’s `fstati64` as a 64bit type) and `off_t` might be *any* width, so the return type would need to be `intmax_t`.
However we only load two types of files:
 - font files, which we’ll need to fully load into memory and thus limit to `SIZE_MAX` anyway
  - ASS styles or full files for which we want to switch to a streaming approach so `get_file_size` will be irrelevant *(i was intending to do that some later date, but if optimising for this case becomes a problem perhaps I should already do that now)*

I’d be surprised if we ever have to deal with more than those types and therefore don't see where `get_file_size > SIZE_MAX` would ever be useful. Do you have some future usecase in mind?

MrSmile: I feel it's incorrect to guess intended use of the result inside a function, it should be caller responsibility. Especially if there's already `bufmax > SIZE_MAX - 2` check outside of it. As for > 2^64 (or 2^63) file sizes, we can safely treat it as error for any foreseeable future. And the simplest use of large result is error reporting: "Cannot load file '...' as it's too large (5.15GB)".

astiob: Isn’t that a contradiction? Treating sizes above 2^64 as an error is exactly what Oneric is doing here, and (from what I understand) it’s impossible to move that handling into the caller except by changing the return type either to `intmax_t` or to a platform-dependent type (or typedeffing `off_t` on Windows).

If the issue is just the fact that the size is being returned in a parameter rather than the return value, how about returning `size_t`, using `SIZE_MAX` as the error marker, and treating files of size exactly `SIZE_MAX` as an error case as well?

astiob: Or you could also use 0 as the error marker.

astiob: Ah, I forgot we need two error modes. So use one of those for one mode and the other for the other.

TheOneric: We want:
 - ideally quit early on empty files
 - definitely quit early on files `>= SIZE_MAX-1` (offset by 1 because as long as this code also handles ASS files we need an additional terminating zero)
 - dynamically increase the buffer for unknown size
 - use an exactly fitting buffer for known size
 
 With `SIZE_MAX` and `0` as error values we can’t handle all of them iinm, but if we sacrifice the early exit on empty files it could work. Though I must say I’m not a fan of using a in-band value which could technically be valid as an error code, I guess it works here (too bad we don’t have `ssize_t` on Win32).
 
 But if the issue is just with the name not reflecting the `size_t` restriction, how about renaming it to `ass_get_file_buffer_size`?

astiob: > too bad we don’t have `ssize_t` on Win32

That would merely _halve_ the valid value space, as opposed to reserving two individual values.

I’m not sure how it isn’t enough, though: you want to handle empty files and huge files the same. In fact, if the upper limit is `>= SIZE_MAX-1`, then you don’t even sacrifice any valid values at all and you even have an extra value left over!

* `0` normally means an empty file, and you want to discard empty files. Return `0` for huge files, too, and they’ll be discarded, too.
* Values from `1` to `SIZE_MAX-2` inclusive are valid sizes of files whose size is known.
* If the size is unknown, either `SIZE_MAX-1` or `SIZE_MAX` can in theory be used as a sentinel value, and the other remains unused.

TheOneric: > `0` normally means an empty file, and you want to discard empty files. Return `0` for huge files, too, and they’ll be discarded, too.

Then we can't log a warning though for too large files (which MrSmile mentioned above and I’m adding locally right now)

TheOneric: > > too bad we don’t have ssize_t on Win32
>
> That would merely halve the valid value space, as opposed to reserving two individual values.

hmm, I assumed `ssize_t` can hold at least `-SIZE_MAX` to `SIZE_MAX`, but apparently this is not guaranteed and POSIX allows for environments where `ssize_t` is only as wide as `long` (identifed by confstr’s `_CS_POSIX_V7_WIDTH_RESTRICTED_ENVS`).
This makes it less useful then, yeah.

astiob: I mean, there’s not much choice on platforms with 64-bit `size_t`. (Some) common compilers only got 128-bit support relatively recently, if indeed at all.

> Then we can't log a warning though for too large files

Why not log it inside this function?

MrSmile: I don't care about platforms with 64-bit `size_t`, such large files would be unfeasible for many years to come. I care about platforms with 32-bit `size_t`, but > 4G files. In such cases file size is perfectly known, but `ass_get_file_size` returns a error, which I think isn't right, especially if that function would be reused in another context in the future.

TheOneric: > especially if that function would be reused in another context in the future.

At least with what I have in mind, it `ass_get_file_size` will only ever be used for files we fully load into memory (font files and atm ASS files), while future streamed ASS files won't use it at all (instead read fixed size chunks like the second commit of this series to have an almost fixed memory usage (if lines are bigger than CHUNK_SIZE realloc will still be needed))

Do you prefer one of the alternative names or have another name suggestion to reflect the limitation?

astiob: It's worth noting that even if we stream a larger file in the future, we still load all events into memory, so unless the file is full of comment lines or embedded fonts, we'll likely run out of 32-bit memory space anyway.

TheOneric:
Renamed the function to `ass_get_file_memory_size` to better reflect its additional limit to what can be loaded into (virtually) continuos memory and added a warning log for too large files.

If this is now acceptable (or this + using in band `SIZE_MAX`/`0` as error codes instead of split return size and error), great.
But otherwise, perhaps I should just already implement the chunk-based parsing for ASS files which will make it clearer how some of this will (not) be used in the future and so we don't over-optimise something getting replaced soon. I  was originally intending to do that sometime later, but given this PR no longer is the initial simple “just change a few types” anyway, I might as well go all the way.

Issue #435, opened 2020-10-18
Issue with n command in drawings


rcombs:
`{\p1\pos(0,0)\an7}m 100 100 l 100 200 l 200 300 n 50 50 l 300 50`
libass:
![image](https://user-images.githubusercontent.com/819638/96355843-35534100-10ac-11eb-97ad-7ae66e4dbc61.png)
vsfilter:
![image](https://user-images.githubusercontent.com/819638/96355845-3ab08b80-10ac-11eb-9ca9-81ccf79e477e.png)

I'm not sure _how_ vsfilter does anything with this command, looking at the code, but apparently it does what appears to be the same was what we do if you replace the `n` with an `l`????

This was discovered by @The0x539; I'm not aware of it existing in any real scripts, so not _super_ worrisome, at least.

rcombs:
At a bit of investigation, it looks like `n` acts like `l` if it's followed by any non-`n` non-`m` command; if it's followed by another `n`, an `m`, or the end of the drawing, it's ignored. I'm not sure if this was deliberate, but it's how it works.
This is because the `_EvaluateLine` function checks if the current cursor position matches the position of the _previous_ point, and if not, draws a line to the previous point (thus updating the cursor) before continuing. This applies to beziers/splines as well, since they're implemented by calling `_EvaluateLine` in a loop. `n` adds a point to the list, but doesn't move the cursor, so it ends up triggering this.

astiob:
Yep, looks the same way to me.

astiob:
_Except_ when `m` is followed by (a number of) `n` and then followed by a cubic curve. Then the (last) `n` overrides the `m`.

Issue #202, opened 2015-10-21
About the `s` drawing command(B-spline)


weizhenye:
I'm writing [ASS.js](https://github.com/weizhenye/ASS) and try to implement the `s` drawing command. As there is no native API for B-spline in SVG or Canvas, I try to translate `s` into several `l` and `b`to simulate it. I tested some cases and found some differences. Here is a example: `m 0 0 s 150 60 150 150 60 150 120 120 90 90 s 60 90 90 120 60 120`

* xy-VSFilter
![xy-vsfilter](https://cloud.githubusercontent.com/assets/1879224/10641847/965f28f8-784d-11e5-8aa1-d4353ba8ea41.png)

* libass
![libjss](https://cloud.githubusercontent.com/assets/1879224/10641866/a9efa6a4-784d-11e5-8791-bf3d5169af7c.png)

* ASS.js
![assjs](https://cloud.githubusercontent.com/assets/1879224/10641870/b256b40e-784d-11e5-9486-c21ad3630926.png)

I'm not very clear about how to drawing B-spline and don't know which is right. The commands before and after the `s` command may affect the result.

astiob:
The right thing is whatever VSFilter does, always. I’ll fix this in libass; thanks for the report! But note that the `s` command does not seem to be used widely in practice if at all, especially not when followed by something other than (an explicit or implied) `p`, which causes the problem.

The trick with the case when `b` follows is to start the Bézier curve at the coordinate that was actually specified rather than the coordinate transformed to give a B-spline.

Speaking of JavaScript ASS renderers, have you heard of [libjass](https://github.com/Arnavion/libjass)? We had a pull request some time ago to list it in our readme file (which we did). I have no idea how good or different it is, but you might want to give it a look.

weizhenye:
I've got it, thanks.

Another issue: `m 0 0 s 0 90 90 90 90 0 233 p 30 40`. In xy-VSFilter, `233` will be ignored for it only specifies x, then expand to (30,40). libass just add `30 40` to the `s` command and then `40` is ignored.

About libjass, when I began ASS.js project, I didn't know there's already a JavaScript ASS renderer. Later I found it and got some help from it, but libjass still had some issues as I tested, for exmaple it's collision detection is not correct. libjass also has some features that ASS.js not implement yet, like karaoke. So each has its advantages.

PR #726, opened 2024-01-04
Reject drawings not starting with m

### LINKED ISSUE #719: Rendering: Drawing without m isn't rendered by VSFilter
### Screenshots

VSFilter
image

Libass 0.16.0
image

### Description of the issue

VSFilter doesn't render at all the drawing

### libass version

0.16.0

### Is it a regression?

No

### ASS Sample

```
[Script Info]
; Script generated by Aegisub 9643-cibuilds-c61f149a3
; http://www.aegisub.org/
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
YCbCr Matrix: None
PlayResX: 1920
PlayResY: 1080

[V4+ Styles]
Style: Default,Arial,20,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,2,2,2,10,10,10,1

[Events]
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\p1\pos(323,63)}l 394 159.75 1009.75 159.75 1009.75 159.75 1009.75 159.75 1007.5 101.25 1007.5
```

### Special Fonts

### System Information

_No response_

### Log

_No response_

### Additional info

I am not sure, but I think VSFilter doesn't show the drawing due to this [if statement](https://sourceforge.net/p/guliverkli2/code/HEAD/tree/src/subtitles/RTS.cpp#l468)

libass doesn't seems to reproduce this behaviour when it parse the string drawing [here](https://github.com/libass/libass/blob/c047dd2ea16f73abb4f448e6db3637158c1226d0/libass/ass_drawing.c#L46-L122) (i'm not 100% sure)

This may be connected to #629

TheOneric:
This adjusts us to current VSFilter behaviour and is one approach to fix #719 

We know this will affect (at least, but from an extensive search *only*) two original typesets, but they aren’t degraded tooo badly.

Since we didn't find *any* samples broken by accepting such drawings, patching VSFilters instead to accept them was also suggested as an alternative on IRC. If done this would codify our somewhat odd existing behaviour on m-less drawings, but no files would be broken.

If someone wants to pursue the VSFilter patch approach, I have no objections to that and we can close this PR once that’s ready. Personally, I think it’s probably fine to degrade those two samples and adopt VSFilter’s current and imho more intuitive behaviour. Either way though, I think we shouldn’t put this off too long to avoid more (future) broken files being created.

### DIFF
--- a/libass/ass_drawing.c
+++ b/libass/ass_drawing.c
@@ -108,8 +108,21 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
                 tail->next = calloc(1, sizeof(ASS_DrawingToken));
                 tail->next->prev = tail;
                 tail = tail->next;
-            } else
+            } else {
+                /* VSFilter compat:
+                 * In guliverkli(2) VSFilter all drawings
+                 * whose first valid command isn't m are rejected.
+                 * xy-VSF and MPC-HC ISR this was (possibly inadvertenly) later relaxed,
+                 * such that all valid commands but n are ignored if there was no m yet.
+                 */
+                if (type == TOKEN_MOVE_NC) {
+                    return NULL;
+                } else if (type != TOKEN_MOVE) {
+                    p++;
+                    continue;
+                }
                 root = tail = calloc(1, sizeof(ASS_DrawingToken));
+            }
             tail->type = type;
             tail->point = point;
             is_set = 0;
--- a/libass/ass_drawing.c
+++ b/libass/ass_drawing.c
@@ -33,7 +33,7 @@
 /*
  * \brief Check whether a number of items on the list is available
  */
-static bool token_check_values(ASS_DrawingToken *token, int i, int type)
+static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType type)
 {
     for (int j = 0; j < i; j++) {
         if (!token || token->type != type) return false;
@@ -50,7 +50,8 @@ static bool token_check_values(ASS_DrawingToken *token, int i, int type)
 static ASS_DrawingToken *drawing_tokenize(const char *str)
 {
     char *p = (char *) str;
-    int type = -1, is_set = 0;
+    ASS_TokenType type = TOKEN_INVALID;
+    int is_set = 0;
     double val;
     ASS_Vector point = {0, 0};
 
@@ -102,7 +103,7 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
         if (!got_coord)
             is_set = 0;
 
-        if (type != -1 && is_set == 2) {
+        if (type != TOKEN_INVALID && is_set == 2) {
             if (root) {
                 tail->next = calloc(1, sizeof(ASS_DrawingToken));
                 tail->next->prev = tail;
--- a/libass/ass_drawing.h
+++ b/libass/ass_drawing.h
@@ -24,6 +24,7 @@
 #include "ass_bitmap.h"
 
 typedef enum {
+    TOKEN_INVALID,
     TOKEN_MOVE,
     TOKEN_MOVE_NC,
     TOKEN_LINE,


astiob (approved):

astiob:
As a possibly unnecessary optimization, you could move the type check into the branch that first allocates `root`. Then you wouldn’t need to parse and then to free the rest of the bad drawing.

arch1t3cht:
It's worth pointing out that this does not match VSFilter exactly. This will completely drop any drawing that does not start with an `m` command, while VSFilter will drop all commands up to the next `m` and start rendering after that. So in a simple shape like
```
{\p1\an7\pos(137,198)}l 27.5 -66 255.5 -69 m 137.5 163 l 283.5 167 336.5 63
```
current libass would render two triangles, VSFilter would render one triangle, and this PR would render nothing.

Now, whether this is a problem is an entirely different question - with the philosophy of "all drawings not starting with `m` are invalid, i.e. 'undefined behavior'" one could argue that libass doesn't need to match VSFilter here - but I wanted to point it out.

astiob:
Bloody hell. Thanks for pointing this out; we’ve been misreading the code.

> VSFilter will drop all commands up to the next `m` and start rendering after that

It’s not quite that simple, either.

Most commands, including `l`, are ignored if they aren’t preceded by enough (non-ignored) points. So here, it ignores the `l` and then it sees the `m` as the first parsed command and proceeds to accept the drawing.

But the commands that are accepted are not only `m` but also `n`. If a drawing starts with some ignored commands, then a valid `n`, then the move won’t be the first parsed command and the whole drawing will be rejected.

TheOneric:
Ohh, thanks for pointing out xy*Filter does that.

There another twist to it though, guliverkli2 *(whose code is linked in #719 and I’ve been using as a reference this whole time)* doesn’t have any “enough prior points” checks and **will** reject `l 27.5 -66 255.5 -69 m 137.5 163 l 283.5 167 336.5 63` :upside_down_face: 

xy*Filter added those new checks together with floating-point support in [xy-VSFilter@7b1e5c6e3](https://github.com/Cyberbeing/xy-VSFilter/commit/7b1e5c6e3b390ca7f0a9803786f1369878f7ae94#diff-ded791ebefee17cb3d6430517d7bec74ca01b482bbdb49df157c013b2d43421dR1010). While it so far seems no files relying on this slightly more lenient parsing exists, given xy intentionally relaxed this we probably should follow this in libass.

astiob:
Oh, so it got even weirder.

> given xy intentionally relaxed this

I really don’t think that’s what happened. The commit message says “add more checks”, and this is a pretty weird way to implement this, so I think he attempted to make the code safer without affecting behaviour and this is an unintentional side effect he didn’t foresee.

In fact, an analogous change is in MPC-HC, which he probably copied: https://github.com/clsid2/mpc-hc/commit/039daf85424e81631d840576e0e159757f1aba14
It says it was needed to fix a crash on `m l 0 0`: https://web.archive.org/web/20190301203552/https://trac.mpc-hc.org/ticket/578

TheOneric:
This should now almost match xy-VSFilter. However there are some points worth bringing up:

Apart from the bounding box, a leading b-spline also acted differently from a b-spline after a move, since in the former case the first node of `s` would set `spline_start` to `NULL` because there is no preceding element.
*(Unrelated, but why does for `c` `token_check_values` only test the next 2 nodes when it wants to readd three points? Is this emulating some VSF quirk?)*

In xy-VSFilter `p` requires at least 3 preceding coordinates. libass doesn’t parse `p` at all resulting in it just being an extension of any previous command. If the drawing is otherwise valid, `p` follows after a already by itself valid `s` and this is equivalent.  
But it leads to e.g. `m 0 0 s 0 0 20 20 p 30 30` being rejected by xy-VSF but not libass. Depending on how it processes the drawing after initial parsing guliverkli(2) *might* have also accepted this *(or maybe it crashed)*.      
Moreover, as pointed out in #202 `m 0 0 s 0 90 90 90 90 0 233 p 30 40` while accepted by both xy-VSFilter and libass, will result in different drawings due to dropping different numbers. `m 0 0 l 10 0 20 5 10 10 p 0 10` will presumably also differ due to `p` just extending the `l` in libass, but presumably being dropped later by VSFilters.
(And also `s` has some more deviations from VSFilter even in valid drawings, as was #202’s main point ... and while we’re at it, so does `n` #435 and a *trailing* `m` #629.)

I’m not sure if it makes sense to bundle `p` into this fix, given it relates to more than just a missing leading `m`.

But please recheck to make sure there are no more lurking issues I missed :)

TheOneric:
> Moreover, as pointed out in https://github.com/libass/libass/issues/202 m 0 0 s 0 90 90 90 90 0 233 p 30 40 while accepted by both xy-VSFilter and libass, will result in different drawings due to dropping different numbers.

Actually, this part might have already been fixed a while before the comment in #202 was made by 0d94ff48077ba683bfd4d0b9ff7ded895ba20919.

TheOneric:
Some more notes: I’ve been considering adding just a stub `p` parsing to handle exclusively its “too few preceding nodes” rejection, but doing this right in turn also depends on VSFilter rejecting `s` commands with less than three points early [here](https://github.com/pinterf/xy-VSFilter/blob/45917d77083d93215738efb140486aba7614947e/src/subtitles/RTS.cpp#L838-L855) and similarly for `b` bezier curves dropping any coords not making up a triple [here](
https://github.com/pinterf/xy-VSFilter/blob/45917d77083d93215738efb140486aba7614947e/src/subtitles/RTS.cpp#L834-836) if I’m interpereting `SetCount` correctly *(GH doesn’t want to display the snippets for some reason)*.

*(But either way the current state should already be an improvement by itself and reviewable)*

astiob (approved):

astiob:
> `m 0 0 l 10 0 20 5 10 10 p 0 10` will presumably also differ due to `p` just extending the `l` in libass

Not only that, but any other unrecognized character will stop number parsing in VSFilter but continue the last parsed command in libass. But I agree these things can/should be fixed in a separate PR.

TheOneric:
Last minute cosmetic change in the first commit: renamed `UNINITIALIZED` to `INVALID` since it’s shorter and fits better to how it might also be used in the future (wrt to unrecognised characters) + i forogt to change `token_check_values` parameter from `int` to `ASS_TokenType` before

Issue #617, opened 2022-04-30
Rendering: Drawings without matching font fail to render


petzku:
### Screenshots

While I can't find a way to test this *against* VSFilter, I was told to file an issue, and this seemed like the most appropriate category.

![correctly rendering subtitles with both text in Gandhi Sans and a red box behind it](https://user-images.githubusercontent.com/5639335/166084222-a86e02e9-cda7-4461-8076-abeb93f03f09.png)

![text rendering but not the red box, as font not found](https://user-images.githubusercontent.com/5639335/166084198-f5957080-a0be-498a-ada9-ae171038078f.png)

### Description of the issue

Drawings require their style- or \fn-specified font to be found to render. When running mpv with `--sub-font-provider=none`, drawings will (apparently) fail to fallback to any font if the specified one isn't muxed, and fail to render anything.

Example where this could come up: Typesetting done with a style defaulting to Arial, but Arial itself is not muxed in, as it's not actually _used_ in any events (Aegisub's font collector, for instance, will not grab it in this case). Users who don't have Arial installed (or at least, those who have neither Arial nor any other font libass would fall back on in this case) will not see the drawings.

In addition, I get the following error in the latter (failing) case:
```
   sub/ass: fontselect: failed to find any fallback with glyph 0x0 for font: (Arial, 400, 0)
```

### libass version

osd/libass: libass source: commit: ac2ddef8841aa2ff37ca9b83f19f15092d0cacc6

### Is it a regression?

As far as I am aware, this has always been the case.

### ASS Sample

Full ASS file used: [font-drawing-rendering.ass.txt](https://github.com/libass/libass/files/8595403/font-drawing-rendering.ass.txt)

```
[Script Info]
; Script generated by Aegisub 9215, Daydream Cafe Edition [Shinon]
; http://www.aegisub.org/
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
YCbCr Matrix: None
PlayResX: 1280
PlayResY: 720

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
[...]

Issue #582, opened 2022-01-01
Rendering: Empty-filling drawing produces weird artifacts


misakikasumi:
### Screenshots

libass:
![image](https://user-images.githubusercontent.com/68379058/147856300-cc4d9965-aead-46ee-8990-2fd817b066df.png)

VSFilter:
![image](https://user-images.githubusercontent.com/68379058/147856302-0fdee5cf-460c-42ca-85ca-01b1dcf0119f.png)

### Description of the issue

Empty-filling drawing (e.g. in this case, `m 0 0 l 312 322 c`) produces weird artifacts.
Empty-filling drawing should be invisible. However, in this case, libass renders a line-like artifact.
It may be difficult to observe for one single drawing, but the artifact darkens as multiple empty-filling drawings stack together.
It is worth noting that the artifact only appear when the line is not horizontal, vertical, or diagonal.
(e.g. `m 0 0 l 312 312 c` does not produce artifact.)
Related to anti-aliasing?

### libass version

0.15.2

### Is it a regression?

I don't know

### ASS Sample

```
[Script Info]
; Script generated by Aegisub 9214, Daydream Cafe Edition [Shinon]
; http://www.aegisub.org/
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
YCbCr Matrix: TV.709
PlayResX: 1920
PlayResY: 1080
Kerning: yes

[Aegisub Project Garbage]
Last Style Storage: Default
Video File: ?dummy:23.976000:40000:1920:1080:47:163:254:
Video AR Value: 1.777778

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Default,Arial,80,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,10,0,1,0,0,5,10,10,10,1

[Events]
Format: Layer, Start, End, Style, Name, MarginL, MarginR, MarginV, Effect, Text
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,fx,{\pos(960,540)\c&H000000&\p1}m 0 0 l 312 322
[...]

Issue #302, opened 2018-03-11
libass draws borders on drawing edges regardless of drawing surface area


logarrhythmic:
Differing behavior: libass seems to draw borders based on the path defined by the vector drawing commands, while VSFilter seems to draw a blob under each pixel in the drawing.

Examples:
- clockwise square with counterclockwise triangle on top
`{\an5\p1\bord10\shad10}m 0 0 l 100 0 l 100 100 l 0 100 l 0 0 l 100 100 l 100 0 l 0 0`
libass
![libass_straight](https://user-images.githubusercontent.com/1952758/37255583-0ca2d826-2557-11e8-8e70-b39923b14ffa.png)
VSFilter
![vsfilter_straight](https://user-images.githubusercontent.com/1952758/37255584-0deaa380-2557-11e8-99e0-5eabcfeeec6b.png)
- clockwise square with 2 curved sides, counterclockwise square with an identical curve on the left and a straight line on the right
`{\an5\p1\bord10\shad10}m 20 0 l 120 0 b 140 20 100 80 120 100 l 20 100 b 0 80 40 20 20 0 b 40 20 0 80 20 100 l 120 100 l 120 0 l 20 0`
libass
![libass_wavy](https://user-images.githubusercontent.com/1952758/37255588-1047ef2a-2557-11e8-9e63-f75c3cd57750.png)
VSFilter
![vsfilter_wavy](https://user-images.githubusercontent.com/1952758/37255586-0f9e0b22-2557-11e8-8b33-fc56015a8953.png)
vsfilter with levels adjusted to highlight the pixels that cause the blobs to be drawn
![vsfilter_wavy_levels](https://user-images.githubusercontent.com/1952758/37255703-c16da546-2558-11e8-89fe-e15c50c9cc1f.png)

The pixels drawn on the left side are probably due to a rounding error which might not be present in all builds of VSFilter, but it makes for a good demonstration of how the border is drawn.

Furthermore, while testing random shapes, I found some ways to make libass render the borders incorrectly. 
1. A counterclockwise shape that is not entirely contained within a clockwise shape will fill the entire intersecting area with the border color
`{\an5\p1\bord5\shad10}m 0 0 l 100 0 l 100 100 l 0 100 l 0 0 m 50 70 l 110 70 110 30 50 30 50 70`
libass
![libass1](https://user-images.githubusercontent.com/1952758/37256037-ecf2458c-255d-11e8-9247-b8914a2c4746.png)
VSFilter
[...]

Issue #118, opened 2014-07-16
VSFilter tries to close single-point contours with bizarre results


Cyberbeing:
Someone linked a screenshot of this Libass issue recently.
![](http://i.imgbox.com/n868FYXu.png)

```
[Script Info]
ScriptType: v4.00+
WrapStyle: 0
PlayResX: 1920
PlayResY: 1080
ScaledBorderAndShadow: yes
YCbCr Matrix: TV.601

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Sign-Mincho,Iwata Mincho Pro M-Kami,60,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,96,100,1,0,1,0,0,5,10,10,10,1

[Events]
Format: Layer, Start, End, Style, Name, MarginL, MarginR, MarginV, Effect, Text
Dialogue: 0,0:00:01.00,0:00:10.00,Sign-Mincho,,0,0,0,,{\pos(0,0)\an7\fscx100\fscy100\c&HF6F5F8&\shad0\bord0\p1\blur1}m -2 -2 1922 -2 1922 1082 -2 1082 -2 -2

```

Note that the original line from the script in circulation had a typo, but that seems unrelated to the missing vector with Libass:

```
m -2 -2 1922 -2 1922 1082-2 1082 -2 -2
```

PR #728, opened 2024-01-19
drawing: tokenise vector drawings like VSFilter

TheOneric:
Our old code deviated from VSFilter in many ways (as discussed in #726) but this should (if I didn’t miss anything) align it with VSFilter. While this already leads to visibly different result in some cases, I don’t think any of the currently open vector drawing issues are fixed by this alone yet.

I decided to replace the whole tokenisation loop in favour of making the final result more readable, although this unfortunately makes the individual compat issues being fixed less visible in the diff. They are spelled out below:

 - VSFilter ignores invalid commands together with their coordinates. Prior to this, libass ignored invalid commands itself but appended their coordinates to the preceding valid command. This was even intentionally exploited to (not-)implement p, resulting in different behaviour if p was following something other than s.
 - VSFilter drops coordinates from b which are not part of a triplet.
 - VSFilter does not check the types of the nodes being re-added by c.
 - VSFilter ignores s commands with less than three points.
 - VSFilter ignores p commands if there aren’t at least 3 nodes yet.

Also drops the tokenised but never-parsed conic bezier `q` (and in the end the recently added `TOKEN_INVALID` now also became redundant and was removed again).

### DIFF
--- a/libass/ass_drawing.c
+++ b/libass/ass_drawing.c
@@ -43,19 +43,32 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
     return true;
 }
 
-static inline void add_node(ASS_DrawingToken **tail, ASS_TokenType type, ASS_Vector point)
+/*
+ * \brief Free a list of tokens
+ */
+static void drawing_free_tokens(ASS_DrawingToken *token)
+{
+    while (token) {
+        ASS_DrawingToken *at = token;
+        token = token->next;
+        free(at);
+    }
+}
+
+static inline bool add_node(ASS_DrawingToken **tail, ASS_TokenType type, ASS_Vector point)
 {
     assert(tail && *tail);
 
     ASS_DrawingToken *new_tail = malloc(sizeof(**tail));
     if (!new_tail)
-        return;
+        return false;
     (*tail)->next = new_tail;
     new_tail->prev = *tail;
     new_tail->next = NULL;
     new_tail->type = type;
     new_tail->point = point;
     *tail = new_tail;
+    return true;
 }
 
 static inline bool get_point(const char **str, ASS_Vector *point)
@@ -72,9 +85,10 @@ static inline bool get_point(const char **str, ASS_Vector *point)
  * Parses and advances the string for exactly 3 points.
  * If the string contains fewer than 3 points,
  * any initial matching coordinates are still consumed.
+ * If an allocation fails, error will be set to true.
  * \return whether three valid points were added
  */
-static bool add_3_points(const char **str, ASS_DrawingToken **tail, ASS_TokenType type)
+static bool add_3_points(const char **str, ASS_DrawingToken **tail, ASS_TokenType type, bool *error)
 {
     ASS_Vector buf[3];
 
@@ -88,9 +102,13 @@ static bool add_3_points(const char **str, ASS_DrawingToken **tail, ASS_TokenTyp
     if (!valid)
         return false;
 
-    add_node(tail, type, buf[0]);
-    add_node(tail, type, buf[1]);
-    add_node(tail, type, buf[2]);
+    valid = add_node(tail, type, buf[0]);
+    valid = valid && add_node(tail, type, buf[1]);
+    valid = valid && add_node(tail, type, buf[2]);
+    if (!valid) {
+        *error = true;
+        return false;
+    }
 
     return true;
 }
@@ -99,10 +117,11 @@ static bool add_3_points(const char **str, ASS_DrawingToken **tail, ASS_TokenTyp
  * Parses and advances the string while it matches points.
  * Each set of batch_size points will be turned into tokens and appended to tail.
  * Partial matches (i.e. an insufficient amount of coordinates) are still consumed.
+ * If an allocation fails, error will be set to true.
  * \return count of added points
  */
 static size_t add_many_points(const char **str, ASS_DrawingToken **tail,
-                              ASS_TokenType type, size_t batch_size)
+                              ASS_TokenType type, size_t batch_size, bool *error)
 {
     ASS_Vector buf[3];
     size_t max_buf = sizeof(buf) / sizeof(*buf);
@@ -125,7 +144,10 @@ static size_t add_many_points(const char **str, ASS_DrawingToken **tail,
             continue;
 
         for (size_t i = 0; i < count_batch; i++)
-            add_node(tail, type, buf[i]);
+            if (!add_node(tail, type, buf[i])) {
+                *error = true;
+                return count_total - count_batch + i;
+            }
         count_batch = 0;
     }
 
@@ -154,6 +176,7 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
     ASS_DrawingToken *root = NULL, *tail = NULL, *spline_start = NULL;
     size_t points = 0;
     bool m_seen = false;
+    bool error = false;
 
     while (p && *p) {
         char cmd = *p;
@@ -177,9 +200,9 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
                 if (!get_point(&p, &point))
                     continue;
                 if (!add_root_node(&root, &tail, &points, point, TOKEN_MOVE))
-                    continue;
+                    return NULL;
             }
-            points += add_many_points(&p, &tail, TOKEN_MOVE, 1);
+            points += add_many_points(&p, &tail, TOKEN_MOVE, 1, &error);
             break;
         case 'n':
             if (!root) {
@@ -189,19 +212,19 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
                 if (!m_seen)
                     return NULL;
                 if (!add_root_node(&root, &tail, &points, point, TOKEN_MOVE_NC))
-                    continue;
+                    return NULL;
             }
-            points += add_many_points(&p, &tail, TOKEN_MOVE_NC, 1);
+            points += add_many_points(&p, &tail, TOKEN_MOVE_NC, 1, &error);
             break;
         case 'l':
             if (!root)
                 continue;
-            points += add_many_points(&p, &tail, TOKEN_LINE, 1);
+            points += add_many_points(&p, &tail, TOKEN_LINE, 1, &error);
             break;
         case 'b':
             if (!root)
                 continue;
-            points += add_many_points(&p, &tail, TOKEN_CUBIC_BEZIER, 3);
+            points += add_many_points(&p, &tail, TOKEN_CUBIC_BEZIER, 3, &error);
             break;
         case 's':
             if (!root)
@@ -209,9 +232,9 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
             // Only the initial 3 points are TOKEN_B_SPLINE,
             // all following ones are TOKEN_EXTEND_SPLINE
             spline_start = tail;
-            if (!add_3_points(&p, &tail, TOKEN_B_SPLINE)) {
+            if (!add_3_points(&p, &tail, TOKEN_B_SPLINE, &error)) {
                 spline_start = NULL;
-                continue;
+                break;
             }
             points += 3;
             //-fallthrough
@@ -219,7 +242,7 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
             if (points < 3)
                 continue;
             // XXX: use TOKEN_EXTEND_SPLINE
-            points += add_many_points(&p, &tail, TOKEN_B_SPLINE, 1);
+            points += add_many_points(&p, &tail, TOKEN_B_SPLINE, 1, &error);
             break;
         case 'c':
             if (!spline_start)
@@ -227,7 +250,10 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
             // Close b-splines: add the first three points of the b-spline back to the end
             for (int i = 0; i < 3; i++) {
                 // XXX: use TOKEN_EXTEND_SPLINE
-                add_node(&tail, TOKEN_B_SPLINE, spline_start->point);
+                if (!add_node(&tail, TOKEN_B_SPLINE, spline_start->point)) {
+                    error = true;
+                    break;
+                }
                 spline_start = spline_start->next;
             }
             spline_start = NULL;
@@ -236,21 +262,15 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
             // Ignore, just search for next valid command
             break;
         }
+        if (error)
+            goto fail;
     }
 
     return root;
-}
 
-/*
- * \brief Free a list of tokens
- */
-static void drawing_free_tokens(ASS_DrawingToken *token)
-{
-    while (token) {
-        ASS_DrawingToken *at = token;
-        token = token->next;
-        free(at);
-    }
+fail:
+    drawing_free_tokens(root);
+    return NULL;
 }
 
 /*
--- a/libass/ass_drawing.c
+++ b/libass/ass_drawing.c
@@ -43,93 +43,199 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
     return true;
 }
 
+static inline void add_node(ASS_DrawingToken **tail, ASS_TokenType type, ASS_Vector point)
+{
+    assert(tail && *tail);
+
+    ASS_DrawingToken *new_tail = malloc(sizeof(**tail));
+    if (!new_tail)
+        return;
+    (*tail)->next = new_tail;
+    new_tail->prev = *tail;
+    new_tail->next = NULL;
+    new_tail->type = type;
+    new_tail->point = point;
+    *tail = new_tail;
+}
+
+static inline bool get_point(const char **str, ASS_Vector *point)
+{
+    double x, y;
+    if (!mystrtod((char **) str, &x) || !mystrtod((char **) str, &y))
+        return false;
+    *point = (ASS_Vector) {double_to_d6(x), double_to_d6(y)};
+    return true;
+}
+
+
+/**
+ * Parses and advances the string for exactly 3 points.
+ * If the string contains fewer than 3 points,
+ * any initial matching coordinates are still consumed.
+ * \return whether three valid points were added
+ */
+static bool add_3_points(const char **str, ASS_DrawingToken **tail, ASS_TokenType type)
+{
+    ASS_Vector buf[3];
+
+    if (!*str)
+        return false;
+
+    bool valid = get_point(str, buf + 0);
+    valid = valid && get_point(str, buf + 1);
+    valid = valid && get_point(str, buf + 2);
+
+    if (!valid)
+        return false;
+
+    add_node(tail, type, buf[0]);
+    add_node(tail, type, buf[1]);
+    add_node(tail, type, buf[2]);
+
+    return true;
+}
+
+/*
+ * Parses and advances the string while it matches points.
+ * Each set of batch_size points will be turned into tokens and appended to tail.
+ * Partial matches (i.e. an insufficient amount of coordinates) are still consumed.
+ * \return count of added points
+ */
+static size_t add_many_points(const char **str, ASS_DrawingToken **tail,
+                              ASS_TokenType type, size_t batch_size)
+{
+    ASS_Vector buf[3];
+    size_t max_buf = sizeof(buf) / sizeof(*buf);
+    assert(batch_size <= max_buf);
+
+    if (!*str)
+        return 0;
+
+    size_t count_total = 0;
+    size_t count_batch = 0;
+    while (**str) {
+        ASS_Vector point;
+        if (!get_point(str, &point))
+            break;
+        buf[count_batch] = point;
+        count_total++;
+        count_batch++;
+
+        if (count_batch != batch_size)
+            continue;
+
+        for (size_t i = 0; i < count_batch; i++)
+            add_node(tail, type, buf[i]);
+        count_batch = 0;
+    }
+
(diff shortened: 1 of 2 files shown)


TheOneric:
As usual, if anyone wants to review this, let me know and this can wait until whatever time suits you. But if no one does, I’ll merge it in a couple days.

libass/ass_drawing.c
```diff
@@ -43,93 +43,148 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
...
-                } else if (type != TOKEN_MOVE) {
-                    p++;
+        char cmd = *p;
+        p++;
+        /* VSFilter compat:
+         * In guliverkli(2) VSFilter all drawings
+         * whose first valid command isn't m are rejected.
+         * xy-VSF and MPC-HC ISR this was (possibly inadvertenly) later relaxed,
+         * such that all valid commands but n are ignored if there was no m yet.
+         * 'p' is further restricted and ignored unless there arealready >= 3 nodes.
```
rcombs, line 127: *already

libass/ass_drawing.c
```diff
@@ -43,93 +43,148 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
...
+            spline_start = tail;
+            size_t new_coords = add_coordinates(&p, &tail, TOKEN_B_SPLINE, 3, 1);
+            coords += new_coords;
+            if (new_coords == 0)
+                spline_start = NULL;
+            break;
+        case 'p':
+            if (coords < 3)
+                continue;
+            // XXX: use TOKEN_EXTEND_B_SPLINE
```
rcombs, line 169: TOKEN_EXTEND_SPLINE?

libass/ass_drawing.c
rcombs: also, clarify if the p restriction is new in xy/ISR or not

rcombs (commented):
A few nits. Code generally looks good, and I like the cleanup.

libass/ass_drawing.c
TheOneric: thx, fixed both

libass/ass_drawing.c
TheOneric: thx, fixed the typo and added *“in newer VSFs, but guliverkli2 just crashed on 'p' with  1 <= n < 3 preceding nodes”* to the `p` description

libass/ass_drawing.c
```diff
@@ -43,93 +43,149 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
...
+            if (coords < 3)
+                continue;
+            // XXX: use TOKEN_EXTEND_SPLINE
+            coords += add_coordinates(&p, &tail, TOKEN_B_SPLINE, 1, 1);
+            break;
+        case 'c':
+            if (!spline_start)
+                continue;
+            // Close b-splines: add the first three points of the b-spline
+            // back to the end (VSF does not check the original node types)
```
astiob, line 177: If I’m reading the code (here and in VSFilter) correctly, then I’m not sure what this parenthetical comment is supposed to mean: the second and third point are guaranteed to be `TOKEN_B_SPLINE`, while the first point can legally be anything, because it’s the point _before_ the spline rather than part of the spline command. But there can be non-spline nodes between `spline_start` and the `c` command, which may be surprising. Well, the whole `c` command seems weird to me, but what can you do.

astiob (commented):
Is there actually any difference between `TOKEN_B_SPLINE` and the hypothetical `TOKEN_EXTEND_SPLINE`? If not, why not drop `TOKEN_EXTEND_SPLINE` completely (from the comments and from the enum)?

libass/ass_drawing.c
astiob: Is there a reason to comment on the crashing for `p` but not the other commands? Older VSFilter crashed on each of these commands if there weren’t enough preceding points; that’s why the guards were added.

Which makes me realize: our handling of “required first `m`” may be still wrong in master and after this PR in its current form. VSFilter actually requires an `m` command _letter_ to be seen at a point where no valid commands have been read yet, but the `m` command itself doesn’t have to be valid (to have valid coordinates)! This is probably most of the reason for those crashes they had to fix, lol.

astiob: Specifically, the `root` can actually be an `n` command:

```
m n 0 0 l 100 0 100 100 0 100 0 0
```

shows a square in VSFilter identical to this:

```
m 0 0 l 100 0 100 100 0 100
```

TheOneric: > Is there a reason to comment on the crashing for p but not the other commands? Older VSFilter crashed on each of these commands if there weren’t enough preceding points

Your second paragraph (good catch!) contains the reason why. If, as I mistakenly thought before, the drawing must start with a *valid* `m`, all those other crashes wouldn’t have been possible, since there already is one `m` and all other commands only need `>= 1` preceding node.

Adjusted the comment and implemented the “invalid `m` before `n` case”, thanks!

libass/ass_drawing.c
```diff
@@ -43,93 +43,149 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
...
+    (*tail)->next = calloc(1, sizeof(**tail));
+    if (!(*tail)->next)
+        return;
+    (*tail)->next->prev = *tail;
+    *tail = (*tail)->next;
+    (*tail)->type = type;
+    (*tail)->point = point;
+}
+
+static inline bool get_coordinate(const char **str, ASS_Vector *coord)
```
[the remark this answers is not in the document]
TheOneric: `add_many_points` sounds good; renamed everything but the few instances actually applying to individual coordinates to something with “point”

libass/ass_drawing.c
TheOneric: You are right, this part of the comment is kinda useless; dropped it.  
And yeah, `c` can ""close"" non-splines (there might even have been an `m` inbetween) and `p` also can ""extend"" non-splines

TheOneric:
> Is there actually any difference between TOKEN_B_SPLINE and the hypothetical TOKEN_EXTEND_SPLINE?

Yes, there is.  `s` will use the preceding node, itself and the next two nodes. While `p` will use the preseding three nodes and itself.

https://github.com/pinterf/xy-VSFilter/blob/45917d77083d93215738efb140486aba7614947e/src/subtitles/Rasterizer.cpp#L3303-L3309

For a "sensible" drawing where an `s` is only followed by any number of `p` and a closing `c` and `p` and `c` only occur after a `s`, this doesn’t matter I think. But something like the following should display differently depending on whether `p` uses `B_SPLINE` or `EXTEND_SPLINE`:
```
m 0 0 l 500 0 500 150 900 200 p 600 800 400 400 200 700
```

----

Annnd this made me noticed I somehow missed another thing in VSFilter: `s` does not add *at least* 3 `B_SPLINE` points, but only exactly 3 or none and all following are `EXTEND_SPLINE` *(the fallthrough is even explicitly commented! duh)*
For implementing this i split out `add_3_points` which in turn simplified `add_many__points` which now only needs to handle `batch_size` and no longer needs to `memcpy`

To sum up the new version:
 - now accepts `n` as root node if there was an invalid `m` before
 - only adds the first three points of `s` as `B_SPLINE`, any following are handled like `p`
 - renamed most `coordinate` to `point`

astiob:
If VSFilter and libass use different points for splines, doesn’t that mean they basically never work in libass? I’m afraid I’m not following this:

> For a "sensible" drawing where an `s` is only followed by any number of `p` and a closing `c` and `p` and `c` only occur after a `s`, this doesn’t matter I think.

FWIW the sample with `p` but without `s` that you posted does render differently: libass (mine is probably a bit outdated) uses only straight lines, but VSFilter has curves.

TheOneric:
> If VSFilter and libass use different points for splines, doesn’t that mean they basically never work in libass?

In the general case there are differences, yes. However, in the specific case where `s` is im- or explicitly followed by `p` after its first three points and `p` only appears after `s` or other `p`, the result ends up being the same. This is because libass currently treats `B_SPLINE` like VSFilter uses `s` — but just without skipping three nodes ahead afterwards.

I’m hoping this visualisation makes it clearer:
libass-vsf_splines


If `p` comes after a non-spline there’ll be a difference still after this PR, until the parser implements `EXTEND_SPLINE` and is fixed. Before this PR `p` didn’t even necessarily create spline-type nodes but just extended whatever came before, which is why you are seeing straight lines when testing without this PR *(first bullet point in OP)*.

For the eventual follow-up PR, the interesting case is then how VSFilter renders the sample from before (a) as is and (b) if replacing `p` with `s`. Barring any more bugs or misunderstandings I’d expect libass with this PR to match (b) but differ from (a).

astiob:
Thanks for the picture! But what happens after? libass still has two unprocessed nodes 🤔

astiob:
Hmm, did you accidentally push a wrong branch? The current code seems to be an _older_ version of this PR that got pushed anew: it’s missing some recent changes from `master` that the last version had, and it’s missing the changes you said you’d done in response to code review.

astiob:
Thanks :) GitHub comparison link (previous non-mistaken push vs current) for convenience: [here](https://github.com/libass/libass/compare/efa78b8deda180634caf9d8598396bdfc1acaf35..18649a629e10499e541a2cedf9f1b1bb373ecf50).

You’ve added a comment explaining `m n`, but it isn’t yet handled by the code. I assume `n` also needs an `if (!root)` section. And the `if (!m_seen) return NULL;` must be conditional on whether the `n` actually succeeds in reading a (`root`) point.

libass/ass_drawing.c
```diff
@@ -43,93 +43,181 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
...
+        /* VSFilter compat:
+         * In guliverkli(2) VSFilter all drawings
+         * whose first known (but potentially invalid) command isn't m are rejected.
+         * xy-VSF and MPC-HC ISR later relaxed this (possibly inadvertenly),
+         * such that all known commands but n are ignored if there was no prior node yet.
+         * If an invalid m preceded n, the latter becomes the root node, otherwise
+         * if n comes before any other not-ignored command the entire drawing is rejected.
+         * 'p' is further restricted and ignored unless there are already >= 3 nodes.
+         * This relaxation was a byproduct of a fix for crashing on drawings
+         * containing commands with less preceding nodes than expected.
```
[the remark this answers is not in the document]
TheOneric: fixed

TheOneric:
> You’ve added a comment explaining m n, but the valid cases aren’t yet allowed by the code. I assume n also needs an if (!root) section. And the if (!m_seen) return NULL; must be conditional on whether the n actually succeeds in reading a (root) point.

Thanks, should be fixed now

astiob (approved):
Ah, I see the observation about `n` has conveniently led to an unchecked allocation being fixed, too. Nice.

Generally LGTM now. But now that I’ve looked closer at the memory handling, would it perhaps make more sense to discard the whole drawing if any point’s allocation fails, rather than skip that point? Or at least do this for moves? Or even just `root`? Not sure.

MrSmile (commented):
A couple of comments from me, but I've not checked logic correctness.

libass/ass_drawing.c
```diff
@@ -43,93 +43,198 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
...
-                    tail->next = calloc(1, sizeof(ASS_DrawingToken));
-                    tail->next->prev = tail;
-                    tail = tail->next;
-                    tail->type = TOKEN_B_SPLINE;
-                    tail->point = spline_start->point;
-                    spline_start = spline_start->next;
-                }
-                spline_start = NULL;
+        char cmd = *p;
+        p++;
```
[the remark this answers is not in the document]
TheOneric: hmm, i left it as two statements for now

moi15moi:
Does this resolve #435, #629, #202 and #302?

TheOneric:
> Does this resolve #435, #629, #202 and #302?

No, it solves some (non)rejection errors and partly `p`, but:

> While this already leads to visibly different result in some cases, I don’t think any of the currently open vector drawing issues are fixed by this alone yet.

It does however ensure we’ll have the same tokenview when tackling the token-to-outline issues, i.e. the first three you mentioned and #118.  
E.g. this prepares a proper `TOKEN_EXTEND_SPLINE` which should solve the `s ... s ...` case from #202.

#302 and #582 are unrelated to the tokenisation changes here and can be solved dropping zero-area parts (or rather since that’s hard, 2-point contours).
And for the sake of completion, there’s also #617 which needs to be tackled higher up in the rendering process.

TheOneric:
Sorry, I forgot to reply to this before:

> But now that I’ve looked closer at the memory handling, would it perhaps make more sense to discard the whole drawing if any point’s allocation fails, rather than skip that point? Or at least do this for moves? Or even just root? Not sure.

In general, not just silently ignoring alloc failures might make sense yeah. However, unfortunately we do this in many places and ideally we would also like to signal the alloc failure to API users  *(so they know the rendering is degraded and memory might be tight)*, but as a I brought up before, we have no way of signalling in the current API.  
As for practical implementation concerns just in the tokeniser, I think this should at least be a new commit as this one already does a lot of things. And being able to detect and properly free the drawing on failures for anything but the root node will at least slightly complicate the code flow.  

If we assume a failure means memory is and will stay tight, dropping the whole drawing makes most sense imho. If we assume it is just temporarily not possible, we’d probably ideally want to ignore just the part from the last to the next `m`, to accurately display as much as possible.

TheOneric:
Added a commit rejecting the entire drawing on alloc failure, using a separate `error` output parameter, since this results in imho cleaner code than negative return values.

libass/ass_drawing.c
```diff
@@ -43,93 +43,198 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
     return true;
 }
 
+static inline void add_node(ASS_DrawingToken **tail, ASS_TokenType type, ASS_Vector point)
+{
+    assert(tail && *tail);
+
+    (*tail)->next = calloc(1, sizeof(**tail));
```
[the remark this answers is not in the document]
MrSmile: I think reverse order is better, i. e.
```c
ASS_DrawingToken *new_tail = malloc(sizeof(**tail));
(*tail)->next = new_tail;
```
It should be easier to understand for compilers too, as it has less memory accesses.

As for `calloc` vs. `malloc`, I think that such "protective" programming is harmful: it would be easier to find a missing initialization (with valgrind, for example) in case of uninitialized value than if it was already zero-initialized. Also it's impossible to say by looking at the code if such zero initialization is intended or accidental and programmer forgot to set a correct value.

libass/ass_drawing.c
```diff
@@ -43,93 +43,198 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
...
+
+    bool valid = get_point(str, buf + 0);
+    valid &= get_point(str, buf + 1);
+    valid &= get_point(str, buf + 2);
+
+    if (valid) {
+        add_node(tail, type, buf[0]);
+        add_node(tail, type, buf[1]);
+        add_node(tail, type, buf[2]);
+        return 3;
```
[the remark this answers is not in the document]
MrSmile: I feel that `bool` variant is clearer as it allows to eliminate intermediate variable `size_t new_points` later.

libass/ass_drawing.c
```diff
@@ -72,9 +85,10 @@ static inline bool get_point(const char **str, ASS_Vector *point)
  * Parses and advances the string for exactly 3 points.
  * If the string does not contain fewer than 3 points,
  * any initial matching coordinates are still consumed.
+ * If an allocation fails, error will be set to a non-zero value.
  * \return count of added points
  */
-static size_t add_3_points(const char **str, ASS_DrawingToken **tail, ASS_TokenType type)
+static size_t add_3_points(const char **str, ASS_DrawingToken **tail, ASS_TokenType type, int *error)
```
MrSmile, line 91: Maybe `bool *error` would be better?

libass/ass_drawing.c
```diff
@@ -88,9 +102,12 @@ static size_t add_3_points(const char **str, ASS_DrawingToken **tail, ASS_TokenT
     if (!valid)
         return 0;
 
-    add_node(tail, type, buf[0]);
-    add_node(tail, type, buf[1]);
-    add_node(tail, type, buf[2]);
+    if (!add_node(tail, type, buf[0]) ||
+            !add_node(tail, type, buf[1]) ||
+            !add_node(tail, type, buf[2])) {
```
MrSmile, line 107: Indentation looks horrible. Maybe the same trick with `valid` would be better?
```c
valid = valid && add_node(tail, type, buf[0]);
valid = valid && add_node(tail, type, buf[1]);
valid = valid && add_node(tail, type, buf[2]);
if (!valid) { ... }
```

libass/ass_drawing.c
```diff
@@ -189,46 +211,47 @@ static ASS_DrawingToken *drawing_tokenize(const char *str)
...
             break;
         case 'c':
             if (!spline_start)
                 continue;
             // Close b-splines: add the first three points of the b-spline back to the end
             for (int i = 0; i < 3; i++) {
                 // XXX: use TOKEN_EXTEND_SPLINE
-                add_node(&tail, TOKEN_B_SPLINE, spline_start->point);
+                if (!add_node(&tail, TOKEN_B_SPLINE, spline_start->point))
+                    goto fail;
```
MrSmile, line 254: I think it's better to reuse `error` approach here to homogenize control flow. In that case error handling can be put directly under `if (error) { ... }` after the switch.

libass/ass_drawing.c
```diff
@@ -43,108 +43,232 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
...
+    if (!mystrtod((char **) str, &x) || !mystrtod((char **) str, &y))
+        return false;
+    *point = (ASS_Vector) {double_to_d6(x), double_to_d6(y)};
+    return true;
+}
+
+
+/**
+ * Parses and advances the string for exactly 3 points.
+ * If the string does not contain fewer than 3 points,
```
astiob, line 86: This is probably meant to say “contains fewer than 3 points” or “does not contain 3 points”

libass/ass_drawing.c
TheOneric: If you mean manually setting `error = 1` inside the `if` block, then it’s a bit more verbose but still fine so sure.  
If you meant `add_node` should take and set an `error` parameter, then i don’t think this is a good idea as we’ll require an `error` variable also in `add_*_points`

libass/ass_drawing.c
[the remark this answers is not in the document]
TheOneric: Adopted the reversed order

libass/ass_drawing.c
```diff
@@ -43,93 +43,198 @@ static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType typ
...
+ */
+static bool add_3_points(const char **str, ASS_DrawingToken **tail, ASS_TokenType type)
+{
+    ASS_Vector buf[3];
+
+    if (!*str)
+        return 0;
+
+    bool valid = get_point(str, buf + 0);
+    valid &= get_point(str, buf + 1);
```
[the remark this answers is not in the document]
astiob: I want to note that this applies short-circuiting and invalidates the promise given in the function comments that the points are always consumed. It shouldn’t actually affect the overall behaviour though, as these unconsumed points will now be skipped by the main loop if I’m not mistaken. It may be good to adjust the comments to avoid future surprises, though.

(Personally, I think `valid &=` is easier to read than `valid = valid &&`, but I don’t mind either way)

libass/ass_drawing.c
TheOneric: fixed thx

libass/ass_drawing.c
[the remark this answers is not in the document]
TheOneric: > I want to note that this applies short-circuiting and invalidates the promise given in the function comments that the points are always consumed.

wait, is it even possible for any subsequent `get_point` calls to consume anything if a prior one failed? iinm, the first char in the string should then be some non-digit, non-dor and thus any attempt at further processing immediately fail, doesn’t it?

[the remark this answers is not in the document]
MrSmile: > I want to note that this applies short-circuiting and invalidates the promise given in the function comments that the points are always consumed.

To preserve that it's possible to rewrite as
```c
valid = get_point(str, buf + 1) && valid;
```

libass/ass_drawing.c
MrSmile: > If you mean manually setting `error = 1` inside the `if` block

Yes, that.

TheOneric: done

libass/ass_drawing.c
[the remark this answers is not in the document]
TheOneric: I don’t mind much, but the `size_t` version is overall more consistent with `add_many_points`

libass/ass_drawing.c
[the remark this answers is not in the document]
astiob: > wait, is it even possible for any subsequent get_point calls to consume anything if a prior one failed?

Ah, no, it is not.

libass/ass_drawing.c
TheOneric: done

libass/ass_drawing.c
TheOneric: done

TheOneric:
Applied one minor, effectively cosmetic change.  
I didn’t spot any more silly mistakes, all comments should be addressed and tests pass; this should now finally be ready for merge.

Issue #634, opened 2022-07-17
Failing to find glyph 0x1F90D in Noto Emoji font, but only when attached


rcombs:
Compact script:
```
[Script Info]
; Script generated by Aegisub 9215, Daydream Cafe Edition [Shinon]
; http://www.aegisub.org/
Title: Call of the Night 02
ScriptType: v4.00+
WrapStyle: 0
PlayResX: 1920
PlayResY: 1080
ScaledBorderAndShadow: yes
YCbCr Matrix: TV.709

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Default,LTFinnegan Medium,75,&H00FFFFFF,&H000019FF,&H00000000,&HB4000000,0,0,0,0,100,100,0,0,1,3.3,1.5,2,135,135,53,1

[Events]
Format: Layer, Start, End, Style, Name, MarginL, MarginR, MarginV, Effect, Text
Dialogue: 5,0:02:15.99,0:02:20.56,Default,,0,0,0,,Seriously, you can't just give me a \N"Drink my blood now{\fnNoto Emoji\b1}🤍{\r}" look like that.
```

Font: [NotoEmoji-Bold.ttf.zip](https://github.com/libass/libass/files/9127089/NotoEmoji-Bold.ttf.zip)

I tested the MKV file (which has this font attached) on libass 7bc0c45dd58de6afa1800f8e8a94285e7535d68d.

With CoreText enabled, I get this:
```
[sub/ass] fontselect: (Noto Emoji, 400, 0) -> NotoEmoji-Bold, 0, NotoEmoji-Bold
[sub/ass] fontselect: (Noto Emoji, 700, 0) -> NotoEmoji-Bold, 0, NotoEmoji-Bold
[sub/ass] Glyph 0x1F90D not found, selecting one more font for (Noto Emoji, 700, 0)
[sub/ass] fontselect: (Noto Emoji, 700, 0) -> /System/Library/Fonts/LastResort.otf, -1, LastResort
```
<img width="1162" alt="image" src="https://user-images.githubusercontent.com/819638/179385921-04663a35-886c-45f0-a5eb-a4e1571eef17.png">

And without, I get this:
```
[...]

PR #729, opened 2024-01-20
font: prefer full-Unicode cmaps to BMP-only ones

astiob:
Fixes: https://github.com/libass/libass/issues/634
Fixes: https://github.com/mpv-player/mpv/issues/13322

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -199,23 +199,35 @@ static uint32_t convert_unicode_to_mb(FT_Encoding encoding, uint32_t codepoint) 
 void ass_charmap_magic(ASS_Library *library, FT_Face face)
 {
     int i;
-    int ms_cmap = -1;
+    int ms_cmap = -1, ms_unicode_cmap = -1;
 
     // Search for a Microsoft Unicode cmap
     for (i = 0; i < face->num_charmaps; ++i) {
         FT_CharMap cmap = face->charmaps[i];
         unsigned pid = cmap->platform_id;
         unsigned eid = cmap->encoding_id;
-        if (pid == 3 /*microsoft */
-            && (eid == 1 /*unicode bmp */
-                || eid == 10 /*full unicode */ )) {
-            FT_Set_Charmap(face, cmap);
-            return;
-        } else if (pid == 3 && ms_cmap < 0)
-            ms_cmap = i;
+        if (pid == TT_PLATFORM_MICROSOFT) {
+            switch (eid) {
+            case TT_MS_ID_UCS_4:
+                // Full Unicode cmap: select this immediately
+                FT_Set_Charmap(face, cmap);
+                return;
+            case TT_MS_ID_UNICODE_CS:
+                // BMP-only Unicode cmap: select this
+                // if no fuller Unicode cmap exists
+                if (ms_unicode_cmap < 0)
+                    ms_unicode_cmap = ms_cmap = i;
+                break;
+            default:
+                // Non-Unicode cmap: select this if no Unicode cmap exists
+                if (ms_cmap < 0)
+                    ms_cmap = i;
+            }
+        }
     }
 
-    // Try the first Microsoft cmap if no Microsoft Unicode cmap was found
+    // Try the first Microsoft BMP cmap if no MS cmap had full Unicode,
+    // or the first MS cmap of any kind if none of them had Unicode at all
     if (ms_cmap >= 0) {
         FT_CharMap cmap = face->charmaps[ms_cmap];
         FT_Set_Charmap(face, cmap);


astiob:
For reference, my (untested) understanding is that if the font has no Microsoft cmaps, our code already ends up using a full-Unicode non-Microsoft cmap if one exists or a less-full non-Microsoft Unicode cmap otherwise, because that’s what FreeType selects automatically upon opening the font face, and only if no Unicode cmap exists, does our last loop execute and select the first of all cmaps.

rcombs:
Hmm, does this match the GDI behavior? At least in older Windows versions (XP), as far as I can tell, there was no special handling for format ID 10, and no preference was given between different encoding IDs at all (it just used the first Microsoft cmap, falling back on a Mac one with plat=1 enc=0 if none existed). Noto Emoji might get symbol-font treatment here since it has a panose familytype of 5 (pictoral) in its OS/2 table.

That said, regardless of GDI behavior, I'm 👍 on this change (it's hard to imagine it making anything *worse*, and clearly improves things for users here).

astiob:
If you mean what I think you mean, there seems to be no support for encoding 10 or 6 (Johab) in that code at all, so it’s definitely outdated. At any rate, we know VSFilter does show supplementary-plane characters, and it couldn’t do it without encoding 10.

But I’ve just run a dedicated test, and it’s… weird. VSFilter doesn’t use the Noto Emoji glyph from #634, substituting it by what looks like the glyph from Segoe UI Emoji (but does use base ASCII glyphs from Noto Emoji). But Segoe UI Emoji is itself affected by this cmap issue (and works fine in VSFilter), and the FiraCode sample from mpv also works in VSFilter.

TheOneric (approved):
Makes sense to me, code looks good and I can confirm it fixes the two linked samples (in `compare`).

I don’t know what exactly GDI/VSFilter does, but since it appear to use UCS-4 cmap at least most of the time, this change should also be an improvement i this regard.

> For reference, my (untested) understanding is that if the font has no Microsoft cmaps, our code already ends up using a full-Unicode non-Microsoft cmap if one exists or a less-full non-Microsoft Unicode cmap otherwise, because that’s what FreeType selects automatically upon opening the font face

I couldn’t find this being promised in the docs, but indeed FreeType’s code currently [calls `find_unicode_charmap` in `open_face`](https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/base/ftobjs.c#L1569)  which [tries to select](https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/base/ftobjs.c#L1410) a full-range Unicode map of either Microsoft or Apple platform encoding or if no such exists any Unicode encoding cmap.

astiob:
> I couldn’t find this being promised in the docs, but indeed FreeType’s code currently [calls `find_unicode_charmap` in `open_face`](https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/base/ftobjs.c#L1569) which [tries to select](https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/base/ftobjs.c#L1410) a full-range Unicode map of either Microsoft or Apple platform encoding or if no such exists any Unicode encoding cmap.

Yup, this is exactly my own main reference. But this is somewhat documented, too:

* [The doc for `FT_Select_Charmap`](https://freetype.org/freetype2/docs/reference/ft2-character_mapping.html#ft_select_charmap) says it selects the fullest Unicode cmap when asked for Unicode.
* [The doc for `FT_CharMap`](https://freetype.org/freetype2/docs/reference/ft2-character_mapping.html#ft_charmap) says newly created faces select a Unicode cmap automatically.

ahaoboy:
After testing, this fix can render 0xF0411 correctly. Place the font file in compare/test/FiraCodeNerdFont-Regular.ttf and modify the font name and rendering text in sub1.ass.

```ass
[Script Info]
PlayResX: 320
PlayResY: 240
ScaledBorderAndShadow: yes

[V4+ Styles]
Style: Default,FiraCode Nerd Font Reg,200,&H000000FF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,8,0,5,10,10,10,1

[Events]
Dialogue: 0,0:00:00.00,0:00:04.00,Default,,0,0,0,,{\t(\frz+180)}󰐑

```
[FiraCodeNerdFont-Regular.zip](https://github.com/libass/libass/files/13999705/FiraCodeNerdFont-Regular.zip)

Branch master
```
Loading font 'FiraCodeNerdFont-Regular.ttf'.
Loading font 'font1.ttf'.
Loading font 'font2.otf'.
Processing 'sub1.ass':
  Time 0:00:00.500 - libass: fontselect: failed to find any fallback with glyph 0xF0411 for font: (FiraCode Nerd Font Reg, 400, 0)
255.000 FAIL
  Time 0:00:01.500 - libass: fontselect: failed to find any fallback with glyph 0xF0411 for font: (FiraCode Nerd Font Reg, 400, 0)
255.000 FAIL
  Time 0:00:02.500 - libass: fontselect: failed to find any fallback with glyph 0xF0411 for font: (FiraCode Nerd Font Reg, 400, 0)
255.000 FAIL
Processing 'sub2.ass':
  Time 0:02:33.000 - 0.728 GOOD
Only 1 of 4 images have passed test (BAD or better)
```
image


Branch full-unicode-cmap
```
Loading font 'FiraCodeNerdFont-Regular.ttf'.
Loading font 'font1 copy.ttf'.
Loading font 'font1.ttf'.
Loading font 'font2.otf'.
Processing 'sub1.ass':
  Time 0:00:00.500 - 255.000 FAIL
  Time 0:00:01.500 - 255.000 FAIL
  Time 0:00:02.500 - 255.000 FAIL
Processing 'sub2.ass':
  Time 0:02:33.000 - 0.728 GOOD
Only 1 of 4 images have passed test (BAD or better)
```
image

PR #735, opened 2024-02-19
Document tonemapping beyond RGB mangling

TheOneric:
Followup to the IRC discussion preceding and https://github.com/mpv-player/mpv/issues/13381 itself.

I’m not sure I got the terminology quite right, please check.

The YCbCr section in the file format guide prob also needs some tweaking. Perhaps something like **(changed section emphasised)*?:
> **YCbCr Matrix**
>
> [...]
> 
> Following these instructions will ensure, the final colours effectively exactly match the sRGB values specified in ASS. Diverging from this, will lead to **unmapped** colours being mangled as described in ass_types.h. **(But colour primaries and tone mapping will always follow the video colourspace regardless)**

### DIFF
--- a/libass/ass_types.h
+++ b/libass/ass_types.h
@@ -229,6 +229,12 @@ typedef struct ass_event {
  * absolutely can't do that, because the video colorspace is required
  * in order to handle this as intended. API users must use the exposed
  * information to perform color mangling as described above.
+ *
+ * Further note all of the above only concerns the RGB values.
+ * Color primaries and transfer charateristics of ASS subtitles
+ * must always match their associated video.
+ * (This indeed has some undesirable effects on HDR videos,
+ * but no mechanism avoiding this is yet standardized.)
  */
 typedef enum ASS_YCbCrMatrix {
     YCBCR_DEFAULT = 0,  // Header missing


astiob (approved):

libass/ass_types.h
```diff
@@ -229,6 +229,12 @@ typedef struct ass_event {
  * absolutely can't do that, because the video colorspace is required
  * in order to handle this as intended. API users must use the exposed
  * information to perform color mangling as described above.
+ *
+ * Further note all of the above only concerns the RGB values.
+ * Gamut mapping and transfer charateristics of ASS subtitles
+ * must always match their associated video.
+ * (This indeed has some undesireable effects on HDR videos,
+ * but no mechansim avoiding this is yet standardized.)
```
arch1t3cht, line 237: Spelling: `mechanism` and `undesirable`

libass/ass_types.h
```diff
@@ -229,6 +229,12 @@ typedef struct ass_event {
  * absolutely can't do that, because the video colorspace is required
  * in order to handle this as intended. API users must use the exposed
  * information to perform color mangling as described above.
+ *
+ * Further note all of the above only concerns the RGB values.
+ * Gamut mapping and transfer charateristics of ASS subtitles
```
arch1t3cht, line 234: As far as I'm aware, "Gamut mapping" specifically refers to the process of mapping colors in one gamut to another gamut (as a color analog of tonemapping), rather than the gamut any given RGB triple is associated with. So maybe "Color primaries" would fit better here.

libass/ass_types.h
TheOneric: fixed, thx

libass/ass_types.h
TheOneric: Indeed “color primaries” seems more appropiate here afaict, thanks!

Issue #744, opened 2024-03-30
[DirectWrite] Does not select the right font when 2 fonts have similar attributes


moi15moi:
### Screenshots

VSFilter
![image](https://github.com/libass/libass/assets/80980684/f8043f40-dda9-431c-8d25-860f8f20f888)
Libass
![image](https://github.com/libass/libass/assets/80980684/2d866f0b-2e1e-4d9a-97b2-701f5a121c74)

### Description of the issue

VSFilter pick up ``SFProDisplay-BoldItalic.ttf``, but libass doesn't find it. You can download the fonts [here](https://github.com/libass/libass/files/14813372/Fonts.zip).

### Step to reproduce (it is very important to respect the order)
1. Install ``SFProDisplay-Bold.ttf``.
2. Install ``SFProDisplay-BoldItalic.ttf``.
3. Select the font by it's fullname (``SF Pro Display Bold Italic``).

Since I am on Windows Desktop, libass call [EnumFontFamiliesW](https://github.com/libass/libass/blob/master/libass/ass_directwrite.c#L900-L901). When libass [CreateFontIndirectW](https://github.com/libass/libass/blob/e51781341493e0312f0c7fee66372f0365f7b962/libass/ass_directwrite.c#L729), it directly use the logfont that the callback give. BUT, the only difference between ``SFProDisplay-Bold.ttf`` and ``SFProDisplay-BoldItalic.ttf`` in the ENUMLOGFONTEXW is ENUMLOGFONTEXW.elfFullName. ENUMLOGFONTEXW.elfLogFont is exactly the same between the 2 fonts ! This is why it pick up ``SFProDisplay-Bold.ttf``, NOT ``SFProDisplay-BoldItalic.ttf``.

### Possible solution
A possible solution is to replace the ``lfFaceName`` with the ``elfFullName`` of the ENUMLOGFONTEXW.

### libass version

9f4e6afeec54c0d240a9b8377f27e6c5b7c2bde0

### Is it a regression?

No

### ASS Sample

```
[Script Info]
; Script generated by Aegisub 9744-Rework-Windows-FontCollector-17a3f487f
; http://www.aegisub.org/
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
YCbCr Matrix: None
PlayResX: 1280
PlayResY: 720

[V4+ Styles]
[...]

PR #745, opened 2024-03-31
directwrite: query GDI-enumerated fonts by full name, not family name

astiob:
Fixes: https://github.com/libass/libass/issues/744

The code is fairly clear, but I guess I want a second opinion on whether ignoring the family name entirely is Good Enough.

### DIFF
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -726,7 +726,21 @@ static int CALLBACK font_enum_proc(const ENUMLOGFONTW *lpelf,
     if (FontType & RASTER_FONTTYPE)
         goto cleanup;
 
-    hFont = CreateFontIndirectW(&lpelf->elfLogFont);
+    LOGFONTW lf = lpelf->elfLogFont;
+    // lf.lfFaceName currently holds the font family name.
+    // The family may contain several fonts with equal numerical parameters
+    // but distinct full names. If we pass lf to CreateFontIndirect as is,
+    // we will miss this variety and merely get the same one font multiple
+    // times. Try to increase our chances of seeing all the right fonts
+    // by replacing the family name by the full name. There's some chance
+    // that this in turn selects some *other* family's fonts and misses
+    // the requested family, but we would rather take this chance than
+    // add every font twice (via the family name and via the full name).
+    // lfFaceName can hold up to LF_FACESIZE wchars; truncate longer names.
+    wcsncpy(lf.lfFaceName, lpelf->elfFullName, LF_FACESIZE - 1);
+    lf.lfFaceName[LF_FACESIZE - 1] = L'\0';
+
+    hFont = CreateFontIndirectW(&lf);
     if (!hFont)
         goto cleanup;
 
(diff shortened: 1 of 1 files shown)


libass/ass_directwrite.c
```diff
@@ -726,7 +726,21 @@ static int CALLBACK font_enum_proc(const ENUMLOGFONTW *lpelf,
...
+    // The family may contain several fonts with equal numerical parameters
+    // but distinct full names. If we pass lf to CreateFontIndirect as is,
+    // we will miss this variety and merely get the same one font multiple
+    // times. Try to increase our chances of seeing all the right fonts
+    // by replacing the family name by the full name. There's some chance
+    // that this in turn selects some *other* family's fonts and misses
+    // the requested family, but we would rather take this chance than
+    // add every font twice (via the family name and via the full name).
+    // lfFaceName can hold up to LF_FACESIZE wchars; truncate longer names.
+    wcsncpy(lf.lfFaceName, lpelf->elfFullName, LF_FACESIZE - 1);
```
moi15moi, line 740: It may be a bit cleaner to do ``wcsncpy_s(lf.lfFaceName, LF_FACESIZE, lpelf->elfFullName, _TRUNCATE);``

TheOneric:
>  I guess I want a second opinion on whether ignoring the family name entirely is Good Enough.

fwiw, I have barely any clue about Windows’ font APIs, but the explanation/motivation seems good to me and it apparently fixes a bug, so afaict LGTM

libass/ass_directwrite.c
moi15moi: > currently our baseline is ISO C99

This information doesn't seem to be written anywhere. It would be nice to have a contributing guide mentioning that.

> afaik all of this code currently still works/can be built for Windows XP

Technically, ``ass_directwrite.c`` is only for Windows Vista SP2 and later, as DWrite doesn't exist before that.

> Are there C11 compilers for Windows XP?

I don't know.

> building for Windows XP still relevant?

[VLC still support Windows XP](https://www.videolan.org/vlc/download-windows.html), so it is relevant.

astiob: It being kinda nonstandard is part of the reason I don’t use it, but to be fair, this file is Windows-specific and it’s existed on Windows for (at least) almost two decades. It’s in Microsoft’s versioned Visual C++ libraries, it’s in UCRT, and it’s also in MSVCRT (so even compiling with 32-bit MinGW-w64 GCC works as I’ve just verified), but I don’t know when exactly it was added to MSVCRT and to MinGW headers.

TheOneric: > > currently our baseline is ISO C99
>
> This information doesn't seem to be written anywhere. It would be nice to have a contributing guide mentioning that.

It’s what our buildsystem sets:

https://github.com/libass/libass/blob/e51781341493e0312f0c7fee66372f0365f7b962/Makefile.am#L3-L6

But admittedly this is not very prominent and buildflags also explicitly enabling GNU extensions is slightly misleading. We enable those since we want to make use of them if available but provide fallbacks if not available.

That being said, if Windows had this function for sufficiently long, using it in a Windows-specific file is absolutely fine.

astiob: > Technically, `ass_directwrite.c` is only for Windows Vista SP2 and later, as DWrite doesn't exist before that.

Not quite. It cannot _run_ on Windows XP, but it can be built on Windows XP and run on Vista+, and binaries built on Vista+ can run on XP without using ass_directwrite at runtime.

astiob: While the code with `_TRUNCATE` does look fairly nice, I feel I remain hesitant to use it for fear of it being absent in older/other compilers/libraries. So I’ll go ahead and merge this as is. But thanks for the suggestion; it may serve as a bit of education and inspiration for future code if/when we upgrade to newer language versions.

PR #746, opened 2024-04-03
Update documentation re subs on HDR

TheOneric:
Revises #735 based on practical experience from mpv and further discussion summarised in https://github.com/libass/libass/issues/297#issuecomment-1991497184

Note:
- ~~the wording implies HDR subs will never be subject to `YCbCr Matrix` colour mangling; this isn’t explicitly stated in the linked comment but was suggested before and seems sensible~~
- this behaviour is now fixed for all current subs and will only change once files opt-in future HDR mechanisms
- this leaves the exact choice of SDR space implementation defined; if *(as some seemed to prefer)* we instead want to choose one single blessed colourspace, [mpv’s current choice](https://github.com/mpv-player/mpv/blob/6179995dd7ef576073e184abd01fd34890e1b59f/video/out/vo_gpu_next.c#L359-L362)  may be a good candidate

### DIFF
--- a/libass/ass_types.h
+++ b/libass/ass_types.h
@@ -232,9 +232,14 @@ typedef struct ass_event {
  *
  * Further note all of the above only concerns the RGB values.
  * Color primaries and transfer charateristics of ASS subtitles
- * must always match their associated video.
- * (This indeed has some undesirable effects on HDR videos,
- * but no mechanism avoiding this is yet standardized.)
+ * must always match their associated video when placed on top of SDR video.
+ *
+ * Subs on HDR video need additional consideration to yield satisfactory results,
+ * but unfortunately no mechanisms for this are standardized yet.
+ * Until such mechanisms are introduced and sub files start to enable them,
+ * all subtitles are to be considered SDR. When placed on top of HDR video
+ * exact color matching is not relevant and the choice of SDR colorspace
+ * to render subs in is left to the presenter.
  */
 typedef enum ASS_YCbCrMatrix {
     YCBCR_DEFAULT = 0,  // Header missing


astiob (approved):
The wording easily seems like an improvement, so +1 👍 But regarding your comments:

> the wording implies HDR subs will never be subject to `YCbCr Matrix` colour mangling; this isn’t explicitly stated in the linked comment but was suggested before and seems sensible

Hmm, I’m not sure about this. I don’t think I intended this in that comment, but I don’t think this has really been discussed/considered much at all and remains, well, unspecified. Maybe we _should_ specify it for SDR-on-HDR at least. I don’t immediately see a reason to specifically exclude it from YCbCr Matrix processing, but what do I know. In fact, specifically for SDR-on-HDR, madshi explicitly suggested that the header _be_ taken into account and also affect the assumed SDR RGB primaries: BT.2020 for BT.2020 matrices, anything older for other values.

> if _(as some seemed to prefer)_ we instead want to choose one single blessed colourspace, [mpv’s current choice](https://github.com/mpv-player/mpv/blob/6179995dd7ef576073e184abd01fd34890e1b59f/video/out/vo_gpu_next.c#L359-L362)  may be a good candidate

FWIW, after the various mentions of sRGB and BT.1886 (2.4 gamma) in the other thread, I’ve ended up confused why mpv prefers one for subtitles but the other for video. If SDR-on-SDR subtitles use BT.1886, why should the same subtitles on HDR video switch to sRGB?… I’ve tried asking in the thread, but my question may have been poorly worded or gone unnoticed. But it hopefully shouldn’t matter much, hence the idea to leave the colorimetry choice up to the presenter.

On the other hand, I imagine WebVTT probably uses sRGB by definition, even on SDR video, through the use of CSS?

TheOneric:
> > the wording implies HDR subs will never be subject to YCbCr Matrix colour mangling; this isn’t explicitly stated in the linked comment but was suggested before and seems sensible
>
> Hmm, I’m not sure about this.

Skimming through the HDR discussion again, requiring `YCbCr Matrix: None` for HDR is mentioned e.g. in https://github.com/libass/libass/issues/297#issuecomment-1988594836 but it still seems to early to actually set this in stone (and requiring an explicit `None` if we decide to, will also make implementations easier rather than making the default value depend on SDR vs HDR)
Reworded to leave it ambiguous.


> On the other hand, I imagine WebVTT probably uses sRGB by definition, even on SDR video, through the use of CSS?

Probably, but I’m not sure. By default CSS uses `sRGB`, but other colourspaces, [e.g. Rec.2020](https://www.w3.org/TR/css-color-4/#predefined-rec2020), are possible and there’s a [CSS draft](https://drafts.csswg.org/css-color-hdr/) for HDR. But then again, WebVTT also aims at support for renderer with no CSS support *(which must deal with a predefined colour subset; all sRGB)* and *limited* CSS support. [Telling renderers with full CSS support](https://www.w3.org/TR/webvtt1/#user-agents-that-support-a-full-html-css-engine):

> However, only a limited set of CSS styles is allowed because [user agents that do not support a full HTML CSS engine](https://www.w3.org/TR/webvtt1/#user-agents-that-do-not-support-a-full-html-css-engine) will need to implement CSS functionality equivalents. User agents that support a full CSS engine must therefore limit the CSS styles they apply for WebVTT so as to enable identical rendering without bleeding in extra CSS styles that are beyond the WebVTT specification.

 I don’t know what this limited CSS subset includes, but presumably not draft extensions.

PR #749, opened 2024-04-19
bitmap: improve shift_bitmap; add vectorized implementation

rcombs:
Followup from #820 (which was originally part of this before being split out).

Rewrites `shift_bitmap` for >20x perf gains. Notable improvements:
- Perform the shift in a single pass instead of splitting into x and y (this also avoids intermediate rounding error)
- Process the shift linearly from top-left to bottom-right, rather than bottom-to-top and right-to-left
- Avoid iterating 2 adjacent rows of the bitmap at once (this confuses the prefetcher); instead use a separate scratch array to store state from the previous row, as in be_blur
- Add ASSUMEs informing the compiler that the stride is 16-byte aligned
- Provide a NEON implementation

The NEON implementation is:
- ~23x as fast as the previous code
- ~2.5x as fast as clang's autovectorized version

This currently only provides C and arm64 implementations; the bitmap_engine code is adjusted to allow for different architectures to have different sets of functions be implemented again. I had some trouble figuring out how this could work on 32-bit x86, since it's extremely SIMD-register-starved; I think in theory it _should_ be doable, but uuuuuuuuugh. Does anyone even care about 32-bit anymore? Anyway, patches welcome, but it shouldn't block merge.

### DIFF
--- a/libass/ass_bitmap_engine.c
+++ b/libass/ass_bitmap_engine.c
@@ -58,6 +58,12 @@
     GENERIC_FUNCTION(mul_bitmaps,  suffix) \
     GENERIC_FUNCTION(be_blur,      suffix)
 
+#define SHIFT_PROTOYPES(suffix) \
+    BitmapShiftFunc ass_shift_bitmap_ ## suffix;
+
+#define SHIFT_FUNCTIONS(suffix) \
+    GENERIC_FUNCTION(shift_bitmap, suffix)
+
 
 #define PARAM_BLUR_SET(suffix) \
     ass_blur4_ ## suffix, \
@@ -102,6 +108,7 @@
     RASTERIZER_PROTOTYPES(16, suffix) \
     RASTERIZER_PROTOTYPES(32, suffix) \
     GENERIC_PROTOTYPES(suffix) \
+    SHIFT_PROTOYPES(suffix) \
     BLUR_PROTOTYPES(alignment, suffix)
 
 #define BASE_FUNCTIONS(align_order_, alignment, suffix) \
@@ -110,7 +117,8 @@
     BLUR_FUNCTIONS(align_order_, alignment, suffix)
 
 #define ALL_FUNCTIONS(align_order_, alignment, suffix) \
-    BASE_FUNCTIONS(align_order_, alignment, suffix)
+    BASE_FUNCTIONS(align_order_, alignment, suffix) \
+    SHIFT_FUNCTIONS(suffix)
 
 
 unsigned ass_get_cpu_flags(unsigned mask)
--- a/libass/ass_bitmap_engine.c
+++ b/libass/ass_bitmap_engine.c
@@ -167,13 +167,17 @@ BitmapEngine ass_bitmap_engine_init(unsigned mask)
     BitmapEngine engine = {0};
     engine.tile_order = mask & ASS_FLAG_LARGE_TILES ? 5 : 4;
 
+    ALL_FUNCTIONS(4, 16, c)
+    if (mask & ASS_FLAG_WIDE_STRIPE) {
+        BLUR_FUNCTIONS(5, 32, c)
+    }
+
 #if CONFIG_ASM
     unsigned flags = ass_get_cpu_flags(mask);
 #if ARCH_X86
     if (flags & ASS_CPU_FLAG_X86_AVX2) {
         ALL_PROTOTYPES(32, avx2)
         ALL_FUNCTIONS(5, 32, avx2)
-        return engine;
     } else if (flags & ASS_CPU_FLAG_X86_SSE2) {
         ALL_PROTOTYPES(16, sse2)
         ALL_FUNCTIONS(4, 16, sse2)
@@ -185,20 +189,14 @@ BitmapEngine ass_bitmap_engine_init(unsigned mask)
             BLUR_FUNCTION(expand_horz, 16, ssse3)
             PARAM_BLUR_FUNCTION(horz, 16, ssse3)
         }
-        return engine;
     }
 #elif ARCH_AARCH64
     if (flags & ASS_CPU_FLAG_ARM_NEON) {
         ALL_PROTOTYPES(16, neon)
         ALL_FUNCTIONS(4, 16, neon)
-        return engine;
     }
 #endif
 #endif
 
-    ALL_FUNCTIONS(4, 16, c)
-    if (mask & ASS_FLAG_WIDE_STRIPE) {
-        BLUR_FUNCTIONS(5, 32, c)
-    }
     return engine;
 }
--- a/libass/ass_bitmap_engine.c
+++ b/libass/ass_bitmap_engine.c
@@ -104,10 +104,13 @@
     GENERIC_PROTOTYPES(suffix) \
     BLUR_PROTOTYPES(alignment, suffix)
 
-#define ALL_FUNCTIONS(align_order_, alignment, suffix) \
+#define BASE_FUNCTIONS(align_order_, alignment, suffix) \
     RASTERIZER_FUNCTIONS(suffix) \
     GENERIC_FUNCTIONS(suffix) \
     BLUR_FUNCTIONS(align_order_, alignment, suffix)
+
+#define ALL_FUNCTIONS(align_order_, alignment, suffix) \
+    BASE_FUNCTIONS(align_order_, alignment, suffix)
 
 
 unsigned ass_get_cpu_flags(unsigned mask)
@@ -177,10 +180,10 @@ BitmapEngine ass_bitmap_engine_init(unsigned mask)
 #if ARCH_X86
     if (flags & ASS_CPU_FLAG_X86_AVX2) {
         ALL_PROTOTYPES(32, avx2)
-        ALL_FUNCTIONS(5, 32, avx2)
+        BASE_FUNCTIONS(5, 32, avx2)
     } else if (flags & ASS_CPU_FLAG_X86_SSE2) {
         ALL_PROTOTYPES(16, sse2)
-        ALL_FUNCTIONS(4, 16, sse2)
+        BASE_FUNCTIONS(4, 16, sse2)
         if (flags & ASS_CPU_FLAG_X86_SSSE3) {
             ALL_PROTOTYPES(16, ssse3)
             RASTERIZER_FUNCTION(fill_generic, ssse3)
@@ -193,7 +196,7 @@ BitmapEngine ass_bitmap_engine_init(unsigned mask)
 #elif ARCH_AARCH64
     if (flags & ASS_CPU_FLAG_ARM_NEON) {
         ALL_PROTOTYPES(16, neon)
-        ALL_FUNCTIONS(4, 16, neon)
+        BASE_FUNCTIONS(4, 16, neon)
     }
 #endif
 #endif
--- a/libass/ass_bitmap_engine.c
+++ b/libass/ass_bitmap_engine.c
@@ -205,6 +205,7 @@ BitmapEngine ass_bitmap_engine_init(unsigned mask)
     if (flags & ASS_CPU_FLAG_ARM_NEON) {
         ALL_PROTOTYPES(16, neon)
         BASE_FUNCTIONS(4, 16, neon)
+        SHIFT_FUNCTIONS(neon)
     }
 #endif
 #endif
--- a/libass/c/c_shift_bitmap.c
+++ b/libass/c/c_shift_bitmap.c
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2024 libass contributors
+ *
+ * This file is part of libass.
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include "config.h"
+#include "ass_compat.h"
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include "ass_utils.h"
+
+
+/**
+ * \brief Shift a bitmap by a fraction of a pixel in x and y directions
+ * expressed in 26.6 fixed point.
+ * Pure C implementation.
+ */
+void ass_shift_bitmap_c(uint8_t *restrict buf, ptrdiff_t s,
+                        size_t w, size_t h,
+                        uint32_t shift_x, uint32_t shift_y, uint16_t *restrict tmp)
+{
+    assert((shift_x & ~63) == 0 && (shift_y & ~63) == 0);
+
+    // Shift in x direction
+    if (shift_x)
+        for (int32_t y = 0; y < h; y++) {
+            for (int32_t x = w - 1; x > 0; x--) {
+                uint8_t b = buf[x + y * s - 1] * shift_x >> 6;
+                buf[x + y * s - 1] -= b;
+                buf[x + y * s] += b;
+            }
+        }
+
+    // Shift in y direction
+    if (shift_y)
+        for (int32_t x = 0; x < w; x++) {
+            for (int32_t y = h - 1; y > 0; y--) {
+                uint8_t b = buf[x + y * s - s] * shift_y >> 6;
+                buf[x + y * s - s] -= b;
+                buf[x + y * s] += b;
+            }
+        }
+}
--- a/libass/c/c_shift_bitmap.c
+++ b/libass/c/c_shift_bitmap.c
@@ -30,29 +30,34 @@
  * expressed in 26.6 fixed point.
  * Pure C implementation.
  */
-void ass_shift_bitmap_c(uint8_t *restrict buf, ptrdiff_t s,
-                        size_t w, size_t h,
+void ass_shift_bitmap_c(uint8_t *restrict buf, ptrdiff_t stride,
+                        size_t width, size_t height,
                         uint32_t shift_x, uint32_t shift_y, uint16_t *restrict tmp)
 {
-    assert((shift_x & ~63) == 0 && (shift_y & ~63) == 0);
+    ASSUME(shift_x < 64 && shift_y < 64 && !(stride % 16) && width > 0 && height > 0 && stride > 0);
 
-    // Shift in x direction
-    if (shift_x)
-        for (int32_t y = 0; y < h; y++) {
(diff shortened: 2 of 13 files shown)


libass/ass_utils.h
```diff
@@ -51,6 +51,14 @@
 
 #define FEATURE_MASK(feat) (((uint32_t) 1) << (feat))
 
+#if defined(__GNUC__)
+    #define ASSUME(x) if (!(x)) __builtin_unreachable()
```
TheOneric, line 55: `__builtin_unreachable` was added in [gcc@468059b](https://gcc.gnu.org/git/?p=gcc.git&a=commit;h=468059bcbb8efdb31fd90369dc2ed122126413fe) and seems to have been first released with 4.5.0 and for clang in [llvm@bf20638966](https://github.com/llvm/llvm-project/commit/bf20638966e205e6dc18451275e5c51264f255e6) released with llvm 2.7.0. For ideal compatibility we’d need to use the following somewhat unwieldy check:

```C
#if (defined(__GNUC__) && ((__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5))) ||
     (defined(__clang__) && __has_builtin(__builtin_unreachable))
```

While clang [got `__has_builtin`](https://github.com/llvm/llvm-project/commit/b6f77af5327f411aa07b81b25b85b016113dace5) before `__builtin_unreachable`, gcc unfortunately only [gained it a decade later](https://gcc.gnu.org/git/?p=gcc.git&a=commit;h=ad1539d5555a161cf6851de8995641d6dfe792d9).

Given this is about optimisation though and even gcc supports `__has_builtin` for around 4 years, stubbing out `__has_builtin` for older compilers and using that for `defined(__GNUC__) || defined(__clang__)` might be fine too

rcombs: Done (using `__has_builtin`).

astiob:
Ah, if this is still blocked on something (32-bit x86 asm), can we merge the non-blocked parts in the meantime?

rcombs:
We can; opened #820.

rcombs:
This is now ready to review. Smaller components were split out into #820 and merged, so this is now just the shift_bitmap rewrite and infrastructure to allow permit C and NEON implementations to merge without an x86 one.

MrSmile:
Note that shifting can be done without a temporary buffer, at the cost of additional multiply per pixel ([see here](https://github.com/MrSmile/libass/commits/shift-tests/)), I'm not sure which approach would be faster.

libass/c/c_shift_bitmap.c
```diff
@@ -0,0 +1,63 @@
...
+                        size_t width, size_t height,
+                        uint32_t shift_x, uint32_t shift_y, uint16_t *restrict tmp)
+{
+    ASSUME(shift_x < 64 && shift_y < 64 && !(stride % 16) && width > 0 && height > 0 && stride > 0);
+
+    for (size_t y = 0; y < height; y++) {
+        uint16_t prev = 0;
+        for (size_t x = 0; x < width; x++) {
+            uint16_t px = buf[x] << 1;
+            uint16_t tmpv = tmp[x];
```
TheOneric, line 43: very much a nit: the horizontal cross-pixel contribution is called `prev` and the vertical contribution `tmpv`. More cohesive names like `prevx`/`prevy` or `tmph`/`tmpv` might be better.

libass/c/c_shift_bitmap.c
```diff
@@ -0,0 +1,63 @@
...
+
+
+/**
+ * \brief Shift a bitmap by a fraction of a pixel in x and y directions
+ * expressed in 26.6 fixed point.
+ * Pure C implementation.
+ */
+void ass_shift_bitmap_c(uint8_t *restrict buf, ptrdiff_t stride,
+                        size_t width, size_t height,
+                        uint32_t shift_x, uint32_t shift_y, uint16_t *restrict tmp)
```
TheOneric, line 35: nit: splitting moving and improving perf into two commits would have been nicer imo and more clearly shows *what* changed

libass/ass_bitmap_engine.c
```diff
@@ -185,20 +200,15 @@ BitmapEngine ass_bitmap_engine_init(unsigned mask)
...
         }
-        return engine;
     }
 #elif ARCH_AARCH64
     if (flags & ASS_CPU_FLAG_ARM_NEON) {
         ALL_PROTOTYPES(16, neon)
-        ALL_FUNCTIONS(4, 16, neon)
-        return engine;
+        BASE_FUNCTIONS(4, 16, neon)
+        SHIFT_FUNCTIONS(neon)
```
TheOneric, line 208: nit: this could use `ALL_FUNCTIONS` making it clear no assembly routines are missing atm

TheOneric (commented):
i didn’t look at the aarch64 assembly in detail and everything else seems fine to me apart from some inconsequential nits

But before merge, we’d ideally benchmark the “extra temporary buffer” and “extra multiply” versions on a variety of common OS and CPU environments to figure out what performs better in practice

libass/c/c_shift_bitmap.c
rcombs: Done; this makes benchmarking simpler as well.

libass/ass_bitmap_engine.c
rcombs: I kinda prefer to be explicit about this until functions get added to the BASE set.

libass/c/c_shift_bitmap.c
rcombs: How about `shifted_from_top` and `shifted_from_left`?

rcombs:
Note that if the malloc ends up taking up a meaningful amount of time (I don't expect it to; allocators are generally pretty good at handling frequent alloc+frees of smallish buffers by the same thread), I think we'd want to solve that by moving `BitmapEngine` into `RenderContext` and adding `scratch_array` + `scratch_size` members, letting them grow as needed, and using them for both this and `be_blur`.

libass/c/c_shift_bitmap.c
TheOneric: seems good

libass/c/c_shift_bitmap.c
```diff
@@ -30,29 +30,34 @@
...
-                buf[x + y * s] += b;
-            }
-        }
+    ASSUME(shift_x < 64 && shift_y < 64 && !(stride % 16) && width > 0 && height > 0 && stride > 0);
+
+    for (size_t y = 0; y < height; y++) {
+        uint16_t shifted_from_top = 0;
+        for (size_t x = 0; x < width; x++) {
+            uint16_t px = buf[x] << 1;
+            uint16_t shifted_from_left = tmp[x];
```
TheOneric, line 43: `shifted_from__(top|left)` names are switched

libass/ass_bitmap_engine.c
TheOneric: i don’t feel strongly about this; feel free to ignore

libass/c/c_shift_bitmap.c
rcombs: Whoops, must've happened during one of the find/replaces (I went through a few options). Fixed.

TheOneric:
> But before merge, we’d ideally benchmark the “extra temporary buffer” and “extra multiply” versions on a variety of common OS and CPU environments to figure out what performs better in practice

Long overdue, but here are benchmarks for x86_64 (Zen+) on glibc-Linux on a single sample both with a realistic (`24`) and an inflated (`240`) FPS value. Results don't paint a clear picture though; which one is faster in the end depends on the used compiler; notably the approach in this PR ends up not being much better or within the uncertainty of current master if compiled with gcc.

With gcc, MrSmile’s approach with an extra multiply performs best by a fair bit; with clang the approach from this PR performs better both relatively and (by a small margin) also absolutely compared to all other results from both compilers.  
Depending on the FPS margins and ordering shuffles a bit, but the above holds true in both.  
*(Note: the relative performance between the compilers cannot be generalised; in other samples from e.g. the `fix_outline` PR GCC performed consistently better on x86_64 before)*

Full results and details: [results.txt](https://github.com/user-attachments/files/19216008/results.txt)  
*(obtained using `hyperfine -w 7 -m 25 -N ...`)*

While this likely doesn't help too much with compiler-dependent changes, perhaps caching the alloced buffer will be enough to always tip the scales in favour of this approach?

MrSmile:
Hmm... looks inconclusive without complete assembly implementations. Maybe I should look into generated assembly...
Can you check allocation impact for your samples (by hardcoding large enough static buffer, for example)?

TheOneric:
It appears to slightly improves things on glibc- and musl-Linux though the difference between the means is within stddev for both.

This are the new results for glibc-Linux again; quite similar to the ones before: [results2.txt](https://github.com/user-attachments/files/19257588/results2.txt)


And the same set of tests on musl-Linux with newer compiler versions: [results2_alpine.txt](https://github.com/user-attachments/files/19257589/results2_alpine.txt)  
Here for 24FPS the extra-multiply version with gcc outperforms the version from this PR with clang just barely beyond the uncertainty range making it the top performer. On 240 FPS this PR, this PR without alloc and extra-multiply are all within uncertainty range if compiled with clang and effectively tied for best performer.

MrSmile:
So, allocation impact is negligible (but I think there were cases with slow alloc implementation?) and `shift_bitmap` is memory bound, so impact of extra multiply is also negligible (but uncertainties much larger in this case). Looks like I should reanimate my branch and add assembly to it.

TheOneric:
To recap: at normal FPS the new C approach here has a small lead when it is compiled with clang, but when compiled with gcc is notably worse then the proposed alternative compiled with either gcc or clang.  At inflated 240FPS this approach and clang in general always perform better than gcc on the single investigated test sample and machine.  
Either approach performs better than the pre-patch C version and ofc the actual assembly is even better.

If someone wants to flesh out an alternative approach in the near future, it would make sense to wait for that. But if not imho this should be fine to merge as is and further C tweaks and x86 assembly can follow later.

Andarwinux:
https://godbolt.org/z/3jcoWch5P

gcc just cannot vectorize it at all, yet clang can vectorize it even without ASSUME.

TheOneric:
Indeed, while the alternative version appears to have some vectorisation in both compilers: <a href="https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gWwA4EsA2ApgE4A0uIEGOBJaq%2BAhgHbABMbAjAAykBmBAC4kAzpQBGIAOTcppcbmaNiAT2mc5CpaoDy4gFaFUg9XOqZCzQTub41MuQBNC4gK7AAwo3z4xUjaTOmCzARNKygbjERoK4AG6EfhGEAB5GrsKm5FaE%2BOFy%2BLjixMoqHujO%2BaSCxLiYWQnE4ugihAAihMGshaz5AJT8TMAiAGq4hADuOtiWWXzoVgDKqN6EIJwALPwLggCqrQAKKVX6IGykTL0A%2BjeoF0VikAPo2LELfgC0OmwA1B/BxFQSAApABmNoAL2YjQArHJ0I1as5HhABq0iMZcAtKJZHOV8K5MMx1qRcQAZRSEAByhPEdE4pGwLVwb2Y%2BMJxIZTJELKxzApzGptPppHRMT5i0EykE7KJJLFmIWkulAqFmDpxHlUuIMvQBLlDJE2sEqpp6vpaPQrkBazYAzyUjhpGJDlI6GkPxSAA4AGwfH0bH4MEI/DgAOk4Ye4PwgAFFHCz0MQfsC2KC7XITA4%2BgMkIRGM5NajSABrEAw3gOjZyF0Rd1SORiXhZ2QDOCwFDUPBEMgUKhYbt0VDATg%2BhkCfDCYhiCCSV1aUpZBe6AwxLLmSzWWz2CLONyeby%2BLJBEJhV0J6KYhJJOSpdKZfxyRTCB0RQrFUrlSqump1BokZpWg6LpQkUUBs0GRhhjGSZplmR9tiWFYwk2RC9kOY5XVOUELhCEAbiuO53xRZ5Xj5T5vj%2BAEgTBSFoRIGE/j4BJjA%2BDAjX%2BCpcloxwVCUTBcFQeFEVwZF4DRXJxWxCBcVlYkcPJSkzQ1dZuC9RlmVZOSQBw7leQWU1hU1Hh1IVVllR1bScLMiVjUM81jLU0VjSs5yVSUozVK9S1rVQW17WkJ1a3hD1vT9AMfmAVBUB%2BUcIxjeNE2TVN0z6TN%2BlzfNC0oAYywrC5pGrUh6h4XhgrdaRGxAZsMoKqQ2DkeoAHZuDDH0vQ2bhQQATjUpruo2JqvXU8r63SiD22QNAuMSwQk3ISgJ0sRgLDmHZllWdYtnmKx9kII4TjOXDrluUVJMVYkZOYPE9Q5EAmtJa77JUjZ1L0rTbrlB73r5Z66Fes6MXMlzPuJB6bKVOyPIckAAaNaVtPBqHBWU/7vNFK0bWkFLFAYVxnBTUEPCNRxnD4MMQVBWNgW4ABBHHmDxgmwWJwQEysCmwWpumUrJykflpxZFl2ABZWMIBSPoflwPgYzASBJalm43CERQrlcZhokYIFGHEIhURp2m4nQUSfkYEQRCuEQkBlwQrgUQRgmwQiIFcZ8vSuQQfgAKmiI1amMH43D4UgfleYgEz4PhPZ%2Bf3RMIUhDZ%2BZOU9TtP05TnkIUIGOJlEwQkFDrOc69vNcGAJBBETumM9ruu3asdMY%2Bt22rhSUOG8EJuvZbvg7ZUDvn1HGPfcSX9A8d7ApcN4EmoAISTlPBeFsWIF7u2UkJjwfgi1MfT32Obb7q4VC3nfAz3g/5bX39mbYRjR2nth9%2Bfn488cAvCdjH5o0v1%2By4rl7LmP8UzPwPnHZmVMf59DBAvHmNcU7zGTGvXA2cY6n1oj/WBPwMFEx%2BAAyu2CVCpgXmwOe0956L3Tp3YePcj7CEcFcPgxAsBXCIH3QmbQsGgjgbTWuSCYzFxjpvTB3BsEiLwe/Au4iSEkIobwuuKcaE%2BhjtgCRXDg7AhhHPFIWiuEsxZrFWBVD65DxUXQ22hBGHMNYXNbAnCfiTy0TovRxj4F8MUT8ZRMdxBtwcWon2h9W4SOplA/ePCZ4IMUQE2R4Igl9ysUwlhmA2GED7m4jx0TN4fEwb43RET3GeK8WYnxJ9/Gb29vE/uX9gHhN4SY2uMSyGYPXokmxKS7EZKKQEnJcTfHEIKfTKJddNHaN0TCfRcSIBNLnrFaeVNgGcAyQ0jOrTrHJNSRw3JbcumeKcWMvRDj%2Bm7PTrPNokTMlp2DqAhecSIGEBOSmJq5yebPIuTPNMuMCSQK8BbEgghOYLPcYIFQMwyaxxqK4QOs8FEp2fN3H47Cq6OJeI8%2BFbBc6h0prM2uAB6XFb986F3wYQcuhDhnh0jtHHut8HmDNTp3D2XtvbBz4CQbBad8X4MCfcoOrgo7steVwueLInbLLpsbU25tLbr3tqKxgztMDTgEvrEVjsFU%2B3EJgUOz4qltx1VYPVKgYGvNhcnfF0r/lQFlSI1%2BAA/H0oJ5nnLib/MBr8170LKQfB1TrOGiJgfSpeQtRbi09cE0BPofi%2BudaIyNB9w3Hwwfax1sbXWBvqcMmWcswBat6bGVlJAnkeFntva%2BNri1GpzCs5O0RBDWmYOKy56Lc5HMwPmiYWK235uxSYylMtqWx27Vze5jzGUj2ubk9tXNC3ECbanEpPdcFcNlbgkteDuqPIERAFtXtl1GJ4Tgs%2B2KcGyLIfImtPxt27s9A4pZh6JHbwmDI5p56nlmrrouxxOBOAOIipU0ZOi/hGImcB2c/LnHZJA1wj4fK%2BCQb0VLSpO6rC0Kljax5tcv2T1%2BJg/9cHIM3KHbBpZoHYPgfg2MojIhgOkZgwRqjJDY6IcCShwQaG9X5I/bXQDUG6MOLYxxiAk9f3kcXVLKA4mYy/GQyJ4DOGcysd3eh3BoTv6PyI6lGpYTMMpzOZe7DP6/2BgAxB7RExaOHNg4BizJHDnIakxhoN6dDPYFw3E/DNnqOWbIwxueFmmM0bs6BhzqHzHoa9Vxy9XnguTK4YJ8LMY5NibC57CTCW0vSdY8l79bnFOha7hiwQKntPqZ9E/WZWmua1N08nSdzS7mPP08M69jciu3tEeIs%2Bz6H1nvIe%2BgzqWvZybwyZvzfGrMxl4z5%2BjvGWMFY4057jKdpuxYE1J4TRmUvscSxAGTuXf0Fe7iV6rUCNNMaqwsnTzmnkvObUNg7xnAkxeg8Bl7/GFuJaWyY97hzMEZeK0lrbMYNv7bk0dorJ2rtlYqz8S7amd7LOeVIAKjoaxVHrNvEQmM/KgPTONVsmUCx0GLFWRqZxuphlBJwGE3UfTcG6pwNg3AYRcHOKNSqopqrVFqg6UE6PXRjR59mAYjQeTYg2EAA%3D">https://godbolt.org/z/x5PhcdoMc</a>

But as mentioned, _either_ version is already doing better than current master and we can still improve it further later on if nobody steps up to already improve it in the immediate future.

MrSmile:
I've added x86 assembly to my [test branch](https://github.com/MrSmile/libass/commits/shift-tests/). I'll create PR after writing aarch64 variant for it.

Andarwinux:
Could you also add a avx512 version? Since clang now generates good auto-vectorized code, asm may likely slow things down for avx512 users.

rcombs:
We don't have any AVX512 code right now, and vectorized `shift_bitmap` is memory-bound, so I'd expect an AVX512 implementation of just this one function to end up causing net performance regression due to downclocking. Most builds can't target AVX512 for autovectorization, since that's a pretty tight limitation on deployment target hardware; for our own perf target purposes, I think it's best to assume that the only AVX code running in libass will usually be our own assembly (though other AVX might exist within the process from eg ffmpeg).

Andarwinux:
Never mind.

```
$ ./checkasm_512 -b -c --test=shift_bitmap
checkasm: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz (000806D1) using random seed 2879136087
 - shift_bitmap.shift_bitmap [OK]
checkasm: all 1 tests passed
nop: 20.2
shift_bitmap_c: 190.9

$ ./checkasm_256 -b -c --test=shift_bitmap
checkasm: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz (000806D1) using random seed 700295472
 - shift_bitmap.shift_bitmap [OK]
checkasm: all 1 tests passed
nop: 20.0
shift_bitmap_c: 295.1

$ ./checkasm_asm -b -c --test=shift
checkasm: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz (000806D1) using random seed 1579670955
 - shift.shift [OK]
SSE2:
 - shift.shift [OK]
SSSE3:
 - shift.shift [OK]
AVX2:
 - shift.shift [OK]
checkasm: all 4 tests passed
nop: 20.2
shift_c: 1622.3
shift_sse2: 248.0
shift_ssse3: 211.2
shift_avx2: 117.4
```

But shift-tests branch's C implementation looks really bad for clang.

MrSmile:
> But shift-tests branch's C implementation looks really bad for clang.

Looks like clang produces dword-vectorized assembly despite all `(int16_t)` casts against it. GCC correctly uses only word instructions. Dunno how to fix that, I'm open for suggestions.

rcombs:
Simply changing the function's arg types to `int16_t` seems to improve the output quite a bit in godbolt, though I haven't benchmarked.

Andarwinux:
> > But shift-tests branch's C implementation looks really bad for clang.
> 
> Looks like clang produces dword-vectorized assembly despite all `(int16_t)` casts against it. GCC correctly uses only word instructions. Dunno how to fix that, I'm open for suggestions.

reported upstream https://github.com/llvm/llvm-project/issues/206379

PR #751, opened 2024-04-20
Various CI and build tweaks

TheOneric:
- automatically use all available cores
- add muon CI
  - muon already fixed e.g. the false-positive warnings we noticed earlier and currently is also tweaking formatting rules. The latter means our buildfiles no longer match muon style, but let’s wait for 0.3.0 before deciding whether to care at all or reformat and make this an actual fail
- ensure meson build works from tarballs
- get rid of unneeded `CONFIG_` macros in both builds
- share library and release version across buildsystems via a toplevel `RELEASEVERSION` file *(exact name and placement can be changed)*


.github/workflows/meson.yml
```diff
@@ -102,3 +113,113 @@ jobs:
...
+        run: |
+          samu -C build
+
+      - name: check binary works
+        run: |
+          build/test/libass_test /tmp/out.png compare/test/sub1.ass 0.5
+
+      - name: lint style (informational)
+        run: |
+          # `muon fmt -q` already let’s us check _if_there’d be changes, but
```
astiob, line 163: (typo in `lets` and missing space after `_if_`)

astiob (approved):
idk meson, lgtm

.github/workflows/meson.yml
TheOneric: fixed, thx

PR #754, opened 2024-04-27
build: add ELF symbol versions

jengelh:
Building ffmpeg with libass-0.17.0 only encodes "libass.so.9" into ffmpeg. This is a problem when the ffmpeg binary is then taken to run on a system with just libass-0.14.0 (unknown symbol ass_track_set_feature).
    
Adding ELF symbol versions makes ld add extra information to ffmpeg that then e.g. rpm can use to specify additional dependencies and prevent installation of an unworkable combination. ld.so also emits a better error message that the wrong libass version is present.

(Haven't tried the Win32 build, I just have seen to it that gen_defs.py generates the same as before.)

### DIFF
--- a/libass/libass.sym
+++ b/libass/libass.sym
@@ -1,46 +1,54 @@
-ass_library_init
-ass_library_done
-ass_library_version
-ass_set_fonts_dir
-ass_set_extract_fonts
-ass_set_style_overrides
-ass_renderer_init
-ass_renderer_done
-ass_set_frame_size
-ass_set_storage_size
-ass_set_margins
-ass_set_use_margins
-ass_set_aspect_ratio
-ass_set_font_scale
-ass_set_hinting
-ass_set_line_spacing
-ass_get_available_font_providers
-ass_set_fonts
-ass_render_frame
-ass_new_track
-ass_free_track
-ass_alloc_style
-ass_alloc_event
-ass_free_style
-ass_free_event
-ass_process_data
-ass_process_codec_private
-ass_process_chunk
-ass_read_file
-ass_read_memory
-ass_read_styles
-ass_add_font
-ass_clear_fonts
-ass_step_sub
-ass_process_force_style
-ass_set_message_cb
-ass_fonts_update
-ass_set_cache_limits
-ass_flush_events
-ass_set_shaper
-ass_set_line_position
-ass_set_pixel_aspect
-ass_set_selective_style_override_enabled
-ass_set_selective_style_override
-ass_set_check_readorder
-ass_track_set_feature
+V_0.13.6 {
+	global:
+	ass_library_init;
+	ass_library_done;
+	ass_library_version;
+	ass_set_fonts_dir;
+	ass_set_extract_fonts;
+	ass_set_style_overrides;
+	ass_renderer_init;
+	ass_renderer_done;
+	ass_set_frame_size;
+	ass_set_storage_size;
+	ass_set_margins;
+	ass_set_use_margins;
+	ass_set_aspect_ratio;
+	ass_set_font_scale;
+	ass_set_hinting;
+	ass_set_line_spacing;
+	ass_get_available_font_providers;
+	ass_set_fonts;
+	ass_render_frame;
+	ass_new_track;
+	ass_free_track;
+	ass_alloc_style;
+	ass_alloc_event;
+	ass_free_style;
+	ass_free_event;
+	ass_process_data;
+	ass_process_codec_private;
+	ass_process_chunk;
+	ass_read_file;
+	ass_read_memory;
+	ass_read_styles;
+	ass_add_font;
+	ass_clear_fonts;
+	ass_step_sub;
+	ass_process_force_style;
+	ass_set_message_cb;
+	ass_fonts_update;
+	ass_set_cache_limits;
+	ass_flush_events;
+	ass_set_shaper;
+	ass_set_line_position;
+	ass_set_pixel_aspect;
+	ass_set_selective_style_override_enabled;
+	ass_set_selective_style_override;
+	ass_set_check_readorder;
+	local:
+	*;
+};
+V_0.16.0 {
+	global:
+	ass_track_set_feature;
+} V_0.13.6;
--- a/configure.ac
+++ b/configure.ac
@@ -9,6 +9,7 @@ AC_CONFIG_HEADERS([config.h])
 AC_PROG_CC
 AM_PROG_CC_C_O
 AM_PROG_AS
+AC_PROG_EGREP
 
 # Checks for header files.
 AC_CHECK_HEADERS_ONCE([iconv.h])
@@ -47,6 +48,8 @@ AC_ARG_VAR([FUZZ_LDFLAGS],
 AC_ARG_VAR([FUZZ_CPPFLAGS],
     [If fuzzing program is enabled, set this to select alternative modes; see fuzzer source for options.])
 FUZZ_CPPFLAGS="${FUZZ_CPPFLAGS:--DASS_FUZZMODE=0}"
+
+AM_CONDITIONAL([WITH_GNU_LD], [test "$with_gnu_ld" = yes])
 
 # Checks for available libraries and define corresponding C Macros
 # Start with system libs, then check everything else via pkg-config
--- a/gen_defs.py
+++ b/gen_defs.py
@@ -1,12 +1,14 @@
 #!/usr/bin/env python3
 
 import sys
+import re
 
 symfile_path = sys.argv[1]
 deffile_path = sys.argv[2]
 
-with open(symfile_path) as symfile:
-    lines = symfile.readlines()
-    lines.insert(0, 'EXPORTS\n')
-    with open(deffile_path, 'w') as deffile:
-        deffile.writelines(lines)
+with open(deffile_path, 'w') as deffile:
+    deffile.write("EXPORTS\n")
+    for line in open(symfile_path, 'r'):
+        a = re.search('(ass_\w+)', line)
+        if a is not None:
+            deffile.write(a.group(1) + "\n")
--- a/libass/Makefile_library.am
+++ b/libass/Makefile_library.am
@@ -65,7 +65,16 @@ libass_libass_la_SOURCES += libass/ass_coretext.h libass/ass_coretext.c
 endif
 
 libass_libass_la_LDFLAGS = -no-undefined -version-info $(LIBASS_LT_CURRENT):$(LIBASS_LT_REVISION):$(LIBASS_LT_AGE)
-libass_libass_la_LDFLAGS += -export-symbols $(top_srcdir)/libass/libass.sym
+if WITH_GNU_LD
+libass_libass_la_LDFLAGS += -Wl,--version-script=$(top_srcdir)/libass/libass.sym
+libass_libass_la_DEPENDENCIES = $(top_srcdir)/libass/libass.sym
+else
+libass_libass_la_LDFLAGS += -export-symbols libass_redux.sym
+libass_libass_la_DEPENDENCIES = libass_redux.sym
+endif
+
+libass_redux.sym: libass.sym
+	$(EGREP) -Eo 'ass([[:alnum:]]|_)+' $@
 
 assheadersdir = $(includedir)/ass
 dist_assheaders_HEADERS = libass/ass_types.h libass/ass.h


TheOneric (commented):
Hi, could you help me understand and elaborate a bit more on which problem this solves and why symbol versioning is used specifically? So far i’ve only encountered symbol versioning in the context of allowing to retain multiple versions of a single interface to preserve backwards ABI compatibility. While looking it up again now, this again seemed to be them main concern main benefit. I doubt we’ll ever (have a need to) do this.

In theory all the info is already here, `libass.9.2.1` can be used by anything linked against `libass.9.x.y` where `x ≤ 2` and anything linking against `libass.9.2.1` can later use any `libass.9.x.y` with `x ≥ 2`.  
I was told before MacOS’ dynamic linker actually uses minor library versions, but indeed on Linux unfortunately only the major version is taken into account. For recording dependencies in RPM however this shouldn’t be a problem; how does symbol versioning help here?

You mentioned version mismatches when shipping a dynamically linked ffmpeg, but what distinguishes libass here from other ffmpeg dependencies. Checking e.g. `libdav1d.so` and `libx264.so` on my system, they also don't use any symbol versioning.
In fact, while ffmpeg libraries themselves do set _a_ symbol version, they appear to jsut indiscriminately set the version of all symbols to `LIB<NAME>_<MAJOR_LIB_VERSION>` providing no benefit afaiu at all over the already encoded major version of the library itself

jengelh:
>In theory all the info is already here, libass.9.2.1 can be used by anything linked against libass.9.x.y

The theory yes. In practice though, the entry is for `libass.so.9` and *there is no record of the x and y values*, which is the gap that symbol versions fill.

>libdav1d.so, libx264.so and libzvbi.so

dav1d has the same issue. I do not go actively looking for these problems, just reacting to [bug reports](https://build.opensuse.org/package/show/multimedia:libs/ffmpeg-7#comment-1929558).
zvbi: same issue, but zvbi has not had a significant API change since about 2008 (so no one could have potentially noticed anything).
x264: does not use minor or micro numbers, and always bumps major. That actually works very well.

>In fact, while ffmpeg libraries themselves do set a symbol version, they appear to jsut indiscriminately set the version of all symbols

which is also wrong, which was reported, and they don't give a fuck. Which…is an option too, though the penalty is that a package then gets the harsh treatment of forced lockstep upgrades in the distro.

>almost all of those symbols are not correctly attributed, not all existed n 0.13.6 and many existed before

that does not matter because it only needs to go back as far as the last new SONAME, and 0.13.6 was the first to produce `libass.so.9`.

Yeah one global: is missing...

rcombs:
Note this is currently not building on macOS or with tcc.

I'm a little confused; generally speaking, people don't expect a binary built against a new version of a shared library to work when run against an older minor version of the same library. As far as I can tell, this change wouldn't make that work either; it'd just result in a different error message when it breaks (or move the error from runtime to install-time, if the package manager happens to check this)? This seems like a lot of infrastructure for very minor (and very situational) benefit. Is there documentation somewhere on how rpm does this? I'm struggling to understand how it's supposed to get any useful information out of this version data that isn't already provided simply by the presence or absence of a symbol.

TheOneric:
> that does not matter because it only needs to go back as far as the last new SONAME, and 0.13.6 was the first to produce libass.so.9

Hmm, there were more ABI and API additions since, but they weren’t functions so fair enough

> x264: does not use minor or micro numbers, and always bumps major. That actually works very well.

While this prevents Linux’ typical linker with an incomplete version check to ever load a mismatching library, i don’t think that’s a “very well” model; it forces unnecessary recompiles and needlessly breaks compatibility with old binaries.

> dav1d has the same issue. I do not go actively looking for these problems, just reacting to [bug reports](https://build.opensuse.org/package/show/multimedia:libs/ffmpeg-7#comment-1929558).

Furthermore, none of libass’ own dependencies does any symbol versioning, neither required not optional dependencies. 

I still don’t understand which benefit this will offer over the standard, full library version. Can you be more elaborate?  
As written before, just because the dynamic linker doesn’t check the full version shouldn’t preclude packaging tools from checking and recording the full library version. Given symbol versioning appears to mainly exists to allow multiple versions of the same symbol, it’s specific to certain platforms and barely anything uses symbol versioning, it seems like an odd choice to rely on for packaging.

Symbol versioning would not suddenly make the ffmpeg binary from the linked bug report work either.

TheOneric:
> there were more ABI and API additions since, but they weren’t functions

but: if non-functions aren’t versioned, versions mismatches can _still_ occur. How would you solve this when relying on symbol versioning? *(while just correctly parsing the full library version already ensures everything will work)*

Furthermore, relying on symbol version to correspond to the public package release version is fraught. It will break in one way or another when multiple symbols are added between a package release but not at the same time. In libass we (at least over the last few releases) try to always adapt the library version immediately when ABI/API changes to ensure indev builds are linked to and work correctly. E.g. FreeType goes even further and also changes it’s pkgconfig version on each ABI/API change. Being coarser in symbol versions would probably not be acceptable.  
This means, packaging must already retain some metadata to map symbol versions to release versions. (Or alternatively metadata about all contained symbols(+version), but if that’d be the case the linked bug qouldn’t have happened).  
If such metadata is already tracked, why not just track the library version in the first place making everything simpler _and_ actually working with most libraries?

jengelh:
>but: if non-functions aren’t versioned

In ELF, data objects and other funky stuff can be versioned too, cf. `readelf -aW libc.so.6 | grep @@ | grep -v FUNC`.

Some software authors might also choose to add explicit getter functions and not expose data objects directly. That has benefits like being in a position to keep an old ABI and potentially convert whatever representation a library has internally to the return type of the unchanged getter function.

>relying on symbol version to correspond

The group names in version-script files can be chosen arbitrarily (like SONAMEs). Uniqueness is the only constraint.

TheOneric:
> not expose data objects directly

libass exclusively exports functions; check `readelf -sW /path/to/libass.so.9.2.1 | grep -v ' UND '`

But structs can gain new members and struct members of return types do not show up n ELF symbols at all. Afaik it’s impossible to version them.
Those structs do not occur in arrays and their `sizeof` is explicitly documented to be subject to change *(this is not unusual and also used in FFmpeg)*. When a new member is added, the minor library version increases ensuring (when honoured) anything possibly accessing those new members can't use old library versions without those members which would cause an OOB read. At the same time, programs linked against older versions never accessing those new members continue to work fine with newer library builds.

If now you ignore the library version and just read symbol versions there’s no way to detect this and OOB accesses and (in the best case) crashes will occur as mentioned before.

I’ll not repeat all questions all over again, but please actually answer everything this time.

jengelh:
>But structs can gain new members and struct members of return types do not show up n ELF symbols at all. Afaik it’s impossible to version them.

Correct. One way to approach that is to not expose the struct layout to programs. openssl3 does that for e.g. EVP_MD_CTX_new().

>packaging must already retain some metadata to map symbol versions to release versions [...]
>If such metadata is already tracked, why not just track the library version in the first place

Aha, but such metadata isn't *guaranteed* to be tracked. Though I mentioned rpm at some point, the packaging environment might be something else. Alpine Linux seems to love tar files. This is why I feel it is important that software projects themselves encode dependencies using ELF features as far as is possible.

>why specifically symbol versioning is used despite its lack of broad adoption

If a car analogy can help me ("why do people drive EVs despite lack of adoption"), then I'd say: it's used because people believe in doing "The Right Thing". Perhaps https://docs.oracle.com/cd/E19683-01/816-5042/solarisabi-8/index.html and https://docs.oracle.com/cd/E26505_01/pdf/E26506.pdf (page 227+) can give you more answers.

But if the question was "why is broad adoption not occurring despite symvers being available", then that's probably the usual reason (few people are motivated enough to do the Right Thing).

astiob:
Could you recap what the goal of this is? I feel misunderstandings about the goal may be the source of the confusion here.

From what I can see:

* Is the goal to prevent running a newer consumer binary on a system with an older libass? But that’s already achieved by symbols missing (on minor updates) or the soname changing (on major updates), and versioning the symbols makes no difference.

* Is the goal to improve the error message when this happens on a minor version mismatch? Personally, I’m not sure the improvement is significant, but I guess this is arguable.

* Is the goal to allow a single libass binary to support multiple consumers that target different ABI versions? This is what symbol versioning is designed for as far as I understand, but this would require libass to actually maintain multiple ABIs in working order simultaneously (which I don’t think we want to do and I’m not sure we _can_ do); just labelling the symbols with the current version wouldn’t be enough.

* Is the goal to improve the error message when the mismatch happens on a major version, e. g. when our public struct layout changes? The idea being that symbol versioning would allow the soname to stay unchanged and just the symbols re-versioned? (If neither the soname nor the symbols were to change, then the ABI break wouldn’t be signalled at all.) Again not sure the improvement is worth it, plus this would introduce additional cross-platform complexity, make the ABI breaks less visible in the file system, and actually make it impossible to have multiple major ABI versions of libass coexist on the same system. This PR doesn’t seem to do this, anyway.

* Is the goal to allow major ABI changes to be less major? If so, how does this help?

    Side note: we’re aware that public structs are bad for ABI and accessors are preferable, but unfortunately we can’t retroactively do this: public structs _already exist_ in our API, and removing them would require both an ABI break and an API break, such that every consumer that does currently use these structs would have to rewrite their code (and either support only newer libass or maintain two versions of the same code).

jengelh:
The goal is to avoid a situation where a package manager allows the system package set to contain an improper mixture of libass and programs, a situation which occurs because tools have *no rules* by default. This expectation:

>generally speaking, people don't expect a binary built against a new version of a shared library to work when run against an older minor version of the same library

is not shared by tools. In other words, whoever maintains mplayer.spec/mplayer.dsc must manually enter into the build recipe that libass is operating on the "people expectation" principle. Though that's a one-time thing, it has to be done for *every build recipe* file in the distro which uses libass, and then repeat for *every distro* that possibly exists. That does not scale very well, agree?

One use of symvers is to convey that very expectation, and it's better than just writing it out in a README, because few people, and even less so tools, read READMEs.

astiob:
Do package managers automatically verify symbol versions?

Or are you just hoping that adding symbol versions would make the issue more obvious to a human package maintainer? I’m not sure I see it: you point out yourself that semantic versioning is the human expectation already; Libtool has mandated it for decades.

jengelh:
>Do package managers automatically verify symbol versions?

Yes.

>Libtool has mandated it for decades.

Be that as it may, rpmbuild/dpkg-buildpackage don't know that any particular file was built with or without libtool (it's not leaving any traces in ELF), so they can't autodetect that they possible should carry forward an implied semver logic. Symvers aren't a panacea (especially with regards to changed struct layouts), but they are what effectively communicates semver.

astiob:
> > Do package managers automatically verify symbol versions?
> 
> Yes.

How is it different from verifying symbol _existence?_

rcombs:
Does anyone have a link to actual documentation from a package manager on what it does with symbol versions?

libass/libass.sym
```diff
@@ -1,46 +1,54 @@
...
+	ass_set_selective_style_override_enabled;
+	ass_set_selective_style_override;
+	ass_set_check_readorder;
+	local:
+	*;
+};
+V_0.16.0 {
+	global:
+	ass_track_set_feature;
+} V_0.13.6;
```
TheOneric, line 54: `ass_track_set_feature` was added in 0.15.0

TheOneric:
> It seems this can be viewed using ldd -v $(which ffmpeg): I don’t know if this is physically stored separately from
 the symbol names, but this does indeed print the library versions that are referenced.

`ldd` resolves the stored dependencies to the actual files which would be used during runtime; try adding `LD_LIBRARY_P
ATH` to change the result. In fact it usually just runs the dynamic linker part of execution, which is why it is not safe to use `ldd` on untrusted binaries *(as pointed out in its manpage)*  
Unfortunately ELF objects currently only store the major version as a side effect of baking it into the soname. New ELF fields and flags are added from time to time (like recently for SHSTK). I’d hope one day there’ll be a "minorVersio" field or similar and linkers learn to copy this into dependent objects, but that’s right now that but a pipedream

> Does anyone have a link to actual documentation from a package manager on what it does with symbol versions?

idk what other packaging infra does, but in Debian (and presumably derivatives), symbol versions aren’t relied upon. There’s however [mechanisms](https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#dependencies-between-the-library-and-other-packages) to sanity check specified version constraints for libraries `shlibs` and [`symbols`](https://www.debian.org/doc/debian-policy/ch-sharedlibs.html#the-symbols-system) the latter of which tracks in which package version a symbol, including its version, was first seen. Generating and updating the symbols files usually only requires a copy-paste from build-tools output.  
This basically is the "tracking symbol existence" approach astiob asked about. It doesn’t require any action on the side of package maintainers of programs using libass and since [as established earlier](https://github.com/libass/libass/pull/754#issuecomment-2091903156) even with symbol versions a similar metadata file is required to map symbol versions to package versions, it shouldn’t make any difference on the part of libass’ package maintainer either. Tracking symbol existence has the upside however of actually being possible at all for every library out there, unlike scarcely used symbol versions. I cannot come up with any upside of relying on symbol versions.

It shares the same pitfall wrt to struct extensions however. But all Debian packages are always built against their target repo. Within a repo libraries are usually only ever upgraded, so there’s no problem. Downgrades only happen on severe issues (use special `+reallyOLDVER` versions) and tbh i’m not sure what happens then, but probably everything will need to get rebuilt.

Why Debian doesn’t just keep track of `soname <-> package version` instead, idk *(maybe because some (very few) libraries don't have so versions at all?)*

---

@jengelh i still don’t understand why you/OpenSUSE wants to use symbol versions and second the request for documentation on why and how symbol versions are used in the packaging infrastructure  
Right now, the impression i got from your comments is if symbol versioning is used, the packaging system will assume it’s sufficient to deduce package version constraints (which is wrong) and we’d actually do you a disservice by adding them

astiob:
> `ldd` resolves the stored dependencies to the actual files which would be used during runtime

Yes, but what matters is what is looked up in the first place. It doesn’t just look up and show a libc; it shows (and therefore I’m assuming it looks up) each referenced version of libc separately.

> Unfortunately ELF objects currently only store the major version as a side effect of baking it into the soname.

To be clear: are you saying that the _only_ information that changes from using symbol versions is the individual symbol names? Because this seems to contradict what OP said; see my quote in the reply that you quote. I’m just confused.

If only the list of symbols differs, then I fail to see how adding a versioned symbol differs from adding an unversioned symbol.

TheOneric:
Sorry, i think i misunderstood what part of `ldd`’s output you were referring to. Do you mean parts like this?:
```
                libc.so.6 (GLIBC_2.14) => /lib/x86_64-linux-gnu/libc.so.6
                libc.so.6 (GLIBC_2.9) => /lib/x86_64-linux-gnu/libc.so.6
```

The part in parentheses is a *symbol version* which does not need to correlate to the library version, public release version or anything else. In this example it doesn’t correlate to the library but (presumably) the public release version prefixed by the project name.

> To be clear: are you saying that the only information that changes from using symbol versions is the individual symbol names?

Not quite. Symbol versions are stored separately from the symbol name. As far as I know (but OP may correct me), the only changes of *providing* versioned symbols are the version of the symbol in the `.gnu.version` section changing from *"unversioned"* to whatever symbol version was provided and for each unique symbol version (without smybol names) an entry is added to `.gnu.version_d` (i guess `d` starnds for "definition"), e.g.:
```
Version definition section '.gnu.version_d' contains 2 entries:
 Addr: 0x00000000000097a0  Offset: 0x000097a0  Link: 4 (.dynstr)
  000000: Rev: 1  Flags: BASE  Index: 1  Cnt: 1  Name: libavutil.so.57
  0x001c: Rev: 1  Flags: none  Index: 2  Cnt: 1  Name: LIBAVUTIL_57
```

When *linking aginst* versioned symbols, the former of the above happens as well and in `.gnu.version_r` *(presumably “require”)*, for each library from which a versioned symbol is used, one or multiple entries are added listing all occurring versions, e.g.:
```
Version needs section '.gnu.version_r' contains 2 entries:
 Addr: 0x0000000000002530  Offset: 0x00002530  Link: 4 (.dynstr)
  000000: Version: 1  File: libc.so.6  Cnt: 5
  0x0010:   Name: GLIBC_2.14  Flags: none  Version: 8
  0x0020:   Name: GLIBC_2.4  Flags: none  Version: 7
  0x0030:   Name: GLIBC_2.3.4  Flags: none  Version: 6
  0x0040:   Name: GLIBC_2.7  Flags: none  Version: 5
  0x0050:   Name: GLIBC_2.2.5  Flags: none  Version: 3
  0x0060: Version: 1  File: libm.so.6  Cnt: 2
  0x0070:   Name: GLIBC_2.29  Flags: none  Version: 4
  0x0080:   Name: GLIBC_2.2.5  Flags: none  Version: 2
```

I assume the latter of both are what makes lookup easier and enables the earlier, more friendly error message. I don’t know whether, but would be surprised if, symbols from the same soname but different symbol version are allowed to reside in a different library files.

If no multiple versions of the same symbol name exist, this doesn’t open up new possibilities wrt to constraint checking compared to just looking up symbol name existence as far as i’m aware.

TheOneric:
Tentatively closing since the requested info wasn’t provided and with the available info it seems like not useful at best to a bad idea at worst (also the current patch is still broken in the same way as pointed out in initial review)

Can be reopened if new insight is provided

PR #756, opened 2024-04-28
ass_shaper improvements extracted from threading PR

rcombs:
Changes to `ass_shaper` made during work on #636 which don't require threading/atomics/locks:
- Keep the `hb_buffer_t` around between calls (it turns out to be fairly expensive to free and re-create in some cases)
- Create the `hb_font_t` eagerly during `add_face`
- Some consolidation of how we access the `FT_Face` and other data during font callbacks

The later changes are required for efficient access to the font object during shaping when running multi-threaded (they allow us to use lock-free cacheing for reads, which avoids most of the locking overhead around the `FT_Face` that we'd otherwise see).

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -424,6 +424,7 @@ static int add_face(ASS_FontSelector *fontsel, ASS_Font *font, uint32_t ch)
     int i, index, uid;
     ASS_FontStream stream = { NULL, NULL };
     FT_Face face;
+    int ret = -1;
 
     if (font->n_faces == ASS_FONT_MAX_FACES)
         return -1;
@@ -457,8 +458,16 @@ static int add_face(ASS_FontSelector *fontsel, ASS_Font *font, uint32_t ch)
     set_font_metrics(face);
 
     font->faces[font->n_faces] = face;
-    font->faces_uid[font->n_faces++] = uid;
-    return font->n_faces - 1;
+    font->faces_uid[font->n_faces] = uid;
+    if (!ass_create_hb_font(font, font->n_faces)) {
+        FT_Done_Face(face);
+        goto fail;
+    }
+
+    ret = font->n_faces++;
+
+fail:
+    return ret;
 }
 
 /**
@@ -483,7 +492,6 @@ size_t ass_font_construct(void *key, void *value, void *priv)
 
     font->library = render_priv->library;
     font->ftlibrary = render_priv->ftlibrary;
-    font->shaper_priv = NULL;
     font->n_faces = 0;
     font->desc.family = desc->family;
     font->desc.bold = desc->bold;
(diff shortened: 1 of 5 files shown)


libass/ass_font.c
```diff
@@ -457,8 +458,16 @@ static int add_face(ASS_FontSelector *fontsel, ASS_Font *font, uint32_t ch)
     set_font_metrics(face);
 
     font->faces[font->n_faces] = face;
-    font->faces_uid[font->n_faces++] = uid;
-    return font->n_faces - 1;
+    font->faces_uid[font->n_faces] = uid;
+    if (!ass_create_hb_font(font, font->n_faces)) {
+        FT_Done_Face(font->faces[font->n_faces]);
```
astiob, line 463: Nit: I feel this would be clearer as `FT_Done_Face(face);`

libass/ass_shaper.c
```diff
@@ -451,54 +455,25 @@ get_contour_point(hb_font_t *font, void *font_data, hb_codepoint_t glyph,
...
-    struct ass_shaper_metrics_data *metrics =
-        font->shaper_priv->metrics_data[info->face_index];
+    update_hb_size(hb_font, font->faces[info->face_index]);
+    metrics->metrics_cache = shaper->metrics_cache;
     metrics->hash_key.font = info->font;
     metrics->hash_key.face_index = info->face_index;
     metrics->hash_key.size = info->font_size;
+    metrics->vertical = info->font->desc.vertical;
+
+    hb_font_set_funcs(hb_font, shaper->font_funcs, metrics, free);
```
astiob, line 474: Is it OK to take a shared `hb_font` from the `ASS_Font` and to apply a unique instance of `ass_shaper_metrics_data` to it via `hb_font_set_funcs`? Will this not interfere with threading?

astiob (commented):
Thanks for these smaller PRs!

libass/ass_font.c
rcombs: Sure; done.

libass/ass_font.c
```diff
@@ -457,8 +458,16 @@ static int add_face(ASS_FontSelector *fontsel, ASS_Font *font, uint32_t ch)
     set_font_metrics(face);
 
     font->faces[font->n_faces] = face;
-    font->faces_uid[font->n_faces++] = uid;
-    return font->n_faces - 1;
+    font->faces_uid[font->n_faces] = uid;
+    if (!ass_create_hb_font(font, font->n_faces)) {
+        FT_Done_Face(face);
+        goto fail;
```
MrSmile, line 464: That looks unnecessary complicated. Something like
```c
    if (ass_create_hb_font(font, font->n_faces))
        return font->n_faces++;
    FT_Done_Face(face);
    return -1;
```
should be better.

libass/ass_shaper.c
rcombs: See `ass_shaper: don't mutate the shared hb_font_t during shaping` in #636 (currently 5ad73525b8834892cf958386c8dc6bc04915d97d); I use `hb_font_create_sub_font` to create a local `hb_font_t` that references the shared one, but has its own funcs font-data.

…though while writing this, I realized I could just cache the sub-font directly and save a decent bit of time for negligible memory cost, so I'm doing that.

libass/ass_font.c
rcombs: The fail path will later be needed for threading (to unlock), and I don't see a particular reason to reorganize this here only to redo it again in the threading branch.

astiob (approved):

PR #757, opened 2024-04-29
drawing: implement EXTEND_SPLINE

TheOneric:
This is the easiest part of our drawing discrepancies and _partially_ fixes the "s ... s ..." case from #202. However the upper left dip is still round and small, instead of a long spike. This probably comes from an unrelated difference in shape closing.

Plus some more cleanup

Relevant [VSFilter code](https://github.com/pinterf/xy-VSFilter/blob/45917d77083d93215738efb140486aba7614947e/src/subtitles/Rasterizer.cpp#L3303-L3309) for reference

Before, After, VSFilter *(from #202)*
202_before
202_partial
202_vsfilter

### DIFF
--- a/libass/ass_drawing.c
+++ b/libass/ass_drawing.c
@@ -378,9 +378,6 @@ bool ass_drawing_parse(ASS_Outline *outline, ASS_Rect *cbox,
             token = token->next;
             started = true;
             break;
-        default:
-            token = token->next;
-            break;
         }
     }
 
--- a/libass/ass_drawing.c
+++ b/libass/ass_drawing.c
@@ -31,19 +31,6 @@
 #define DRAWING_INITIAL_SEGMENTS 100
 
 /*
- * \brief Check whether a number of items on the list is available
- */
-static bool token_check_values(ASS_DrawingToken *token, int i, ASS_TokenType type)
-{
-    for (int j = 0; j < i; j++) {
-        if (!token || token->type != type) return false;
-        token = token->next;
-    }
-
-    return true;
-}
-
-/*
  * \brief Free a list of tokens
  */
 static void drawing_free_tokens(ASS_DrawingToken *token)
@@ -367,16 +354,13 @@ bool ass_drawing_parse(ASS_Outline *outline, ASS_Rect *cbox,
             break;
         }
         case TOKEN_CUBIC_BEZIER:
-            if (token_check_values(token, 3, TOKEN_CUBIC_BEZIER) &&
-                token->prev) {
-                if (!drawing_add_curve(outline, cbox, token->prev, false, started))
-                    goto error;
-                token = token->next;
-                token = token->next;
-                token = token->next;
-                started = true;
-            } else
-                token = token->next;
+            assert_3_forward(token);
+            if (!drawing_add_curve(outline, cbox, token->prev, false, started))
+                goto error;
+            token = token->next;
+            token = token->next;
+            token = token->next;
+            started = true;
             break;
         case TOKEN_B_SPLINE:
             assert_3_forward(token);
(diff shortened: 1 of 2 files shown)


rcombs (commented):
Seems reasonable enough? I still don't fully understand all of this code, and my reflex is to say "we should probably have more test cases for this stuff".

TheOneric:
> "we should probably have more test cases for this stuff"

more tests are definitely a good thing, though right now we deviate from VSFilter in a [number of cases](https://github.com/libass/libass/labels/drawing) and idk how much sense it makes to create regression tests for known-bad outputs

Right now we have each drawing command at least once in crash tests iinm and a dedicated `m 0 0 l ...` regression test *(iirc the latter was pulled from some issue)*

rcombs:
I do think it'd be worth having test cases even when the output is known not to match vsfilter, as long as we flag them as such and know to check to see whether changes move them closer or further from reference output.

Meanwhile, can we construct a test case that reproduces the issue fixed here, but not the shape-closing issue that still persists? I don't think we need to block on this regardless, but it seems worth having.

astiob (approved):
LGTM; thanks.

I’m not sure I see how the merged line is clearer, but don’t let that stop you.

astiob:
> However the upper left dip is still round and small, instead of a long spike. This probably comes from an unrelated difference in shape closing.

It is related in the sense that it is related to B-splines (`s` and `p`, but not Bézier curves `b`). When the first command after `m` is a B-spline, VSFilter saves [the spline’s starting coordinates](https://github.com/pinterf/xy-VSFilter/blob/45917d77083d93215738efb140486aba7614947e/src/subtitles/Rasterizer.cpp#L3178) as the whole contour’s starting point that it reads back when it eventually tries to close the contour. I assume we use the actual `m` coordinate instead. For any other kind of segment, the starting coordinate is the same as the `m` coordinate, but not for a B-spline.

astiob:
Except… that doesn’t actually seems to change our output. Perhaps MrSmile already takes this into account in the rasterizer (maybe even by accident). This is still broken:

```
m 125 65 n 0 0 s 150 60 150 150 60 150 120 120 90 90 s 60 90 90 120 60 120
```

(with or without an added `l 125 65` at the end)

The difference _actually_ comes from [this VSFilter block](https://github.com/pinterf/xy-VSFilter/blob/45917d77083d93215738efb140486aba7614947e/src/subtitles/Rasterizer.cpp#L3199-L3202), which is unrelated to contour closure. Indeed, adding the missing line segment manually produces the desired output:

```
m 0 0 s 150 60 150 150 60 150 120 120 90 90 l 70 95 n 90 90 s 60 90 90 120 60 120
```

TheOneric:
> The difference actually comes from [this VSFilter block](https://github.com/pinterf/xy-VSFilter/blob/45917d77083d93215738efb140486aba7614947e/src/subtitles/Rasterizer.cpp#L3199-L3202), which is unrelated to contour closure.

ahh, so i was searching in the wrong place! Earlier today i tried to get an unaffected test sample but noticed even without closing `s ... s ...` often deviates a bit; this explains why and ig means there’s no bug in the current commits, whew 
Thanks!

Also, now knowing what to look for this was alerady discovered previously in #435 in the context of `n`.
Afaict, only `n` and b-splines can produce such extra line segments since for lines *(obviously)* and bezier curves the final point is always identical to the input coordinate and `m` closes the shape. When a shape consists entirely of one b-spline closed by `c` there probably is no (visible) difference.

> I’m not sure I see how the merged line is clearer, but don’t let that stop you.

Several times while looking over our code, this line break and folllowing indentation confused me into thinking the `started ||` part applied to all point additions for some moments

astiob:
> Several times while looking over our code, this line break and folllowing indentation confused me into thinking the `started ||` part applied to all point additions for some moments

Ah, see, I triple-checked the line before commenting and I still actually missed the surrounding parentheses 🤦 Yeah, that’s certainly better, then.

PR #760, opened 2024-05-11
Match VSFilter’s faux italics on PostScript-outline fonts, too

astiob:
#676 previously adjusted our faux italics to match GDI/VSFilter on TrueType fonts, but we later discovered that PostScript-outline fonts behave differently (font driver differences strike again) and our rendering became _less_ similar to VSFilter’s after that change. This should fix it.

I derived the coefficient experimentally using Fansub Block. The first-approximation value I got was awfully close to 10° (9.998°), so I figure it’s probably exactly 10°. Thus, I’ve hardcoded the closest fixed-point representation of tan(10°) in this commit.

TrueType:
A screenshot shows a parallelogram whose interior angle is about 71.2°.

PostScript:
A screenshot shows a parallelogram whose interior angle is about 80.0°.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -27,6 +27,7 @@
 #include FT_TRUETYPE_TABLES_H
 #include FT_OUTLINE_H
 #include FT_TRUETYPE_IDS_H
+#include FT_TYPE1_TABLES_H
 #include <limits.h>
 
 #include "ass.h"
@@ -515,6 +516,12 @@ void ass_face_set_size(FT_Face face, double size)
     FT_Request_Size(face, &rq);
 }
 
+bool ass_face_is_postscript(FT_Face face)
+{
+    PS_FontInfoRec postscript_info;
+    return !FT_Get_PS_Font_Info(face, &postscript_info);
+}
+
 /**
  * \brief Get face weight
  **/
@@ -602,16 +609,18 @@ static void ass_glyph_embolden(FT_GlyphSlot slot)
 /**
  * Slightly italicize a glyph
  */
-static void ass_glyph_italicize(FT_GlyphSlot slot)
+static void ass_glyph_italicize(FT_Face face)
 {
     FT_Matrix xfrm = {
         .xx = 0x10000L,
         .yx = 0x00000L,
-        .xy = 0x05700L,
+        .xy = ass_face_is_postscript(face)
+            ? 0x02d24L /* tan(10 deg) */
+            : 0x05700L /* matches GDI; effectively tan(18.77 deg) */,
         .yy = 0x10000L,
     };
 
-    FT_Outline_Transform(&slot->outline, &xfrm);
+    FT_Outline_Transform(&face->glyph->outline, &xfrm);
 }
 
 /**
@@ -719,7 +728,7 @@ bool ass_font_get_glyph(ASS_Font *font, int face_index, int index,
 
     FT_Long style_flags = ass_face_get_style_flags(face);
     if (!(style_flags & FT_STYLE_FLAG_ITALIC) && (font->desc.italic > 55))
-        ass_glyph_italicize(face->glyph);
+        ass_glyph_italicize(face);
     if (!(style_flags & FT_STYLE_FLAG_BOLD) &&
         font->desc.bold > ass_face_get_weight(face) + 150)
         ass_glyph_embolden(face->glyph);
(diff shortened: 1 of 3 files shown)


libass/ass_font.c
```diff
@@ -515,6 +516,12 @@ void ass_face_set_size(FT_Face face, double size)
     FT_Request_Size(face, &rq);
 }
 
+bool ass_face_is_postscript(FT_Face face)
+{
+    PS_FontInfoRec postscript_info;
+    return !FT_Get_PS_Font_Info(face, &postscript_info);
```
TheOneric, line 522: i’m trusting you to have verified FreeType doesn’t synthesise values here for pure TrueType fonts, though it would probably good to add regrioons test samples to libass-tests to ensure this doesn't change in the future without us noticing

FreeType 2.1.2’s changelog also [mentions](https://gitlab.freedesktop.org/freetype/freetype/-/blob/b875924a6fc176d46fc9f82c616db631fc4d665c/docs/CHANGES#L3899) the function now working with CID and Type42 fonts. The latter appear to be a compatibility wrapper around TrueType fonts iiuc, are we sure GDI treats them as PostScript outline fonts?

astiob: This is just our existing implementation moved from `ass_fontselect`. It has worked thus far.

astiob: As for Type42 fonts, I’ve never seen one and I don’t know what they look like; if you have an example, I’m happy to test it on Windows. Our tests until now have been with TrueType fonts with or without OpenType extensions; OpenType/CFF fonts (not sure if we’ve explicitly tested fonts using the `CFF2` table, but I assume they should behave the same as those using `CFF`); and Type 1 fonts (which enjoy limited support on some systems and are completely unsupported on others).

TheOneric: I converted DejaVu Sans to a Type42 font using `ttftotype42` from [LCDF Typetools](https://www.lcdf.org/type/) *(conversion of the Jester font from the original issue failed with checksum errors)*. In libass with this patch they do indeed get different slant values using `mpv --sub-font-provider=none --sub-fonts-dir=/tmp/empty`: [type42_sample.zip](https://github.com/libass/libass/files/15283911/type42_sample.zip)

libass/ass_font.c
```diff
@@ -579,16 +586,18 @@ static void ass_glyph_embolden(FT_GlyphSlot slot)
...
-static void ass_glyph_italicize(FT_GlyphSlot slot)
+static void ass_glyph_italicize(FT_Face face)
 {
     FT_Matrix xfrm = {
         .xx = 0x10000L,
         .yx = 0x00000L,
-        .xy = 0x05700L,
+        .xy = ass_face_is_postscript(face)
+            ? 0x02d24L /* tan(10 deg) */
+            : 0x05700L /* from GDI; roughly 0.34; tan(18.77 deg) */,
```
moi15moi, line 596: Technically, it is more ``sin(20 deg)`` instead of ``tan(18.77 deg)``

astiob: Thanks for the pointer. However, sin has no physical meaning in this context, whereas tan refers to the slant angle. I can add a comment that it may have apparently originated as sin(20°), but it doesn’t seem very useful—nor is it even a very good approximation of sin(20°), although certainly better than of tan(20°) for what it’s worth.

libass/ass_font.c
astiob: I can’t tell whether VSFilter uses this font because I have a system font of the same name, but extracting the attachment and trying to open it produces an error (“not a font file”). I strongly suspect this format isn’t supported at all.

libass/ass_font.c
rcombs: It's sin(20°)*256, truncated to an integer, and then shifted left 8 to expand to 8.8 form.

astiob: Well, yeah. “Truncated to 1/8th” being “not a very good approximation” (even _rounding_ to 1/8th would have produced a different value). Anyway, is this relevant enough that you want it mentioned?

libass/ass_font.c
TheOneric: verified VSFilter/GDI doesn't use such files on IRC using a renamed font sample

TheOneric (approved):

libass/ass_font.c
astiob: Reworded to add a mention of sin(20°). Does this look good?

astiob: Aaand removed again and reworded some more to avoid unofficial data.

rcombs (approved):

PR #764, opened 2024-05-19
fix order of arguments in calloc()

vtorri:


TheOneric:
Thanks! merged with some more details added to the commit message

PR #765, opened 2024-05-19
readme: move related projects into wiki; update VSFMod status

TheOneric:
Information about the format and to a lesser degree also other ASS implementations seem more relevant than a selection of some libass users to me

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,28 +34,6 @@ Windows. It is suited for static-only builds on any platform well supported by
 Meson and as a Meson subproject.
 Notably, Meson supports MSVC and generation of VS project files.
 
-Related Links
-=============
-The following projects/companies use libass:
-
-- [MPlayer](http://www.mplayerhq.hu/)
-- [mplayer2](http://www.mplayer2.org/)
-- [mpv](http://mpv.io/)
-- [VLC](http://www.videolan.org/)
-- [GStreamer](http://gstreamer.freedesktop.org/) (assrender plugin)
-- [FFmpeg](http://ffmpeg.org/)
-  - [Libav](http://libav.org/) (unmaintained)
-- [Aegisub](http://www.aegisub.org/)
-- [Kodi (XBMC)](http://kodi.tv/)
-- [avidemux](http://fixounet.free.fr/avidemux/)
-- [PunkGraphicsStream (BD subtitle encoder)](http://code.google.com/p/punkgraphicstream/)
-- [HandBrake](http://handbrake.fr/)
-- [MX Player](https://play.google.com/store/apps/details?id=com.mxtech.videoplayer.ad)
-- [QMPlay2](http://zaps166.sourceforge.net/?app=QMPlay2)
-- [SubtitlesOctopus](https://github.com/libass/JavascriptSubtitlesOctopus) (JavaScript ASS via wasm libass)
-  - [Crunchyroll](http://www.crunchyroll.com/) uses SubtitlesOctopus
-- [AssRender for Avisynth(+)](http://avisynth.nl/index.php/AssRender)
-
 Information about the ASS format:
 =================================
 - [ASS format overview](https://github.com/libass/libass/wiki/ASS-File-Format-Guide)
--- a/README.md
+++ b/README.md
@@ -46,8 +46,9 @@ Other ASS/SSA implementations:
   - [xy-VSFilter/XySubFilter](https://github.com/Cyberbeing/xy-VSFilter/)
     - [pfmod](https://github.com/pinterf/xy-VSFilter/)
   - VSFilter in [MPC-HC](https://github.com/clsid2/mpc-hc/tree/develop/src/filters/transform/VSFilter/)
-  - [VSFilterMod](https://code.google.com/archive/p/vsfiltermod/) (with custom format extensions)
-    - [sorayuki fork](https://github.com/sorayuki/VSFilterMod/) with some bugfixes
+  - [VSFilterMod](https://code.google.com/archive/p/vsfiltermod/) with custom format extensions (defunct, subsumed by forks)
+    - [sorayuki fork](https://github.com/sorayuki/VSFilterMod/) with some bugfixes (defunct)
+    - various forks focussing on internal use
   - [Threaded VSFilter](https://code.google.com/p/threaded-vsfilter/) (defunct)
   - VSFilter in [Guliverkli2](http://sourceforge.net/projects/guliverkli2/) (defunct, subsumed by all of the above)
   - VSFilter in [guliverkli](http://sourceforge.net/projects/guliverkli/) (defunct, forked as Guliverkli2)


rcombs (approved):
Seems reasonable

TheOneric:
hmm, looking at it, the “Other ASS/SSA implementations” list might need some updates too, but not sure what exactly to

The original VSFilterMod is also long defunct but not yet marked such and sorayuki [points people to forks for future updates](https://github.com/sorayuki/VSFilterMod/commit/643b3def0970eb9dedaa5e6d9be6c3c492ce970d). It’s not clear to me if or which fork emerged as the main successor. Both Masaiki and kedaitnh12 forks start out from the last sorayuki commit (minus the Readme update), the latter seems to be based on yet another intermediate fork which doesn't show up (only the most recent commits are from kedaitnh12, almost everything before since  diverging from another identity).
passerbya branched off 5 commits earlier.

Also not sure how to consider and list Cyberbeing/xy-VSFilter (even more so since i technically have push access)

ASS.js has been inactive on master for 4 years and inactive on the “next” branch for 2 years, but there’s some more recent activity in issues and in the presumably related [ass-compiler](https://github.com/weizhenye/ass-compiler)

MIRIMIRIM:
For VSFilterMod, I’d recommend the [computerfan fork](https://github.com/computerfan/VSFilterMod), because vmoe and it’s members are the primary users and developers of it, though lately, they've been a bit quiet. Within the computerfan fork, there are two forks: the [Asd-g fork](https://github.com/Asd-g/VSFilterMod/tree/master-1) and the [AmusementClub fork](https://github.com/AmusementClub/VSFilterMod). The former focuses more on Avisynth-related functionalities, while the latter leans towards bug fixes since it doesn’t utilize Avisynth. It’s worth noting that Masaiki is one of the maintainers of the AmusementClub fork.

Speaking of which, let’s touch on assrender again. I’m not entirely sure about [pinterf](https://github.com/pinterf/assrender)’s plan, but the assrender for Avisytnh hasn’t seen updates for quite some time (the last update was for libass 0.15.0). I’m uncertain if anyone else is still using Avisynth to encode hardsub videos. The assrender in the [AmusementClub (Masaiki) fork](https://github.com/AmusementClub/assrender), however, has relatively active maintainers, primarily focusing on bug fixes and upstream updates like libass. Would you consider adding the assrender Vapoursynth version to the Related Links?

TheOneric:
Thanks for the pointer re VSFilterMod. Weird computerfan’s fork and most but not all of its forks don't show up in the overview. The computerfan repo too branched off from the latests sorayuki commit sans README update and indeed kedaitnh12 is based on computerfan’s fork with a few patches from asd-g and AmusementClub on top. Except for computerfan, which itself has been inactive for two years, none of them allows creating issues though so i’m not sure if they even want to be considered for general use, or focus on their own needs. Do you know more?

Regarding *synth assrender, we already point to the AviSynth wiki page instead of a particular fork. I’ve seen some filters also list their VapourSynth version in the wiki, but not sure how common this is given it’s a wiki for AviSynth(+). Though, I also noticed it has more input parameters than pinterf’s AviSynth version, so i guess this warrants a separate listing.
I’ll add it once the rest is figured out.

MIRIMIRIM:
The computerfan fork is more for internal use by Vmoe, but I can only say that based on my experience, it is difficult to determine the general usage of VSFilterMod by users. In most cases, they do not know which version of VSFilterMod they are using. Most of them are privately disseminated version (such as computerfan / Vmoe version) or sorayuki version (may not necessarily be the latest). However, I’m not quite sure how people other than Chinese fansubbers use VSFilterMod. It seems that the TypesettingTools members are not interested in this. Maybe you can ask kedaitinh12?

astiob (approved):

cubicibo:
If you do want to list a BD subtitle encoder, I would replace PunkGraphicStream with Masaiki's ass2bdnxml (which is based on avs2bdnxml). Neither PunkGraphicStream or a*s2bdnxml produce compliant BD SUP files... but ass2bdnxml does offer Sony BDN as an output format, the common input format for BD authoring suites.

TheOneric:
Added more commits to list the requested and update (only) VSFMod status. Decided against linking any particular fork since if they only focus on internal use and don't even have issues enabled, they might not appreciate getting external users with different needs and bug reports and feature requests redirect their way.

---

>  I’m not quite sure how people other than Chinese fansubbers use VSFilterMod

In general VSFilterMod is only suitable for hardsubbing though (as the original project prominently pointed out). Not only are its extensions incompatible with all other renderers, even standard tags can render differently in VSFMod and on top of that different versions of the same VSFMod branch can also exhibit breaking rendering changes. Having various independent forks means even more differences.  
Thus if softsubs were created for VSFMod, they’d be bound to the particular VSFMod version they were authored with for all eternity and users would need to swap out the installed VSFMod version per release. So typically only hardsubs (often kfx) are/were created with it and viewers never obtain a copy of VSFilterMod. Hardsubbing in general became less common though; otoh i don't recall anyone doing hardsubs recently but i don’t know all communities.

> If you do want to list a BD subtitle encoder, I would replace PunkGraphicStream with Masaiki's ass2bdnxml (which is based on avs2bdnxml).

Sure i can add it; since i don’t know in which scenarios either one works well or not though, i didn’t remove PunkGraphicStream.

arch1t3cht:
It's worth noting that VapourSynth also has its own "more official" (i.e. part of the VapourSynth org) libass subtitle rendering plugin [here](https://github.com/vapoursynth/subtext/tree/master). However, I can't say how it compares to assrender. I think I recall people recommending assrender over subtext, but don't quote me on that. (And either way I don't remember the reasons, if there were any.)

TheOneric:
>  I think I recall people recommending assrender over subtext, but don't quote me on that. (And either way I don't remember the reasons, if there were any.)

idk what other differences there might be, but at least subtext doesn’t call `ass_set_storage_size` while VapourSynth assrenderer does

TheOneric:
> PunkGraphicStream unfortunately never really worked beyond some software players.

Is it really so useless as to be removed completely?
I have no experience with either, but iiuc PunkGraphicStream and ass2bdnxml produce different output formats and PunkGraphicStream has been listed since the very first version of the README

arch1t3cht:
After searching logs and doing some (very) limited testing, it seems that assrender doesn't support blending onto 32-bit video while subtext does (exactly contrary to what the only relevant message I found in chat logs claimed...), but that subtext crashes on full range input while assrender doesn't. But neither of these are very relevant to the actual libass usage.

cubicibo:
>Is it really so useless as to be removed completely?

Basically anything that goes beyond the basic "one subtitle line at a time" will not display properly on Blu-ray Disc players. More complex output may also not work on software players.

>I have no experience with either, but iiuc PunkGraphicStream and ass2bdnxml produce different output formats¹ [...]

a*s2bdnxml can output SUP, but it is only _remotely_ compliant if image splitting is not enabled in any form.

TheOneric:
iiuc compliant SUPs only support non-overlapping events and depending on configuration a\*s2bdnxml (when outputting SUP directly instead of BDN) and supposedly PGS may omit overlapping SUP events causing problems; additionally (considering your “*remotely* compliant”) a\*sbdnxml’s SUP output (unknown for PGS) has further other spec violation, at least sometimes. o did i get this correct?

Looking at PGS’ source though, it seems like it already [aggregates simultaneous events](https://github.com/slartibardfast/punkgraphicstream/blame/fef21a9ce98c3e3dcccf6dde6a9fb57a9718f8ed/src/main/java/name/connolly/david/pgs/SubtitleEvent.java#L37) and later [turns them into a single SUP event](https://github.com/slartibardfast/punkgraphicstream/blob/fef21a9ce98c3e3dcccf6dde6a9fb57a9718f8ed/src/main/java/name/connolly/david/pgs/SupGenerator.java#L175) although there appear to be some limits on how many simultaneous events the SUP generator actually supports.

Questions being:
- if the simultaneous event limits are met, are there _some_ scenarios where PGS’ SUP output is more or fully compliant than a*s2bdnxml’s SUP output?
- how easy is it and which options are available to turn ass2bdnxml’s BDN output into a fully compliant SUP?

cubicibo:
Your summary is roughly correct. However, overlapping events are not the issue per-se. The fact that neither solution respects the target decoder model is. If you are familiar with Transport Streams and T-STD decoders, you will get it instantly: clock speed, buffer management and decoding delays are paramount. 

PunkGraphicStream assumes a perfect decoder without latency and infinite memory. Ideal decoders do not exist, so in reality a (real) decoder will write OOB or read-only memory. Furthermore, the DTS may not be monotonic, something the ITU H.222 forbids fiercely.

a*s2bdnxml does try to manage the memory, set decoding timestamps, and perform proper assets versioning: a must for seamless branching in HDMV streams.

>if the simultaneous event limits are met, are there some scenarios where PGS’ SUP output is more or fully compliant than a*s2bdnxml’s SUP output?

- Anything frame-by-frame is a no-go for PunkGraphicStream as the DTS is no longer monotonic. Any change in the bitmap shape with contiguous events will also crash a decoder due to OOB writes. 
- a*s2bdnxml goes out of spec whenever an event cannot be decoded and overlaid in due time. This margin depends on the framerate, screen layout,  

> how easy is it and which options are available to turn ass2bdnxml’s BDN output into a fully compliant SUP?

You are never guaranteed to get a fully compliant SUP out of a*s2bdnxml. Those settings may increase your chances: `-a1 -p1 -b0 -u0` and optionally `-z1`. 
If you desire a fully compliant SUP out of an ASS script, you must use Sony BDN as an intermediate format. The HDMV PGS encoder you choose must shape the output according to a target decoder model. This is just like VBV constrained video encoding.
Anime distributor Dynit [[1](https://twitter.com/NLScavenger/status/1287134106141196288)] [[2](https://twitter.com/NLScavenger/status/1476858214830514181)] and Discotek Media [[1](https://www.youtube.com/watch?v=3YkrvtaO868)] [[2](https://x.com/Schmullus1/status/1736841182326694169/video/1)] go through that two-steps conversion process (ASS -> BDN -> HDMV PGS) to render ASS scripts on Blu-ray discs.

TheOneric:
Thanks for all the input; PunkGraphiStream is now replaced by ass2bdnxml. I also added VapourSynth’s first-party subtext (and filed an issue there re storage size).

Though after more discussion on IRC we decided to move this list out of the README into the wiki; it now lives at https://github.com/libass/libass/wiki/Projects-using-libass

PR #766, opened 2024-05-19
Replace checkasm Haiku build patch with upstream solution

TheOneric:
https://code.videolan.org/videolan/x86inc.asm/-/commit/b6ba1e3045d758fd6c6e24591dac21a3dc812e1d approaches the issue more broadly and replaces 0a63b95ce6dbe6fc3a6470c8b88b242341c8f961


rcombs (approved):
Seems fine, assuming we've tested and it works.

TheOneric:
> assuming we've tested and it works

yep, it works at least on both 64- and 32-bit Haiku, illumos, NetBSD and everything in our CI

PR #767, opened 2024-05-20
compare: add support for a wider PNG range

MrSmile:
I've tried to trigger all necessary colorspace conversions. It works on several samples that I created, but it certainly not a full set of possible PNG variants. So, more testing is welcome, especially with different paletted versions.

### DIFF
--- a/compare/image.c
+++ b/compare/image.c
@@ -60,14 +60,22 @@ bool read_png(const char *path, Image16 *img)
 
     uint32_t w = png_get_image_width(png, info);
     uint32_t h = png_get_image_height(png, info);
+    uint32_t a = png_get_valid(png, info, PNG_INFO_tRNS);
     int type   = png_get_color_type(png, info);
     int depth  = png_get_bit_depth(png, info);
 
-    if (w > 0xFFFF || h > 0xFFFF || type != PNG_COLOR_TYPE_RGBA) {
+    if (w > 0xFFFF || h > 0xFFFF) {
         png_destroy_read_struct(&png, &info, NULL);
         fclose(fp);
         return false;
     }
+
+    if (!(type & PNG_COLOR_MASK_COLOR))
+        png_set_gray_to_rgb(png);
+    if ((type & PNG_COLOR_MASK_PALETTE) || a)
+        png_set_expand(png);
+    if (!a && !(type & PNG_COLOR_MASK_ALPHA))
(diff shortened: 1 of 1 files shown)


compare/image.c
```diff
@@ -60,15 +60,20 @@ bool read_png(const char *path, Image16 *img)
...
-    if (w > 0xFFFF || h > 0xFFFF || type != PNG_COLOR_TYPE_RGBA) {
+    if (w > 0xFFFF || h > 0xFFFF) {
         png_destroy_read_struct(&png, &info, NULL);
         fclose(fp);
         return false;
     }
 
+    if(!(type & PNG_COLOR_MASK_COLOR))png_set_gray_to_rgb(png);
+    if((type & PNG_COLOR_MASK_PALETTE) || a)png_set_expand(png);
+    if(!a && !(type & PNG_COLOR_MASK_ALPHA))png_set_filler(png, -1, PNG_FILLER_AFTER);
```
rcombs, line 75: Whitespace on all these lines (preferably the conditionalized statements on separate lines from the conditions)

MrSmile: Done.

TheOneric (approved):
haven’t tested decoding samples, but code seems good

TheOneric:
Here are some tests with 8bpc RGB, 8bpc GRAY+ALPHA and palletised (with alpha). All are accepted as identical to `compare`’s own output. Notably when palettising with GIMP the exported result appear to have the green box shifted by 1px, but palletising with `optipng` gave an identical result.

[pngfmt_test.zip](https://github.com/user-attachments/files/15879766/pngfmt_test.zip)

TheOneric:
@MrSmile: do any more formats need testing, or is it now good for merge?

MrSmile:
Rebased.

> do any more formats need testing

Both my and yours samples are from GIMP, so I feel that's not fully representative, especially in tRNS case. I've tried `pngcrush` at a couple of samples, but failed to produce a new color format with it (maybe I don't know how to use it correctly). Well, we can always merge as it is and fix errors afterwards, it's debug functionality anyway.

TheOneric:
> Both my and yours samples are from GIMP, so I feel that's not fully representative, especially in tRNS case.

My palletised alpha case was generated by `optipng` *(since GIMPs output was shifted for some reason judging by imagemagicks diff view)*. `optipng`’s inpout was an RGBA image from GIMP, but afaik this shouldn't matter.

PR #768, opened 2024-05-21
Constify API parameters where possible

TheOneric:
We constified memory font data before in #448. Prompted by running into filepaths not being `const`, this now constifies most `char*`-related API paramters if they’re already effectively treated as const anyway and the new signature is pointer-compatible with the existing one.

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -1222,7 +1222,7 @@ static int process_text(ASS_Track *track, char *str)
  * \param data string to parse
  * \param size length of data
 */
-void ass_process_data(ASS_Track *track, char *data, int size)
+void ass_process_data(ASS_Track *track, const char *data, int size)
 {
     char *str = malloc(size + 1);
     if (!str)
@@ -1243,7 +1243,7 @@ void ass_process_data(ASS_Track *track, char *data, int size)
  * \param size length of data
  CodecPrivate section contains [Stream Info] and [V4+ Styles] ([V4 Styles] for SSA) sections
 */
-void ass_process_codec_private(ASS_Track *track, char *data, int size)
+void ass_process_codec_private(ASS_Track *track, const char *data, int size)
 {
     ass_process_data(track, data, size);
 
@@ -1279,7 +1279,7 @@ void ass_set_check_readorder(ASS_Track *track, int check_readorder)
  * \param timecode starting time of the event (milliseconds)
  * \param duration duration of the event (milliseconds)
 */
-void ass_process_chunk(ASS_Track *track, char *data, int size,
+void ass_process_chunk(ASS_Track *track, const char *data, int size,
                        long long timecode, long long duration)
 {
     char *str = NULL;
@@ -1367,7 +1367,7 @@ void ass_flush_events(ASS_Track *track)
  * \return a pointer to recoded buffer, caller is responsible for freeing it
 **/
 static char *sub_recode(ASS_Library *library, char *data, size_t size,
-                        char *codepage)
+                        const char *codepage)
 {
     iconv_t icdsc;
     char *tocp = "UTF-8";
@@ -1535,7 +1535,7 @@ static ASS_Track *parse_memory(ASS_Library *library, char *buf)
  * \return newly allocated track
 */
 ASS_Track *ass_read_memory(ASS_Library *library, char *buf,
-                           size_t bufsize, char *codepage)
+                           size_t bufsize, const char *codepage)
 {
     ASS_Track *track;
     int copied = 0;
@@ -1571,8 +1571,8 @@ ASS_Track *ass_read_memory(ASS_Library *library, char *buf,
     return track;
 }
 
-static char *read_file_recode(ASS_Library *library, char *fname,
-                              char *codepage, size_t *size)
+static char *read_file_recode(ASS_Library *library, const char *fname,
+                              const char *codepage, size_t *size)
 {
     char *buf;
     size_t bufsize;
@@ -1600,8 +1600,8 @@ static char *read_file_recode(ASS_Library *library, char *fname,
  * \param codepage recode buffer contents from given codepage
  * \return newly allocated track
 */
-ASS_Track *ass_read_file(ASS_Library *library, char *fname,
-                         char *codepage)
+ASS_Track *ass_read_file(ASS_Library *library, const char *fname,
+                         const char *codepage)
 {
     char *buf;
     ASS_Track *track;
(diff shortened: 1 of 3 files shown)


rcombs (approved):

libass/ass.c
```diff
@@ -1535,8 +1535,9 @@ static ASS_Track *parse_memory(ASS_Library *library, char *buf)
  * \return newly allocated track
 */
 ASS_Track *ass_read_memory(ASS_Library *library, char *buf,
-                           size_t bufsize, char *codepage)
+                           size_t bufsize, const char *codepage)
 {
+    // Documentation promises buf isn't modified other than being passed to iconv
```
astiob, line 1540: I’d prefer to declare `const char *buf`. The `iconv` call can stay an implementation detail, and the caller shouldn’t need to care.

It may be worth mentioning that older versions of the spec [did in fact include `const`](https://pubs.opengroup.org/onlinepubs/007908799/xsh/iconv.html) in the `iconv` prototype but this was [changed after complaints](http://www.opengroup.org/infosrv/XPG4_Brand/resolutions/XoTGBase94-001.ps.Z) that this was inconsistent with other text in the spec that mentioned “The type of _inbuf_ and _outbuf_, **char \*\***” which implied that applications should be able to pass non-const pointers in. Of course, the working group’s resolution text doesn’t seem to even consider that the change breaks applications that supply const pointers instead, but there’s no way to allow both in a single prototype anyway. The resolution also explicitly says:

> For conformance to the current version of XSH4 either prototype is allowed.

Anyway, my point is: although the definition of `iconv` doesn’t explicitly say this, this history corroborates the common sense assumption that `iconv` never writes through this pointer and should accept a const pointer just fine.

libass/ass.h
```diff
@@ -324,7 +324,8 @@ void ass_set_extract_fonts(ASS_Library *priv, int extract);
  *   ScaledBorderAndShadow=yes
  *
  * \param priv library handle
- * \param list NULL-terminated list of strings
+ * \param list NULL-terminated list of strings,
+ *             copied and never modified by libass
  */
 void ass_set_style_overrides(ASS_Library *priv, char **list);
```
astiob, line 330: Consider also:

```suggestion
#ifdef __cplusplus
void ass_set_style_overrides(ASS_Library *priv, const char *const *list);
#else
void ass_set_style_overrides(ASS_Library *priv, char **list);
#endif
```

rcombs: Isn't this non-compatible?

astiob: Only in C—hence the ifdef.

astiob: To elaborate: C++ allows implicitly constifying any number of outermost layers of a pointer. Hence `const char *const *list`, not `const char **list`. With the latter, a call like `char *x; foo(&x);` could reassign `x = <pointer to some truly immutable memory>;`, which the caller would then treat as writable memory and crash. With the outer layer constified as well, the called function can’t reassign the caller’s pointer, so there’s nothing to worry about. For some reason, C has never implemented this (IMO it totally should, but I assume some kind of inertia prevents it).

TheOneric: does this interact in any way with the enclosing `extern "C"` block?

libass/ass.c
TheOneric: > older versions of the spec [did in fact include const](https://pubs.opengroup.org/onlinepubs/007908799/xsh/iconv.html)
> Of course, the working group’s resolution text doesn’t seem to even consider that the change broke applications that supply const pointers instead, but there’s no way to allow both in a single prototype anyway.

The version in the header is declared to be the normative form, thus `const char**` was to my understanding technically always incorrect (but the discrepancy quite confusing indeed). See [here](https://pubs.opengroup.org/onlinepubs/7908799/xsh/interfaces.html):

> will make all necessary symbols visible as described in the Headers section of this specification. 

I didn’t know both forms are allowed; though the current spec contains no mention of this. Are both allowed by the standard, or was this just a remark about their validation testsuite (for SUSv2)? I’m also not sure if an implementations is required to satisfy constraints of both variants (i.e. never modify input data) or if only those actually using `const` have to.

I’d love to be able to assume iconv never modifies the input memory, but there doesn't seem to be anything explicitly corroborating this. (Not does the description mention modifying inbuf data, which maybe ca be seen as implying its constness)

libass/ass.h
astiob: No (as far as I know). `extern "C"` affects symbol linkage but not const/volatile conversions.

libass/ass.c
astiob: > I didn’t know both forms are allowed; though the current spec contains no mention of this. Are both allowed by the standard, or was this just a remark about their validation testsuite (for SUSv2)?

My reading is that both forms aren’t allowed when claiming compliance with newer versions of the standard but both forms are (were) allowed if one wants to claim compliance with the older version(s) specifically.

> I’d love to be able to assume iconv never modifies the input memory, but there doesn't seem to be anything explicitly corroborating this. (Not does the description mention modifying inbuf data, which maybe ca be seen as implying its constness)

Yes, it’s frustrating that this isn’t actually specified, but I do think this is a risk I’m willing to take, personally, all things considered.

libass/ass.h
astiob: Just to be super sure, I’ve verified that this compiles without any issues using `g++ -Wall -Wextra -pedantic`:

```c++
extern "C" void foo(const char *const *bar);
char *x;
int main() {
    foo(&x);
}
```

TheOneric: seems good then; adopted

libass/ass.c
TheOneric: By putting `const` into our API and leaving the iconv call an unmentioned implementation detail, we make this decision for all API users though and promise something which isn’t actually under out control, so I’m not sure we should.

By documenting users can decide whether to take their chances with `iconv` or not and by default (if not reading docs) the compiler will guide towards the guaranteed safe approach.
But if we constify this users might pass in read-only memory without ever taking a look at docs. If there’s some cursed iconv implementation actually modifying the input, this then turns into a direct path into UB.

astiob:
@TheOneric Check #506 out. We completely forgot about that one, didn’t we.

TheOneric:
heh, porting #506’s VSF fixes and alloc eliminations was something i wanted to do eventually, but i indeed completely forgot it also touched API already

libass/ass.c
TheOneric: Given how truly cursed an input-modifying iconv would be, `const`ifying our API parameter _and_ noting potential iconv calls might also be ok, but `const`ifying without documenting iconv use...

TheOneric:
Squahed commits and added a `LIBASS_VERSION` bump in case anyone wants to conditionally cast parameters to avoid harmless compiler warnings and dropped the `iconv` comment.

PR #769, opened 2024-05-22
ass_filesystem: Fix add_separator compilation warning 

cubicibo:
```
In function 'open_dir_wtf8',
    inlined from 'ass_open_dir' at ../libass-0.17.2/libass/ass_filesystem.c:339:9:
../libass-0.17.2/libass/ass_filesystem.c:325:8: warning: 'add_separator' may be used uninitialized [-Wmaybe-uninitialized]
  325 |     if (add_separator)
      |        ^
../libass-0.17.2/libass/ass_filesystem.c: In function 'ass_open_dir':
../libass-0.17.2/libass/ass_filesystem.c:305:10: note: 'add_separator' was declared here
  305 |     bool add_separator;
```

`add_separator` will never be used uninitialized as the function starts with the assert. But that's not ideal nonetheless.

### DIFF
--- a/libass/ass_filesystem.c
+++ b/libass/ass_filesystem.c
@@ -302,13 +302,16 @@ static bool open_dir_wtf8(ASS_Dir *dir, ASS_StringView path)
     if (!wpath)
         return false;
 
-    bool add_separator;
     WIN32_FIND_DATAW data;
     WCHAR *end = convert_wtf8to16(wpath, path);
-    if (end) {
-        add_separator = append_tail(wpath, end - wpath);
-        dir->handle = FindFirstFileExW(wpath, FindExInfoBasic, &data, FindExSearchNameMatch, NULL, 0);
+    if (!end) {
+        free(wpath);
+        return false;
     }
+
+    bool add_separator = append_tail(wpath, end - wpath);
+    dir->handle = FindFirstFileExW(wpath, FindExInfoBasic, &data, FindExSearchNameMatch, NULL, 0);
+
     free(wpath);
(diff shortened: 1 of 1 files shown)


astiob (commented):
Um, does this really suppress the warning? Now, the goto skips the initializer and the later reference still seems to use an unititialized value until you remember the `assert(dir->handle == INVALID_HANDLE_VALUE)`, so I’d actually expect the warning to stay.

If we want to get rid of the warning _and_ make the code easier to follow to a human, too, I think it may be best to just do:

```c
if (!end) {
    free(wpath);
    return false;
}
```

TheOneric:
> Um, does this really suppress the warning?

it does not, see: https://github.com/libass/libass/actions/runs/9197292825/job/25341567905?pr=769#step:12:22

(the muon failure is unrelated though)

cubicibo:
_facepalm_
That's what I get for using different compilers and configs between finding out and testing.

MrSmile:
I think we only discussed code correctness, but nobody bothered to fix an error that doesn't exist afterwards.

MrSmile (approved):

astiob (approved):
You might want to reword the commit message now that the goto is gone, but anyway, the code seems good to me.

TheOneric:
`^^ looks like you accidentally also rolled back the change to its initial insufficient version while changing the commit message

I took the liberty to restore the change from 73d275bf47bf02a856f280e2fea1efc9b001f5d0 with an adapted commit message and will merge this shortly. Thanks for finally addressing this warning!

PR #773, opened 2024-05-26
ass_parse: use memchr in ass_parse_tags

rcombs:
I don't have a sample that's affected by this, but I'd expect it to be a little faster in some cases relating to clips. Hard to imagine this *hurts*, at any rate.


TheOneric (approved):
there might be a little bit of overhead from the function call, but without testing i’m guessing this doesn't matter too much in the grand scheme of things for small parenthesised sections and even if not vectorised not much at all for long section 

seems good

PR #774, opened 2024-05-26
ci/ghi: use pacman deps for MSVC builds as well

rcombs:
Should save a lot of CI time on rebuilding all the wrapdb stuff.

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -19,10 +19,10 @@ jobs:
             name: Windows MSVC Release,
             os: windows-latest,
             msvc: true,
-            pkgmanager: choco,
+            msystem: MINGW64,
             shell: 'powershell',
             buildtype: release,
-            args: '-Ddefault_library=static --wrap-mode=forcefallback'
+            args: '-Ddefault_library=static'
           }
           - {
             name: Windows MinGW Release,
@@ -57,7 +57,7 @@ jobs:
           fetch-depth: '0'
 
       - name: Setup MSys2
-        if: matrix.config.pkgmanager == 'msys'
+        if: startsWith(matrix.config.os, 'windows')
         uses: msys2/setup-msys2@v2
         with:
           msystem: ${{ matrix.config.msystem }}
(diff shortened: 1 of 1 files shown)


.github/workflows/meson.yml
```diff
@@ -19,7 +19,8 @@ jobs:
             name: Windows MSVC Release,
             os: windows-latest,
             msvc: true,
-            pkgmanager: choco,
+            pkgmanager: msys,
```
TheOneric, line 22: `pkgmanager=msys` is never used iinm and is technically not entirely correct as a combination of choco and msys is used

.github/workflows/meson.yml
```diff
@@ -19,7 +19,8 @@ jobs:
...
+            msystem: MINGW64,
             shell: 'powershell',
             buildtype: release,
             args: '-Ddefault_library=static --wrap-mode=forcefallback'
```
TheOneric, line 26: `--wrap-mode=forcefalback` is now entirely obsolete; though since it still compiles only libass atm i guess this already became obsolete when explicit wrap-fallbacks got removed from our `meson.build`

.github/workflows/meson.yml
```diff
@@ -68,15 +69,27 @@ jobs:
...
+        if: matrix.config.shell == 'powershell'
         run: |
           # for whatever reason choco insist on a reboot if it installs meson
-          choco install ninja nasm python
+          choco install ninja nasm python pkgconfiglite
           python -m pip install --upgrade pip
           pip install meson
 
+      - name: Install dependecies with msys
+        if: matrix.config.shell == 'powershell'
```
TheOneric, line 80: I’m not sure about this and other `== 'powershell'` checks. At the one hand, all current `powershell` checks actually target “compiling with MSVC” and checking the latter would be more idiomatic. On the other hand, _if_ for whatever reason we ever add a “gcc build in powershell” runner it would probably need the same or a very similar treatment.

Another nit, to avoid confusion with native msys2 package installation “Install msys dependencies for PowerShell” might be a better step name

TheOneric:
@rcombs here’s a version of this with review suggestions applied; does it look alright?: https://github.com/TheOneric/libass/commit/110ff5a24748c5fb919227ab022323d5aba06859

As for time savings, we gain a bit more than 3 minutes in the compile step, but now spent an additional minute installing things from MSys2; so it’s a net positive but MSVC still remains by far the slowest CI job

rcombs:
Looks good; I think the msys2 time can probably be mitigated by setting up cacheing (which we'll probably have to do ourselves for this, since we're installing stuff into a custom location). Shouldn't block though, since this is already a net win.

PR #775, opened 2024-05-27
ass_utils: add ass_malloc and ass_free routines

rcombs:
These are simple wrappers around the libc malloc/free routines. Exposing them allows for safe heap usage across shared library boundaries in cases where there may be multiple instances of the libc functions present in the same process (e.g. libc statically linked into a shared lib). This mostly affects Windows.

### DIFF
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -24,7 +24,7 @@
 #include <stdarg.h>
 #include "ass_types.h"
 
-#define LIBASS_VERSION 0x01702010
+#define LIBASS_VERSION 0x01702020
 
 #ifdef __cplusplus
 extern "C" {
@@ -806,6 +806,22 @@ void ass_clear_fonts(ASS_Library *library);
  * \return timeshift in milliseconds
  */
 long long ass_step_sub(ASS_Track *track, long long now, int movement);
+
+/**
+ * \brief Allocates memory that can be safely freed by libass later.
+ * Use this to allocate buffers you'll use to manually modify ASS_Track events
+ * or related structures.
+ * \param size number of bytes to allocate
+ * \return pointer to the allocated buffer, or NULL on failure
+ */
+void *ass_malloc(size_t size);
+
+/**
+ * \brief Releases memory previously allocated within libass.
+ * Use this to free memory allocated using ass_malloc.
+ * \param ptr pointer to the buffer to release; passing NULL is a no-op
+ */
+void ass_free(void *ptr);
 
 #undef ASS_DEPRECATED
 #undef ASS_DEPRECATED_ENUM
--- a/libass/Makefile_library.am
+++ b/libass/Makefile_library.am
@@ -1,6 +1,6 @@
-LIBASS_LT_CURRENT = 11
-LIBASS_LT_REVISION = 2
-LIBASS_LT_AGE = 2
+LIBASS_LT_CURRENT = 12
+LIBASS_LT_REVISION = 0
+LIBASS_LT_AGE = 3
 
 .asm.lo:
 	$(nasm_verbose)$(LIBTOOL) $(AM_V_lt) --tag=CC --mode=compile $(top_srcdir)/ltnasm.sh $(AS) $(ASFLAGS) -I$(top_srcdir)/libass/ -Dprivate_prefix=ass -o $@ $<
--- a/libass/ass_library.c
+++ b/libass/ass_library.c
@@ -158,3 +158,13 @@ void ass_set_message_cb(ASS_Library *priv,
         priv->msg_callback_data = data;
     }
 }
+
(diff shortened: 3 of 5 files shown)


astiob (approved):

libass/ass.h
```diff
@@ -24,7 +24,7 @@
 #include <stdarg.h>
 #include "ass_types.h"
 
-#define LIBASS_VERSION 0x01702000
+#define LIBASS_VERSION 0x01702010
```
TheOneric, line 27: This bump is exactly as it should be; but unless we want to include them in the hotfix 0.17.3 merging this now will again bring up the question of how to name the release.

`LIBASS_VERSION` has the lowest nibble reserved for hotfixes, so it’s technically fine — but we have no scheme to map this to public release numbers and since usually `LIBASS_VERSION` corresponds to the public release number, naming `0x01702001` 0.17.3 will be a bit confusing

TheOneric (commented):
Code seems good, probably should be advertised more in docs though

Possible conflict with a Fonctconfig-unregression release though

libass/ass.h
rcombs: I'm not in any rush with this; happy to delay if needed. Alternately, if we want to just include it in the hotfix (seems fairly low-risk?), we can increment to `0x01703000` at that time.

libass/ass.h
```diff
@@ -802,6 +802,22 @@ void ass_clear_fonts(ASS_Library *library);
  */
 long long ass_step_sub(ASS_Track *track, long long now, int movement);
 
+/**
+ * \brief Allocates memory that can be safely freed by libass later.
+ * Use this to allocate buffers you'll use to manually modify ASS_Track events
+ * or related structures.
```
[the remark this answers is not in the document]
rcombs: How's this?
```
 * If modifying a libass struct to contain a pointer to a user-allocated buffer,
 * the buffer should be allocated using ass_malloc(). Failing to do so may result
 * in undefined behavior if the buffer is later freed within libass, and the caller
 * is linked against a different libc instance than libass itself.
```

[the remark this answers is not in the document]
astiob: Looks fine, but looking at the latest commit, that may not be the best place to put it. The next sentence mentions “this possibility”, which is meant to reference the ability to edit structs at all but now sounds like something about different libcs.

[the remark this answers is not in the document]
TheOneric: In a pedantic reading this doesn’t require users to free already present data they replace with `ass_free`, but this can most likely be inferred.

TheOneric:
I agree with the changes just documenting previously unmentioned pitfalls and not really imposing new restrictions *(brought up on IRC)*

libass/ass.h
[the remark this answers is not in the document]
rcombs: Moved to end of paragraph; added note re: `ass_free`.

libass/ass_utils.c
```diff
@@ -271,3 +271,13 @@ int ass_lookup_style(ASS_Track *track, char *name)
             track, name, track->styles[i].Name);
     return i;
 }
+
+void *ass_malloc(size_t size)
```
TheOneric, line 275: nit: didn’t notice before, but i think this fits better into `ass_library.c`. We have several other public API functions there which don't belong to a more specific file, while `ass_utils` is more for internal utilities

libass/ass.h
TheOneric: so current non-cosmetic, non-CI changes from 0.17.2 are:
 - upgrading x86inc to be more generic about calls (replacing checkasm-specific Haiku fix)
 - explicit AC_CANONICAL_HOST
 -  use memchr for pasing tags
 - fontconfig refcounting and ordering

And to be added at least: fix shared-only checkasm

Which seems fine for tagging things on `master` again rather than a release branch. In that case this and `-fno-errno-math` are prob fine too for emrge ahead of 0.17.2

libass/ass_utils.c
astiob: Agreed. I guess the counterpoint is that unlike other public API functions, this one is actually used internally throughout the library?

TheOneric: We just use `(m|c)alloc`  and `free` directly internally; not `ass_*`

TheOneric:
@rcombs: pushed a rebased version with `ass_free`/`malloc` moved to `ass_library.c`
Feel free to merge if it looks good or revert to https://github.com/libass/libass/commit/fa1f2ee5e083b63186246a6ce2f288521ce07b5c if you don’t like the changes

rcombs:
Fine by me; I think the only question is whether we want to hold until after the release hotfixing the fontconfig issue?

TheOneric:
Should be fine, especially now that API constification was already merged

Issue #770, opened 2024-05-22
Rendering: Incorrect font variant being selected


neeshy:
### Screenshots

mpv w/libass 0.17.1
![Wolf Children (2012)  1080p x265 HEVC 10bit BluRay Dual Audio AAC 5 1   Prof mkv_snapshot_00:01:00_ 2024 05 22_16 36 24](https://github.com/libass/libass/assets/60193883/a79fea42-d578-42c7-bae0-5f7bc06b669d)

mpv w/ libass 0.17.2
![Wolf Children (2012)  1080p x265 HEVC 10bit BluRay Dual Audio AAC 5 1   Prof mkv_snapshot_00:01:00_ 2024 05 22_16 37 38](https://github.com/libass/libass/assets/60193883/30966fbf-bd5f-4556-a84d-7b1e8d487db8)

### Description of the issue

Starting with version 0.17.2, alternative versions of fonts are erroneously being selected. I've configured fontconfig to assign `sans-serif` to `Noto Sans` and up until recently this would have selected the Regular variant. Now, the Condensed variant is being selected. I'm not sure if this is an issue with how mpv is invoking libass.

### libass version

0.17.2

### Is it a regression?

Yes. The regression originates with commit 3f153be286934f66d0f73f93eb0bb90fa1a14918.

### Special Fonts

The issue doesn't depend on a specific font

### System Information

Linux on x86-64

### Log

mpv w/ libass 0.17.1
`[osd/libass] fontselect: (sans-serif, 400, 0) -> /usr/share/fonts/noto/NotoSans-Regular.ttf, 0, NotoSans-Regular`

mpv w/ libass 0.17.2
`[osd/libass] fontselect: (sans-serif, 400, 0) -> /usr/share/fonts/noto/NotoSans-Condensed.ttf, 0, NotoSans-Condensed`

kimhbryan:
I'm experiencing the exact same thing, and can confirm the issue seems like a regression with commit https://github.com/libass/libass/commit/3f153be286934f66d0f73f93eb0bb90fa1a14918.

I built mpv with mpv-build on libass commit https://github.com/libass/libass/commit/0a63b95ce6dbe6fc3a6470c8b88b242341c8f961 and https://github.com/libass/libass/commit/3f153be286934f66d0f73f93eb0bb90fa1a14918, and fontselect exhibits this incorrect behaviour starting from the latter.

And I also do have fontconfig assigning 'sans-serif' to Noto Sans Regular.
```
[...]

PR #777, opened 2024-05-28
fontconfig: sort full system font list by width

TheOneric:
Temporary “at least not worse than before” workaround for #770 until we properly fix this in a larger overhaul.

Other approaches were also discussed on IRC but turned out to create new problems or make existing problems (e.g. related to fontconfig substitutions) worse.

<details>
<summary>Debug Logs</summary>

```diff
diff --git a/libass/ass_fontconfig.c b/libass/ass_fontconfig.c
index 924a78d5..e529ce11 100644
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -117,9 +117,41 @@ static void merge_sorted_segments(FcPattern **data, FcPattern **insert,
     }
 }
 
+static void dbg_order(FcPattern **fonts, size_t n)
+{
+    for (size_t i = 0; i < n; ++i) {
+        FcPattern *pat = fonts[i];
+        char *name = NULL;
+
+        if (FcPatternGetString(pat, FC_FAMILY, 0, (FcChar8 **) &name) != FcResultMatch)
+            continue;
+        if (strncmp(name, "Noto Sans", 9))
+            continue;
+
+        int slant;
+        double weight;
+        if (FcPatternGetInteger(pat, FC_SLANT, 0, &slant) != FcResultMatch)
+            slant = -1;
+        if (FcPatternGetDouble(pat, FC_WEIGHT, 0, &weight) != FcResultMatch)
+            weight = -1;;
+
+        printf("%4zu: %s (%.3f, %d)", i, name, weight, slant);
+
+        int idx = 1;
+        while (FcPatternGetString(pat, FC_FAMILY, idx++, (FcChar8 **)&name) == FcResultMatch)
+            printf(", %s", name);
+        printf("\n");
+    }
+}
+
 // stable sort to preserve original order for equally regular fonts
 static void sort_fonts(FcPattern **fonts, size_t n)
 {
+    // DBG
+    printf("Before sorting (%4zu):\n", n);
+    printf("======================\n");
+    dbg_order(fonts, n);
+
     // loop var and malloc arg overflows
     if (n > SIZE_MAX / FFMAX(sizeof(*fonts), 2))
         return;
@@ -130,6 +162,11 @@ static void sort_fonts(FcPattern **fonts, size_t n)
 
     FcPattern **data_buf = fonts;
 
+    printf("\n\n\n");
+    printf("font list: %p\n", fonts);
+    printf("tmp space: %p\n", insert_buf);
+    printf("Sorting now...\n");
+
     for (size_t w = 1; w < n; w <<= 1) {
         for (size_t i = 0; i < n; i += w << 1)
             merge_sorted_segments(data_buf, insert_buf, i,
@@ -140,12 +177,21 @@ static void sort_fonts(FcPattern **fonts, size_t n)
         insert_buf = tmp;
     }
 
+    printf("Done. Cleaning up...\n");
+    printf("insert_buf: %p\n", insert_buf);
+    printf("data_buf:   %p\n", data_buf);
     if (data_buf != fonts) {
         memcpy(fonts, data_buf, n);
         free(data_buf);
     } else {
         free(insert_buf);
     }
+
+    // DBG
+    printf("\n\n\n");
+    printf("After sorting:\n");
+    printf("==============\n");
+    dbg_order(fonts, n);
 }
 
 static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)

```
</details>

### DIFF
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -83,6 +83,68 @@ static void destroy(void *priv)
     free(fc);
 }
 
+static int compare_patterns(const FcPattern *p1, const FcPattern *p2)
+{
+    // Temporary workaround for fontconfig family name matching being too lenient.
+    // Ensure regular variants come first, so if the score is tied, they'll be selected.
+    // Weight and slant already factor into scoring for regular selection and fallback
+    // selection uses a different list already sorted by FcFontSort; thus just width suffices.
+    FcResult res;
+    int wi1, wi2;
+    res = FcPatternGetInteger(p1, FC_WIDTH, 0, &wi1);
+    res |= FcPatternGetInteger(p2, FC_WIDTH, 0, &wi2);
+    if (res != FcResultMatch)
+        return 0;
+    wi1 = abs(wi1 - FC_WIDTH_NORMAL);
+    wi2 = abs(wi2 - FC_WIDTH_NORMAL);
+    return wi1 - wi2;
+}
+
+static void merge_sorted_segments(FcPattern **data, FcPattern **insert,
+                                  size_t start, size_t mid, size_t end)
+{
+    size_t s1 = start;
+    size_t s2 = mid;
+
+    for (size_t i = start; i < end; i++) {
+        if (s1 < mid && (s2 >= end || compare_patterns(data[s1], data[s2]) <= 0))
+            insert[i] = data[s1++];
+        else
+            insert[i] = data[s2++];
+    }
+}
+
+// stable sort to preserve original order for equally regular fonts
+static void sort_fonts(FcPattern **fonts, size_t n)
+{
+    // loop var and malloc arg overflows
+    if (n > SIZE_MAX / FFMAX(sizeof(*fonts), 2))
+        return;
+
+    FcPattern **insert_buf = malloc(n * sizeof(*fonts));
+    if (!insert_buf)
+        return;
+
+    FcPattern **data_buf = fonts;
+
+    for (size_t w = 1; w < n; w <<= 1) {
+        for (size_t i = 0; i < n; i += w << 1)
+            merge_sorted_segments(data_buf, insert_buf, i,
+                    FFMIN(n, i + w), FFMIN(n, i + (w<<1)));
+
+        FcPattern **tmp = data_buf;
+        data_buf = insert_buf;
+        insert_buf = tmp;
+    }
+
+    if (data_buf != fonts) {
+        memcpy(fonts, data_buf, n * sizeof(*fonts));
+        free(data_buf);
+    } else {
+        free(insert_buf);
+    }
+}
+
 static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
 {
     int i;
@@ -91,6 +153,9 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
 
     // get list of fonts
     fonts = FcConfigGetFonts(config, FcSetSystem);
+
+    // sort list to prefer regular variants
+    sort_fonts(fonts->fonts, fonts->nfont);
 
     // fill font_info list
     for (i = 0; i < fonts->nfont; i++) {


libass/ass_fontconfig.c
```diff
@@ -83,6 +83,71 @@ static void destroy(void *priv)
...
+    res = FcPatternGetInteger(p1, FC_WIDTH, 0, &wi1);
+    res |= FcPatternGetInteger(p2, FC_WIDTH, 0, &wi2);
+    if (res != FcResultMatch)
+        return 0;
+    wi1 = abs(wi1 - FC_WIDTH_NORMAL);
+    wi2 = abs(wi2 - FC_WIDTH_NORMAL);
+    if (wi1 != wi2)
+        return wi1 - wi2;
+
+    return 0;
```
astiob, line 103: Nit: this if/return/return seems silly and could just be `return wi1 - wi2;`

It might be intended to allow adding more sort criteria later, but the comment suggests we shouldn’t ever have any more.

libass/ass_fontconfig.c
```diff
@@ -83,6 +83,71 @@ static void destroy(void *priv)
     free(fc);
 }
 
+static int compare_patterns(const FcPattern *p1, const FcPattern *p2)
+{
+    // Temporary workaround for fontconfig family name matching being too lenient.
+    // Ensure regular variants come first, so if the score is tied, they'll be selected.
+    // Weight and slant alread y factor into scoring for regular selection and fallback
```
rcombs, line 90: typo: already

libass/ass_fontconfig.c
```diff
@@ -83,6 +83,71 @@ static void destroy(void *priv)
...
+            merge_sorted_segments(data_buf, insert_buf, i,
+                    FFMIN(n, i + w), FFMIN(n, i + (w<<1)));
+
+        FcPattern **tmp = data_buf;
+        data_buf = insert_buf;
+        insert_buf = tmp;
+    }
+
+    if (data_buf != fonts) {
+        memcpy(fonts, data_buf, n);
```
MrSmile, line 144: Missing `sizeof`.

astiob: should be `n * sizeof(*fonts)`

TheOneric: fixed, thanks

rcombs (commented):
Why not just use `FcFontSort` (with a default pattern) instead of `FcConfigGetFonts`?

libass/ass_fontconfig.c
```diff
@@ -92,6 +157,9 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
     // get list of fonts
     fonts = FcConfigGetFonts(config, FcSetSystem);
 
+    // temporary workaround: sort list to prefer regular variants
```
astiob, line 160: Not sure how useful this comment is. Might be better either to paste the explanation from the commit message (perhaps shortened, perhaps not) or to remove the “temporary workaround” part.

libass/ass_fontconfig.c
```diff
@@ -83,6 +83,71 @@ static void destroy(void *priv)
...
+    wi1 = abs(wi1 - FC_WIDTH_NORMAL);
+    wi2 = abs(wi2 - FC_WIDTH_NORMAL);
+    if (wi1 != wi2)
+        return wi1 - wi2;
+
+    return 0;
+}
+
+static void merge_sorted_segments(FcPattern **data, FcPattern **insert,
+                                size_t start, size_t mid, size_t end)
```
astiob, line 107: Nit: indentation mismatch

TheOneric:
> Why not just use FcFontSort (with a default pattern) instead of FcConfigGetFonts?

summarising from earlier IRC discussion: `FcFontSort` may introduce other unwanted sorting criteria and maybe completely eliminate some unfit entries

astiob (commented):
How’s the performance of the repeated `FcPatternGetInteger`s? If appropriate, we could do a single linear pass to read out all the widths and base the sort on those.

(Let’s see what GitHub does to my inline comment that is already outdated and has disappeared from my own UI but seemingly continues to be counted on the “Finish your review” button’s badge…)

astiob:
> Why not just use `FcFontSort` (with a default pattern) instead of `FcConfigGetFonts`?

> `FcFontSort` may introduce other unwanted sorting criteria and maybe completely eliminate some unfit entries

I don’t think there can be any unwanted criteria here (in fact, if Fontconfig does add more ordering preferences like, say, the system locale, that sounds desirable to me!), but I do fear that it may eliminate some fonts completely. If anyone has a default pattern that is guaranteed to return all fonts, I’d welcome that.

libass/ass_fontconfig.c
TheOneric: fixed

libass/ass_fontconfig.c
TheOneric: yeah, it’s a leftover from other tested versions with more criteria; shortened it now

libass/ass_fontconfig.c
TheOneric: dropped "temporary workaround"

libass/ass_fontconfig.c
TheOneric: fixed

MrSmile:
> If appropriate, we could do a single linear pass to read out all the widths and base the sort on those.

Actually, we can sort (width, index) pairs — that eliminates the need to use stable sort, so usual `qsort` is sufficient.

TheOneric:
superseded by #778

Issue #737, opened 2024-03-09
Rendering: Wrong font used for mpv OSD on Fedora 39+


luk1337:
### Screenshots

See https://github.com/mpv-player/mpv/issues/13396

### Description of the issue

After [this](https://src.fedoraproject.org/rpms/google-noto-fonts/c/609f6bc1323476289beda736ce3912b73b8900ad?branch=rawhide) change on Fedora end, mpv is using "Noto Sans Arabic" instead of regular "Noto Sans" for rendering numbers in its OSD.

---

I originally reported it [here](https://github.com/mpv-player/mpv/issues/13396), then moved to [RedHat Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2262410) where the person handling my bug report said that it actually originated from libass, thus I'm reporting it here as well.

### libass version

0.17.1-4.fc40

### Is it a regression?

No

### Special Fonts

I uploaded or linked to the required font

### System Information

See above.

astiob:
You said in the mpv issue that you were seeing issues in other applications besides mpv. It would be very good to mention this on the Fedora tracker.

luk1337:
> You said in the mpv issue that you were seeing issues in other applications besides mpv. It would be very good to mention this on the Fedora tracker.

Thanks for the reminder, I mentioned it.

PR #778, opened 2024-05-29
fontconfig: query a sorted list

TheOneric:
Based on IRC discussion an alternative to #777.
Posted as a new PR to make it easier to compare both approaches

Notably this adds language as a sorting parameter, which our existing `cache_fallbacks` logic makes an explicit effort to remove, but discussion in the RedHat issue linked from #737 suggests we might actually want this.
Also this omits any `FcConfigSubstitute` calls even though Fontconfig docs require it

Given the question of efficiency (re repeated width queries) was raised over in #777: i’d expect, but haven’t verified, this approach to be more perf costly given it does more (than we actually need). Not sure if the impact is relevant though

### DIFF
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -94,9 +94,18 @@ static bool scan_fonts(FcConfig *config, ASS_FontProvider *provider)
     FcFontSet *fonts;
     ASS_FontProviderMetaData meta = {0};
 
-    // get list of fonts
-    fonts = FcConfigGetFonts(config, FcSetSystem);
-    if (!fonts)
+    // get list of fonts;
+    // sorting by default pattern prefers regular variants
+    FcPattern *pat = FcPatternCreate();
+    if (!pat)
+        return false;
+
+    FcDefaultSubstitute(pat);
+    FcResult res;
+    // trim=FcFalse returns all system fonts
+    fonts = FcFontSort(config, pat, FcFalse, NULL, &res);
+    FcPatternDestroy(pat);
+    if (res != FcResultMatch)
         return false;
 
     // fill font_info list
@@ -200,6 +209,7 @@ static bool scan_fonts(FcConfig *config, ASS_FontProvider *provider)
         ass_font_provider_add_font(provider, &meta, path, index, (void *)pat);
     }
 
+    FcFontSetDestroy(fonts);
     return true;
 }
 
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -351,7 +351,7 @@ ass_fontconfig_add_provider(ASS_Library *lib, ASS_FontSelector *selector,
         rc = FcConfigBuildFonts(fc->config);
 
     if (!rc || !fc->config) {
-        ass_msg(lib, MSGL_FATAL,
+        ass_msg(lib, MSGL_ERR,
                 "No valid fontconfig configuration found!");
         FcConfigDestroy(fc->config);
         free(fc);
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -71,6 +71,11 @@ static bool check_glyph(void *priv, uint32_t code)
     return false;
 }
 
+static void destroy_font(void *priv)
+{
+    FcPatternDestroy((FcPattern *) priv);
+}
+
 static void destroy(void *priv)
 {
     ProviderPrivate *fc = (ProviderPrivate *)priv;
@@ -189,6 +194,7 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
         if (result != FcResultMatch)
             meta.postscript_name = NULL;
 
+        FcPatternReference(pat);
         ass_font_provider_add_font(provider, &meta, path, index, (void *)pat);
     }
 }
@@ -314,6 +320,7 @@ cleanup:
 static ASS_FontProviderFuncs fontconfig_callbacks = {
     .check_postscript   = check_postscript,
     .check_glyph        = check_glyph,
+    .destroy_font       = destroy_font,
     .destroy_provider   = destroy,
     .get_substitutions  = get_substitutions,
     .get_fallback       = get_fallback,
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -88,7 +88,7 @@ static void destroy(void *priv)
     free(fc);
 }
 
-static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
+static bool scan_fonts(FcConfig *config, ASS_FontProvider *provider)
 {
     int i;
     FcFontSet *fonts;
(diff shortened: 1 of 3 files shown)


libass/ass_fontconfig.c
```diff
@@ -191,6 +202,11 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
 
         ass_font_provider_add_font(provider, &meta, path, index, (void *)pat);
     }
+
+    // FcFontSort's, but not FcConfigGetFOnt's!, FOntSet needs to be cleaned up explicitly
```
rcombs, line 206: typos: FcConfigGetFont's; FontSet

libass/ass_fontconfig.c
```diff
@@ -191,6 +202,11 @@ static void scan_fonts(FcConfig *config, ASS_FontProvider *provider)
...
+    if (res == FcResultMatch) {
+        FcFontSetDestroy(fonts);
```
rcombs, line 208: Digging into this, I'm arriving at the conclusion that this code is currently technically safe, but only by chance: `FcFontSetDestroy` unrefs all the `FcPattern`s that we've just inserted into our font list. All of those patterns ultimately come straight out of our `FcConfig`, which will hold a ref to them until we exit, but I don't think there's any guarantee that that'll always be true (and it's also kinda fragile even in the existing code). Instead, we should either:
- skip destroying `fonts` here and instead hold onto it in `ProviderPrivate` until `destroy` is called, or
- `FcPatternReference` each pattern before calling `ass_font_provider_add_font`, and introduce a `destroy_font` callback that releases it

I'm slightly partial to option 2, since it'd lay some nice groundwork for a lazy-loading fontconfig provider future, and ensures we're not relying on any potentially-non-guaranteed lifetimes in the `FcConfigGetFonts` fallback case.

rcombs (commented):
A couple notes; I'm not inclined to block merge on any of this, but I'm mildly concerned about lifetimes here.

libass/ass_fontconfig.c
TheOneric: very good point; implemented referencing and `destroy_font`

libass/ass_fontconfig.c
```diff
@@ -71,6 +71,11 @@ static bool check_glyph(void *priv, uint32_t code)
     return false;
 }
 
+static void destroy_font(void *priv)
```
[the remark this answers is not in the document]
TheOneric: ah that explains it; thanks

also gosh I’m really making quite a lot of typos in those two PRs

rcombs (approved):

libass/ass_fontconfig.c
```diff
@@ -88,14 +88,25 @@ static void destroy(void *priv)
...
 {
     int i;
     FcFontSet *fonts;
     ASS_FontProviderMetaData meta = {0};
 
-    // get list of fonts
-    fonts = FcConfigGetFonts(config, FcSetSystem);
+    // get list of fonts;
+    // sorting by default pattern prefers regular variants
+    FcPattern *pat = FcPatternCreate();
```
astiob, line 99: Can this fail and return `NULL`?

libass/ass_fontconfig.c
```diff
@@ -88,14 +88,25 @@ static void destroy(void *priv)
...
+    FcDefaultSubstitute(pat);
+    FcResult res;
+    // trim=FcFalse returns all system fonts
+    fonts = FcFontSort(config, pat, FcFalse, NULL, &res);
```
astiob, line 103: Can this return `NULL` without also setting `res != FcResultMatch`?

libass/ass_fontconfig.c
```diff
@@ -88,14 +88,25 @@ static void destroy(void *priv)
...
+    FcPatternDestroy(pat);
+    if (res != FcResultMatch) {
+        // fallback to unsorted full list
+        ass_msg(lib, MSGL_WARN, "fontconfig: failed to sort list of system fonts!");
+        fonts = FcConfigGetFonts(config, FcSetSystem);
```
astiob, line 108: Can this return `NULL`?

astiob (commented):
I’ve left a few comments about the apparent lack of null checks. It seems like the existing code may have already lacked them, but this could be a good opportunity to add them.

libass/ass_fontconfig.c
TheOneric: no mention in docs, but yes indeed it can, good catch!
This also affects `cache_fallbacks`

libass/ass_fontconfig.c
TheOneric: I’d expect no, since why else have result return value and looking at current code seems to support this. However one peculiar comment suggests at least the underlying `FcFontSetSort` may have done so in the past:
```C
 /* There are some implementation that relying on the result of
 * "result" to check if the return value of FcFontSetSort
 * is valid or not.
 * So we should initialize it to the conservative way since
 * this function doesn't return NULL anymore.
 */
```

libass/ass_fontconfig.c
TheOneric: again not documented, but upon checking the code, yes if and only if `config` is `NULL` and obtaining/creating the default config fails.

However there *are* somet other relevant bits in docs:

> FcConfigReference must be called before using this  and  then  FcConfigDestroy when the return value is no longer referenced.

which we didn't before and still do after, oops
(The function is already explicitly documented as not MT safe, but i believe pending threading patches already take care of this anyway)

TheOneric: Event hough docs technically require it, i think we don't actually need it since we hold on to the config object anyway even without refs. But either way, given it too can fail (and it makes cleanup more complicated), using it as a fallback no longer seemed to provide much benefit to me, so i dropped it entirely now

libass/ass_fontconfig.c
TheOneric: This comment was introduced in [fontconfig@a18ca17b](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/commit/a18ca17b6211f62fbd1d893811b94b8c83db4cc0) and released with 2.9.0 *(twelve years ago)*.
We require fontconfig `>= 2.10.92` and therefore can rely on `res`.

libass/ass_fontconfig.c
```diff
@@ -362,7 +372,8 @@ ass_fontconfig_add_provider(ASS_Library *lib, ASS_FontSelector *selector,
     provider = ass_font_provider_new(selector, &fontconfig_callbacks, fc);
 
     // build database from system fonts
-    scan_fonts(fc->config, provider);
+    if (!scan_fonts(fc->config, provider))
+        ass_msg(lib, MSGL_FATAL, "Failed to load fonctconfig fonts!");
```
astiob, line 386: Not a blocker:

This matches the level of the “No valid fontconfig configuration found!” message, but perhaps both should be `MSGL_ERROR`, seeing as libass continues to function, just without system fonts?

astiob (approved):

libass/ass_fontconfig.c
TheOneric: done

PR #779, opened 2024-06-01
Deal with shared-only checkasm

TheOneric:
Take this PR more as a place for discussion about the right approach rather than pushing for the approach of the current patch

The simple way to fix this: just don’t run `checkasm` if there’s no static lib. *(what this PR currently does)*

iirc, originally (before merge) all relevant libass source files were also included in checkasm sources and there was no library link. This would have avoided this issue but introduced possible mismatches between lib and checkasm compile flags and thus changed.

To avoid both we could build just the bitmap engine + assembly as a always static-only internal lib and reuse it for checkasm. ~~However, at first glance I’m not sure how to tell libtool to build a library always static and if we use regular automake static libs, some flags (like PIC) might not be set automatically for source files of this internal static lib... *(but on the upside, i believe we could drop `ltnasm.sh`)*~~
`noinst_LTLIBRARIES` ~~or `EXTRA_LTLIBRARIES`~~ maybe work?

### DIFF
--- a/checkasm/checkasm.h
+++ b/checkasm/checkasm.h
@@ -30,7 +30,7 @@
 
 #include "config.h"
 
-#include "ass_bitmap.h"
+#include "ass_bitmap_engine.h"
 
 #include <stdint.h>
 #include <stdlib.h>
--- a/Makefile.am
+++ b/Makefile.am
@@ -20,6 +20,7 @@ nasm_verbose_ = $(nasm_verbose_$(AM_DEFAULT_VERBOSITY))
 nasm_verbose_0 = @echo "  NASM    " $@;
 
 lib_LTLIBRARIES =
+noinst_LTLIBRARIES =
 noinst_PROGRAMS =
 check_PROGRAMS =
 TESTS =
--- a/Makefile_util.am
+++ b/Makefile_util.am
@@ -54,7 +54,8 @@ checkasm_checkasm_SOURCES = \
     checkasm/blend_bitmaps.c \
     checkasm/be_blur.c \
     checkasm/blur.c \
-    checkasm/checkasm.h checkasm/checkasm.c
+    checkasm/checkasm.h checkasm/checkasm.c \
+    libass/ass_rasterizer.h libass/ass_utils.h
 
 checkasm_checkasm_CPPFLAGS = -I$(top_srcdir)/libass
 checkasm_checkasm_LDADD = libass/libass.la
--- a/Makefile_util.am
+++ b/Makefile_util.am
@@ -58,7 +58,7 @@ checkasm_checkasm_SOURCES = \
     libass/ass_rasterizer.h libass/ass_utils.h
 
 checkasm_checkasm_CPPFLAGS = -I$(top_srcdir)/libass
-checkasm_checkasm_LDADD = libass/libass.la
+checkasm_checkasm_LDADD = libass/libass_internal.la
 checkasm_checkasm_LDFLAGS = $(AM_LDFLAGS) -static
 
 if X86
--- a/checkasm/be_blur.c
+++ b/checkasm/be_blur.c
@@ -18,6 +18,8 @@
 
 #include "checkasm.h"
 
+#include <string.h>
+
 #define HEIGHT 8
 #define STRIDE 64
 #define MIN_WIDTH 2
--- a/checkasm/blend_bitmaps.c
+++ b/checkasm/blend_bitmaps.c
@@ -18,6 +18,8 @@
 
 #include "checkasm.h"
 
+#include <string.h>
+
 #define HEIGHT 8
 #define DST_STRIDE 64
 #define MIN_WIDTH  1
--- a/checkasm/blur.c
+++ b/checkasm/blur.c
@@ -16,7 +16,10 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+#include "ass_utils.h"
 #include "checkasm.h"
+
+#include <string.h>
 
 #define HEIGHT 13
 #define STRIDE 64
--- a/checkasm/checkasm.c
+++ b/checkasm/checkasm.c
@@ -25,6 +25,7 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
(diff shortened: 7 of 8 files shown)


TheOneric:
`noinst_LTLIBRARIES` provides the automatic matching flags and static-only output we need for checkasm to actually always build and run successfully. Updated commit to use that.

Since it also uses types from other libass headers, the split is a bit akward though.
Alternatives:
  - just don't run checkasm in shared-only builds ([`19acb3e`](https://github.com/libass/libass/commit/19acb3e61461c6f89869d74d338b725a6c203037))
  - wrap _all libass sources_ into a `noinst_LTLIBRARIES` and pass this around and make the actual libass library an otherwise empty wrapper around it

MrSmile:
Why even `ass_bitmap.c` in that library? I think for this approach we should separate bitmap engine more:
- move `struct segment` into `ass_bitmap_engine.h` to eliminate dependency on `ass_rasterizer.h` (and `ass_bitmap.h`, `ass_outline.h`, etc. within it);
- manually inline `FFMIN` in `c_blend_bitmaps.c` to eliminate dependency on `ass_utils.h`.

Maybe we don't even need `ass_compat.h` for bitmap functions?

rcombs:
It doesn't matter what .h files are used, though?

TheOneric:
> Why even ass_bitmap.c in that library?

checkasm included `ass_bitmap.h`, but as it turns out it doesn't actually need it so i reduced it to `ass_bitmap_engine.h` now.

This cuts down headers included in both libass and the bitmapengine lib to
- `ass.h` *(definitions in libass)*
- `ass_utils.h` *(definitions in bitmapengine)*
- `ass_compat.h` *(mostly macro defines, but fallback functions defined in `ass_utils.h`, i.e. bitmapengine)*

which seems reasonable enough to me

Additionally checkasm needs types from `ass_rasterizer.h`

TheOneric:
Actually `ass_utils.c` includes more headers i missed before as well as a vendored declaration:
```C
/* defined in ass_strtod.c */
double ass_strtod(const char *string, char **endPtr);
```

I guess this currently works becuase relevant things don'T end up used in checkasm, but this needs more work afterall if we go with this approach

rcombs:
Including headers that declare functions that won't be linked is harmless as long as they aren't used.

TheOneric:
yeah, but e.g. `ass_strtod` gets used and isn’t a public symbol; checkasm just doesn't use the function which uses `ass_strtod` so omitting this would be fragile wrt to future changes and it needs to be part of libbitmapengine. Haven’t checkd what’s used from other headers yet

headers only included for macros or types are fine, but prob should be specified in the source list (with dependency tracking it might even be fine (after the first compilation pass) without listing any headers, but without dep tracking it might cause problems and explicit listing is what we currently do already)

MrSmile:
I'm certainly against putting unrelated things into bitmap engine library. The only used things from `ass_utils.h` are `FFMIN/MAX` macros, so bundling the whole `ass_utils.c` looks like overkill. Sadly, that macros are also used in `rasterizer_template.h`, so inlining them would be quite cumbersome. Maybe it would be better to move them into separate `ass_minmax.h` header?

And `struct segment` should definitely be moved into `ass_bitmap_engine.h`. Maybe it's not perfectly logically sound, but I think it's much less evil overall.

> Including headers that declare functions that won't be linked is harmless as long as they aren't used.

Maybe it's harmless, but it's certainly more confusing and error-prone if that header changed.
So I vote for removing `#include "ass_compat.h"` from bitmap engine files too.

astiob:
Frankly, I don’t really see the point of spending all this effort on splitting; I’m inclined to recommend this if it’s easier:

> wrap _all libass sources_ into a `noinst_LTLIBRARIES` and pass this around and make the actual libass library an otherwise empty wrapper around it

MrSmile:
> Frankly, I don’t really see the point of spending all this effort on splitting

It's much simpler to work with assembly for obscure architectures. It's why I've done that for initial checkasm tests in the first place.

astiob:
I… don’t follow. How are assembly architectures affected by which C files are included in the internal .a file?

MrSmile:
> How are assembly architectures affected by which C files are included in the internal .a file?

Checkasm can be built as a single self-contained executable without any dependencies.  So I can simply run it with `qemu` without the need to setup the whole virtual machine with full libass dependencies.

TheOneric:
> So I vote for removing #include "ass_compat.h" from bitmap engine files too.

`ass_compat.h` sets things up such that the same set of C features and (POSIX) functions are available everywhere.
Omitting this and having to always remember to avoid e.g. `inline` in bitmapengine else some compilers blow up is imho a bad idea and will only cause pain in the long run. *At least* ass_compat.h and all fallback functions ought to be available. I’m also not a fan of duplicating internal utility functions just to avoid some extra file in `libbitmapengine`; especially if `FFMIN`/`MAX` will get more complex as suggested in the math-flag discussion

> Checkasm can be built as a single self-contained executable without any dependencies.

Note that the current splitting doesn't enable this; for this to work adjustments to `libbitmapengine`’s CFLAGS and linkflags as well as configure chnges would be needed.  
I remain sceptical how much value there is in testing only assembly routines on a new/exotic platform without verifying everything else too builds and works as intended

TheOneric:
Switched to “just put all sources into the internal static lib” approach. This should basically be the same as what we had before just with one extra (static) link step and `checkasm` now working with `--disable-static`. Imho this should be good enough for 0.17.3 and if we want to split things further but can't find a consensuns right now we can still discuss and change this later.

*If* we want to split things up then as stated above i believe `ass_compat.h` should still be used everywhere. We could probably split fallback function definitions and the subset of `ass_utils` the bitmapengine needs into `ass_utils_minimal` or similar and include `ass_utils_minimal.h` in the full `ass_utils.h`.

checkasm/checkasm.h
```diff
@@ -30,7 +30,7 @@
 
 #include "config.h"
 
-#include "ass_bitmap.h"
+#include "ass_bitmap_engine.h"
```
astiob, line 33: left over in the latest push

TheOneric: i left it intentionally since regardless of splitting or not, checkasm doesn't use anything from `ass_bitmpa.h`; though if desired i can drop it anyway

PR #781, opened 2024-06-04
configure.ac: add missing AC_CANONICAL_HOST

orbea:
Fixes detection for nasm when using slibtoolize instead of GNU libtoolize.

Fix from chimera-linux: https://github.com/chimera-linux/cports/commit/fa524593a52c057fe30215245298c2913a729ff6

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,9 +1,11 @@
 AC_INIT(libass, m4_normalize(m4_include([RELEASEVERSION])))
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 AC_CONFIG_MACRO_DIR([m4])
-LT_INIT
 AC_CONFIG_SRCDIR([libass/ass.c])
 AC_CONFIG_HEADERS([config.h])
+AC_CANONICAL_HOST
+
+LT_INIT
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -5,6 +5,8 @@ LT_INIT
 AC_CONFIG_SRCDIR([libass/ass.c])
 AC_CONFIG_HEADERS([config.h])
 
+AC_CANONICAL_HOST
+
```
TheOneric, line 9: nit: placing this immediately before or after `LT_INIT` imho makes more sense. Though possibly moving `LT_INIT` itself after `AC_CONFIG_HEADERS` might also be a good idea to avoid switching between autoconf setup and libtool

unrelated: while it still shows p as “verified”,  GitHub informs me the key you used to sign the commit has expired

rcombs (approved):

configure.ac
orbea: I believe I changed it as you suggested and thanks for pointing out my GPG key was expired, I updated it locally a while ago and forgot to update it for github too.

TheOneric:
Thanks!

PR #783, opened 2024-06-06
Cache changes extracted from threading PR

rcombs:
I'm continuing to break up #636 into more-digestible chunks. This is a small portion of the changes to cache infrastructure, but includes the bulk of the changes to the cache's actual externally-visible behavior, which will make other chunks easier to break out from here.

- Remove unused stats functionality
- Don't increment refcount on cache hits

Issue #566, opened 2021-10-19
Audit HarfBuzz allocation checks & leaks


astiob:
I started this in [astiob@`harfbuzz-memchecks`](https://github.com/astiob/libass/commit/harfbuzz-memchecks). Reminder to self to finish it.

One example is https://github.com/libass/libass/pull/553#discussion_r715981040, but there are more.

Also, HarfBuzz allocation APIs never return `NULL`, but instead, on error, return valid pointers to dummy, immutable objects that do nothing or produce errors when used in subsequent API calls. This allows some allocation checks to be dropped entirely; meanwhile, the ones that are kept, if any, must be rewritten.

PR #784, opened 2024-06-06
Cache per-size metrics in shaper

rcombs:
More extracts from #636. This is on top of #783.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -83,7 +83,7 @@ static bool render_context_init(RenderContext *state, ASS_Renderer *priv)
     if (!text_info_init(&state->text_info))
         return false;
 
-    if (!(state->shaper = ass_shaper_new(priv->cache.metrics_cache)))
+    if (!(state->shaper = ass_shaper_new(priv->cache.metrics_cache, priv->cache.face_size_metrics_cache)))
         return false;
 
     return ass_rasterizer_init(&priv->engine, &state->rasterizer, RASTERIZER_PRECISION);
@@ -139,8 +139,11 @@ ASS_Renderer *ass_renderer_init(ASS_Library *library)
     priv->cache.bitmap_cache = ass_bitmap_cache_create();
     priv->cache.composite_cache = ass_composite_cache_create();
     priv->cache.outline_cache = ass_outline_cache_create();
+    priv->cache.face_size_metrics_cache = ass_face_size_metrics_cache_create();
     priv->cache.metrics_cache = ass_glyph_metrics_cache_create();
-    if (!priv->cache.font_cache || !priv->cache.bitmap_cache || !priv->cache.composite_cache || !priv->cache.outline_cache || !priv->cache.metrics_cache)
+    if (!priv->cache.font_cache || !priv->cache.bitmap_cache ||
+        !priv->cache.composite_cache || !priv->cache.outline_cache ||
+        !priv->cache.face_size_metrics_cache || !priv->cache.metrics_cache)
         goto fail;
 
     priv->cache.glyph_max = GLYPH_CACHE_MAX;
@@ -180,6 +183,7 @@ void ass_renderer_done(ASS_Renderer *render_priv)
     ass_cache_done(render_priv->cache.composite_cache);
     ass_cache_done(render_priv->cache.bitmap_cache);
     ass_cache_done(render_priv->cache.outline_cache);
+    ass_cache_done(render_priv->cache.face_size_metrics_cache);
     ass_cache_done(render_priv->cache.metrics_cache);
     ass_cache_done(render_priv->cache.font_cache);
 
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -62,7 +62,8 @@ struct ass_shaper {
     hb_feature_t *features;
     hb_language_t language;
 
-    // Glyph metrics cache, to speed up shaping
+    // Glyph and face-size metrics caches, to speed up shaping
+    Cache *face_size_metrics_cache;
     Cache *metrics_cache;
 
     hb_font_funcs_t *font_funcs;
@@ -191,13 +192,12 @@ static void set_run_features(ASS_Shaper *shaper, GlyphInfo *info)
  * \param hb_font HarfBuzz font
  * \param face associated FreeType font face
  */
-static void update_hb_size(hb_font_t *hb_font, FT_Face face)
+static void update_hb_size(hb_font_t *hb_font, FT_Face face, FT_Size_Metrics *m)
 {
     hb_font_set_scale (hb_font,
-            ((uint64_t) face->size->metrics.x_scale * (uint64_t) face->units_per_EM) >> 16,
-            ((uint64_t) face->size->metrics.y_scale * (uint64_t) face->units_per_EM) >> 16);
-    hb_font_set_ppem (hb_font, face->size->metrics.x_ppem,
-            face->size->metrics.y_ppem);
+            ((uint64_t) m->x_scale * (uint64_t) face->units_per_EM) >> 16,
+            ((uint64_t) m->y_scale * (uint64_t) face->units_per_EM) >> 16);
+    hb_font_set_ppem (hb_font, m->x_ppem, m->y_ppem);
 }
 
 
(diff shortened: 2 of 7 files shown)


libass/ass_shaper.c
```diff
@@ -466,10 +466,6 @@ bool ass_create_hb_font(ASS_Font *font, int index)
 static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
 {
     ASS_Font *font = info->font;
-    hb_font_t *hb_font = font->hb_fonts[info->face_index];
```
TheOneric, line 448: docs a few line above need to be updated; it no longer returns a cache-owned font instance

libass/ass_shaper.c
```diff
@@ -480,10 +476,16 @@ static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
     if (!m)
         return NULL;
 
+    hb_font_t *hb_font = hb_font_create_sub_font(font->hb_fonts[info->face_index]);
+    if (!hb_font)
+        return NULL;
+
```
TheOneric, line 482: i’m not sure about this error check

previously `font->hb_fonts[info->face_index]` was checked for `NULL`, now it is directly passed to hb_font_create_subfont. This is indeed safe and always has been save [since this function was introduced](https://github.com/harfbuzz/harfbuzz/commit/defc45be6d75aba4a67fa7814b91b73bad953fe6#diff-98b5fb69fc328126ab634ef52c824b6146b0fd6505c12f51de7e830485d580daR488), but it will return a dummy font instead like most HarfBuzz API (see also: #566)

Currently the returned dummy object results from a `&Null` cast and the check likely work as intended, but this might change at any point

libass/ass_shaper.c
```diff
@@ -234,6 +234,20 @@ get_cached_metrics(struct ass_shaper_metrics_data *metrics,
...
 }
 
+size_t ass_face_size_metrics_construct(void *key, void *value, void *priv)
+{
+    FaceSizeMetricsHashKey *k = key;
+    FT_Size_Metrics *v = value;
+
+    FT_Face face = k->font->faces[k->face_index];
+
+    ass_face_set_size(face, k->size);
```
TheOneric, line 244: `glyph` and `face_size` construct functions now mutating the passed in face object might warrant a doc hint?

libass/ass_shaper.c
```diff
@@ -221,8 +221,13 @@ get_cached_metrics(struct ass_shaper_metrics_data *metrics,
...
 
-    metrics->hash_key.glyph_index = glyph;
-    FT_Glyph_Metrics *val = ass_cache_get(metrics->metrics_cache, &metrics->hash_key,
+    GlyphMetricsHashKey key = {
+        .font = metrics->hash_key.font,
+        .face_index = metrics->hash_key.face_index,
+        .size = metrics->hash_key.size,
+        .glyph_index = glyph,
+    };
+    FT_Glyph_Metrics *val = ass_cache_get(metrics->metrics_cache, &key,
```
TheOneric, line 229: there are two `get_cached_metrics` calls in `get_glyph_nominal` and `get_glyph_variation` with a leading comment suggesting it might or at some point might have relied on side effects from this mutation:
```C
    // rotate glyph advances for @fonts while we still know the Unicode
codepoints
    get_cached_metrics(metrics_priv, unicode, *glyph);
```

They were added in 1636a551c7d7b05ea15c63a4ad041f2e8d8eb79f and dbb76629145567458c683823a1c20a3c6c2aae64 and respectively

libass/ass_shaper.c
rcombs: This can no longer be null since 1dfb934affee2a4152619bb671a8c82ec5ccad43; if `ass_create_hb_font` returns null, `add_face` will fail.

libass/ass_shaper.c
rcombs: Hmm, reasonable enough; where do you think such a comment should go? (I generally think of the face size as transient, as we always set it and then immediately perform any operations that rely on it; if anything it was unusual that previously we were setting the size and performing operations depending on it so far apart from each other)

libass/ass_shaper.c
rcombs: That's not about mutation of the cache key; it's about making sure that the first call to `get_cached_metrics` on a given event for a given key will always have the correct value for `unicode` (whereas later calls may pass 0). `unicode` is used to determine the value passed to the `priv` arg of `ass_cache_get`, and thus to `ass_glyph_metrics_construct` on cache miss (and misses can only happen on the first call of a frame, or in a future threaded world, the first of an event).

libass/ass_shaper.c
TheOneric: hmm, `ass_create_hb_font` errors checks likely also don’t work

And to correct my previous message: the `Null` hb’s dummy objects are derived from is not `NULL`/`nullptr` but a custom `Null` object iiuc, so checks will never work
Admittedly this ties in to the preexisting issue #566, but adding new know-to-be-useless checks also doesn't seem right

libass/ass_shaper.c
TheOneric: The mutation ultimately happens when calling `ass_cache_get` so at one hand documenting it here would make sense — but it only affects a specific cache.
Probably best handled like the ownership docs we discussed on IRC: add a generic note and pointer to where to find each specific cache’s properties to `ass_cache_get` and then at the more specific place for the glyph and face size metrics caches note what get’s mutated etc

libass/ass_shaper.c
rcombs: Added a `hb_font_is_immutable` check, mirroring how harfbuzz internally checks for this condition; also added the same in `ass_create_hb_font`.

libass/ass_shaper.c
rcombs: Added in ass_cache.h and ass_cache_template.h.

Not going to write up the more general docs on each function's behavior right now, since that's about to be simplified substantially in the threading PR.

libass/ass_render.c
```diff
@@ -139,8 +139,9 @@ ASS_Renderer *ass_renderer_init(ASS_Library *library)
     priv->cache.bitmap_cache = ass_bitmap_cache_create();
     priv->cache.composite_cache = ass_composite_cache_create();
     priv->cache.outline_cache = ass_outline_cache_create();
+    priv->cache.face_size_metrics_cache = ass_face_size_metrics_cache_create();
     priv->cache.metrics_cache = ass_glyph_metrics_cache_create();
-    if (!priv->cache.font_cache || !priv->cache.bitmap_cache || !priv->cache.composite_cache || !priv->cache.outline_cache || !priv->cache.metrics_cache)
+    if (!priv->cache.font_cache || !priv->cache.bitmap_cache || !priv->cache.composite_cache || !priv->cache.outline_cache || !priv->cache.face_size_metrics_cache || !priv->cache.metrics_cache)
```
MrSmile, line 144: This line is getting too long. Maybe it's better to move definition of `CacheStore` struct into `ass_cache.h` and coalesce all cache creation functions into one?

libass/ass_shaper.c
```diff
@@ -486,9 +486,7 @@ static hb_font_t *get_hb_font(ASS_Shaper *shaper, GlyphInfo *info)
         return NULL;
     }
     metrics->metrics_cache = shaper->metrics_cache;
-    metrics->hash_key.font = info->font;
-    metrics->hash_key.face_index = info->face_index;
-    metrics->hash_key.size = info->font_size;
+    memcpy(&metrics->hash_key, &key, sizeof(key));
```
MrSmile, line 489: Wouldn't simple `metrics->hash_key = key;` sufficient here?

MrSmile (commented):
Now `GlyphMetricsHashKey` is basically `FaceSizeMetricsHashKey` + `glyph_index`. I think it would be logically correct to put `FaceSizeMetricsHashKey` member at the start directly, but that breaks automatic function generators, so I'm conflicted here.

libass/ass_shaper.c
TheOneric: yup, that’s fine; thanks!

libass/ass_shaper.c
rcombs: Yup.

libass/ass_render.c
rcombs: This seems like a good idea, but I'm kinda inclined to defer it to a larger cache refactor post-threading (featuring e.g. a single shared LRU queue).

TheOneric (approved):

libass/ass_render.c
rcombs: Added breaks.

PR #786, opened 2024-06-14
ass_shaper: check for singleton returns from harfbuzz; fixes #566

### LINKED ISSUE #566: Audit HarfBuzz allocation checks & leaks

rcombs:
This was pointed out in #784, and I realized the remaining work that needed doing here was pretty minimal.

Most of our allocation and destruction of harfbuzz objects has been streamlined since #566 was filed, so checking for singleton returns was all that was left.

### DIFF
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -271,7 +271,7 @@ get_reference_table(hb_face_t *hbface, hb_tag_t tag, void *font_data)
   }
 
   hb_blob_t *blob = hb_blob_create(buf, len, HB_MEMORY_MODE_WRITABLE, buf, free);
-  if (!blob)
+  if (len > 0 && hb_blob_is_immutable(blob))
       free(buf);
 
   return blob;
@@ -420,7 +420,7 @@ bool ass_create_hb_font(ASS_Font *font, int index)
 {
     FT_Face face = font->faces[index];
     hb_face_t *hb_face = hb_face_create_for_tables(get_reference_table, face, NULL);
-    if (!hb_face)
+    if (hb_face_is_immutable(hb_face))
         return false;
 
     hb_face_set_index(hb_face, face->face_index);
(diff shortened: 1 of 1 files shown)


libass/ass_shaper.c
```diff
@@ -271,7 +271,7 @@ get_reference_table(hb_face_t *hbface, hb_tag_t tag, void *font_data)
   }
 
   hb_blob_t *blob = hb_blob_create(buf, len, HB_MEMORY_MODE_WRITABLE, buf, free);
-  if (!blob)
+  if (!blob || (len > 0 && hb_blob_is_immutable(blob)))
```
TheOneric, line 274: afaik HarfBuzz _never_ returns `NULL` but always statically allocated dummy objects making the `!blob` check redundant

rcombs: I'm inclined to keep the checks since that behavior isn't actually documented anywhere afaik; it's not like this is hot-path code.

TheOneric: it’s a conscious design decision; all API which can return NULL ends with `_or_fail` and explicitly documents a `NULL` return is possible, see e.g. [hb_subset_input_create_or_fail](https://harfbuzz.github.io/harfbuzz-hb-subset.html#hb-subset-input-create-or-fail)

While `hb_bloc_create` is documented to return a new blob (and nothing else, i.e. no `NULL`)

rcombs: Alrighty, fair enough; done.

TheOneric (approved):

PR #787, opened 2024-06-14
ass_parse: simplify range check in dtoi32

rcombs:
This is equivalent, but compilers seem to handle it better (and never emit a call to isnan, which occasionally happens otherwise).


libass/ass_parse.c
```diff
@@ -120,8 +120,9 @@ void ass_update_font(RenderContext *state)
  */
 static inline int32_t dtoi32(double val)
 {
-    if (isnan(val) || val <= INT32_MIN || val >= INT32_MAX + 1LL)
+    if (!(val > INT32_MIN - 1LL && val < INT32_MAX + 1LL))
```
astiob, line 123: Any particular reason to add the `- 1LL`?

Also, maybe invert the conditional control flow to avoid the slightly-confusing negation?

rcombs: Hmm, I'd hoped the 1LL might provide a better hint to the compiler here, but it looks like it doesn't. I'll remove.

Code generated is identical regardless of the negation, so sure, I'll swap that as well.

Wondering if a call to `__builtin_ia32_cvttsd2si` is worthwhile here (it's the only way I've found to convince the compiler to actually just do the conversion unchecked on x86 without invoking UB in the out-of-range case), but I'm a bit concerned that it might constrain optimizations that could've otherwise skipped the cast altogether.

MrSmile (approved):

libass/ass_parse.c
```diff
@@ -120,9 +120,10 @@ void ass_update_font(RenderContext *state)
  */
 static inline int32_t dtoi32(double val)
 {
-    if (isnan(val) || val <= INT32_MIN || val >= INT32_MAX + 1LL)
+    if (val > INT32_MIN && val < INT32_MAX + 1LL)
```
TheOneric, line 123: For reference, in general using `>`/`<` in float-int comparisons is not safe.
All values will be implicitly cast to floating-point type before comparison; but the precision of the floating -point type may not be sufficient to accurately represent the integer value resulting in implementation-defined rounding to an adjacent represnetable float value. This may lead to the cast integer bounds having a greater magnitude than originally allowing too large values to pass the check. Afterwards casting those too large values to the integer type invokes UB.

However, if my understanding is correct this specific case is still safe. Integer operands are `±2^31 ≈ 2.1475e+09` and `double` is iiuc guaranteed to be able to at least represent 10 decimal digits accurately *(see `DBL_DIG` (and `DECIMAL_DIG`, but the latter doesn't necessarily apply to `double`))*
(And `long long` is guaranteed to cover at least the complement-of-two 64-bit range making the `+ 1LL` safe)

This isn’t obvious though and i had to look up the minimum precision values, so i’d prefer a comment added; e.g.:
```C
// floating-point imprecision not relevant in cast; see DBL_DIG
```

astiob: Precision doesn’t matter here, as long as the int-to-float conversion produces the closest representable value. If the true value of `(float) INT32_MIN` is equal to `INT32_MIN` or closer to zero, `val > INT32_MIN` trivially implies `(int)val >= INT32_MIN`. If it’s further away from zero, `val > INT32_MIN` still implies that `val` is at least one ulp closer to zero than that, and there’s no representable value that’s closer to zero than `val` but still further away than `INT32_MIN`: if there were such a value, then `(float) INT32_MIN` would have been that value.

astiob: (I’ve edited the preceding comment to fix some typos and add more clarity.)

You might be confusing `>` with `>=`, although I would still caution that each instance be considered separately (e. g. we have `>=` here before this PR, but the control flow is inverted, so it’s still safe by the same reasoning). Indeed, this is precisely why the conversion path uses strictly-unequal bounds and why the upper bound is written as `< INT32_MAX + 1LL` rather than `<= INT32_MAX` (as mentioned in the previous commit that touched this line, 84928b1fb100004686ac61851720da9247cad34b). If you feel a comment is needed, that’s fair, but I don’t believe `DBL_DIG` is relevant.

astiob: And that’s enough. As long as it can’t be a more _distant_ value, everything works.

TheOneric: i didn't notice the control flow got iverted :facepalm...
nvm then everything is fine, though `DBL_DIG` suggest we even _could_ use `<=` checks here but leaving it as is is probably fine and always obviously *(when not misreading...)* correct

TheOneric, line 123: On the risk of missing something obvious again, is omission of the NaN check really safe?

Looking at ISO C99 7.12.14 §1:
> The relational and equality operators support the usual mathematical relationships between numeric values. For any ordered pair of numeric values exactly one of the relationships — *less, greater*, and *equal* — is true. Relational operators may raise the ‘‘invalid’’ floating-point exception when argument values are NaNs. For a NaN and a numeric value, or for two NaNs, just the *unordered* relationship is true. <sup>215)</sup> The following subclauses provide macros that are *quiet* (non floating-point exception raising) versions of the relational operators, and other comparison macros that facilitate writing efficient code that accounts for NaNs without suffering the ‘‘invalid’’ floating-point exception. In the synopses in this subclause, *real-floating* indicates that the argument shall be an expression of real floating type.
>
> [followed by definitons of `isgreater` etc macros]
>
> 215) IEC 60559 requires that the built-in relational operators raise the ‘‘invalid’’ floating-point exception if the operands compare unordered, as an error indicator for programs written without consideration of NaNs; the result in these cases is false.

Comparisons with `NaN` will *always* raise a floating point exception. *If* we deal with IEEE floats the results *(if/when normal control flow resumes)* will be false and the check will work as intended regardless. But while IEEE only *requires* some flag be raised, it afaik allows for additional error handling like traps etc; are we comfortable with possibly invoking those?   
If we don't deal with IEEE floats the program may just immediately terminate. So far when the question of whether or not we want to assume IEE floats came up, iiirc we always ended up avoiding relying on IEEE behaviour.

astiob: Ah, these comparison operators should be replaced by `isless` and `isgreater` to avoid exceptions from signalling NaNs.

TheOneric: Will the result of comparsions with NaN always be `false` though when we don't have IEEE floats?

astiob: Yes. This is given in your own quote:

> For a NaN and a numeric value, or for two NaNs, just the unordered relationship is true.

No mention of IEEE float.

rcombs: > Ah, these comparison operators should be replaced by `isless` and `isgreater` to avoid exceptions from NaNs.

Hmm, I pushed this change after testing and verifying it produces identical code on macOS and glibc, but now that I test further, I see it *does* produce substantially worse code on musl (which apparently doesn't use the gcc builtins for `__isnan` and friends), so I've reverted to the previous version. ICX also emits external calls to `__isless` and `__isgreater`. As far as I'm aware, it's not actually possible to get a NaN here, and this change came out of a discussion around adding `-ffinite-math-only`, so it seems probably-irrelevant.

Side-note: are we aware of any actual environment we'd plausibly run in where operations on NaNs actually trap? I'm pretty sure this isn't the only place that would be susceptible, but I'm also not aware of any real-world situation where it matters.

MrSmile: I think libass isn't safe to FP exceptions in general, so caring about them only here seems counterproductive.

TheOneric: > Yes. This is given in your own quote:

Where does it define what result “unordered” relationships yield in comparisons? I missed it
The footnote to my understanding also only clarifies *IEEE floats* wil yield false *(after exception handling)*

>  As far as I'm aware, it's not actually possible to get a NaN here, and this change came out of a discussion around adding -ffinite-math-only, so it seems probably-irrelevant.

As i mentioned in the general math flag issue already, we currently _rely_ on infinite and NaN math to catch or belatedly cleaning up after some undesireable cases like “squished to zero” glyphs.
Thus i don’t think disregarding the possibility of NaNs is safe atm

astiob: > Where does it define what result “unordered” relationships yield in comparisons? I missed it

§6.5.8 Relational operators:

> Each of the operators `<` (less than), `>` (greater than), `<=` (less than or equal to), and `>=` (greater than or equal to) shall yield 1 if the specified relation is true and 0 if it is false.

§7.12.14 Comparison macros (your earlier quote):

> For any ordered pair of numeric values exactly one of the relationships — <i>less</i>, <i>greater</i>, and <i>equal</i> — is true. […] For a NaN and a numeric value, or for two NaNs, just the <i>unordered</i> relationship is true.

Quite simply, for a NaN, the <i>less</i> or <i>greater</i> relationship (as appropriate) isn’t true, so the relational operator doesn’t yield 1. (And the macros are defined to yield the same value as the operators.)

astiob: > As far as I'm aware, it's not actually possible to get a NaN here, and this change came out of a discussion around adding `-ffinite-math-only`, so it seems probably-irrelevant.

I’ve just checked; it’s possible to get a NaN using a negative acceleration in `\t`:

```
Dialogue: 0,0:00:00.00,0:00:10.00,Default,,0,0,0,,{\t(0,1000,-1,\c&H808080&)}Black on the first frame
```

rcombs: Wait, do we actually care about floating-point exceptions at all? They're not a POSIX signal or a C++-style exception or anything, they're just a flag in a register we never read. For our purposes, comparisons with NaNs just return false, and we don't have to care about the side-effects.

astiob: > Wait, do we actually care about floating-point exceptions at all? They're not a POSIX signal

`SIGFPE`? And when it occurs, it defaults to aborting the process with a core dump, and to add insult to injury, “According to POSIX, the behavior of a process is undefined after it ignores a SIGFPE […] that was not generated by kill(2) or raise(3)”.

It’s just that (some?) modern architectures seem to disable triggering this signal by default. For example, on my Linux, `fegetexcept()` in a dummy test program returns 0. I can explicitly `feenableexcept(FE_INVALID)` and get a core dump on `pow(0, -1)`. Ironically enough, I can’t seem to make it emit a signal on `NaN < 0` (which was the crux of this discussion) even with a manually memcpy’d signaling NaN and `-fsignaling-nans`.

rcombs: SIGFPE is misleadingly-named; it's used for math exceptions in general (more commonly integer than floating-point ones), and some math floating-point traps raise a signal other than SIGFPE (eg on macOS they can raise SIGILL). IEC 60559 mandates that trapping behavior be disabled at process startup, and in practice this is true on all realistic target platforms I'm aware of. I don't think it'd be reasonable to try to support every possible non-default floating-point environment (including alternate rounding modes, etc), and if we *did* want to do so, I think the only sane solution would be to set the environment to all-defaults at entry to libass and then back to the previous values before return.

PR #788, opened 2024-06-14
ass_fontselect: add LIBASS_FORCE_FONTCONFIG env var

rcombs:
This simplifies testing fontconfig when on a platform with another preferred font provider.

Also improves docs for ASS_DefaultFontProvider.

### DIFF
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -209,11 +209,12 @@ int ass_library_version(void);
  *
  * NONE don't use any default font provider for font lookup
  * AUTODETECT use the first available font provider
- * CORETEXT force a CoreText based font provider (OS X only)
+ * CORETEXT force a CoreText based font provider (Apple platforms only)
  * DIRECTWRITE force a DirectWrite based font provider (Microsoft Win32 only)
  * FONTCONFIG force a Fontconfig based font provider
  *
- * libass uses the best shaper available by default.
+ * In the AUTODETECT case, fontconfig will be preferred over other available providers
+ * if the environment variable "LIBASS_FORCE_FONTCONFIG" is set to any value.
  */
 typedef enum {
     ASS_FONTPROVIDER_NONE       = 0,
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -1095,6 +1095,11 @@ ass_fontselect_init(ASS_Library *library, FT_Library ftlibrary, size_t *num_emfo
         goto fail;
     }
 
+#ifdef CONFIG_FONTCONFIG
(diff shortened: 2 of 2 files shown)


astiob:
Not a fan. What are you using for testing? mpv already provides `--sub-font-provider=fontconfig`; our test utility could do the same (although I’ve never found this necessary myself; you can just recompile with other providers disabled).

rcombs:
This comes up for me when using the test utility (and especially `make check`); particularly when working on fontselect code, it's much quicker and easier than recompiling every time I want to test both the fontconfig path and the native one. I considered adding this to each of our internal tools (test, profile, fuzz…), but it'd have required a bunch of code duplication to achieve the same result, and IMO this is reasonably within the bounds of what `auto` can mean.

TheOneric:
I’m kinda indifferent to (not) adding an env var for this.

At least the singling out of fontconfig seems a bit odd though. CoreText and DirectWrite can only occur on a single platform each and already are the highest prio provider there, so admittedly no `AUTO` nudge for them is needed (atm). Perhaps nudging `AUTO` to always select no system provider at all and using one common `LIBASS_FONTPROVIDER_AUTO` env var might make sense?

astiob:
> Perhaps nudging `AUTO` to always select no system provider at all and using one common `LIBASS_FONTPROVIDER_AUTO` env var might make sense?

I don’t follow; could you rephrase that?

rcombs’s idea is that on systems that have _both_ Fontconfig and another provider, this would let more easily test the non-default Fontconfig, which is normally overridden by the other provider.

TheOneric:
> rcombs’s idea is that on systems that have both Fontconfig and another provider, this would let us more easily test the non-default Fontconfig, which is normally overridden by the other provider.

Despite being named `FORCE_FONTCONFIG` the env var only takes effect if `AUTODETECT` was im- or explicitly set by the API user. If fontconfig support was *not* built-in, no system font provider gets selected. So the env var effectively nudges, or considering it can’t fallback to another provider “restricts”, autodetection into a particular direction.

Currently fontconfig is the only system font provider existing on (some) platforms as the non-default choice. But we might also want to test for things working correctly with no system font provider at all *(e.g. nothing crashes when no fallback can be found)* or (albeit it seems unlikely atm) we might gain more multip-platform providers or additional platform-specific providers for an already existing platform *(e.g. a GDI-only fontprovider)*.  
Having a `LIBASS_FORCE_XXX` env var for each of them raises questions about which takes priority if multiple are set and seems like a more messy interface. Thus the suggestion to use one shared `LIBASS_FONTPROVIDER_AUTO` env var to select the default provider or none at all for `AUTODETECTION`

PR #789, opened 2024-06-14
ass_fontselect: lazy-load GDI-compatible metadata for matching

### LINKED ISSUE #315: Incorrect Font selection
We have a user reporting incorrect font-selection that appears to have the required front on their system. 

https://github.com/HandBrake/HandBrake/issues/1615

fc-list | grep 'Ong Do Gia'
/usr/share/fonts/transcoder/utm-fonts/UTM Ong Do Gia.ttf: UTM Ong Do Gia:style=Regular
/usr/share/fonts/transcoder/unicode-fonts/UTM Ong Do Gia.ttf: UTM Ong Do Gia:style=Regular

Results:
[09:26:01] [ass] fontselect: (UTM Ong Do Gia, 700, 0) -> **/usr/share/fonts/transcoder/svn-fonts/SVN-Student.ttf, 0, SVN-Student**
[09:26:01] [ass] fontselect: (UTM Swiss Condensed, 700, 0) -> /usr/share/fonts/transcoder/unicode-fonts/UTM Swiss CondensedBold.ttf, 0, UTMSwissCondensed-Bold

First one was incorrect, second one was.

Utilising 0.14.0

I'm not convinced this is a bug, but more something funny about the source that's triggering the font choice  to go the wrong way.  Any suggestions would be appreciated.

rcombs:
Fixes #315.

Eager-loading font providers can provide more names, but if we get a match on one, we'll load the GDI data and repeat the check using that.

### DIFF
--- a/libass/ass_fontselect.c
+++ b/libass/ass_fontselect.c
@@ -67,6 +67,9 @@ struct font_info {
 
     // Match properties
     ASS_FontProviderMetaData meta;
+
+    // whether attempting to load properties of this font has failed previously
+    bool failed;
 };
 
 struct font_selector {
@@ -289,6 +292,8 @@ get_font_info(FT_Library lib, FT_Face face, const char *fallback_family_name,
     }
     info->n_fullname = num_fullname;
 
+    info->loaded_from_file = true;
+
     return true;
 
 error:
@@ -345,7 +350,8 @@ static void ass_font_provider_free_fontinfo(ASS_FontInfo *info)
 /**
  * \brief Fill a partly-populated ASS_FontInfo with GDI-compatible metadata.
  * Opens the font file, reads in its metadata, and populates font->meta appropriately.
- * On failure, font is unmodified.
+ * On success, font->meta.loaded_from_file is set to true.
+ * On failure, font->failed is set to true and font->meta is unmodified.
  *
  * \param font FontInfo struct to fill
  * \return whether the operation completed successfully
@@ -389,6 +395,7 @@ static bool fill_font_info(ASS_FontInfo *font)
                        &local_meta))
         goto cleanup;
 
+    free_font_info(meta);
     free(meta->postscript_name);
     local_meta.postscript_name = local_meta.postscript_name ? strdup(local_meta.postscript_name) : NULL;
 
@@ -398,6 +405,7 @@ static bool fill_font_info(ASS_FontInfo *font)
 
 cleanup:
     FT_Done_Face(face);
+    font->failed = !success;
     return success;
 }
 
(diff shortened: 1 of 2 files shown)


libass/ass_fontselect.c
```diff
@@ -342,6 +342,65 @@ static void ass_font_provider_free_fontinfo(ASS_FontInfo *info)
...
+        goto fail;
+
+    free(meta->postscript_name);
+    local_meta.postscript_name = local_meta.postscript_name ? strdup(local_meta.postscript_name) : NULL;
+
+    memcpy(meta, &local_meta, sizeof(local_meta));
+
+    success = true;
+
+fail:
```
astiob, line 406: Nit, subjective: I’d rather name this `cleanup`, because the code below this label executes on both failed and successful paths.

libass/ass_fontselect.c
```diff
@@ -342,6 +342,65 @@ static void ass_font_provider_free_fontinfo(ASS_FontInfo *info)
...
+    ASS_FontProviderMetaData local_meta = *meta;
+
+    if (!get_font_info(selector->ftlibrary, face, meta->extended_family,
+                       &local_meta))
+        goto fail;
+
+    free(meta->postscript_name);
+    local_meta.postscript_name = local_meta.postscript_name ? strdup(local_meta.postscript_name) : NULL;
+
+    memcpy(meta, &local_meta, sizeof(local_meta));
```
astiob, line 402: I know this kinda mirrors existing code, but wouldn’t a simple assignment `*meta = local_meta` suffice?

astiob (commented):
We may want to rename `free_font_info` to something like `free_font_name_arrays` for more clarity.

Anyway, this is great, but what about your own concerns about GDI-incompatible names in Fontconfig’s substitutes, as well as font fallback and user configuration? (Note: font fallback supports `extended_family`, but it’s currently only a single name and not filled at all for Fontconfig.)

libass/ass_fontselect.c
rcombs: Sure.

libass/ass_fontselect.c
rcombs: I never remember when direct assignment to aggregates is and isn't legal; apparently it's fine here, so sure. (The compiler is aware that these are equivalent and can emit either a `memcpy` call or an inlined move in either case.)

rcombs:
Hmmmmmmmmmmmmmmmmmmm, re: substitutions, I *think* the solution is going to be:
- Put the first fontconfig-specified family in `extended_family` (this will *not* be overridden by )
- Before running `get_substitutions`, run `find_font` once with only the specified family, and skip substitutions entirely if it finds anything
- After running `get_substitutions`, always call `find_font` with `match_extended_family=true`

That sound reasonable?

libass/ass_fontselect.c
astiob: Arrays can’t be assigned; structs (even if they contain array fields) can.

TheOneric:
superseded by #792

PR #790, opened 2024-06-14
ass_fontconfig: ignore named instances of variable fonts

rcombs:
These aren't GDI-compatible.

Fixes part of #386 and #315.


astiob (approved):

astiob:
But “not GDI-compatible” is a bit misleading: the specific names used by Fontconfig may not be the same ones used by GDI, but GDI does support named instances of variable fonts and the names are likely the same in many cases.

rcombs:
Hmm, the issue addressed here might be more fully resolved by #789, and we'll want the GDI-compatible names (whichever those might be) to be usable in searches, so I'll hold off on merging this for now until we have a better understanding of how all this works.

TheOneric:
superseded by #792

PR #792, opened 2024-06-20
Variable fonts

rcombs:
Supersedes #789 and #790, and also adds variable font support, along with some other refactoring of font handling.

TODO:
- Confirm GDI behavior on some edge cases around the STAT table (there are a couple cases where we believe the current behavior here is incorrect); cc @moi15moi 
- Consider cacheing a flag indicating that a given font has already been inserted; just requires a good key:
  - On CoreText and Fontconfig, path+index+psname works, but on DirectWrite this gets more complicated; LOGFONT name works sometimes, but not in the IDWriteFontSet case… maybe add a cache like the fontconfig one?
- Consider inserting all GDI-compatible font names into a cache at load-time (similar to the fontconfig one) and searching that in `find_font` instead of iterating the full global list (probably should be a separate PR after this one)
  - Would require a downcase function but that should be fine


libass/ass_fontselect.c
```diff
@@ -256,6 +260,404 @@ static ASS_FontInfo *allocate_font_info(ASS_FontSelector *selector)
...
+
+        a->index = i;
+        a->tag = read_be(p, 4);
+        a->ordering = read_be(p + 6, 2);
+
+        bool matched = false;
+
+        for (j = 0; j < variants->num_axis; j++) {
+            if (a->tag == variants->axis[j].tag) {
+                a->fvar_idx = j;
```
moi15moi, line 381: This is incorrect.
For example, [Alegreya - Original.zip](https://github.com/user-attachments/files/15962083/Alegreya.-.Original.zip)
has 2 DesignAxisRecord (wght and ital) in the STAT table, but the fvar table only has 1 axis (wght).

libass/ass_fontselect.c
```diff
@@ -256,6 +260,404 @@ static ASS_FontInfo *allocate_font_info(ASS_FontSelector *selector)
...
+/**
+ * \brief Compute GDI-compatible configuration for a given named instance of a font.
+ *
+ * \param data structure to fill
+ * \param variant FreeType fvar data for the variant
+ * \param stat parsed OpenType STAT table
+ * \return success
+ */
+static void
+compute_variant_data(struct VariantData *data, const FT_Var_Named_Style *variant,
```
moi15moi, line 567: You don't use the GDI score algorithm.
Here is the algorithm: https://github.com/moi15moi/FontCollector/blob/7d82835aab9de817a4953f50bf69c9e688acb7ce/font_collector/font/font_parser.py#L96-L141

libass/ass_fontselect.c
```diff
@@ -256,6 +260,404 @@ static ASS_FontInfo *allocate_font_info(ASS_FontSelector *selector)
...
+    free(data->appended_names);
+}
+
+/**
+ * \brief Compute GDI-compatible configuration for a given named instance of a font.
+ *
+ * \param data structure to fill
+ * \param variant FreeType fvar data for the variant
+ * \param stat parsed OpenType STAT table
+ * \return success
```
moi15moi, line 564: The doc doesn't represent the function.
It says ``\return success``, but the method actually return nothing.

moi15moi (changes requested):
Here is the first check.

Also, if I request ``Bahnschrift Semibold``, libass will actually select ``Bahnschrift Regular``. I haven't checked exactly why it happen, but it may be because the provider extended_family is ``Bahnschrift Semibold`` and will be used for all the face of Bahnschrift which shouldn't be the case.

I used DirectWrite.

libass/ass_fontselect.c
rcombs: The sample font alone doesn't tell me much without knowing how it behaves in GDI. What's this supposed to do?

libass/ass_fontselect.c
```diff
@@ -256,6 +260,404 @@ static ASS_FontInfo *allocate_font_info(ASS_FontSelector *selector)
...
+    uint16_t minorVersion = read_be(data + 2, 2);
+
+    uint16_t axis_size = read_be(data + 4, 2);
+    stat->nb_axes = read_be(data + 6, 2);
+    uint32_t axes_offset = read_be(data + 8, 4);
+
+    stat->nb_value_groups = read_be(data + 12, 2);
+    uint32_t values_offset = read_be(data + 14, 4);
+
+    if (majorVersion > 1 || minorVersion >= 1)
```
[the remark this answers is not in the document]
rcombs: Good catch; corrected.

libass/ass_fontselect.c
rcombs: Good catch; corrected.

libass/ass_fontselect.c
rcombs: I'm hoping we can document behavior in specific edge cases here (in particular, the documented behavior for format-4 tables differs from that code, and I'd like to confirm exactly what happens in those cases).

MrSmile (commented):
I'm not an expert on font selection, so mostly comments about cosmetics.

PR #793, opened 2024-06-22
Threading

rcombs:
Replaces #636, as it'd been through several major cache redesigns and the thread was too long to be practical review-wise.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,7 @@
 ACLOCAL_AMFLAGS = -I m4
 AM_CPPFLAGS = -I$(top_srcdir)/libass
 
-# Using gnu99 to gain compatibility between const and
+# Using GNU C (set by autoconf) to gain compatibility between const and
 # non-const array types without needing to jump to ISO C23.
 # Otherwise intended to be compatible with ISO C.
 #
@@ -11,7 +11,7 @@ AM_CPPFLAGS = -I$(top_srcdir)/libass
 # non-POSIX interfaces even if their respective feature macro is set too.
 # However, POSIX interfaces are usually implicitly enabled
 # with other feature macros, thus omit the POSIX one.
-AM_CFLAGS = -std=gnu99 -D_GNU_SOURCE -D_XPLATFORM_SOURCE
+AM_CFLAGS = -D_GNU_SOURCE -D_XPLATFORM_SOURCE
 
 EXTRA_DIST = libass.pc.in Changelog MAINTAINERS RELEASEVERSION ltnasm.sh
 
--- a/configure.ac
+++ b/configure.ac
@@ -66,6 +66,7 @@ AX_APPEND_COMPILE_FLAGS([ \
     -Wno-missing-field-initializers \
     dnl clang-specific, gcc currently has no fine-grained toggle for this, only -Wpedantic
     -Wformat-non-iso -Werror=format-non-iso \
+    -Watomic-implicit-seq-cst \
 ])
 
 # Checks for available libraries and define corresponding C Macros
--- a/configure.ac
+++ b/configure.ac
@@ -98,8 +98,8 @@ PKG_CHECK_MODULES([FREETYPE], [freetype2 >= 9.17.3], [
     LIBS="$LIBS $FREETYPE_LIBS"
 ])
 
-PKG_CHECK_MODULES([FRIBIDI], [fribidi >= 0.19.1], [
-    pkg_requires="fribidi >= 0.19.1, ${pkg_requires}"
+PKG_CHECK_MODULES([FRIBIDI], [fribidi >= 0.19.7], [
+    pkg_requires="fribidi >= 0.19.7, ${pkg_requires}"
     CFLAGS="$CFLAGS $FRIBIDI_CFLAGS"
     LIBS="$LIBS $FRIBIDI_LIBS"
 ])
--- a/configure.ac
+++ b/configure.ac
@@ -13,10 +13,10 @@ AM_PROG_CC_C_O
 AM_PROG_AS
 
 # Checks for header files.
-AC_CHECK_HEADERS_ONCE([iconv.h])
+AC_CHECK_HEADERS_ONCE([iconv.h sched.h stdatomic.h unistd.h windows.h])
 
 # Checks for library functions.
-AC_CHECK_FUNCS([strdup strndup])
+AC_CHECK_FUNCS([sched_getaffinity strdup strndup pthread_setname_np])
 
 # Query configuration parameters and set their description
 AC_ARG_ENABLE([test], AS_HELP_STRING([--enable-test],
@@ -39,6 +39,10 @@ AC_ARG_ENABLE([require-system-font-provider], AS_HELP_STRING([--disable-require-
     [allow compilation even if no system font provider was found @<:@default=enabled:>@]))
 AC_ARG_ENABLE([asm], AS_HELP_STRING([--disable-asm],
     [disable compiling with ASM @<:@default=check@:>@]))
+AC_ARG_ENABLE([pthreads], AS_HELP_STRING([--disable-pthreads],
+    [disable threading support via pthreads @<:@default=check@:>@]))
+AC_ARG_ENABLE([w32threads], AS_HELP_STRING([--disable-w32threads],
+    [disable threading support via win32 threads @<:@default=check@:>@]))
 AC_ARG_ENABLE([large-tiles], AS_HELP_STRING([--enable-large-tiles],
     [use larger tiles in the rasterizer (better performance, slightly worse quality) @<:@default=disabled@:>@]))
 
@@ -90,6 +94,42 @@ AC_SEARCH_LIBS([lrint], [m], [
     AC_MSG_ERROR([Unable to locate math functions!])
 ])
 pkg_libs="$LIBS"
+
+AS_IF([test "x$enable_pthreads" = xyes && test "x$enable_w32threads" = xyes], [
+    AC_MSG_ERROR([both pthreads and w32threads were requested.])
+])
+
+AS_IF([test "x$enable_pthreads" = xyes && test "x$ac_cv_header_stdatomic_h" != xyes], [
+    AC_MSG_ERROR([pthread support was requested, but <stdatomic.h> was not found.])
+])
+
+thread_api="none"
+
+AS_IF([test "x$enable_w32threads" != xno], [
+    AS_IF([test "x$ac_cv_header_windows_h" = xyes], [
+        AC_DEFINE(CONFIG_W32THREAD, 1, [Have Win32 threads])
+        thread_api=w32threads
+    ], [test "x$enable_w32threads" = xyes], [
+        AC_MSG_ERROR([win32 threading support was requested, but was not found.])
+    ])
+])
+
+AS_IF([test $thread_api == none && test "x$enable_pthreads" != xno && test "x$ac_cv_header_stdatomic_h" = xyes], [
+    AX_PTHREAD([
+        pkg_libs="$pkg_libs $PTHREAD_LIBS"
+        CFLAGS="$CFLAGS $PTHREAD_CFLAGS"
+        LIBS="$LIBS $PTHREAD_LIBS"
+        CC="$PTHREAD_CC"
+        AC_DEFINE(CONFIG_PTHREAD, 1, [Have POSIX threads])
+        thread_api=pthreads
+    ], [
+        AS_IF([test "x$enable_pthreads" = xyes], [
+            AC_MSG_ERROR([pthread support was requested, but was not found.])
+        ])
+    ])
+])
+
+AC_MSG_NOTICE([Chosen threading model: $thread_api])
 
 ## Check for libraries via pkg-config and add to pkg_requires as needed
 PKG_CHECK_MODULES([FREETYPE], [freetype2 >= 9.17.3], [
@@ -196,9 +236,7 @@ AS_IF([test "x$enable_coretext" != xno], [
 AS_IF([test "x$enable_directwrite" != xno], [
     # Linking to DirectWrite directly only works from Windows
     AC_MSG_CHECKING([for DirectWrite])
-    AC_COMPILE_IFELSE([
-        AC_LANG_PROGRAM([[#include <windows.h>]], [[;]])
-    ], [
+    AS_IF([test "x$ac_cv_header_windows_h" = xyes], [
         directwrite=true
         AC_MSG_RESULT([yes])
         AC_MSG_CHECKING([for Win32 desktop APIs])
--- a/libass/ass_cache.c
+++ b/libass/ass_cache.c
@@ -61,7 +61,13 @@ static bool font_key_move(void *dst, void *src)
     return d->family.str;
 }
 
-static void font_destruct(void *key, void *value)
+static void font_key_destruct(void *k)
+{
+    ASS_FontDesc *key = k;
+    free((void *) key->family.str);
+}
+
+static void font_destruct(void *value)
 {
     ass_font_clear(value);
 }
@@ -72,8 +78,9 @@ const CacheDesc font_cache_desc = {
     .hash_func = font_hash,
     .compare_func = font_compare,
     .key_move_func = font_key_move,
+    .key_destruct_func = font_key_destruct,
     .construct_func = ass_font_construct,
-    .destruct_func = font_destruct,
+    .value_destruct_func = font_destruct,
     .key_size = sizeof(ASS_FontDesc),
     .value_size = sizeof(ASS_Font)
 };
@@ -90,11 +97,15 @@ static bool bitmap_key_move(void *dst, void *src)
     return true;
 }
 
-static void bitmap_destruct(void *key, void *value)
+static void bitmap_key_destruct(void *key)
 {
     BitmapHashKey *k = key;
-    ass_free_bitmap(value);
     ass_cache_dec_ref(k->outline);
+}
+
+static void bitmap_destruct(void *value)
+{
+    ass_free_bitmap(value);
 }
 
 size_t ass_bitmap_construct(void *key, void *value, void *priv);
@@ -103,8 +114,9 @@ const CacheDesc bitmap_cache_desc = {
     .hash_func = bitmap_hash,
     .compare_func = bitmap_compare,
     .key_move_func = bitmap_key_move,
+    .key_destruct_func = bitmap_key_destruct,
     .construct_func = ass_bitmap_construct,
-    .destruct_func = bitmap_destruct,
+    .value_destruct_func = bitmap_destruct,
     .key_size = sizeof(BitmapHashKey),
     .value_size = sizeof(Bitmap)
 };
@@ -150,18 +162,22 @@ static bool composite_key_move(void *dst, void *src)
     return true;
 }
 
-static void composite_destruct(void *key, void *value)
+static void composite_key_destruct(void *key)
 {
-    CompositeHashValue *v = value;
     CompositeHashKey *k = key;
-    ass_free_bitmap(&v->bm);
-    ass_free_bitmap(&v->bm_o);
-    ass_free_bitmap(&v->bm_s);
     for (size_t i = 0; i < k->bitmap_count; i++) {
         ass_cache_dec_ref(k->bitmaps[i].bm);
         ass_cache_dec_ref(k->bitmaps[i].bm_o);
     }
     free(k->bitmaps);
+}
+
+static void composite_destruct(void *value)
+{
+    CompositeHashValue *v = value;
+    ass_free_bitmap(&v->bm);
+    ass_free_bitmap(&v->bm_o);
+    ass_free_bitmap(&v->bm_s);
 }
 
 size_t ass_composite_construct(void *key, void *value, void *priv);
@@ -170,8 +186,9 @@ const CacheDesc composite_cache_desc = {
     .hash_func = composite_hash,
     .compare_func = composite_compare,
     .key_move_func = composite_key_move,
+    .key_destruct_func = composite_key_destruct,
     .construct_func = ass_composite_construct,
-    .destruct_func = composite_destruct,
+    .value_destruct_func = composite_destruct,
     .key_size = sizeof(CompositeHashKey),
     .value_size = sizeof(CompositeHashValue)
 };
@@ -229,12 +246,9 @@ static bool outline_key_move(void *dst, void *src)
     return true;
 }
 
-static void outline_destruct(void *key, void *value)
+static void outline_key_destruct(void *key)
 {
-    OutlineHashValue *v = value;
     OutlineHashKey *k = key;
-    ass_outline_free(&v->outline[0]);
-    ass_outline_free(&v->outline[1]);
     switch (k->type) {
     case OUTLINE_GLYPH:
         ass_cache_dec_ref(k->u.glyph.font);
@@ -250,14 +264,22 @@ static void outline_destruct(void *key, void *value)
     }
 }
 
+static void outline_destruct(void *value)
+{
+    OutlineHashValue *v = value;
+    ass_outline_free(&v->outline[0]);
+    ass_outline_free(&v->outline[1]);
+}
+
 size_t ass_outline_construct(void *key, void *value, void *priv);
 
 const CacheDesc outline_cache_desc = {
     .hash_func = outline_hash,
     .compare_func = outline_compare,
     .key_move_func = outline_key_move,
+    .key_destruct_func = outline_key_destruct,
     .construct_func = ass_outline_construct,
-    .destruct_func = outline_destruct,
+    .value_destruct_func = outline_destruct,
     .key_size = sizeof(OutlineHashKey),
     .value_size = sizeof(OutlineHashValue)
 };
@@ -275,7 +297,7 @@ static bool face_size_metrics_key_move(void *dst, void *src)
     return true;
 }
 
-static void face_size_metrics_destruct(void *key, void *value)
+static void face_size_metrics_key_destruct(void *key)
 {
     FaceSizeMetricsHashKey *k = key;
     ass_cache_dec_ref(k->font);
@@ -287,8 +309,8 @@ const CacheDesc face_size_metrics_cache_desc = {
     .hash_func = face_size_metrics_hash,
     .compare_func = face_size_metrics_compare,
     .key_move_func = face_size_metrics_key_move,
+    .key_destruct_func = face_size_metrics_key_destruct,
     .construct_func = ass_face_size_metrics_construct,
-    .destruct_func = face_size_metrics_destruct,
     .key_size = sizeof(FaceSizeMetricsHashKey),
     .value_size = sizeof(FT_Size_Metrics)
 };
@@ -306,7 +328,7 @@ static bool glyph_metrics_key_move(void *dst, void *src)
     return true;
 }
 
-static void glyph_metrics_destruct(void *key, void *value)
+static void glyph_metrics_key_destruct(void *key)
 {
     GlyphMetricsHashKey *k = key;
     ass_cache_dec_ref(k->font);
@@ -318,8 +340,8 @@ const CacheDesc glyph_metrics_cache_desc = {
     .hash_func = glyph_metrics_hash,
     .compare_func = glyph_metrics_compare,
     .key_move_func = glyph_metrics_key_move,
+    .key_destruct_func = glyph_metrics_key_destruct,
     .construct_func = ass_glyph_metrics_construct,
-    .destruct_func = glyph_metrics_destruct,
     .key_size = sizeof(GlyphMetricsHashKey),
     .value_size = sizeof(FT_Glyph_Metrics)
 };
@@ -497,7 +519,10 @@ static inline void destroy_item(const CacheDesc *desc, CacheItem *item)
 {
     assert(item->desc == desc);
     char *value = (char *) item + CACHE_ITEM_SIZE;
-    desc->destruct_func(value + align_cache(desc->value_size), value);
+    if (desc->key_destruct_func)
+        desc->key_destruct_func(value + align_cache(desc->value_size));
+    if (desc->value_destruct_func)
+        desc->value_destruct_func(value);
     free(item);
 }
 
--- a/libass/ass_cache.c
+++ b/libass/ass_cache.c
@@ -345,6 +345,10 @@ struct cache {
     size_t cache_size;
 };
 
+struct cache_client {
+    struct cache_client *next;
+};
+
 #define CACHE_ALIGN 8
 #define CACHE_ITEM_SIZE ((sizeof(CacheItem) + (CACHE_ALIGN - 1)) & ~(CACHE_ALIGN - 1))
 
@@ -377,11 +381,53 @@ Cache *ass_cache_create(const CacheDesc *desc)
     return cache;
 }
 
+bool ass_cache_client_set_init(CacheClientSet *set)
+{
+    memset(set, 0, sizeof(*set));
+
+    return true;
+}
+
+static void cache_client_done(CacheClient *client)
+{
+    free(client);
+}
+
+void ass_cache_client_set_clear(CacheClientSet *set)
+{
+    CacheClient *client = set->first_client;
+
+    while (client) {
+        CacheClient *next = client->next;
+        cache_client_done(client);
+        client = next;
+    }
+
+    set->first_client = NULL;
+}
+
+void ass_cache_client_set_done(CacheClientSet *set)
+{
+    ass_cache_client_set_clear(set);
+}
+
+CacheClient *ass_cache_client_create(CacheClientSet *set)
+{
+    CacheClient *client = calloc(1, sizeof(*client));
+    if (!client)
+        return NULL;
+
+    client->next = set->first_client;
+    set->first_client = client;
+
+    return client;
+}
+
 // Retrieve a value corresponding to a particular cache key,
 // creating one if it does not already exist.
 // The returned item is guaranteed to be valid until the next ass_cache_cut call;
 // to extend its lifetime further, call ass_cache_inc_ref().
-void *ass_cache_get(Cache *cache, void *key, void *priv)
+void *ass_cache_get(Cache *cache, CacheClient *client, void *key, void *priv)
 {
     const CacheDesc *desc = cache->desc;
     size_t key_offs = CACHE_ITEM_SIZE + align_cache(desc->value_size);
--- a/libass/ass_cache.c
+++ b/libass/ass_cache.c
@@ -23,12 +23,12 @@
 #include <inttypes.h>
 #include <ft2build.h>
 #include FT_OUTLINE_H
-#include <assert.h>
 
 #include "ass_utils.h"
 #include "ass_font.h"
 #include "ass_outline.h"
 #include "ass_cache.h"
+#include "ass_threading.h"
 
 // Always enable native-endian mode, since we don't care about cross-platform consistency of the hash
 #define WYHASH_LITTLE_ENDIAN 1
@@ -352,23 +352,40 @@ const CacheDesc glyph_metrics_cache_desc = {
 typedef struct cache_item {
     Cache *cache;
     const CacheDesc *desc;
-    struct cache_item *next, **prev;
+    struct cache_item *_Atomic next, *_Atomic *prev;
     struct cache_item *queue_next, **queue_prev;
-    size_t size, ref_count;
+    struct cache_item *promote_next;
+    _Atomic AtomicInt size, ref_count;
+    ass_hashcode hash;
+
+    _Atomic AtomicInt last_used_frame;
+
+#if ENABLE_THREADS
+    struct cache_client *creating_client;
+#endif
 } CacheItem;
 
 struct cache {
     unsigned buckets;
-    CacheItem **map;
+    CacheItem *_Atomic *map;
     CacheItem *queue_first, **queue_last;
 
     const CacheDesc *desc;
 
-    size_t cache_size;
+    _Atomic AtomicInt cache_size;
+
+    uintptr_t cur_frame;
 };
 
 struct cache_client {
+    CacheItem *promote_first;
+
     struct cache_client *next;
+
+#if ENABLE_THREADS
+    pthread_mutex_t mutex;
+    pthread_cond_t cond;
+#endif
 };
 
 #define CACHE_ALIGN 8
@@ -400,6 +417,13 @@ Cache *ass_cache_create(const CacheDesc *desc)
         return NULL;
     }
 
+#if ENABLE_THREADS
+    for (size_t i = 0; i < cache->buckets; i++)
+        atomic_init(&cache->map[i], NULL);
+
+    atomic_init(&cache->cache_size, 0);
+#endif
+
     return cache;
 }
 
@@ -407,11 +431,21 @@ bool ass_cache_client_set_init(CacheClientSet *set)
 {
     memset(set, 0, sizeof(*set));
 
+#if ENABLE_THREADS
+    if (pthread_mutex_init(&set->mutex, NULL) != 0)
+        return false;
+#endif
+
     return true;
 }
 
 static void cache_client_done(CacheClient *client)
 {
+#if ENABLE_THREADS
+    pthread_mutex_destroy(&client->mutex);
+    pthread_cond_destroy(&client->cond);
+#endif
+
     free(client);
 }
 
@@ -431,6 +465,10 @@ void ass_cache_client_set_clear(CacheClientSet *set)
 void ass_cache_client_set_done(CacheClientSet *set)
 {
     ass_cache_client_set_clear(set);
+
+#if ENABLE_THREADS
+    pthread_mutex_destroy(&set->mutex);
+#endif
 }
 
 CacheClient *ass_cache_client_create(CacheClientSet *set)
@@ -439,8 +477,27 @@ CacheClient *ass_cache_client_create(CacheClientSet *set)
     if (!client)
         return NULL;
 
+#if ENABLE_THREADS
+    if (pthread_mutex_init(&client->mutex, NULL) != 0) {
+        free(client);
+        return NULL;
+    }
+
+    if (pthread_cond_init(&client->cond, NULL) != 0) {
+        pthread_cond_destroy(&client->cond);
+        free(client);
+        return NULL;
+    }
+
+    pthread_mutex_lock(&set->mutex);
+#endif
+
     client->next = set->first_client;
     set->first_client = client;
+
+#if ENABLE_THREADS
+    pthread_mutex_unlock(&set->mutex);
+#endif
 
     return client;
 }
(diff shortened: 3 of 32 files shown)


Makefile.am
```diff
@@ -1,6 +1,6 @@
 ACLOCAL_AMFLAGS = -I m4
 AM_CPPFLAGS = -I$(top_srcdir)/libass
-AM_CFLAGS = -std=gnu99 -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter \
+AM_CFLAGS = -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter \
```
TheOneric, line 3: No longer requesting an explicit standard version has undesireable effects:
 - our expected baseline is no longer documented at all
 - we now depend on compiler and/or autotools defaults which might change at any time and differ across compilers
 - with the current C11 default we’ll no longer test our actual default in CI, nor on developer machines
 
 This makes it likely we’ll accidentally break our supposed baseline at some point by inadvertenly using a newer feature *(see e.g. what happened with an `{}` between 0.17.1 and 0.17.2 because compilers didn’t warn about it)*.
 
 On the other hand, the reason for dropping this is some libcs otherwise hiding atomics; which libcs do that?
 
I think instead of just dropping the standard version we should do one of the following *(ordered by my personal preference)*:
- continue to explicitly request GNU99 **if** no libcs actually hide atomics
- raise our baseline to C11 and explicitly request GNU/ISO C11
- explicitly request _either_ GNU/ISO C11 _or_ GNU/ISO C99 and make sure at least one CI job always uses C99
- drop the explicit version like now, but ensure one CI job explicitly runs with our baseline version
 
*(Having a meson buildsystem and meson not _yet_ being changed actually helps out here, since it still explicitly requests ISO C99, but once meson will change too so this is only temporary)*

libass/ass_utils.c
```diff
@@ -116,10 +117,20 @@ void *ass_try_realloc_array(void *ptr, size_t nmemb, size_t size)
 
 void ass_msg(ASS_Library *priv, int lvl, const char *fmt, ...)
 {
+#if ENABLE_THREADS
+    if (!priv->thread_safe_cb)
+        pthread_mutex_lock(&priv->log_mutex);
+#endif
```
TheOneric, line 123: Potential cosmetic issue: our default log handler uses several printf calls and is listed as thread safe. Individual calls from different messages might end up interleaved.

Strictly speaking, albeit only a theoretical concern for any remotely relevant platform as far as i’m aware: ISO C99 to my knowledge doesn't guarantee thread-safety of `(f))printf`. [POSIX explicitly guarantees this](https://pubs.opengroup.org/onlinepubs/9699919799/functions/flockfile.html) and provides `_unlocked` non-thread-safe variants for when custom locking is used and print perf relevant.

TheOneric (commented):
Moving over still relevant comments from the old PR

profile/profile.c
```diff
@@ -57,6 +57,7 @@ static void init(int frame_w, int frame_h)
         exit(1);
     }
 
+    ass_set_threads(ass_renderer, 0);
```
TheOneric, line 60: At least `fuzz`, possibly all other utilities too, should opt into thread-safe logging

configure.ac
```diff
@@ -80,6 +84,40 @@ AC_SEARCH_LIBS([lrint], [m], [
...
+        CC="$PTHREAD_CC"
+        AC_DEFINE(CONFIG_PTHREAD, 1, [Have POSIX threads])
+        threads_found=true
+    ], [
+        AS_IF([test "x$enable_pthreads" = xyes], [
+            AC_MSG_ERROR([pthread support was requested, but was not found.])
+        ])
+    ])
+])
+
```
TheOneric, line 120: nit: atm AX_PTHREAD produces some output indicating whether pthread was found, but if Win32 is used or there’s no `stdatomic.h` no indicators wrt threading are printed at all. Might be nice to add `AC_MSG_CHECKING` to the Win32 tests or alternatively print the overall result at the end:
```autoconf
thread_api="none" # instead of threads_found
# ....
AC_MSG_NOTICE([Chosen threading model: $thread_api])
```

libass/ass_library.c
```diff
@@ -45,8 +45,13 @@ static void ass_msg_handler(int level, const char *fmt, va_list va, void *data)
 ASS_Library *ass_library_init(void)
 {
     ASS_Library* lib = calloc(1, sizeof(*lib));
-    if (lib)
+    if (lib) {
         lib->msg_callback = ass_msg_handler;
+#if ENABLE_THREADS
+        lib->thread_safe_cb = -1;
+        pthread_mutex_init(&lib->log_mutex, NULL);
```
TheOneric, line 52: What if mutex init fails?

libass/ass_cache.c
```diff
@@ -17,6 +17,12 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+// We have some asserts on highly-contended atomics in here,
+// so they're disabled if not requested
+#if !defined(DEBUG) && !defined(NDEBUG)
+#define NDEBUG 1
```
TheOneric, line 23: This disables cache asserts in almost all builds, including our own CI builds. If this is kept, all or most of our CI builds should be changed to explicitly set `-DDEBUG`

libass/ass_font.c
```diff
@@ -427,20 +427,23 @@ static int add_face(ASS_FontSelector *fontsel, ASS_Font *font, uint32_t ch)
...
-        return -1;
+        goto fail;
 
     for (i = 0; i < font->n_faces; i++) {
         if (font->faces_uid[i] == uid) {
             ass_msg(font->library, MSGL_INFO,
                     "Got a font face that already is available! Skipping.");
-            return i;
+            ret = i;
+            goto fail;
```
TheOneric, line 446: nit: this label is now used for a non-failing condition here; may be better to rename

TheOneric (commented):
And previously unreleased pending comments from an unfinished review

Makefile.am
rcombs: `AC_PROG_CC` automatically selects the latest known non-strict-conformance (i.e. `gnu`) standard version the compiler supports. If we want, we can check `ac_prog_cc_stdc` and warn if it's `c89`, but I don't really see the need; if an older compiler fails on our code due to not supporting the minimum required standard, the user will be notified by, uh… compilation failing (which is already what happens now, just slightly faster via `-std=gnu99` being unrecognized). I think the only actual potential issue here is that, say, a future C version could come along and remove something we use? But generally the C standard provides years of deprecation notice, so I'm not really worried about that either.

Regardless of whether atomics are hidden in C99 mode, I see no benefit in unconditionally explicitly requesting an older version than what we're using features from.

Running a build with an explicit `-std=gnu[earliest supported version]` seems reasonable, though I think it'd be more useful to do one with the earliest *gcc version* we're targeting (which is a much more meaningful test than using a newer compiler in an older mode).

Documenting the minimum requirements somewhere seems reasonable (README?); as of this PR it's "99, or 11 for threading", and I expect there's a good chance it'll move to "11, unless we have an explicit compat shim for your system" in the near future (assuming we go forwards with requiring atomics for publicly-refcounted frames).

libass/ass_utils.c
rcombs: Yeah, this interleaving does happen in practice, but I'm not particularly concerned about it (since as you say it's cosmetic). I've found the overhead of locking to actually be meaningful in some cases (esp. when people put nonstandard stuff in the `Effect` field, since that logs on every frame).

libass/ass_library.c
rcombs: Added check.

profile/profile.c
rcombs: Done.

libass/ass_library.c
```diff
@@ -156,6 +164,10 @@ void ass_set_message_cb(ASS_Library *priv,
     if (msg_cb) {
         priv->msg_callback = msg_cb;
         priv->msg_callback_data = data;
+#if ENABLE_THREADS
+        if (priv->thread_safe_cb == -1)
+            priv->thread_safe_cb = 0;
+#endif
```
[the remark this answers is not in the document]
rcombs: Honestly, I originally brought up this possibility mostly as a strawman to illustrate an argument against locking around these log calls at all, and I mostly regret it; I don't think this is actually a realistic concern, and we've never remotely documented any of the relevant threading behavior, so I'd argue that in the extremely unlikely case that somebody *is* relying on the existing behavior, anything bad that happens to them is their own fault.

configure.ac
rcombs: Sure, done.

libass/ass_font.c
rcombs: Sure, done.

libass/ass_cache.c
rcombs: I think we should either do that, or explicitly put the asserts on newly-atomic values behind an `#ifdef DEBUG` or the like; the perf implications can be pretty major at times, and disabling them in almost all builds is deliberate.

libass/ass_utils.c
TheOneric: yeah, i’m not too concerned either; mostly just brought it up again since the old thread ended without a conclusion on `flockfile`.  
If interleaving turns out too annoying in practice we can still change this later

libass/ass_library.c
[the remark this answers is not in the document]
TheOneric: I don’t fully agree on the “is their own fault” part, but either way there is no reason to _not_ at least document this change; all i’m asking for is a notice in the commit message (and later the changelog / release notes)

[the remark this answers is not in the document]
rcombs: Wrote something up; thoughts?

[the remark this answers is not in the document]
TheOneric: yep, the note added to the “add support for per-event threaded rendering” commit seems good; thanks

rcombs:
After a bit of poking around, I have this working in Meson and MSVC (via the Win32 Interlocked* API). Muon is now failing because of the `threads` dep getting applied to nasm for reasons that are completely beyond me (maybe it's just on an old buggy version of meson?), and we don't appear to have the test suite set up for the meson builds yet (definitely would be good to have for this).

Other changes:
- Added `-Watomic-implicit-seq-cst`, avoiding accidental use of arithmetic operators on atomics (these would otherwise silently degrade to unsafe non-atomic accesses on MSVC, which we certainly don't want, and they're also slower).
- Explicitly specified memory ordering everywhere; this should improve perf a bit on ARM.
- Shored up the end-of-frame handling to make sure we can't have any threads go straight from one frame to the next without passing through a full memory barrier.
- Introduced `AtomicInt`, a type suitable for atomic ops (this simplifies the MSVC shims).
- Adjusted cache code to always use the promotion queue for new items, rather than inserting them into the LRU queue during creation; this simplifies the code a bit, and reduces a bit of atomic contention in favor of more use of thread-sharded data; the MRU queue is now entirely non-atomic.
- Always use `atomic_init` to initialize dynamically-allocated atomics (i.e. all of them); this is never actually required on any of our target platforms, but it's essentially free, and is technically required by the standard.
- Update meson build configuration to better match autotools:
  - `_GNU_SOURCE` (a looser mode) instead of `_POSIX_C_SOURCE` (a stricter one, which disabled some functionality we want, here in the form of thread name setters)
  - fribidi bump
  - prefer C11 instead of C99
    - I tried to make this optional, but the Muon build is on too old a Meson to support that; personally I'm fine mandating this, particularly since I'm hoping to make atomics mandatory soon anyway.

meson.build
```diff
@@ -135,6 +141,37 @@ if fontconfig
     font_providers += ['Fontconfig']
 endif
 
+threads_opt = get_option('threads')
+if not threads_opt.disabled()
+    thread_dep = dependency('threads')
+    if thread_dep.found()
```
arch1t3cht, line 147: Note that `dependency('threads').found()` is currently [always true](https://github.com/mesonbuild/meson/blob/7ec1fc507ea6f80e819ac46b56125b955073de7f/mesonbuild/dependencies/misc.py#L149), even if the compiler does not actually support the `-pthread` argument (see also https://github.com/mesonbuild/meson/issues/7775).
Of course, it's always possible for this to change in the future, but I wanted to point this out in case you weren't aware.

rcombs: Good to know, though I'm not sure if there's much to be done about it at the moment.

TheOneric:
> muon is now failing because of the threads dep getting applied to nasm for reasons that are completely beyond me (maybe it's just on an old buggy version of meson?),

We do (now) use a somewhat old 0.2.0 in CI, but this also happens in current git master. I filed an issue upstream: https://todo.sr.ht/~lattis/muon/129

rcombs:
@debugzxcv I don't believe that crash can possibly occur on the current version of the branch, but it's hard to tell without source line information, or at least a disassembly of the crashing function. What git hash are you running?

TheOneric:
> [7d5208a](https://github.com/libass/libass/commit/7d5208a6eccf3b83dabfc2cf497e9a4dd5b25268) + this branch

What commit from this PR are you using? This branch keeps changing

kasper93:
> which makes mpv stats.lua actually usable without stuttering.

It doesn't solve anything.

meson.build
```diff
@@ -71,6 +72,11 @@ if (
     conf.set('HAVE_FSTAT', 1)
 endif
 
+have_stdatomic = cc.check_header('stdatomic.h')
+if have_stdatomic
+    conf.set('HAVE_STDATOMIC_H', 1)
+endif
```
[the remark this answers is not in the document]
TheOneric: `have_stdatomic` is used later on and thus cannot be inlined into `conf.set()`, but it’ſ true the other two checks/defines are currently missing in meson

libass/ass_render.c
```diff
@@ -3049,6 +3130,67 @@ ass_render_event(RenderContext *state, ASS_Event *event,
...
+        priv->thread_start_failed = 1;
+        goto fail;
+    }
+
+    priv->started_threads++;
+
+    pthread_cond_broadcast(&priv->main_cond);
+
+    for (;;) {
+        while (!priv->shutting_down && atomic_load_explicit(&priv->next_eimg, memory_order_relaxed) >= atomic_load_explicit(&priv->sent_eimgs, memory_order_relaxed))
```
MrSmile, line 3157: Too long line.

libass/ass_render.c
```diff
@@ -3374,6 +3505,42 @@ static int ass_detect_change(ASS_Renderer *priv)
     return diff;
 }
 
+#if ENABLE_THREADS
+static bool start_threads(ASS_Renderer *priv)
+{
+    if (priv->n_threads != priv->settings.threads && !priv->thread_start_failed) {
```
MrSmile, line 3511: Excess indentation. I think it's better to use
```c
     if (priv->n_threads == priv->settings.threads || priv->thread_start_failed)
        return true;
```

libass/ass_render.h
```diff
@@ -342,6 +349,21 @@ struct ass_renderer {
...
+#if ENABLE_THREADS
+    pthread_mutex_t mutex;
+    int mutex_inited;
+    pthread_cond_t main_cond, pool_cond;
+    int main_cond_inited, pool_cond_inited;
+
+    unsigned n_threads, started_threads, active_threads;
+    pthread_t *threads;
+
+    int thread_start_failed;
```
MrSmile, line 362: Failure modes are total mess here: 7 status variables instead of 1&ndash;2. I feel it can be greatly simplified.

MrSmile (commented):
I think that ideal solution for logging would be to pin in to the main thread only. But that would be quite substantial and intrusive change without much return...

libass/ass_cache.c
MrSmile: I think instead of locally redefining `NDEBUG`, it would be better to create a specialized time-critical assert function for use in performance-sensitive contexts. It's not exclusive to the cache system, potential tiled rendering can benefit from it too, for example.

libass/ass_cache.c
```diff
@@ -461,80 +661,115 @@ void ass_cache_inc_ref(void *value)
...
-        destroy_item(cache->desc, item);
-    } while (cache->cache_size > max_size);
-    if (cache->queue_first)
-        cache->queue_first->queue_prev = &cache->queue_first;
-    else
-        cache->queue_last = &cache->queue_first;
+        CacheItem *item_next = atomic_load_explicit(&item->next, memory_order_relaxed);
+        if (item_next)
+            item_next->prev = item->prev;
+        atomic_store_explicit(item->prev, item_next, memory_order_relaxed);
```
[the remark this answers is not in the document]
rcombs: That isn't really actionable on its own; do you have a full all-thread stack trace? The rest of the state of `*item`?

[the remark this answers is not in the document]
rcombs: I'm not sure if it's possible to debug with a pdb but no corresponding executable? (If there is, I don't know how to do it)

libass/ass_threading.h
```diff
@@ -0,0 +1,348 @@
...
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#endif
+
+#ifdef HAVE_STDATOMIC_H
+
+#include <stdatomic.h>
+#define HAVE_ATOMICS 1
+
+typedef intptr_t AtomicInt;
```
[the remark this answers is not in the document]
rcombs: I'd kinda prefer that all else being equal, but the Windows atomic functions used by the compat shims all operate on `LONG`/`LONG64`, which are signed, so I picked `intptr_t` for consistency with those.

libass/ass_threading.h
```diff
@@ -0,0 +1,348 @@
...
+#define memory_order_relaxed NoFence
+
+#define IS_RELAXED(order) ((#order)[0] == 'N')
+#define MAYBE_FENCE(order) (!IS_RELAXED(order) ? MemoryBarrier() : (void)0)
+
+#define DoCompareExchangeStrong(order) \
+static inline bool INTERLOCKED(DoCompareExchangePointer, order)(void *obj, void *expected, void *desired) \
+{ \
+    void *old = *(void**)expected; \
+    return (*(void**)expected = INTERLOCKED(CompareExchangePointer, order)((void**)obj, expected, desired)) == old; \
```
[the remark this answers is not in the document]
rcombs: We have a CI job that builds for win64 and it does work, but you have a point; I have no idea *why* this works. Changed to just do `InterlockedCompareExchangePointer ## order`.

libass/ass_threading.h
```diff
@@ -0,0 +1,348 @@
...
+}
+
+#endif /* CONFIG_PTHREAD */
+
+#endif /* HAVE_ATOMICS */
+
+#ifdef ENABLE_THREADS
+
+#define inc_ref(x) (void) (atomic_fetch_add_explicit(x, 1, memory_order_relaxed))
+#define dec_ref(x) (atomic_fetch_sub_explicit(x, 1, memory_order_acq_rel) - 1)
```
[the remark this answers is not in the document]
rcombs: These are meant specifically for refcounts, where release is correct for incrementing and acquire-release is correct for decrementing. Other usage can call the underlying functions directly.

libass/ass_shaper.c
```diff
@@ -425,25 +452,42 @@ get_contour_point(hb_font_t *font, void *font_data, hb_codepoint_t glyph,
...
+fail:
+    ass_font_unlock(metrics_priv->hash_key.font);
+    return ret;
 }
 
 bool ass_create_hb_font(ASS_Font *font, int index)
 {
-    FT_Face face = font->faces[index];
-    hb_face_t *hb_face = hb_face_create_for_tables(get_reference_table, face, NULL);
+    struct font_ref *key = malloc(sizeof(struct font_ref));
```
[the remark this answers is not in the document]
rcombs: Seems like a premature optimization to me; this only runs once per font face, it's not hot-path by any stretch of the imagination.

libass/ass_cache.c
```diff
@@ -377,11 +381,55 @@ Cache *ass_cache_create(const CacheDesc *desc)
...
+void ass_cache_client_set_done(CacheClientSet *set)
+{
+    ass_cache_client_set_clear(set);
+}
+
+CacheClient *ass_cache_client_create(CacheClientSet *set)
+{
+    CacheClient *client = NULL;
+
+    client = calloc(1, sizeof(*client));
```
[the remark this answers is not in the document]
rcombs: Just good practice in general; the extra cost of a cold-path `calloc` vs `malloc` is extremely negligible.

libass/ass_cache.c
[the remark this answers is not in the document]
rcombs: Can you reproduce this crash routinely, or is it rare and random? Does it happen with a particular sample file?

I've added some new debug assertions that may help catch this kind of issue sooner; if you pull the latest changes and build with `-DDEBUG_LEVEL=2`, hopefully you'll hit an assert before a crash like this happens, and the dump from that earlier crash should be useful in diagnosing the problem.

Side-note, what build system and compiler did you use to build the libass used in that dump? I noticed that the atomic variables behave somewhat strangely in the debug symbols; this probably isn't related to the cause of the crash, but it'd be useful to know what causes that.

libass/ass_threading.h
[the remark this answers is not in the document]
astiob: Could it be that our CI has stdatomic.h?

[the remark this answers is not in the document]
astiob: It seems [there was](https://github.com/libass/libass/actions/runs/12193986212/job/34017107542#step:10:316) an undefined-symbol warning:

```
'InterlockedCompareExchangePointer64' undefined; assuming extern returning int
```

It looks like in the end it only builds a static `libass.a`, so I guess the presence of external symbols isn’t verified. @TheOneric?

(Meanwhile in MinGW-w64 builds, stdatomic.h is indeed present.)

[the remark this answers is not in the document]
rcombs: Oh, do we not have tests in the meson builds yet?

libass/ass_cache.c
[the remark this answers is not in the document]
rcombs: I've pushed a change that I believe should fix this. It was reproducible by rendering multiple frames that each produced only one bitmap (i.e. a single event with a single line with `{\bord0\shad0}`). Thanks for the report; please confirm if it still recurs on the latest version of the branch.

libass/ass_threading.h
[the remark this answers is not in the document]
TheOneric: > do we not have tests in the meson builds yet?
nope, meson doesn’t even know how to build `compare`, `fuzz` or `checkasm`

Though we probably should switch the meson windows builds to `default_library=both` to make sure linking works. This should also catch undefined symbols though it’d be good to enable error and warning options in MSVC similar to the set we enable for GCC-like compilers if anyone knows which options MSVC has for this.

libass/ass_cache.c
rcombs: Went ahead and did this (piggybacking on `DEBUG_LEVEL`).

libass/ass_threading.h
```diff
@@ -0,0 +1,348 @@
...
+#define memory_order_acq_rel
+#define memory_order_acquire Acquire
+#define memory_order_release Release
+#define memory_order_consume Release
+#define memory_order_relaxed NoFence
+
+#define IS_RELAXED(order) ((#order)[0] == 'N')
+#define MAYBE_FENCE(order) (!IS_RELAXED(order) ? MemoryBarrier() : (void)0)
+
+#define DoCompareExchangeStrong(order) \
```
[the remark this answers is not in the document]
rcombs: Capitalized; not `#undef`-ing since these are internal and the names are pretty specific to atomics.

libass/ass_cache.h
```diff
@@ -52,7 +52,8 @@ typedef ass_hashcode (*HashFunction)(void *key, ass_hashcode hval);
 typedef bool (*HashCompare)(void *a, void *b);
 typedef bool (*CacheKeyMove)(void *dst, void *src);
 typedef size_t (*CacheValueConstructor)(void *key, void *value, void *priv);
-typedef void (*CacheItemDestructor)(void *key, void *value);
+typedef void (*CacheValueDestructor)(void *value);
+typedef void (*CacheKeyDestructor)(void *key);
```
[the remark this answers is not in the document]
rcombs: Done.

meson.build
```diff
@@ -52,7 +53,7 @@ add_project_arguments(
 
 # Configuration
 
-str_check_functions = ['strdup', 'strndup']
+str_check_functions = ['strdup', 'strndup', 'sched_getaffinity']
```
TheOneric, line 56: this won't work; functions in this array are cheked for existence _in the `string.h` header_

rcombs:
This is now rebased on latest upstream master; it's building on MSVC, and the tests are passing (getting the utils built did turn up a couple issues, which were trivial to fix; the tests themselves passed without any further changes).

rcombs:
Not sure why the MSVC build is broken after a rebase; my guess is it's something to do with the changes to the build configuration that have recently merged. Patches welcome from windows-touchers, or I can just turn threading back off for MSVC builds.

rcombs:
This was an entirely legitimate use-before-init on a mutex; no idea why it didn't crash previously, or was never caught by any of our sanitizers. Fixed; thanks much for the stack.

libass/ass_threading.h
```diff
@@ -0,0 +1,360 @@
...
+
+typedef struct ThreadStruct *pthread_t;
+
+typedef CRITICAL_SECTION   pthread_mutex_t;
+typedef CONDITION_VARIABLE pthread_cond_t;
+
+static inline int pthread_mutex_init(pthread_mutex_t *mtx, const void *attr)
+{
+    assert(!attr);
+    InitializeCriticalSection(mtx);
```
kasper93, line 212: If recursive mutexes are not needed, consider using SWR locks instead. They are lighter and faster.

rcombs: Interesting; it looks like SRW locks perform better in *most* (but not all) cases in this benchmark, but the results appear fairly noisy, and are highly dependent on exactly how much contention is happening. I'd still definitely be curious to see a measured impact in a real-world libass scenario.

ThaUnknown:
image

works as advertised

Andarwinux:
Yes, this allows fancy special effects subtitles to be rendered at an acceptable speed without any dropped frames on CPUs without excellent single-threaded performance. Unfortunately, this PR seems to have stalled.

ThaUnknown:
I think I found a problem

Please take this bug report with a heavy grain of salt, I might be doing something majorly wrong, as my implementation is quite "non-standard". So most of the things here are a "I think".

There appears to be fringing happening on the first character, symbol or glyph of each line. This happens on the start of the character/symbol usually only for the first 5-ish pixels, then everything else continues normally. If there are multiple lines in the presented subtitle, and they don't line up this issue shows up on both lines, so its like not a "5px of the left side of the bitmap" but "5px of the left side of the text line", I've put together a collage which shows this happening in many characters, symbols etc, all of which I doubled to show where the problem is visible:
image
It would be nice if some1 else could reproduce the issue, from what I can tell the font that causes this doesn't matter, but I'm not sure if things like outline etc matter, I can provide extra information if this is non-reproducible, but for me it happens 100% of the time and is 100% reproducible and consistent.


There's a non-0 chance this is also a bug in my implementation, as I'm using experimental vulkan features, an intel arc gpu and some quite stupid hacks to render this, but from my albeit limited testing I don't think that's the case.


Another problem I found is the thread count allocation, this could entirely be an emscripten/wasm bug, but I was not able to find information on this being the case so I'll report it just in case:
The thread count allocation seems weird, if you call `ass_set_threads` with `1`, it will only use the one main thread and no extra threads, but calling it with `2` makes it want to use the main thread +2 pthreads, which means `3`, which means you can never use 2 threads for rendering, only 1, 3, 4, 5 etc, also the entire library froze up when it wanted to spawn a pthread but there were no more available threads, but this could be a wasm problem

rcombs:
Re: the antialiasing oddities in that screenshot, I can't repro locally, and all test cases pass here (as well as in the CI). Does the test suite pass in your build configuration? To ask the obvious question, can you reproduce this *without* the "experimental vulkan features" and "stupid hacks"?

Re: thread count, you're describing the intended behavior. Setting the thread count to 2 means that 2 threads will be used for rendering (while the thread in `ass_render_frame` simply waits for them to finish). This is standard pool-size-configuration behavior; the point isn't to request that a particular number of total threads *exist*, but instead to request that no more than a particular number of cores are loaded at any given time.

> the entire library froze up when it wanted to spawn a pthread but there were no more available threads

What does it mean for there to be "no more available threads"? What is `pthread_create` returning? When you say the library "froze up", which call is blocking?

ThaUnknown:
Welp then just ignore me then if you can't reproduce this issue, I was hopeful that it wasn't something on my end but it must be.

> Re: the antialiasing oddities in that screenshot, I can't repro locally, and all test cases pass here (as well as in the CI). Does the test suite pass in your build configuration?

Ah you see here in 'ye ole' WASM land things aren't so simple, so I don't really know.don

> To ask the obvious question, can you reproduce this _without_ the "experimental vulkan features" and "stupid hacks"?

If it was easy I'd have done it in the first place, I spent 4 days making the last renderer, 2 of which I didn't sleep so I'll assume it's some bug in my pipeline somewhere.

> Re: thread count, you're describing the intended behavior. Setting the thread count to 2 means that 2 threads will be used for rendering (while the thread in `ass_render_frame` simply waits for them to finish). This is standard pool-size-configuration behavior; the point isn't to request that a particular number of total threads _exist_, but instead to request that no more than a particular number of cores are loaded at any given time.

Ah that make sense, unfortunate since it wastes quite a bit of resources in WASM land, but yeah if that's the norm I understand

> What does it mean for there to be "no more available threads"? What is `pthread_create` returning? When you say the library "froze up", which call is blocking?

When running WASM with pthreads for emscripten you specify how many pthreads it can create, so exceeding that value hangs. I can give you the WASM call id if you want to, but that's useless, can't do much more than that since libass+DWARF doesn't rly go well together idk why.

ThaUnknown:
yeah, sorry for spamming this thread with useless comments, the code from this PR works flawless, I fucked up floating point precision in the vulkan shader, my bad!

profile/profile.c
```diff
@@ -62,6 +68,20 @@ static void init(int frame_w, int frame_h)
...
+{
+#ifdef _WIN32
+    LARGE_INTEGER counter, frequency;
+    QueryPerformanceCounter(&counter);
+    QueryPerformanceFrequency(&frequency);
+    return counter.QuadPart / (double)frequency.QuadPart;
+#else
+    struct timespec spec;
+    clock_gettime(CLOCK_MONOTONIC, &spec);
+    return spec.tv_sec + spec.tv_nsec / 1000000000.;
```
MrSmile, line 84:
```c
return spec.tv_sec + spec.tv_nsec * 1e-9;
```
Division is slow and compilers cannot replace it with multiplication without `-ffast-math` (but they happily replace `a + b * c` with FMA with about the same discrepancy, where's logic in it?).

Not sure about returning floating point from timing functions though, I generally consider that a bad practice. Maybe it's better to use a more precise approach with separate multiplier applied later:
```c
static double get_time_factor(void)
{
#ifdef _WIN32
    LARGE_INTEGER frequency;
    QueryPerformanceFrequency(&frequency);
    return 1.0 / frequency.QuadPart;
#else
    return 1e-9;
#endif
}

static uint64_t get_time(void)
{
#ifdef _WIN32
    LARGE_INTEGER counter;
    QueryPerformanceCounter(&counter);
    return counter.QuadPart;
#else
    struct timespec spec;
    clock_gettime(CLOCK_MONOTONIC, &spec);
    return 1000000000 * spec.tv_sec + spec.tv_nsec;
#endif
}
```

profile/profile.c
```diff
@@ -102,6 +104,11 @@ int main(int argc, char *argv[])
         exit(1);
     }
 
+    const char *log_level_var = getenv("LIBASS_LOG_LEVEL");
```
MrSmile, line 108: Maybe command line parameter is better?

libass/ass_utils.h
```diff
@@ -55,6 +55,18 @@
 
 #define FEATURE_MASK(feat) (((uint32_t) 1) << (feat))
 
+#if DEBUG_LEVEL >= 1
+    #define ass_assert1(x) assert(x)
+#else
+    #define ass_assert1(x) (void)0
```
MrSmile, line 61: Maybe it's safer to add extra brackets `((void) 0)`?
Also I think it's better to be consistent with libass macro style and name it `ASSERT1`.

libass/ass_threading.h
```diff
@@ -0,0 +1,360 @@
...
+}
+
+#else /* ENABLE_THREADS */
+
+#define ENABLE_THREADS 0
+
+#define pthread_mutex_lock(x) do {} while(0)
+#define pthread_mutex_unlock(x) do {} while(0)
+#define pthread_cond_signal(x) do {} while(0)
+#define pthread_cond_broadcast(x) do {} while(0)
```
MrSmile, line 356: Earlier for no-op operations were used `((void) 0)`, better to be consistent.

libass/ass_font.c
```diff
@@ -498,17 +498,27 @@ size_t ass_font_construct(void *key, void *value, void *priv)
...
+
+    font->library = render_priv->library;
+
     int error = add_face(render_priv->fontselect, font, 0);
-    if (error == -1)
+    if (error == -1) {
         font->library = NULL;
+#if ENABLE_THREADS
+        pthread_mutex_destroy(&font->mutex);
+#endif
```
MrSmile, line 520: Maybe it's better to create no-op stub for `pthread_mutex_destroy` too?
Possibly even for `pthread_mutex_init`.

libass/ass_cache.c
```diff
@@ -453,59 +510,117 @@ void *ass_cache_get(Cache *cache, CacheClient *client, void *key, void *priv)
...
+
+retry:
+    for (item = start; item && item != stop_at; item = atomic_load_explicit(&item->next, memory_order_acquire)) {
+        if (item->hash == hash && desc->compare_func(key, (char *)item + key_offs))
+            break;
+    }
+
+    if (item != NULL && item != stop_at) {
+        if (atomic_load_explicit(&item->last_used_frame, memory_order_consume) != cache->cur_frame) {
+            uintptr_t last_used = atomic_exchange_explicit(&item->last_used_frame, cache->cur_frame, memory_order_consume);
```
MrSmile: `memory_order_consume` is considered somewhat problematic, so maybe it's better to use `memory_order_acquire` instead?

libass/ass_cache.c
```diff
@@ -453,59 +510,117 @@ void *ass_cache_get(Cache *cache, CacheClient *client, void *key, void *priv)
...
-
-    *cache->queue_last = item;
-    item->queue_prev = cache->queue_last;
-    cache->queue_last = &item->queue_next;
-    item->queue_next = NULL;
-    item->ref_count = 1;
-
-    cache->cache_size += item->size + (item->size == 1 ? 0 : CACHE_ITEM_SIZE);
+    size_t size = desc->construct_func(new_key, value, priv);
+    ass_assert1(size);
```
MrSmile, line 610: This assert is slow? Looks like it should be dominated by indirect function call earlier.

libass/ass_cache.c
```diff
@@ -453,59 +510,117 @@ void *ass_cache_get(Cache *cache, CacheClient *client, void *key, void *priv)
...
+
+    atomic_fetch_add_explicit(&cache->cache_size, size + (size == 1 ? 0 : CACHE_ITEM_SIZE), memory_order_relaxed);
+
+#if ENABLE_THREADS
```
MrSmile, line 614: `#if` should be unnecessary.

libass/ass_cache.c
```diff
@@ -453,59 +510,117 @@ void *ass_cache_get(Cache *cache, CacheClient *client, void *key, void *priv)
...
+    pthread_mutex_lock(&client->mutex);
+#endif
+
+    atomic_store_explicit(&item->size, size, memory_order_release);
+
+#if ENABLE_THREADS
```
MrSmile, line 620: `#if` should be unnecessary.

libass/ass_cache.c
```diff
@@ -531,81 +647,134 @@ void ass_cache_inc_ref(void *value)
     if (!value)
         return;
     CacheItem *item = value_to_item(value);
-    assert(item->size && item->ref_count);
-    item->ref_count++;
+    ass_assert2(atomic_load_explicit(&item->size, memory_order_acquire) && atomic_load_explicit(&item->ref_count, memory_order_acquire));
```
MrSmile: Better to split into two individual asserts.

libass/ass_cache.c
```diff
@@ -531,81 +647,134 @@ void ass_cache_inc_ref(void *value)
...
+    inc_ref(&item->ref_count);
+}
+
+static void dec_ref_item(CacheItem *item)
+{
+    ass_assert2(atomic_load_explicit(&item->size, memory_order_acquire) && atomic_load_explicit(&item->ref_count, memory_order_acquire));
```
MrSmile: Here too.

libass/ass_threading.h
```diff
@@ -0,0 +1,360 @@
...
+#else // defined(__APPLE__)
+    pthread_setname_np(pthread_self(), name);
+#endif // defined(__APPLE__)
+#endif // HAVE_PTHREAD_SETNAME_NP
+}
+
+#endif /* CONFIG_PTHREAD */
+
+#define inc_ref(x) (void) (atomic_fetch_add_explicit(x, 1, memory_order_relaxed))
+#define dec_ref(x) (atomic_fetch_sub_explicit(x, 1, memory_order_acq_rel) - 1)
```
MrSmile, line 314: Not sure about usefulness of this. I think it's better to keep memory order explicit.

MrSmile (commented):
Mostly cosmetic nits. Substantial code should be fairly tested by now, I think. What do we do about pathological cases (memory bound) though? Ignore for now?

profile/profile.c
TheOneric: I like the `get_factor_version`. To avoid potential precision loss though, we could return the integer value directly as `uint64_t get_time_denominator(void)`

profile/profile.c
```diff
@@ -82,21 +102,66 @@ int main(int argc, char *argv[])
...
 
+    double cleanup_time = gettime();
+
+    printf("Timing:\n");
+    printf("           init: %f\n", init_time - start_time);
+    printf("           read: %f\n", read_time - init_time);
+    printf("   total render: %f\n", last_frame_time - read_time);
+    if (frames > 0) {
+        printf("    first frame: %f\n", first_frame_time - read_time);
+        printf("     post-first: %f\n", last_frame_time - first_frame_time);
```
TheOneric, line 164: as is, this seems redundant with the already logged “total render” and “first frame”. Instead I think it would be good to log the count of rendered frames (perhaps differentiating empty frames without any `ASS_Image`s from those with actual results)

profile/profile.c
```diff
@@ -82,21 +102,66 @@ int main(int argc, char *argv[])
...
+        printf("    worst frame: %f\n", worst_frame);
+    }
+    printf("        cleanup: %f\n", cleanup_time - last_frame_time);
+    if (frames > 0) {
+        printf("      total fps: %f\n", frames / (last_frame_time - read_time));
+        printf("     post-1 fps: %f\n", frames / (last_frame_time - first_frame_time));
+        printf("     post-1 fps: %f\n", frames / (last_frame_time - first_frame_time));
```
TheOneric: “post-1 fps” log is duplicated

profile/profile.c
```diff
@@ -35,9 +41,9 @@ void msg_callback(int level, const char *fmt, va_list va, void *data)
 {
     if (level > 6)
         return;
-    printf("libass: ");
-    vprintf(fmt, va);
-    printf("\n");
+    char fmt_buf[1024];
+    snprintf(fmt_buf, sizeof(fmt_buf), "libass: %s\n", fmt);
+    vfprintf(stderr, fmt_buf, va);
```
TheOneric, line 48: The `snprintf` result is used unchecked (and thus might miss the terminating zero byte).
Given it only needs to accommodate _format_ strings from our own code and `assert` might suffice, but there should be some form of check

TheOneric (commented):
The first two commits could already have been merged much earlier like the other prep work, wonder why we didn’t do that yet

libass/ass_cache.c
TheOneric: `client->mutex` only exists if `ENABLE_THREADS` is defined thruthy.
*EDIT:* ah, reading your other comments, you probably mean it won’t be necessary anymore with a noop mutex creation/destruction macro for threadless builds

ThaUnknown:
would it be somehow possible to reduce the amt of calls for system time when WebAssembly build target is detected? they are quite expensive since they cross the WASM boundary, that said I don't know if that even is possible with the current architecture of this PR

other than that I've been running this PR in my software, and after a few hundred thousand hours of usage no issues arose, good job on this

TheOneric:
if you’re talking about calls to the `get_time` function discussed in preceding set of comments, they are part of the `profile` utility not the library itself. Thus I don’t think performance of it in WASM amtters; unless you want to use the profile utility inside WASM, in which case you’re welcome to contribute a more efficient way to get execution times in WASM builds. But even then this doesn’t seem like a blocker to me

Andarwinux:
To improve WASM performance, it may be necessary to add WASM simd128 asm. On x86, even with clang avx512 automatic vectorization, the C implementation in libass is still extremely slow.

ThaUnknown:
> calls to the get_time function

no, I _**think**_ SleepConditionVariableSRW calls system time, do it enough times and it stacks up

>  you’re welcome to contribute a more efficient way to get execution times in WASM builds

I wish I was good enough at C to do that, I'd have submitted a PR to this PR by now, unfortunately I struggled, so I came here to ask if reducing the amt of system calls for time even is possible at all

> improve WASM performance

[libass in WASM performs fine](https://jassub.pages.dev/beastars/jassub/), this was strictly about the [relatively] frequent system time calls when a LOOOT of rendering is being done

rcombs:
Well given that SleepConditionVariableSRW is a WinAPI function, it seems *pretty* unlikely that it would be called in a wasm build. Can we take the unrelated chatter elsewhere, please?

profile/profile.c
rcombs: > and thus might miss the terminating zero byte

Not sure what this means? `snprintf` always 0-terminates its output even in the insufficient-space case. Might miss the trailing newline, though.

libass/ass_threading.h
rcombs: The memory ordering required for refcounting is always the same; having it specified in a single consistent place avoids confusion.

profile/profile.c
rcombs: Is this realistically ever going to matter at the precisions we're working with here?

libass/ass_cache.c
rcombs: Hmmm, it looks like `memory_order_consume` was deprecated in C++, and has never been implemented anywhere (it's effectively identical to `memory_order_acquire` in all real implementations), so sure, might as well make that explicit here.

libass/ass_cache.c
rcombs: I don't expect it to be particularly slow; I'm just trying to avoid 0-level asserts in cases where failure is still *safe*.

libass/ass_cache.c
rcombs: Personally, I think it's a little icky to have code that appears to reference struct members that might not exist (and then just have it not be compiled because it's a macro arg).

profile/profile.c
TheOneric: you’re right, my bad. Mixed it up with other APIs like `strncpy`.
Might still be nice to have something loudly complain in case the buffer ever ends up too small, so we can bump it. But nothing critical

profile/profile.c
```diff
@@ -82,21 +105,70 @@ int main(int argc, char *argv[])
...
+    printf("   total render: %f\n", last_frame_time - read_time);
+    if (frames > 0) {
+        printf("    first frame: %f\n", first_frame_time - read_time);
+        printf("     post-first: %f\n", last_frame_time - first_frame_time);
+        printf("    worst frame: %f\n", worst_frame);
+    }
+    printf("        cleanup: %f\n", cleanup_time - last_frame_time);
+    if (frames > 0) {
+        printf("      total fps: %f\n", frames / (last_frame_time - read_time));
+        printf("     post-1 fps: %f\n", frames / (last_frame_time - first_frame_time));
```
TheOneric, line 170: It’s a false warning since it’s gated behind `frames  > 0`, but my compiler complains about `first_frame_time` being used here while being possibly uninitialised.  
Though given `last_frame_time` also already gets a default initialisation, we could just initialise `first_frame_time` too for consistency avoiding the warning

profile/profile.c
TheOneric: this is what i’m thinking of *(`average frame` is just include for a convenient comparison; strictly speaking it’s redundant)*:
```diff
diff --git a/profile/profile.c b/profile/profile.c
index e8aed7f2..89955d8a 100644
--- a/profile/profile.c
+++ b/profile/profile.c
@@ -128,6 +128,7 @@ int main(int argc, char *argv[])
     double worst_frame = 0;
     bool got_frame = false;
     int frames = 0;
+    int nonempty_frames = 0;
 
     while (tm < end_time) {
         bool got = ass_render_frame(ass_renderer, track, (int) (tm * 1000), NULL) != NULL;
@@ -146,6 +147,7 @@ int main(int argc, char *argv[])
         last_frame_time = current_time;
 
         frames++;
+        if (got) nonempty_frames++;
     }
 
     ass_free_track(track);
@@ -154,14 +156,17 @@ int main(int argc, char *argv[])
 
     double cleanup_time = gettime();
 
-    printf("Timing:\n");
+    printf("Timing [seconds]:\n");
     printf("           init: %f\n", init_time - start_time);
     printf("           read: %f\n", read_time - init_time);
     printf("   total render: %f\n", last_frame_time - read_time);
     if (frames > 0) {
         printf("    first frame: %f\n", first_frame_time - read_time);
-        printf("     post-first: %f\n", last_frame_time - first_frame_time);
         printf("    worst frame: %f\n", worst_frame);
+        printf("  average frame: %f\n", (last_frame_time - read_time) / frames);
+        printf("     all frames: %f\n", last_frame_time - read_time);
+        printf("total frame cnt: %d\n", frames);
+        printf("nonempty frames: %d\n", nonempty_frames);
     }
     printf("        cleanup: %f\n", cleanup_time - last_frame_time);
     if (frames > 0) {

```

PR #794, opened 2024-06-22
Use IDWriteFontFallback when it is available

moi15moi:
IDWriteFontFallback is available on Windows 8.1 and more.

There isn't any major advantage over the IDWriteTextLayout "technique". It is just less "hacky".

### DIFF
--- a/libass/ass_directwrite.c
+++ b/libass/ass_directwrite.c
@@ -253,6 +253,145 @@ static void init_FallbackLogTextRenderer(FallbackLogTextRenderer *r,
     };
 }
 
+typedef struct FallbackLogTextAnalysisSource {
+    IDWriteTextAnalysisSource iface;
+    IDWriteTextAnalysisSourceVtbl vtbl;
+    wchar_t* locale;
+    UINT locale_len;
+    wchar_t* text;
+    UINT text_len;
+    LONG ref_count;
+} FallbackLogTextAnalysisSource;
+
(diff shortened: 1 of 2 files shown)


moi15moi:
Will this PR eventually get reviewed?

astiob:
At a very quick glance: this seems to add quite a lot of code, but the old code is preserved and you say the new code has no advantage. What’s the point, then? Is there really nothing to gain here?

moi15moi:
> What’s the point, then?

My first message pretty much say it. It is just less "hacky". That's the only point.
I compared the performance with IDWriteFontFallback and IDWriteTextLayout and they are pretty much the same.

I would totally understand that you would reject the PR.

astiob:
Thanks. In that case:

I’d like a second opinion from other maintainers, but personally, I think it’s good to have this patch here on GitHub for posterity—for whenever (if ever) we decide to drop support for Windows versions prior to 8.1—but I’d rather avoid merging it, because it seems to increase maintenance burden (and attack surface) for no real benefit: even if the new code is cleaner, the old code is still there and continues to be maintained.

TheOneric:
I’m not familiar with DirectWrite API, so i’m just basing this off descriptions from prior comments here:  if there’s no benefit for users and it only increases code complexity atm, i agree merging doesn’t make sense until(if) it can replace the current "hacky" code path

PR #795, opened 2024-06-24
build: move warning flags to configure.ac

rcombs:
This lets us check for warning flags at configure-time, avoiding introducing errors if a compiler doesn't support a particular flag.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -53,6 +53,13 @@ FUZZ_CPPFLAGS="${FUZZ_CPPFLAGS:--DASS_FUZZMODE=0}"
 # Checks for flags support
 AX_APPEND_COMPILE_FLAGS([-fno-math-errno])
 
+# Enable some warnings
+AX_APPEND_COMPILE_FLAGS([ \
+    -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter -Wstrict-prototypes \
+    -Wpointer-arith -Werror-implicit-function-declaration -Wredundant-decls \
+    -Wno-missing-field-initializers \
+])
+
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -53,6 +53,11 @@ FUZZ_CPPFLAGS="${FUZZ_CPPFLAGS:--DASS_FUZZMODE=0}"
 # Checks for flags support
 AX_APPEND_COMPILE_FLAGS([-fno-math-errno])
 
+# Enable some warnings
+AX_APPEND_COMPILE_FLAGS([-Wall -Wextra -Wno-sign-compare -Wno-unused-parameter])
+AX_APPEND_COMPILE_FLAGS([-Werror-implicit-function-declaration -Wstrict-prototypes])
+AX_APPEND_COMPILE_FLAGS([-Wpointer-arith -Wredundant-decls -Wno-missing-field-initializers])
+
```
TheOneric, line 60: Was wondering why this is split into three macro calls, but the need for escaping the initial and final newline might be surprising and unintuitive with C-like syntax in mind
```autoconf
AX_APPEND_COMPILE_FLAGS([ \
    -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter \
    -Werror-implicit-function-declaration -Wstrict-prototypes \
    -Wpointer-arith -Wredundant-decls -Wno-missing-field-initializers \
])
```

Feel free to keep the current version or use the above if you think it’s clear enough

TheOneric (approved):

configure.ac
rcombs: Sure, fair enough.

PR #796, opened 2024-06-25
c/blur: use ptrdiff_t to avoid unsigned underflow

rcombs:
Found by coverity-scan, 462876-462879


MrSmile:
Unsigned overflow is intended here, so I'm not sure about that diagnostic. Basically this replaces `size_t` wrap-around with `ptrdiff_t` → `size_t` conversion (with that same wrap-around). I think that resulting code is even less safe as it involves signed integers, where overflows are UB instead. To fully silence static analyzers I think we should convert argument of `get_line()` into `ptrdiff_t` too, with manual cast to `size_t` in comparison later.

But I'm not sure about fixing not broken. Next it nags about hash calculations or what? I think the same problem with overzealous analyzers recently arises in Linux kernel too.

rcombs:
The arg to `get_line` is already changed in this PR. I don't believe signed overflow is possible here (that would require a >2GB image, which would trip earlier max size checks), and the behavior is clearer to the reader (whether human or analyzer) when using a signed arg than when relying on unsigned underflow/wraparound.

MrSmile:
> The arg to `get_line` is already changed in this PR.

Ah, I missed that. But in this case `offs < size` should be replaced with `(size_t) offs < size`, as comparison of mixed signedness is a much bigger problem than unsigned overflow.

rcombs:
I just realized I'd misunderstood what the `get_line` code was supposed to do; I've added an explicit check for `offs < 0`. If my understanding of this code is correct, this substantially clarifies the intent here.

MrSmile:
That makes code significantly slower. `get_line()` is on the hot path and additional check adds extra `cmov` in `clang` and even degrades to branches in `gcc`.

rcombs:
This code produces identical output to the existing version on clang and gcc, and is clearer imo:
```
const int16_t *get_line(const int16_t *ptr, ptrdiff_t offs, size_t size)
{
    if (size > PTRDIFF_MAX) // This precondition was already required for the existing code to work
        unreachable();

    return (offs >= 0 && offs < (ptrdiff_t)size) ? ptr + offs : zero_line;
}
```
(This requires a backport of the `unreachable` macro from C23, but that's simple enough, and probably worthwhile for other optimization hints as well.)

This still produces branches in MSVC, but MSVC *already* uses branches for this function on x86 and x64, so I'm not sure how big of a deal that is. (Also, MSVC exclusively supports platforms where we have assembly for this code.)

If we *do* decide to keep the existing wraparound code, I'd say we should at least add comments clarifying the intent.

MrSmile:
Maybe that diagnostic has triggered due to `size_t - int` operation and `size_t - size_t` would be perfectly ok?
There're errors for `offs - n * STRIPE_WIDTH` but not for `offs - n * step` with `size_t step`.
So it would be enough to redefine
```c
#define STRIPE_WIDTH  ((size_t) ALIGNMENT / 2)
```

Alternatively, something like this could be a good compromise too:
```c
inline static const int16_t *get_line(const int16_t *ptr, ptrdiff_t offs, size_t size)
{
    // single comparison version to help performance
    // it is enough as negatives wrap around to large values
    return (size_t) offs < size ? ptr + offs : zero_line;
}
```
but I prefer to not deal with signed intermediates.

Issue #529, opened 2021-07-08
Enable GCC and Clang warnings about nonstandard `printf` format strings


astiob:
GCC needs `-Wpedantic`, which also displays a few other warnings that we may want to fix. Clang also supports this flag and displays a few other warnings (mostly the same as GCC), but nonstandard format string warnings are not among them.

Clang needs `-Wformat-non-iso`, which GCC doesn’t support (it exits with an error about an unrecognized flag).

We need a `configure` check for `-Wformat-non-iso` at the very least. We probably want to do configure checks for both of these and also for those options that we currently unconditionally add in `Makefile.am`.

rcombs:
I don't think `-Wpedantic` is a great call here; it's extremely picky about plenty of stuff that's entirely well-defined and sensible on all supported platforms.

What specific cases are we actually worried about here? Using C11 format string modifiers while building for GNU99? I think the more appropriate route at this point is to just bump the build to GNU11 and note that we require a sufficiently recent (i.e. ten years!) compiler and standard library.

astiob:
> Using C11 format string modifiers while building for GNU99?

No, just using formats that aren’t standard C at all, like `%m`. And the only way to get warned about `%m` is to use GCC and give it `-Wpedantic`.

(FWIW I don’t think standard `printf` formats have been extended since C99.)

> it's extremely picky about plenty of stuff that's entirely well-defined and sensible on all supported platforms

You’re not wrong. But then again, it gives only a few warnings for current libass, and they’re easy to fix, even if the fixes do introduce a little noise. And it’s the only way to get nonstandard-format warnings.

> bump the build to GNU11 and note that we require a sufficiently recent (i.e. ten years!) compiler and standard library.

For that matter, it’s not just about the standard library. **Our log messages do not go to the standard library. They are part of our own API.** They’re parsed by the log callback, which is implemented by the consumer, so whenever we start using any new format specifier, we’re breaking the API in a way that can’t even be automatically detected at compile-time.

Even now, we neither explicitly document what formats we use nor are consistent about it: we use `%zu`, and at the same time we use `PRId64` (which may be `%I64d`) with `(int64_t) long_long_value`.

It seems we have a majority consensus that using standard C99 formats is OK, but I’d prefer that we document this, stick to it consistently, and then avoid any future standard formats until/unless we really need them, even if we eagerly start using `-std=gnu23+` to compile ourselves.

rcombs:
[...]

PR #798, opened 2024-07-03
build: warn about non-ISO format strings with clang

TheOneric:
With `-Wpedantic` instead to also cover GCC, we get three warnings about missing `(void *)` casts for printf arguments in both GCC and clang *(since different pointer types are allowed to have different representations, `%p` only taking `void *` and due to varargs no implicit cast can occur)*. While afaik no modern arch actually has different representations, the warning is entirely valid and easy enough to fix; imho we should add a cast here.

Only with GCC we additionally get 5 instances of `invalid use of pointers to arrays with different qualifiers in ISO C before C2X`. I haven’t looked into this but C23 adopting this as safe suggests there might not be much of at all problems in practice.

If `-Wpedantic` continues to be too contentious, we can at least enable this warning for clang; with `AX_APPEND_COMPILE_FLAG` this is now trivial enough.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -59,6 +59,8 @@ AX_APPEND_COMPILE_FLAGS([ \
     -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter -Wstrict-prototypes \
     -Wpointer-arith -Werror-implicit-function-declaration -Wredundant-decls \
     -Wno-missing-field-initializers \
+    dnl clang-specific, gcc currently has no fine-grained toggle for this, only -Wpedantic
+    -Wformat-non-iso -Werror=format-non-iso \
 ])
 
 # Checks for available libraries and define corresponding C Macros
--- a/meson.build
+++ b/meson.build
@@ -40,6 +40,8 @@ if cc.get_argument_syntax() == 'gcc'
         '-Wpointer-arith',
         '-Wredundant-decls',
         '-Wno-missing-field-initializers',
+        '-Wformat-non-iso',
+        '-Werror=format-non-iso',
     ]
 endif
 


rcombs:
Would we get more useful results out of adding the `_Printf_format_string_` attr? What cases is this flag actually capable of catching?

Re: `-Wpedantic`, I don't find the idea of someone ever trying to run libass on a system where pointers have different representations particularly credible.

TheOneric:
> Would we get more useful results out of adding the _Printf_format_string_ attr?

`printf` attributes already got added to `ass_msg` in #514 and adapted for Windows in #523
Did more specific ISO-only attributes emerge since?

> What cases is this flag actually capable of catching?

Accidental use of non-standard format specifiers or modifiers; see #529 for prior discussion and examples for when this matters.

TheOneric:
From #529:
> On outdated MinGW-w64 (and IIRC on current suddenly-revived MinGW32), PRId64 is still "I64d" even with -std=gnu99 or whatever we use.
>
> But glibc has its own I flag, which enables locale-specific digits (à la Windows).
> 
> So GCC [...] warns that int arguments are expected but long long values are passed in.
> 
> Of course, the root of the problem is our use of system-specific PRId64 with system-independent gnu_printf; but even so, this particular warning is especially disruptive, as it claims there’s a legitimate problem when there isn’t. If we forced GCC to recognize only standard C formats, it would instead complain that I is unrecognized, which is more easily shrugged off by the user.

At least for clang 15’s `format-non-iso` this doesn’t appear to actually pan out like hoped. It instead issues only a `-Wformat` warning similar to what’s described for `-Wpednatic`-less gcc. Furthermore, using POSIX’s `'` modifier doesn't result in any warning at all, while `n$` leads to a `format-non-iso` warning as expected.

This isn’t ideal, but at least will catch some mistakes. Since there don’t seem to be any expected harmless mismatch scenarios involving clang, i turned `format-non-iso` into an error now to avoid the (limited) cases this actually works for being overllooked when using gcc locally.

PR #799, opened 2024-07-07
fuzz: allow outputting parsed file content

TheOneric:
This was useful for OSS Fuzz’ timeout sample from 2 years ago and again for the  current one to figure out what’s (not) going on, so here’s a cleaned up version. Of course, if we actually have an infinite loop or very expensive operation in initially parsing, the writeout won’t halt (in reasonable time) either but this in itself already provides some insight

### DIFF
--- a/fuzz/fuzz.c
+++ b/fuzz/fuzz.c
@@ -117,7 +117,7 @@ static bool init(void)
 }
 
 
-static inline void process_image(ASS_Image* imgs)
+static inline void process_image(ASS_Image *imgs)
 {
     for (; imgs; imgs = imgs->next) {
         assert(imgs->w >= 0 && imgs->h >= 0 &&
@@ -156,7 +156,7 @@ static void consume_track(ASS_Renderer *renderer, ASS_Track *track)
 static ASS_Track *read_track_from_stdin(void)
 {
     size_t smax = 4096;
-    char* buf = malloc(smax);
+    char *buf = malloc(smax);
     if (!buf)
         goto error;
     size_t s = 0;
@@ -168,7 +168,7 @@ static ASS_Track *read_track_from_stdin(void)
         s += read_b > 0 ? read_b : 0;
         if (s == smax) {
             size_t new_smax = smax > SIZE_MAX / 2 ? SIZE_MAX : smax * 2;
-            char* new_buf = realloc(buf, new_smax);
+            char *new_buf = realloc(buf, new_smax);
             if (!new_buf || new_smax <= smax) {
                 free(new_buf ? new_buf : buf);
                 goto error;
--- a/fuzz/fuzz.c
+++ b/fuzz/fuzz.c
@@ -153,6 +153,18 @@ static void consume_track(ASS_Renderer *renderer, ASS_Track *track)
 }
 
 #if ASS_FUZZMODE == FUZZMODE_STANDALONE
+#include "writeout.h"
+
+struct settings {
+    enum {
+        CONSUME_INPUT,
+        WRITEOUT_TRACK
+    } mode;
+    const char *input; // path or "-" for stdin
+    const char *output; // path or NULL for tmp file
+};
+
+
 static ASS_Track *read_track_from_stdin(void)
 {
     size_t smax = 4096;
@@ -186,6 +198,64 @@ error:
     return NULL;
 }
 
+/**
+ * \param argc
+ * \param argv
+ * \param settings will be filled according to parsed args or defaults
+ * \return whether CLI args could be parsed successfully
+ */
+static bool parse_cmdline(int argc, char *argv[], struct settings *settings)
+{
+    // defaults
+    settings->mode = CONSUME_INPUT;
+    settings->input = NULL;
+    settings->output = NULL;
+
+    int i;
+    for (i = 1; i < argc; i++) {
+        const char *param = argv[i];
+        if (!param || param[0] != '-' || !param[1])
+            goto no_more_args;
+
+        switch (param[1]) {
+        case 'q':
+            quiet = true;
+            break;
+
+        case 'o':
+            settings->mode = WRITEOUT_TRACK;
+            // optional argument
+            if (argc - i > 1 && argv[i + 1][0] != '-') {
+                settings->output = argv[i + 1];
+                i++;
+            }
+            break;
+
+        case '-':
+            if (param[2]) {
+                return false;
+            } else {
+                i++;
+                goto no_more_args;
+            }
+
+        default:
+            return false;
+        }
+
+        continue;
+
+no_more_args:
+        break;
+    }
+
+    if (argc < 2 || argc - i > 1 || argc == i)
+        return false;
+
+    settings->input = argv[argc - 1];
+    return !!settings->input;
+}
+
 int main(int argc, char *argv[])
 {
     /* Default failure code of sanitisers is 1, unless
@@ -206,16 +276,18 @@ int main(int argc, char *argv[])
     ASS_Track *track = NULL;
     int retval = FUZZ_OK;
 
-    if (argc < 2 || argc > 3 ||
-            (argc == 3 && strcmp(argv[1], "-q")) ) {
-        printf("usage: %s [-q] <subtitle file>\n", argc ? argv[0] : "fuzz");
+    struct settings settings;
+    if (!parse_cmdline(argc, argv, &settings)) {
+        printf("usage: %s [-q] [-o [output_file]] [--] <subtitle file>\n"
+               "  -q:\n"
+               "    Hide libass log messages\n"
+               "\n"
+               "  -o [FILE]:\n"
+               "    Write out parsed file content in a standardized form\n"
+               "    into FILE or if omitted a generated temproary file.\n"
+               "    If used the input file will not be processed, only parsed.\n",
+               argc ? argv[0] : "fuzz");
         return FUZZ_BAD_USAGE;
-    }
-
-    size_t fileidx = 1;
-    if (argc == 3) {
-        quiet = true;
-        fileidx = 2;
     }
 
     if (!init()) {
@@ -224,8 +296,8 @@ int main(int argc, char *argv[])
         goto cleanup;
     }
 
-    if (strcmp(argv[fileidx], "-"))
-        track = ass_read_file(ass_library, argv[fileidx], NULL);
+    if (strcmp(settings.input, "-"))
+        track = ass_read_file(ass_library, settings.input, NULL);
     else
         track = read_track_from_stdin();
 
@@ -235,7 +307,15 @@ int main(int argc, char *argv[])
         goto cleanup;
     }
 
-    consume_track(ass_renderer, track);
+    switch (settings.mode) {
+    case CONSUME_INPUT:
+        consume_track(ass_renderer, track);
+        break;
+
+    case WRITEOUT_TRACK:
+        write_out_track(track, settings.output);
+        break;
+    }
 
 cleanup:
     if (track)        ass_free_track(track);
--- a/fuzz/writeout.c
+++ b/fuzz/writeout.c
@@ -1,0 +1,216 @@
+/*
+ * Copyright (C) 2024 libass contributors
+ *
+ * This file is part of libass.
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <math.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "writeout.h"
+
+
+static const char *ycbcr_to_str(ASS_YCbCrMatrix ycbcr_val)
+{
+    switch (ycbcr_val) {
+        case YCBCR_NONE:         return "None";
+        case YCBCR_BT601_TV:     return "TV.601";
+        case YCBCR_BT601_PC:     return "PC.601";
+        case YCBCR_BT709_TV:     return "TV.709";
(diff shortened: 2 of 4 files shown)


fuzz/writeout.c
```diff
@@ -0,0 +1,217 @@
...
+{
+    time /= 10; // ASS files can only have centi-second precision
+    int sign = time < 0 ? -1 : 1;
+    time = labs(time);
+
+    // Without chars for hours, except for sign indicator
+    size_t min_size = sign == -1 ? 14 : 10;
+    long ntime;
+    int cs, s, m;
+    #define DIVMOD(div,mod) ntime = time / div; mod = time - div * ntime; time = ntime
```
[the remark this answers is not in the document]
TheOneric: I don’t recall why i did it this way originally, but even if there’s a perf difference, this isn’t in a perf-critical fuzzing path. I’ll switch it to the simpler version

fuzz/fuzz.c
```diff
@@ -206,26 +276,28 @@ int main(int argc, char *argv[])
...
-            (argc == 3 && strcmp(argv[1], "-q")) ) {
-        printf("usage: %s [-q] <subtitle file>\n", argc ? argv[0] : "fuzz");
+    struct settings settings;
+    if (!parse_cmdline(argc, argv, &settings)) {
+        printf("usage: %s [-q] [-o [output_file]] [--] <subtitle file>\n"
+               "  -q:\n"
+               "    Hide libass log messages\n"
+               "\n"
+               "  -o [FILE]:\n"
+               "    Write out parsed file content in a standardized from\n"
```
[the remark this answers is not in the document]
TheOneric: fixed, thx

fuzz/writeout.c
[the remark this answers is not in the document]
MrSmile: I recall there was a platform without remainder operation, so it was implemented through integer FMA like this.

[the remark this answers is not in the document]
rcombs: As mentioned, this isn't hot-path code anyway, but for reference: I checked several compilers and architectures in godbolt, and every version I saw either uses a native remainder operation for `/` + `%`, or does this same multiply-and-subtract operation (sometimes with a dedicated combined instruction, if available).

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+        char* nbuf = realloc(*buf, min_size + 14);
+        if (!nbuf) {
+            return false;
+        }
+        *buf = nbuf;
+        *buf_size = min_size + 14;
+    }
+
+    int ret = snprintf(*buf, *buf_size, "%ld:%02d:%02d.%02d",
+                       time * sign, m * sign, s * sign, cs * sign);
```
MrSmile, line 73: I think it would be better to make `m`, `s` and `cs` strictly positive and only `h` potentially negative.

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+    if (ret < 0 || ret >= *buf_size)
+        return false;
+    return true;
```
MrSmile, line 76: I think simple `return ret >= 0 && res < *buf_size` would be better.

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+static int ssa2ass_align(int ssa_align)
+{
+    return ((ssa_align & 0xC) >> 2) * 3 + (ssa_align & 0x3);
+}
+
+static const char *tracktype_to_str(int track_type)
+{
+    return track_type == TRACK_TYPE_ASS ? "ASS"
+        : track_type == TRACK_TYPE_SSA ? "SSA"
+        : "other";
```
MrSmile, line 88: Maybe `switch` would be better?

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+}
+
+
+static void write_header(FILE *f, ASS_Track *track, const char* originalformat_name)
```
MrSmile, line 92: Style nit: space `char *originalformat_name`.

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+{
+    fprintf(f, "\n[V4+ Styles]\n");
+    fprintf(f, "Format: "
+        "Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, "
+        "OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, "
+        "ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, "
+        "Alignment, MarginL, MarginR, MarginV, Encoding\n");
+
+    // don't print out builtin fallback
+    for (int i = 1; i < track->n_styles; ++i) {
```
MrSmile, line 132: I think we decided to use `i++` in this context.

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+        ASS_Style *s = track->styles + i;
+        fprintf(f, "Style: "
+                   "%s,%s,%.3f,&H%08X,&H%08X,&H%08X,&H%08X,%d,%d,%d,%d,"
+                   "%.3f,%.3f,%.3f,%.3f,%d,%.3f,%.3f,%d,%d,%d,%d,%d\n",
+            s->Name, s->FontName, s->FontSize, s->PrimaryColour, s->SecondaryColour,
+            s->OutlineColour, s->BackColour, s->Bold, s->Italic, s->Underline, s->StrikeOut,
+            s->ScaleX, s->ScaleY, s->Spacing, s->Angle, s->BorderStyle, s->Outline, s->Shadow,
+            ssa2ass_align(s->Alignment), s->MarginL, s->MarginR, s->MarginV, s->Encoding
+            );
```
MrSmile, line 141: I think per style guide `);` should be in the previous line.

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+    }
+}
+
+static void write_events(FILE *f, ASS_Track *track)
+{
+    size_t start_size = 128;
+    char* start = calloc(1, start_size);
```
MrSmile, line 148: Style nit: space `char *start`.

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+    if (!start) {
+        printf("Failure to alloc start time buffer!\n");
+        return;
+    }
+
+    size_t end_size = 128;
+    char* end = calloc(1, end_size);
```
MrSmile, line 155: Here too.

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+        }
+        fprintf(f, "Dialogue: %d,%s,%s,%s,%s,%03d,%03d,%03d,%s,%s\n",
+            e->Layer, start, end, (track->styles + e->Style)->Name, e->Name,
+            e->MarginL, e->MarginR, e->MarginV, e->Effect, e->Text);
+    }
+}
+
+void write_out_track(ASS_Track *track, const char *outpath)
+{
+    const char* originalformat_name = tracktype_to_str(track->track_type);
```
MrSmile, line 183: Here too.

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+        f = fopen(outpath, "w");
+        printf("Parsed File will be written to:  %s\n", outpath);
+    } else {
+        char filename[] = "/tmp/parsedSubs_XXXXXX";
+        int fd = mkstemp(filename);
+        if (fd == -1) {
+            printf("Failed to acquire temporary file!\n");
+            return;
+        }
+        FILE* f = fdopen(fd, "w");
```
MrSmile, line 201: Here too.

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+        if (!f) {
+            printf("Failed to associate fd with a stream!\n");
+            return;
+        }
+        printf("Parsed File will be written to:  %.*s\n", (int)sizeof(filename), filename);
```
MrSmile, line 206: Style nit: space `(int) sizeof(filename)`.

fuzz/fuzz.c
```diff
@@ -186,6 +198,64 @@ static ASS_Track *read_track_from_stdin(void)
...
+static bool parse_cmdline(int argc, char *argv[], struct settings *settings)
+{
+    // defaults
+    settings->mode = CONSUME_INPUT;
+    settings->input = NULL;
+    settings->output = NULL;
+
+    int i;
+    for (i = 1; i < argc; i++) {
+        const char* param = argv[i];
```
MrSmile, line 216: Style nit: space `const char *param`.

fuzz/writeout.c
TheOneric: fixed here and at another place

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+    DIVMOD(60, m);
+    #undef DIVMOD
+
+    min_size += time >= 10 ? ceil(log10(time + 1)) : 1;
+    if (*buf_size < min_size) {
+        // +14 are arbitrary padding to avoid too many reallocs
+        char* nbuf = realloc(*buf, min_size + 14);
+        if (!nbuf) {
+            return false;
+        }
```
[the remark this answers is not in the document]
TheOneric: dropped

fuzz/fuzz.c
```diff
@@ -153,6 +153,18 @@ static void consume_track(ASS_Renderer *renderer, ASS_Track *track)
 }
 
 #if ASS_FUZZMODE == FUZZMODE_STANDALONE
+#include "writeout.h"
+
+struct settings {
+    enum {
+        CONSUME_INPUT = 0,
+        WRITEOUT_TRACK
```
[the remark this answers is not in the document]
TheOneric: done

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+        case YCBCR_FCC_TV:       return "TV.fcc";
+        case YCBCR_FCC_PC:       return "PC.fcc";
+        case YCBCR_DEFAULT:      return "ThereWasNoHeader";
+        case YCBCR_UNKNOWN:
+        default:
+            return "InvalidUnknownValue";
+    }
+}
+
+static bool time_to_str(long time, char** buf, size_t* buf_size)
```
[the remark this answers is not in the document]
TheOneric: fixed the style issue here and in other mentioned and a few additionall places

For fixed size buffer see the other comment about it

fuzz/writeout.c
TheOneric: fixed

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+    time = labs(time);
+
+    // Without chars for hours, except for sign indicator
+    size_t min_size = sign == -1 ? 14 : 10;
+    int cs, s, m;
+    #define DIVMOD(div,mod) mod = time % div; time /= div
+    DIVMOD(100, cs);
+    DIVMOD(60, s);
+    DIVMOD(60, m);
+    #undef DIVMOD
```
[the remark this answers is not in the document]
TheOneric: yeah, with the simplified form the macro doesn't help that much anymore and just writing it out might be clearer

changed it now

fuzz/writeout.c
TheOneric: For this we’d need to bump the absolute hour value and recalculate the other values to retain the same overall time if it’s negative. Imho this isn’t worth it for a minor cosmetic improvement in the output

fuzz/writeout.c
TheOneric: changed

fuzz/writeout.c
TheOneric: yeah, in theory a `switch` could also enable us to get compiler warnings when a new track type value (`ASS2` most likely) gets added, but since this is an inlined anonymous enum and `track_type` thus here only an `int` i don’t think this actually works out.
Alas, changed it to a `switch` and added an explicit distinction between `UNKNOWN` and an out of sync track type list

fuzz/writeout.c
TheOneric: changed

fuzz/fuzz.c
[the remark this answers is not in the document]
TheOneric: for reference, we do the same for all enums in `ass_types.h` and some in `ass.h`

fuzz/writeout.c
```diff
@@ -0,0 +1,216 @@
...
+
+    min_size += time >= 10 ? ceil(log10(time + 1)) : 1;
```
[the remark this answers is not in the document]
TheOneric: it turns out with a common 64-bit `long` this gives `+= 13` so in practice the later “`+14` arbitrary padding” already prevented more than two reallocs from ever needing to happen on such platforms.With the actual initial size of 128 used in `write_events` no reallocs occur in practice. There’s a (probably theoretical) catch due to differences in absolute value between `LONG_MAX` and `LONG_MIN`, but this is solvable

Can you explain more what you have in mind for assuming a fixed-size buffer?

Afaict for being able to pass in a guaranteed safe, fixed-size buffer and never needing to do any size checks or allocs here, I think we’d need to (a) just trust callers or (b) hardcode the ceiled log10 value *(with an assert on `LONG_MAX`; even C23 constexpr doesn't help due to needing a function call for the logarithm)* neither of which i particularly like

Still checking a passed-in size parameter and just failing instead of reallocing if the buffer is too small would work ig, but is this really desirable or a notable simplification?

fuzz/fuzz.c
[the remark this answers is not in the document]
MrSmile: It's API, so exact values matter (but I prefer to set all values for that, in case we decide to deprecate value in the middle).

fuzz/writeout.c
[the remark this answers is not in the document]
MrSmile: Actually, `long` type is incorrect here, should be `long long` or `int64_t`. And maximal value of `time` is `3661010 * (long long) INT32_MAX`, see [here](https://github.com/libass/libass/blob/master/libass/ass.c#L229-L240). So, we can calculate maximum length based on that value. Or assume maximum 64-bit or even 128-bit value, not really a big difference.

Fixed buffer without caller trust can be implemented with
```c
typedef struct {
     char value[64];
} TimeBuffer;
```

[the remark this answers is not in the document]
TheOneric: > Actually, `long` type is incorrect here, should be `long long`

yep, good catch

> Fixed buffer without caller trust can be implemented with

This still hardcodes the assumed to be safe value derived from `log10`


> And maximal value of time is 3661010 * (long long) INT32_MAX, see [here](https://github.com/libass/libass/blob/master/libass/ass.c#L229-L240).

Right, we changed this to explicit 32-bit components recent'ish. In that case, there’s no risk of the max value growing further on other platforms and  hardcoding miiight be fine.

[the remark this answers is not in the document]
TheOneric: Pushed a fix for `long long` and a separate commit for fixed-size buffers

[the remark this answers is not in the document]
astiob: > (b) hardcode the ceiled log10 value _(with an assert on `LONG_MAX`; even C23 constexpr doesn't help due to needing a function call for the logarithm)_

As little as C23 constexpr may currently matter, would this not work?

```c
constexpr size_t maxlen = (int) (CHOSENTYPENAME_WIDTH * 0.30102999566398119521373889472449 + 1);
```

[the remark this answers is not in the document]
TheOneric: yeah you’re right. With constexpr we could just hardcode the platform-independent `log10(2)` and use `log2(LLONG_MAX)  ≤ sizeof(long long) * CHAR_BITS)`

Issue #442, opened 2020-10-26
Stacked blur reveals patterns in low bits of output not present in vsfilter


rcombs:
```
[Script Info]
; Script generated by Aegisub 9004-rcombs-local-53affe134
; http://www.aegisub.org/
Title: Default Aegisub file
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
YCbCr Matrix: None
PlayResX: 1920
PlayResY: 1080

[Aegisub Project Garbage]
Video File: ?dummy:25.000000:50000:1920:1080:255:0:0:
Video AR Value: 1.777778
Video Zoom Percent: 1.500000

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Default,Slate Pro Medium,51,&H00FFFFFF,&H000000FF,&H00472A27,&HCC000000,0,0,0,0,100,100,0,0,1,3.17333,1.88889,2,200,200,43,1

[Events]
Format: Layer, Start, End, Style, Name, MarginL, MarginR, MarginV, Effect, Text
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\an5\bord0\shad0\c&H153BA1&\p1\blur40\pos(952,561.333)}m 0 0 l 0 300 300 300 300 0{\p0}
```
Repeat that event 70 times and you get this in vsfilter:
![image](https://user-images.githubusercontent.com/819638/97123614-aa50f700-16fa-11eb-991d-d8496659ade8.png)

But this in libass:
![image](https://user-images.githubusercontent.com/819638/97123622-b341c880-16fa-11eb-9447-682abbd81571.png)

This came up as a result of investigating this (less-absurd) script:
[BrokeBlur.ass.zip](https://github.com/libass/libass/files/5436267/BrokeBlur.ass.zip)

vsfilter:
![image](https://user-images.githubusercontent.com/819638/97123729-15023280-16fb-11eb-87cc-9aa082b75ea7.png)

libass:
![image](https://user-images.githubusercontent.com/819638/97123736-1d5a6d80-16fb-11eb-92a1-ec5771e460fd.png)

MrSmile:
[...]

PR #800, opened 2024-07-09
Use transparent background in test’s output images

TheOneric:
We discussed this a while ago in IRC wrt how creation of new regression tests can be made easier. With this patch a 320×180 reference image at 1000ms can be generated by running `~1/test sample-1000.png sample.ass 1000  320 180`.
Without this patch first a placeholder image with the desired dimension needs to be created, then `compare` run with output enabled.

(The latter used to be even more painful since only limited input formats were accepted and regenerating reference images needed manual ``compare` invocation. #767 and https://github.com/libass/libass-tests/pull/5 already made this easier)

However, there are caveats considering our utilities also serve as examples for API users and `test` is/was the most approachable one actually blending images:

1. unless i’m missing an obvious mistake in my simple blending code, the output noticeably differs from what e.g. Aegisub or mpv produce. This might mislead some into being subtly incompatible with other renderers
2. the blending procedure is more complicated and fixes inverted alpha late; which might even confuse those familiar with simpler blend methods wrt what’s correct

I imagine `compare`’s blending was intentionally chosen for best quality-comparison results since it’ſ not constrained by compatiblity concerns and we don’t want to adjust it. Alternatively we could leave `test` slightly incompatible with `compare` and rely only on the previous qol improvement mentioned above

<details>
<summary>For reference the simpler blending code i tested against</summary>

Extending the existing opaque variant, but moving the division by 255 out of `k` to avoid loss of precision by truncating intermediate values.
This appears to match mpv, and except for intermediate truncation Aegisub.
```C
static void blend_single(image_t * frame, ASS_Image *img)
{

    unsigned char r = img->color >> 24;
    unsigned char g = (img->color >> 16) & 0xFF;
    unsigned char b = (img->color >> 8) & 0xFF;
    unsigned char a = 255 - (img->color & 0xFF);

    unsigned char *src = img->bitmap;
    unsigned char *dst = frame->buffer + img->dst_y * frame->stride + img->dst_x * 4;

    for (int y = 0; y < img->h; ++y) {
        for (int x = 0; x < img->w; ++x) {
            unsigned k = ((unsigned) src[x]) * a;
            dst[x * 4 + 0] = (k *   r + (255 * 255 - k) * dst[x * 4 + 0]) / (255 * 255);
            dst[x * 4 + 1] = (k *   g + (255 * 255 - k) * dst[x * 4 + 1]) / (255 * 255);
            dst[x * 4 + 2] = (k *   b + (255 * 255 - k) * dst[x * 4 + 2]) / (255 * 255);
            dst[x * 4 + 3] = (k * 255 + (255 * 255 - k) * dst[x * 4 + 3]) / (255 * 255);
        }
        src += img->stride;
        dst += frame->stride;
    }
}
```
</details>

### DIFF
--- a/compare/image.c
+++ b/compare/image.c
@@ -194,6 +194,8 @@ static bool write_png(const char *path, uint32_t width, uint32_t height,
     }
 
     png_init_io(png, fp);
+    png_set_compression_level(png, 9);
+
     png_set_IHDR(png, info, width, height, depth,
                  PNG_COLOR_TYPE_RGBA, PNG_INTERLACE_NONE,
                  PNG_COMPRESSION_TYPE_DEFAULT, PNG_FILTER_TYPE_DEFAULT);
--- a/test/test.c
+++ b/test/test.c
@@ -80,7 +80,7 @@ static void write_png(char *fname, image_t *img)
     }
 
     png_init_io(png_ptr, fp);
-    png_set_compression_level(png_ptr, 0);
+    png_set_compression_level(png_ptr, 9);
 
     png_set_IHDR(png_ptr, info_ptr, img->width, img->height,
                  8, PNG_COLOR_TYPE_RGB, PNG_INTERLACE_NONE,
--- a/test/test.c
+++ b/test/test.c
@@ -43,27 +43,40 @@ void msg_callback(int level, const char *fmt, va_list va, void *data)
 
 static void write_png(char *fname, image_t *img)
 {
-    FILE *fp;
-    png_structp png_ptr;
-    png_infop info_ptr;
-    png_byte **row_pointers;
-    int k;
+    png_structp png_ptr = NULL;
+    png_infop info_ptr = NULL;
+    png_byte **volatile row_pointers = NULL;
 
-    png_ptr =
-        png_create_write_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
-    info_ptr = png_create_info_struct(png_ptr);
-    fp = NULL;
-
-    if (setjmp(png_jmpbuf(png_ptr))) {
-        png_destroy_write_struct(&png_ptr, &info_ptr);
-        fclose(fp);
-        return;
-    }
-
-    fp = fopen(fname, "wb");
+    FILE *fp = fopen(fname, "wb");
     if (fp == NULL) {
         printf("PNG Error opening %s for writing!\n", fname);
-        return;
+        goto fail;
+    }
+
+    png_ptr = png_create_write_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
+    if (!png_ptr) {
+        printf("PNG Error creating write struct!\n");
+        goto fail;
+    }
+
+    info_ptr = png_create_info_struct(png_ptr);
+    if (!info_ptr) {
+        printf("PNG Error creating info struct!\n");
+        goto fail;
+    }
+
+    row_pointers = malloc(img->height * sizeof(png_byte *));
+    if (!row_pointers) {
+        printf("PNG Failed to allocate row pointers!\n");
+        goto fail;
+    }
+    for (int k = 0; k < img->height; k++)
+        row_pointers[k] = img->buffer + img->stride * k;
+
+
+    if (setjmp(png_jmpbuf(png_ptr))) {
+        printf("PNG unknown error!\n");
+        goto fail;
     }
 
     png_init_io(png_ptr, fp);
@@ -77,17 +90,19 @@ static void write_png(char *fname, image_t *img)
 
     png_set_bgr(png_ptr);
 
-    row_pointers = malloc(img->height * sizeof(png_byte *));
-    for (k = 0; k < img->height; k++)
-        row_pointers[k] = img->buffer + img->stride * k;
-
     png_write_image(png_ptr, row_pointers);
     png_write_end(png_ptr, info_ptr);
-    png_destroy_write_struct(&png_ptr, &info_ptr);
 
+fail:
     free(row_pointers);
 
-    fclose(fp);
+    if (png_ptr && !info_ptr)
+        png_destroy_write_struct(&png_ptr, NULL);
+    else if (png_ptr && info_ptr)
+        png_destroy_write_struct(&png_ptr, &info_ptr);
+
+    if (fp)
+        fclose(fp);
 }
 
 static void init(int frame_w, int frame_h)
--- a/test/test.c
+++ b/test/test.c
@@ -26,7 +26,7 @@
 
 typedef struct image_s {
     int width, height, stride;
-    unsigned char *buffer;      // RGB24
+    unsigned char *buffer;      // RGBA32
 } image_t;
 
 ASS_Library *ass_library;
@@ -83,12 +83,10 @@ static void write_png(char *fname, image_t *img)
     png_set_compression_level(png_ptr, 9);
 
     png_set_IHDR(png_ptr, info_ptr, img->width, img->height,
-                 8, PNG_COLOR_TYPE_RGB, PNG_INTERLACE_NONE,
+                 8, PNG_COLOR_TYPE_RGBA, PNG_INTERLACE_NONE,
                  PNG_COMPRESSION_TYPE_DEFAULT, PNG_FILTER_TYPE_DEFAULT);
 
     png_write_info(png_ptr, info_ptr);
-
-    png_set_bgr(png_ptr);
 
     png_write_image(png_ptr, row_pointers);
     png_write_end(png_ptr, info_ptr);
@@ -133,39 +131,33 @@ static image_t *gen_image(int width, int height)
     image_t *img = malloc(sizeof(image_t));
     img->width = width;
     img->height = height;
-    img->stride = width * 3;
-    img->buffer = calloc(1, height * width * 3);
-    memset(img->buffer, 63, img->stride * img->height);
-    //for (int i = 0; i < height * width * 3; ++i)
-    // img->buffer[i] = (i/3/50) % 100;
+    img->stride = width * 4;
+    img->buffer = calloc(1, height * width * 4);
     return img;
 }
 
-#define _r(c)  ((c)>>24)
-#define _g(c)  (((c)>>16)&0xFF)
-#define _b(c)  (((c)>>8)&0xFF)
-#define _a(c)  ((c)&0xFF)
-
 static void blend_single(image_t * frame, ASS_Image *img)
 {
-    int x, y;
-    unsigned char opacity = 255 - _a(img->color);
-    unsigned char r = _r(img->color);
-    unsigned char g = _g(img->color);
-    unsigned char b = _b(img->color);
+    unsigned char r = img->color >> 24;
+    unsigned char g = (img->color >> 16) & 0xFF;
+    unsigned char b = (img->color >> 8) & 0xFF;
+    unsigned char a = 255 - (img->color & 0xFF);
 
-    unsigned char *src;
-    unsigned char *dst;
+    unsigned char *src = img->bitmap;
+    unsigned char *dst = frame->buffer + img->dst_y * frame->stride + img->dst_x * 4;
 
-    src = img->bitmap;
-    dst = frame->buffer + img->dst_y * frame->stride + img->dst_x * 3;
-    for (y = 0; y < img->h; ++y) {
-        for (x = 0; x < img->w; ++x) {
-            unsigned k = ((unsigned) src[x]) * opacity / 255;
-            // possible endianness problems
-            dst[x * 3] = (k * b + (255 - k) * dst[x * 3]) / 255;
-            dst[x * 3 + 1] = (k * g + (255 - k) * dst[x * 3 + 1]) / 255;
-            dst[x * 3 + 2] = (k * r + (255 - k) * dst[x * 3 + 2]) / 255;
+    for (int y = 0; y < img->h; ++y) {
+        for (int x = 0; x < img->w; ++x) {
+            unsigned k = ((unsigned) src[x]) * a;
+            // For high-quality output consider using dithering instead;
+            // this static offset results in biased rounding but is faster
+            unsigned rounding_offset = 255 * 255 / 2;
+            // If the original frame is not in premultiplied alpha, convert it beforehand or adjust
+            // the blending code. For fully-opaque output frames there's no difference either way.
+            dst[x * 4 + 0] = (k *   r + (255 * 255 - k) * dst[x * 4 + 0] + rounding_offset) / (255 * 255);
+            dst[x * 4 + 1] = (k *   g + (255 * 255 - k) * dst[x * 4 + 1] + rounding_offset) / (255 * 255);
+            dst[x * 4 + 2] = (k *   b + (255 * 255 - k) * dst[x * 4 + 2] + rounding_offset) / (255 * 255);
+            dst[x * 4 + 3] = (k * 255 + (255 * 255 - k) * dst[x * 4 + 3] + rounding_offset) / (255 * 255);
         }
         src += img->stride;
         dst += frame->stride;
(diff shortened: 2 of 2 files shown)


astiob:
> unless i’m missing an obvious mistake in my simple blending code, the output noticeably differs from what e.g. Aegisub or mpv produce

Uh. Could you describe this in more detail and show an example?

TheOneric:
Anything which is neither fully-transparent nor fully-opaque differs. E.g. using a freshly generated `compare/test/sub2-153000.png` *(the version committed in the repo intentionally differs from the correct output for illustration purposes)*

This version of the patch yields an image bit identical to `comapre`’s output image:
sub2-153000_test_cmpBlend

The simpler and commonly used blend method yields this *(pay attention to letter edges)*
sub2-153000_test_simpleBlend

And imagemagick’s `compare compare_output.png simble_blend_output.png -compose src diff.png` highlights this:
diff_simpleBlend

If the difference is hard to spot in GH preview, try comparing the results locally enlarged with linear upscaling, e.g.: `feh -Z -. --force-alias simple_blend.png cmp_blend.png`

astiob:
Neither piece of code that I see in the diff (old and new) seems correct. The old code rounds an intermediate fraction, introducing extra rounding error. The new code approximates a divisor that should be 8,388,225 by 8,388,608, producing a global brightness/tint shift.

astiob:
Here’s an example of correct blending: https://gist.github.com/astiob/7e0dc1fa94fb4c8cf633

> I imagine compare’s blending was intentionally chosen for best quality-comparison results

If I understand correctly which code you’re referring to, then it seems to be optimized for performance, not accuracy, given all the bitwise arithmetic that I see.

TheOneric:
> >  I imagine compare’s blending was intentionally chosen for best quality-comparison results
>
> If I understand correctly which code you’re referring to, then it seems to be optimized for performance, not accuracy, given all the bitwise arithmetic that I see.

cc @MrSmile 

> The old code rounds an intermediate fraction, introducing extra rounding error.

Yep, as mentioned above. If i recall correctly the reference image from my previous reply was generated without this rounding error, i.e.:

```C
static void blend_single(image_t * frame, ASS_Image *img)
{
    unsigned char r = img->color >> 24;
    unsigned char g = (img->color >> 16) & 0xFF;
    unsigned char b = (img->color >> 8) & 0xFF;
    unsigned char a = 255 - (img->color & 0xFF);

    unsigned char *src = img->bitmap;
    unsigned char *dst = frame->buffer + img->dst_y * frame->stride + img->dst_x * 4;

    for (int y = 0; y < img->h; ++y) {
        for (int x = 0; x < img->w; ++x) {
            unsigned k = ((unsigned) src[x]) * a;
            dst[x * 4 + 0] = (k *   r + (255 * 255 - k) * dst[x * 4 + 0]) / (255 * 255);
            dst[x * 4 + 1] = (k *   g + (255 * 255 - k) * dst[x * 4 + 1]) / (255 * 255);
            dst[x * 4 + 2] = (k *   b + (255 * 255 - k) * dst[x * 4 + 2]) / (255 * 255);
            dst[x * 4 + 3] = (k * 255 + (255 * 255 - k) * dst[x * 4 + 3]) / (255 * 255);
        }
        src += img->stride;
        dst += frame->stride;
    }
}
```

Random dithering will conflict with pixel-exact comparisons later, but i guess we could use `double` and unbiased rounding if relevant.

MrSmile:
> The new code approximates a divisor that should be 255×255×129 = 8,388,225 by 223 = 8,388,608, producing a global brightness/tint shift.

No, my approximation should be exact for the given range of input values. I designed it to be such.

> Random dithering will conflict with pixel-exact comparisons later, but i guess we could use `double` and unbiased rounding if relevant.

Dithering is psychovisual optimization, I think it's not appropriate for testing tools, better to use 16-bit output instead.

TheOneric:
> No, my approximation should be exact for the given range of input values. I designed it to be such.

Do you have any clues where the observed difference is coming from?

MrSmile:
From this variant?
```c
            unsigned k = ((unsigned) src[x]) * a;
            dst[x * 4 + 0] = (k *   r + (255 * 255 - k) * dst[x * 4 + 0]) / (255 * 255);
            dst[x * 4 + 1] = (k *   g + (255 * 255 - k) * dst[x * 4 + 1]) / (255 * 255);
            dst[x * 4 + 2] = (k *   b + (255 * 255 - k) * dst[x * 4 + 2]) / (255 * 255);
            dst[x * 4 + 3] = (k * 255 + (255 * 255 - k) * dst[x * 4 + 3]) / (255 * 255);
```
Looks like bad rounding. Try to add `255 * 255 / 2` before division, should be the same as my variant if I don't miss something.

Actually, maybe it's a good time to unify `compare` and `test` code? After this PR most of the work they do is the same.

TheOneric:
> Try to add 255 * 255 / 2 before division, should be the same as my variant if I don't miss something.

I’d expect this to only make a minor difference and in fact applying the below makes no differnce at all in this particular example. However, the difference to `compare`’s output is noticeable even by eye.

```diff
diff --git a/test/test.c b/test/test.c
index 90e0e8aa..91ab72e6 100644
--- a/test/test.c
+++ b/test/test.c
@@ -149,10 +149,11 @@ static void blend_single(image_t * frame, ASS_Image *img)
     for (int y = 0; y < img->h; ++y) {
         for (int x = 0; x < img->w; ++x) {
             unsigned k = ((unsigned) src[x]) * a;
-            dst[x * 4 + 0] = (k *   r + (255 * 255 - k) * dst[x * 4 + 0]) / (255 * 255);
-            dst[x * 4 + 1] = (k *   g + (255 * 255 - k) * dst[x * 4 + 1]) / (255 * 255);
-            dst[x * 4 + 2] = (k *   b + (255 * 255 - k) * dst[x * 4 + 2]) / (255 * 255);
-            dst[x * 4 + 3] = (k * 255 + (255 * 255 - k) * dst[x * 4 + 3]) / (255 * 255);
+            unsigned rounding_offset = 255 * 255 / 2;
+            dst[x * 4 + 0] = (k *   r + (255 * 255 - k) * dst[x * 4 + 0] + rounding_offset) / (255 * 255);
+            dst[x * 4 + 1] = (k *   g + (255 * 255 - k) * dst[x * 4 + 1] + rounding_offset) / (255 * 255);
+            dst[x * 4 + 2] = (k *   b + (255 * 255 - k) * dst[x * 4 + 2] + rounding_offset) / (255 * 255);
+            dst[x * 4 + 3] = (k * 255 + (255 * 255 - k) * dst[x * 4 + 3] + rounding_offset) / (255 * 255);
         }
         src += img->stride;
         dst += frame->stride;
```

hmm.. is it maybe related to alpha (non-)premultiplication?

astiob:
> my approximation should be exact for the given range of input values. I designed it to be such.

This isn’t at all obvious from the code, so I’d appreciate having a comment in the code saying as much.

> Dithering is psychovisual optimization, I think it's not appropriate for testing tools, better to use 16-bit output instead.

I agree that higher bit depths are always better, but FWIW I can’t agree with calling dithering a “psychovisual optimization”: regardless of whether the viewer is human, dithering is vitally necessary to preserve gradients and the overall energy, countering quantization error. See for example the banding artefacts in #442. The _lack_ of dithering is an optimization (a common one), and it comes at a correctness cost.

Also FWIW, the dithering in the code I posted earlier would indeed be especially problematic for testing tools because it’s randomized, but dithering doesn’t _have_ to be randomized. That said, for our testing purposes, I don’t think it matters whether dithering is applied at all.

MrSmile:
> hmm.. is it maybe related to alpha (non-)premultiplication?

Looks like it's wrong formula in the alpha part, should be
```c
dst[x * 4 + 3] = (k * 0 + (255 * 255 - k) * dst[x * 4 + 3] + rounding_offset) / (255 * 255);
```
Inverse alpha is transmission coefficient, overlaying image should never add anything to it.

> This isn’t at all obvious from the code, so I’d appreciate having a comment in the code saying as much.

Looks like I was wrong and exact approximation was for something else (as I usually add comments for that indeed). 23 is simply maximum possible shift to not overflow `int32_t` intermediate values. Anyway, compared to `double` calculation, maximum error is around `0.51`, so it's fairly good approximation. Also, it's locally optimal in the sense as it's not improve until the shift of 31. BTW, exact approximation in this case would be the shift of 38 (with `int64_t` intermediates, obviously) and compilers automatically generate an analog with shift 47. But `compare` can be slow, especially with `-s` switch, so I'm not sure about using `int64_t` for the 0.01 pixel value improvement.

> I can’t agree with calling dithering a “psychovisual optimization”

It's a trade-off between low and high frequency noise. Humans perceive low frequencies better, so it's valid to reduce error there at the cost of increasing high-frequency noise.

TheOneric:
> Looks like it's wrong formula in the alpha part, should be
> 
> ```
> dst[x * 4 + 3] = (k * 0 + (255 * 255 - k) * dst[x * 4 + 3] + rounding_offset) / (255 * 255);
> ```
> 
> Inverse alpha is transmission coefficient, overlaying image should never add anything to it.

This doesn’t seem right. Consider some fully transparent (`k=255*255`) gets overlayed. With the suggestion above this will result in:
```
alpha_output = k * 0 + (255 * 255 - k) * alpha_bg_in + rounding_offset) / (255 * 255)
             = 0 * k + 0 * alpha_bg_in + rounding_offset) / (255 * 255)
             = rounding_offset / (255 * 255)
             = (int) 0
```

For an initially empty, fully transparent background like we have here, even semi-transparent overlay pixels end up fully transparent and nothing can ever be displayed at all.

fwiw, mpv [agrees with the current formula](https://github.com/mpv-player/mpv/blob/d384a6b79303ef21b7330555e968a99d59a7c5b0/sub/ass_mp.c#L190C71-L191C1) here.


> >  I can’t agree with calling dithering a “psychovisual optimization”
>
> It's a trade-off between low and high frequency noise. Humans perceive low frequencies better, so it's valid to reduce error there at the cost of increasing high-frequency noise.

Everyone seems in agreement we don’t need to dither here anyway, but as a side note:  
It’s not just about visual perception; dithering can preserve some attributes better. Consider e.g. a simple case where after floating-point blending all pixels end up with a value of exactly `10.4` in all colour channels. Any rounding scheme will result in a uniformly-coloured `10` everywhere, reducing the average luminosity a bit. With random dithering the resulting average average colour is a normal distribution around `10.4` and the narrow the more pixels there are.  
But there are indeed perf and high-frequency noise tradeoffs; what’s more important will depend on the usecase.

TheOneric:
> 

Changing the formula for colour channels to
```C
unsigned alpha = (k * 255 + (255 * 255 - k) * dst[x * 4 + 3] + rounding_offset) / (255 * 255);
            unsigned alpha_correction = alpha ? alpha : 255;
            dst[x * 4 + 0] = (k *   r + (255 * 255 - k) * dst[x * 4 + 0] + rounding_offset) / (255 * 255) * 255 / alp
ha_correction;
```
yields `compare`-identical output for this sample *(in general rounding errors may differ)*. However using
```C
            unsigned alpha = (k * 255 + (255 * 255 - k) * dst[x * 4 + 3] + rounding_offset) / (255 * 255);
            unsigned alpha_correction = alpha ? alpha : 255;
            dst[x * 4 + 0] = (k *   r + (255 * 255 - k) * dst[x * 4 + 0] + rounding_offset) / (255 * alpha_correction
);
```
which should theoretically be identical and in practice slightly less error-prone iinm again is noticeably a tad different.

It seems `compare` applies this premultiplication at the very end together with alpha inversion however and this might lead to diverging results for overlapping events [`compare/image.c@write_png8`](https://github.com/libass/libass/blob/c131a95c443f894bd95a05dfa3d0a8dc67a22d5e/compare/image.c#L213):
```C
bool write_png8(const char *path, Image8 *img)
{
    uint8_t *ptr = img->buffer;
    size_t size = (size_t) img->width * img->height;
    for (size_t i = 0; i < size; i++) {
        uint8_t alpha = ~ptr[3];
        if (alpha) {
            const uint32_t offs = (uint32_t) 1 << 15;
            uint32_t inv = ((uint32_t) 255 << 16) / alpha + 1;
            // equivalent to (255 * ptr[k] + alpha / 2) / alpha
            ptr[0] = (ptr[0] * inv + offs) >> 16;
            ptr[1] = (ptr[1] * inv + offs) >> 16;
            ptr[2] = (ptr[2] * inv + offs) >> 16;
        }
        ptr[3] = alpha;
        ptr += 4;
    }
    return write_png(path, img->width, img->height,
                     4 * img->width, img->buffer, 8);
}
```

TheOneric:
> Actually, maybe it's a good time to unify compare and test code? After this PR most of the work they do is the same.

Sharing some code or even merging the utilities into one qould certainly help avoiding slightly different rounding artefacts etc.

But we also have to keep in mind, our utilities serve a double-duty as examples for how to use libass for prospective users. I think it’s a good idea to retain a *simple* example of how to blend `ASS_Image`s; *“simple”* being key here. `fuzz` and `profile` don’t actually blend rendering results and current `compare` is too complex especially with all the bit blending arithmetic and late alpha inversion for optimisation purposes.   
The trouble we have here figuring out where the difference comes from and what we should ideally do imho reinforces this

Maybe if `compare` blending is simplified and reorganised so one contiguous section has all relevant blending bits?

MrSmile:
> It seems `compare` applies this premultiplication at the very end together with alpha inversion however and this might lead to diverging results for overlapping events

All image filtering should be done in premultiplied alpha (inverted premultiplied alpha is even better, gamma-corrected inverted premultiplied alpha is ideal, but sadly never supported anywhere). Use of non-premultiplied alpha is almost always wrong. I think it would be better to separate actual physical blending from adaptations for badly designed image format, and compare is doing exactly that.

> But we also have to keep in mind, our utilities serve a double-duty as examples for how to use libass for prospective users.

Maybe it would be better to write dedicated docs for that?

TheOneric:
> All image filtering should be done in premultiplied alpha (inverted alpha is even better, gamma-corrected inverted alpha is ideal, but sadly never supported anywhere). Use of non-premultiplied alpha is almost always wrong.

I’m not sure which alpha inversion you’re referring to *(i.e.: keeping ASS alpha which is inherently inverted compared to prevalent convention, or inverting ASS alpha early to match prevalent convention)*  or why either choice makes a conceptual difference. The direction of alpha seems to me like an arbitrary choice with either option being equivalent apart from (mis)matching prevalent convention.  
But otherwise iiuc i agree.

Updated the PR and added some comments; it now matches compare for this example and i used the same rounding method as compare for premult-to-straight conversion, but iirc there might still be minor discrepancies due to `compare`’s divisor approximation, right?

TheOneric:
> But compare can be slow, especially with -s switch, so I'm not sure about using int64_t for the 0.01 pixel value improvement.

Did you run tests which showed blending as the bottleneck for `comapre -s`? Intuitively I’d expect rasterizing and `ASS_Image` rendering to take significantly more time than blending at large resolutions, especially since most our test samples only have a few small `ASS_Image`s anyway
If it turns out to not make much of a difference, imho it’d be best to use simpler and more precise blending logic in `compare` and then merging `test` and `compare`

MrSmile:
> keeping ASS alpha which is inherently inverted compared to prevalent convention

I mean that. ASS alpha is physically natural, compared to prevalent convention.

> why either choice makes a conceptual difference

Blending equations are simplest and most symmetric when written in natural variables. Also when considering gamma correction, it's ASS alpha that should be gamma-compressed, not its inverted value (it's where all GPU vendors and standard committees failed, sadly without a good workaround), but it's not relevant for libass, as it's gamma-incorrect.

> there might still be minor discrepancies due to `compare`’s divisor approximation, right?

Yes, in about 0.1% cases result is ±1 of exact division.

> Did you run tests which showed blending as the bottleneck for `comapre -s`?

Yes. Exact division is about 60% slower than my approximation, and for some of my tests it's 24% → 43% of the full `compare` time.

TheOneric:
Given this already contains some improvements and changes imo good to have in git history, i intend to merge this in a couple days if there are no objections eventhough it’s not bit-identical to compare.

To achieve bit-identical results we’ll either need to
- let `compare` and `test` share blending routines, BUT
   - this is akward with the current dir structure; merging `test` and `compare` into a common directory would make this better but when i proposed this before during work on non-recursive make there was pushback to changing locations
   - we still loose an easily understandable example for API users
      - adding a dedicated example for documentation purposes was suggested, but to make sure doc example code actually works we’d prob like to build it which basically amounts to recreating `test` as it is now — it’s already the simplest possible libass user by rendering one frame
- merge `compare` and `test`
  - same considerations as above apply
  - and also i’ve seen `--enable-test` used in some packaging scripts (optional in Gentoo)
  
It might be simpler to just add another mode to `compare` and not bother with bit-exact matches in `test`

Issue #747, opened 2024-04-08
API to discard older events from memory


astiob:
When playing a Matroska file, the ASS track is usually streamed in an incremental fashion: new events are added via `ass_process_chunk` when the current playback timestamp reaches them. This avoids having to read the whole file at once. Playback proceeds continuously with a monotonically increasing timestamp.

Currently, libass keeps all events that are added in this manner in memory until `ass_flush_events` is called to discard all events at once or `ass_free_track` is called to discard the entire track. This causes the memory usage to rise steadily as playback continues. With reasonably small subtitle tracks, this is unnoticeable, but DataHoarder’s monstrous bitmap (1.1 TiB ASS file!) and vector-art (hundreds of MiB per file) conversions are an easy example of where this matters a lot. It’s not unfathomable that this might matter in slightly more realistic cases, too.

However, ASS events whose ending timestamp has passed are actually no longer necessary to render the subtitles, so they could be discarded, freeing up memory and speeding up event lookup. This probably needs a new API by which the media player could negotiate the cutoff timestamp with libass.

When seeking (jumping) to an earlier timestamp, this will necessitate reloading the earlier subtitles, but this already happens in general, as the seek may target a timestamp that’s earlier than any timestamp the user has already played. I’m not sure whether players implement optimizations to avoid reloading subtitles when it’s known that they’re already cached, but I suspect not.

CastagnaIT:
On Kodi we use different ways to meet different needs, since we use libass especially to display others subtitles formats
main cases:

1) For others subtitles formats, we convert in full the original file to ass format,
then we use `ass_alloc_event` to add converted sub data as events,
this is done all-in-one (usually) at the beginning of playback.
In this case, we need to disable the new "auto-discarding of old events", since data are converted and added as events only once.

2) For video streaming (especially live streams) or however demuxed data, where the demuxer provide packets, we always use `ass_alloc_event` for converted subs (except for ass data format we use `ass_process_chunk`) in this case usually the "video seek" execute `ass_flush_events` to discard old events, and new fresh data is provider by the demuxer after the seek.
But on live streaming the "seek" is not always available, and currently we lack a way to delete old events,
"auto-discarding of old events" can help to free memory.

so auto-discarding old events in general would be fine

we need attention regarding point 1, that must avoid to use the "auto-discarding", otherwise it becomes a breaking change that forces us to overhaul the entire subtitles management on Kodi that it is too large a job,
[...]

PR #804, opened 2024-07-11
Add event pruning system and API.

cubicibo:
An attempt to address #747, and to manage bloated scripts embedded in mkvs, like DataHoarder's. 

Motivations supporting the proposed solution:
- The default behaviour matches the current one: no collection is ever performed. It has to be explicitly enabled.
- The user is free to prune events at will, or rely on the automatic process after every render.
- Can be configured per track.

Point open for discussion:
- The prune function is not thread safe [to changes made to the events array].

### DIFF
--- a/libass/ass.c
+++ b/libass/ass.c
@@ -190,12 +190,27 @@ static int test_and_set_read_order_bit(ASS_Track *track, int id)
 {
     if (resize_read_order_bitmap(track, id) < 0)
         return -1;
-    int index = id / 32;
-    uint32_t bit = 1u << (id % 32);
+    int index = id >> 5;
+    uint32_t bit = 1u << (id & 0x1F);
     if (track->parser_priv->read_order_bitmap[index] & bit)
         return 1;
     track->parser_priv->read_order_bitmap[index] |= bit;
     return 0;
+}
+
+static inline void clear_read_order_bit(ASS_Track *track, int id)
+{
+    int index = id >> 5;
+    if (index < track->parser_priv->read_order_elems) {
+        uint32_t mask = ~(1u << (id & 0x1F));
+        track->parser_priv->read_order_bitmap[index] &= mask;
+    }
+}
+
+static inline void update_prune_ts(ASS_Track *track, const long long ts)
+{
+    if (ts < track->parser_priv->prune_next_ts)
+        track->parser_priv->prune_next_ts = ts;
 }
 
 // ==============================================================================================
@@ -1015,8 +1030,10 @@ static int process_events_line(ASS_Track *track, char *str)
         event = track->events + eid;
 
         int ret = process_event_tail(track, event, str, 0);
-        if (!ret)
+        if (!ret) {
+            update_prune_ts(track, event->Start + event->Duration);
             return 0;
+        }
         // If something went wrong, discard the useless Event
         ass_free_event(track, eid);
         track->n_events--;
@@ -1330,7 +1347,7 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
 
         event->Start = timecode;
         event->Duration = duration;
-
+        update_prune_ts(track, event->Start + event->Duration);
         goto cleanup;
 //              dump_events(tid);
     } while (0);
@@ -1357,6 +1374,46 @@ void ass_flush_events(ASS_Track *track)
     free(track->parser_priv->read_order_bitmap);
     track->parser_priv->read_order_bitmap = NULL;
     track->parser_priv->read_order_elems = 0;
+}
+
+void ass_configure_prune(ASS_Track *track, long long delay)
+{
+    track->parser_priv->prune_delay = delay;
+}
+
+void ass_prune_events(ASS_Track *track, long long deadline)
+{
+    if (deadline < track->parser_priv->prune_next_ts)
+        return;
+
+    const bool check_readorder = track->parser_priv->check_readorder;
+    const int old_n_events = track->n_events;
+
+    int n_kept = 0;
+    ASS_Event *events = track->events;
+
+    track->parser_priv->prune_next_ts = LLONG_MAX;
+    for (int k = 0; k < old_n_events;) {
+        // discardable sequence
+        for (; k < old_n_events && events[k].Start + events[k].Duration < deadline; k++) {
+            if (check_readorder)
+                clear_read_order_bit(track, events[k].ReadOrder);
+            ass_free_event(track, k);
+        }
+
+        // to-be-kept sequence
+        int move_from = k;
+        for (long long ts; k < old_n_events && (ts = events[k].Start + events[k].Duration) >= deadline; k++)
+            update_prune_ts(track, ts);
+
+        // Relocate kept events
+        if (move_from < k) {
+            int cnt = k - move_from;
+            memmove(events + n_kept, events + move_from, cnt * sizeof(*track->events));
+            n_kept += cnt;
+        }
+    }
+    track->n_events = n_kept;
 }
 
 #ifdef CONFIG_ICONV
@@ -1725,6 +1782,8 @@ ASS_Track *ass_new_track(ASS_Library *library)
     if (!track->styles[def_sid].Name || !track->styles[def_sid].FontName)
         goto fail;
     track->parser_priv->check_readorder = 1;
+    track->parser_priv->prune_delay = -1;
+    track->parser_priv->prune_next_ts = LLONG_MAX;
     return track;
 
 fail:
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -24,7 +24,7 @@
 #include <stdarg.h>
 #include "ass_types.h"
 
-#define LIBASS_VERSION 0x01703000
+#define LIBASS_VERSION 0x01703010
 
 #ifdef __cplusplus
 extern "C" {
@@ -733,6 +733,24 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
  * If this function is not called, the default value is 1.
  */
 void ass_set_check_readorder(ASS_Track *track, int check_readorder);
+
+/**
+ * \brief Prune events that preceed deadline.
+ * \param track track
+ * \param deadline cut-off timestamp in milliseconds.
+*/
+void ass_prune_events(ASS_Track *track, long long deadline);
+
+/**
+ * \brief Configure automatic pruning of events.
+ * \param track track
+ * \param delay delay from "now" (ass_render_frame) in milliseconds.
+ * After every render, events whose undisplay timestamp predate
+ * "now - delay" will be deleted. A delay of 0 prunes aggressively.
+ * Negative delays disable automatic pruning.
+ * Disabled by default (no removal of events from memory).
+ */
+void ass_configure_prune(ASS_Track *track, long long delay);
 
 /**
  * \brief Flush buffered events.
--- a/libass/ass_priv.h
+++ b/libass/ass_priv.h
@@ -65,6 +65,9 @@ struct parser_priv {
     uint32_t header_flags;
 
     uint32_t feature_flags;
+
+    long long prune_delay;
+    long long prune_next_ts;
 };
 
 #endif /* LIBASS_PRIV_H */
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -3428,6 +3428,9 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
     ass_frame_unref(priv->prev_images_root);
     priv->prev_images_root = NULL;
 
+    if (track->parser_priv->prune_delay >= 0)
+        ass_prune_events(track, now - track->parser_priv->prune_delay);
+
     return priv->images_root;
 }
 
--- a/libass/Makefile_library.am
+++ b/libass/Makefile_library.am
@@ -1,6 +1,6 @@
-LIBASS_LT_CURRENT = 12
-LIBASS_LT_REVISION = 1
-LIBASS_LT_AGE = 3
+LIBASS_LT_CURRENT = 13
+LIBASS_LT_REVISION = 0
+LIBASS_LT_AGE = 4
 
 .asm.lo:
 	$(nasm_verbose)$(LIBTOOL) $(AM_V_lt) --tag=CC --mode=compile $(top_srcdir)/ltnasm.sh $(AS) $(ASFLAGS) -I$(top_srcdir)/libass/ -Dprivate_prefix=ass -o $@ $<


libass/ass.c
```diff
@@ -186,6 +191,32 @@ static int resize_read_order_bitmap(ASS_Track *track, int max_id)
...
+        //undisplay time descending order
+        qsort((void*)track->events, track->n_events, sizeof(ASS_Event), compare_event_deadline);
+        ASS_Event *ev = track->events + track->n_events - 1;
+        while (track->n_events > 0 && ev->Start + ev->Duration < deadline) {
+            ass_free_event(track, --track->n_events);
+            --ev;
+        }
+    }
+    return n_events_ini-track->n_events;
+}
```
TheOneric, line 218: There are several issues with this sorting:

 - `qsort` isn’t (guaranteed to be) a stable sort, yet rendering depends on the order of events, introducing random display differences
 - even a stable sort by end time might mix this order up in a detectable way if two events start simultaneously but the originally later-defined event end earlier
 - *(sorting by start time should be side-effect-free for a continuous watch, but this doesn't help much with this cleanup and i’m not fully sure if there might be visible differences when skipping around or if we should even care about the latter)*

libass/ass.h
```diff
@@ -734,6 +734,27 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
...
 
+/**
+ * \brief Prune events that preceed deadline if the cache exceeds size_threshold
+ * \param track track
+ * \param deadline cut-off timestamp
+ * \param size_threshold Minimum KiB size to trigger the garbage collection.
+ * Negative values skip the logic.
+ * \return number of events pruned.
+*/
+int ass_prune_history(ASS_Track *track, long long deadline, int size_threshold);
```
TheOneric, line 745: i think `size_t` and `0` for skipping would be more correct type-wise

In general, does tracking by accumulated event text size bring a real benefit over just number of events?

libass/ass.h
```diff
@@ -734,6 +734,27 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
...
+/**
+ * \brief Set the minimum size to trigger automatic garbage collection of events.
+ * \param track track
+ * \param size KiB threshold to trigger garbage collection.
+ * -1 is disable. 0 eagerly attempts collection after every render.
+ * libass may use more memory to store events than the threshold, if required.
+ * The default value is -1 (no garbage collection, keep all parsed events).
+ * Any other negative value is undefined.
+ */
+void ass_set_gc_threshold(ASS_Track *track, int size);
```
TheOneric, line 756: similarly here `size_t`

libass/ass_priv.h
```diff
@@ -65,6 +65,9 @@ struct parser_priv {
     uint32_t header_flags;
 
     uint32_t feature_flags;
+
+    int gc_threshold;
+    int64_t events_size;
```
TheOneric, line 70: And here also `size_t`

libass/ass.c
```diff
@@ -485,6 +516,7 @@ static int process_event_tail(ASS_Track *track, ASS_Event *event,
                 while (end > event->Text &&
                        (end[-1] == '\r' || end[-1] == '\t' || end[-1] == ' '))
                     *--end = 0;
+                track->parser_priv->events_size += (int64_t)(end - event->Text);
```
TheOneric, line 519: style nit: `(int54_t) (...)`

libass/ass.c
```diff
@@ -90,6 +84,14 @@ void ass_free_track(ASS_Track *track)
...
     }
     free(track->events);
+
+    if (track->parser_priv) {
+        free(track->parser_priv->read_order_bitmap);
+        free(track->parser_priv->fontname);
+        free(track->parser_priv->fontdata);
+        free(track->parser_priv);
+    }
+
```
TheOneric, line 94: Could you explain what the double-free issue was and why reordering fixed it?

libass/ass_render.c
```diff
@@ -3428,6 +3428,8 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
     ass_frame_unref(priv->prev_images_root);
     priv->prev_images_root = NULL;
 
+    ass_prune_history(track, now, track->parser_priv->gc_threshold);
```
TheOneric, line 3431: Directly discarding anything prior to “now” breaks even simple, short rewinds which might be too harsh for most. Perhaps auto-gc should also take a time as an offset from “now”

libass/ass.h
cubicibo: A bitmap/vector event would consume a lot more memory than innocent dialogue. A memory surge could happen if there's a spike of such events within a narrow time window, or the distribution of events type changes vastly as the stream goes. I would say it is more robust and flexible.

libass/ass.c
cubicibo: I could sort considering both the event id and undisplay timestamp... or just go for thepiecewise memmove or memcpy to ensure the event order is strictly preserved.

libass/ass.c
cubicibo: `ass_free_track` does not mirror the allocations done in `ass_new_track`. The first element allocated in `ass_new_track` should be the last free'd in `ass_free_track`.

The code in the first commit assume this mirrored symmetry, and ass_free_event access `track->parser_priv->events_size`, which would not exist due to the `free(track->parser_priv);` happening too early.

cubicibo:
I presume the bit flag in `track->parser_priv->read_order_bitmap` should also be cleared?

libass/ass.h
```diff
@@ -734,6 +734,27 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
...
+ * \param deadline cut-off timestamp.
+ * \return number of bytes dropped.
+*/
+size_t ass_prune_events(ASS_Track *track, size_t threshold, long long deadline);
+
+/**
+ * \brief Configure the automatic garbage collection of events.
+ * \param track track
+ * \param threshold KiB threshold to trigger garbage collection. disabled if 0.
+ * \param lag delay from "now" within which events are kept for back-seeking.
```
CastagnaIT, line 750: lag parameter, what "delay" value format should be used? its not specified

TheOneric:
Sorry this takes a while; since this patch adds public API we’ll need to support whatever shape it ends up being merged as will need to be supported for a long time, possibly forever.  I’m still undecided on what the best path forward is and am hoping to collect more comments.

I’m still not sure whether using an imprecise space measure **or** just count of events as one of the triggers makes more sense. On the one hand, tracking consumed space allows freeing up memory faster if there was a short spike of one or a few very large events, on the other hand it’s imprecise anyway yet more costly, peak memory consumption still needs to be accommodated anyway and it may end up being basically always eligible if just all events within a section are large.
Event count + delay is simpler, but may end up with more collection cycles for section with many small events even if there’s no memory pressure.

And if we go with a memory metric, while indeed it seems unlikely anyone actually wants to configure a limit down to a single byte of memory consumed *(which wouldn’t work anyway due to imprecisions in tracked consumption)*, using KiB in the API might be surprising as it doesn’t match existing chance limits and needs overflow checking when converting to bytes later.

I’ve been thinking whether it makes sense to keep a binary tree, e.g. a red-black tree, around of event IDs sorted by their end time to speed up garbage collection. This might also lessen concerns about spurious gc cycles for either eligibility metric. BUT it adds complexity and memory costs of its own. Not sure. Probably should be profiled first.

cubicibo:
>I’m still not sure whether using an imprecise space measure or just count of events as one of the triggers makes more sense

That choice could also be left to the user, and so would the trigger (either or both conditions). Probably over-engineered, but the implementation would be a detail regardless.

----

> Maybe for first version of the new API option, just make it a boolean option that sets whether GC is allowed or not. You can always expand it later with additional parameters to fine-tune its behavior. Leaving the initial behavior up to libass gives you the freedom to experiment a bit to find best GC decision making.

A single entry point, with the config passed in as a struct? Fair enough.

>Since the big memory consumptions mainly come from bitmaps/drawings, maybe restrict (size) counting to such events? Count big events and their approximate size. And just set regular events to 0 size. Do GC once a certain size threshold is passed and then discard old events.

Easier to treat all events identically at the parser level of abstraction. Your suggestion would fail with bloated text dialogue events.

libass/ass.c
```diff
@@ -186,6 +189,25 @@ static int resize_read_order_bitmap(ASS_Track *track, int max_id)
     return -1;
 }
 
+#define TS_OUT(ev) (ev.Start + ev.Duration)
+
+// This function is not thread safe
+size_t ass_prune_events(ASS_Track *track, size_t threshold, long long deadline)
+{
+    size_t n_size_ini = track->parser_priv->gc_events_size;
+    if (threshold << 10 < track->parser_priv->gc_events_size) {
```
MrSmile: Events should be removed uniformly every frame and not periodically in large batches. I've already eliminated similar behavior in cache system and don't want to deal with another. _libass_ is realtime application so latency is more important than throughput.

cubicibo: Mhh, I see a few ways to solve this:
- Drop the size threshold: the events would be collected solely based on `deadline`. The user would be responsible to define `lag` appropriately (or `deadline`, if they call `ass_prune_events` directly) to steer the memory footprint.
- Keep the memory footprint threshold, but stop the collection process as soon as $\sum_n SIZE(event_n) < k*gc\textunderscore{}threshold$ is true. $k \in [0; 1.0]$ would be either user-defined, or hardcoded.
- The collection may be aborted whenever $N$ events ($N$ user-defined or hardcoded) have been collected in a pass. This will not protect libass from attacks where more events are added than removed in a frame.

astiob:
> I’m still not sure whether using an imprecise space measure **or** just count of events as one of the triggers makes more sense.

IMO, as a user, neither the size nor the count matters; only the supported back-seekable time (is this what `lag` is?) does. I would expect libass to eagerly prune all events whose end timestamp has passed `now - lag`, because I won’t be needing them, and I certainly won’t be needing an arbitrary subset of them. Players that configure libass to clear its cache on every seek would set `lag` to exactly 0; others could tune this value to optimize small seeks.

> I’ve been thinking whether it makes sense to keep a binary tree, e.g. a red-black tree, around of event IDs sorted by their end time to speed up garbage collection.

To identify expired events efficiently, an end-timestamp-ordered binary heap would be enough. Binary heaps are pretty trivial, and red-black trees are unnecessary.

The bigger problem may be the primary list of events in the `ASS_Track`, which is an array and has to stay an array due to being publicly exposed as part of the API and ABI.

TheOneric:
> the supported back-seekable time (is this what lag is?)

yes

> IMO, as a user, neither the size nor the count matters; only the supported back-seekable time

True, unless a users spends extra effort on tracking each events memory size a size-based limit isn’t really meaningful for them and if they spend this effort also invoking prune calls manually is barely any further complexity.  
And while users themselves might cache the last N video packets etc, this doesn't work out for ASS anyway since each timestamp might have an arbitrary amount of packets, so time indeed seems like the only meaningful metric.

libass/ass.c
```diff
@@ -1359,6 +1364,34 @@ void ass_flush_events(ASS_Track *track)
...
 
+void ass_configure_gc(ASS_Track *track, int enabled, long long delay)
+{
+    track->parser_priv->gc_enabled = enabled;
+
+    if (0ll <= delay)
+        track->parser_priv->gc_delay = delay;
+}
+
+#define TS_OUT(ev) (ev.Start + ev.Duration)
```
TheOneric, line 1375: The macro is only used once and can be inlined

libass/ass.c
```diff
@@ -1359,6 +1364,34 @@ void ass_flush_events(ASS_Track *track)
...
+
+    for (int k = 0, eid = 0; eid < track->n_events; eid = ++k) {
+        while (k < track->n_events && TS_OUT(track->events[k]) < deadline) {
+            if (check_readorder)
+                test_and_write_read_order_bit(track, track->events[k].ReadOrder, false);
+            ass_free_event(track, k++);
+        }
+        if (k > eid) {
+            memmove(&track->events[eid], &track->events[k], sizeof(ASS_Event)*(track->n_events - k));
+            track->n_events -= (k - eid);
```
TheOneric, line 1390: When $N$ non-consecutive events get freed this will move data after the last freed event $N$-times. This can be optimised to only move each remaining chunk once by only moving data after encountering the _next_ chunk of to-be-freed events or the end of the list

libass/ass.h
```diff
@@ -734,6 +734,24 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
...
+/**
+ * \brief Configure the automatic garbage collection of events.
+ * \param track track
+ * \param enabled enable automatic garbage collection with the given delay.
+ * \param delay delay from "now" (ass_render_frame) in milliseconds.
+ * After every succesful render, events whose undisplay timestamp predate
+ * "now - delay" will be deleted. Delay of 0 performs aggressive collection.
+ * Disabled by default (no collection, keep all parsed events).
+ */
+void ass_configure_gc(ASS_Track *track, int enabled, long long delay);
```
TheOneric, line 753: Not sure but maybe `backbuffer`is more intuitive than `delay`?

TheOneric, line 753: There’s currently no way to disable automatic gc again without recreating the track. negative delay/backbuffer values can be used and documented for this

libass/ass.c
```diff
@@ -186,15 +188,18 @@ static int resize_read_order_bitmap(ASS_Track *track, int max_id)
...
+    int index = id >> 5; //div32
+    uint32_t bit = 1u << (id & 0x1F); //mod32
+    if (set && track->parser_priv->read_order_bitmap[index] & bit)
         return 1;
-    track->parser_priv->read_order_bitmap[index] |= bit;
+    if (set)
+        track->parser_priv->read_order_bitmap[index] |= bit;
+    else
+        track->parser_priv->read_order_bitmap[index] &= ~bit;
     return 0;
```
TheOneric, line 203: Optimisation of the division and modulo suggests you view this as perf-critical, but `set` introduces new conditions and branching options. `set` is always passed as a constant though, so if perf is relevant we could split out computaton of `index` and `bit` as a `static inline` function and have two different functions for test-and-set and clearing of a bit.

TheOneric (commented):
Alright since there were no more comments arguing for more complex criteria, i think we can move ahead with a time duration as the only parameter for auto gc

libass/ass.h
```diff
@@ -734,6 +734,24 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
  */
 void ass_set_check_readorder(ASS_Track *track, int check_readorder);
 
+/**
+ * \brief Garbage collect events that preceed deadline
+ * \param track track
+ * \param deadline cut-off timestamp in milliseconds.
+*/
+void ass_prune_events(ASS_Track *track, long long deadline);
```
TheOneric, line 742: SInce this adds new interfaces the `LIBASS_VERSION` macro in `ass.h` needs to be incremented on the second-least-significant nibble (the least-significant one is reserved for hotfixes) and the libtool version in `libass/Makefile_library.am` also needs to be bumped by zeroing `revision` and bumping both `current` and `age`

libass/ass.c
```diff
@@ -71,12 +71,6 @@ void ass_free_track(ASS_Track *track)
     if (!track)
         return;
 
-    if (track->parser_priv) {
-        free(track->parser_priv->read_order_bitmap);
-        free(track->parser_priv->fontname);
-        free(track->parser_priv->fontdata);
-        free(track->parser_priv);
-    }
```
TheOneric, line 79: Since `events_size` no longer exists there’s no more reason to move this code, right?

libass/ass.h
cubicibo: Isn't `gc_enabled`, set by `int enabled` of `ass_configure_gc` letting the user change the behaviour on the fly, or am I missing something?

libass/ass.c
cubicibo: I changed those to bitwise operation solely based on the context of the function: working on bits. Division and module are better candidates for other contexts. No performance considerations were made otherwise. I will split the set and clear functionalities in two separate functions, as it reduces the complexity anyway.

libass/ass.h
TheOneric: nope you’re not; *I* was missing/forgetting `enabled`

Most of our other API doesn’t take a dedicated `enabled` using zero and/or negative values instead, but having a clear, explicit toggle isn’ŧ a bad thing

libass/ass.c
```diff
@@ -1359,6 +1364,34 @@ void ass_flush_events(ASS_Track *track)
     track->parser_priv->read_order_elems = 0;
 }
 
+void ass_configure_gc(ASS_Track *track, int enabled, long long delay)
+{
+    track->parser_priv->gc_enabled = enabled;
+
+    if (0ll <= delay)
+        track->parser_priv->gc_delay = delay;
```
TheOneric, line 1372: This may behave surprisingly for a sequence like e.g. the following:

```C
ass_configure_gc(track, true, 5000);
ass_configure_gc(track, false, LONG_MAX);
ass_configure_gc(track, true, -1);
```

This will end up with a delay of `LONG_MAX` eventhough it wasn't requested during activation and not in use at any point in time before. Otoh the following options come to mind:
- ignore `delay` when `enabled == false` and explicitly document either the current behaviour for negative values or declare negative values to be illegal and subject to change
- refuse to change anything for illegal values and return `bool` to indicate whether configuration was successfull
- merge `enabled` and `delay` with negative values disabling gc

libass/ass.h
cubicibo: Frankly, I found the term `lag` better, but astiob question on its purpose made me doubt that choice. I don't think the term `backbuffer` helps. It does not convey the unit or what it is unless you really read the docu, unlike `now`, `delay` or `lag`. I am open to another suggestion.

libass/ass.c
```diff
@@ -1330,6 +1341,7 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
 
         event->Start = timecode;
         event->Duration = duration;
+        track->parser_priv->gc_next_undisplay_ts = FFMIN(track->parser_priv->gc_next_undisplay_ts, event->Start + event->Duration);
```
cubicibo, line 1344: The start and duration fix-ups above prevent this line to be in `process_event_tail`, preventing a clean implementation.

TheOneric: We could move time assigns here before `process_event_tail`, but in the pathological case where
- custom format lines are used
- `Start` and/or `End` appear (again?) as or after the fourth element
- one of these value differs from packet timestamps

However this case is arguably already somewhat broken and might not be worth worrying about

libass/ass.c
```diff
@@ -190,14 +190,23 @@ static int test_and_set_read_order_bit(ASS_Track *track, int id)
...
     if (track->parser_priv->read_order_bitmap[index] & bit)
         return 1;
     track->parser_priv->read_order_bitmap[index] |= bit;
     return 0;
 }
 
+static void clear_read_order_bit(ASS_Track *track, int id)
+{
+    int index = id >> 5;
+    if (index < track->parser_priv->read_order_elems) {
```
TheOneric, line 204: Given `test_and_set_read_order_bimap` doesn't bother checking whether the id is in range, is there any reason why it’s needed here?

libass/ass.c
```diff
@@ -1359,6 +1371,48 @@ void ass_flush_events(ASS_Track *track)
...
+            ++k;
+        }
+        diff = k - stop;
+        if (stop > start)
+            memmove(&track->events[start], &track->events[stop], sizeof(ASS_Event)*diff);
+        start += diff;
+    }
+    track->parser_priv->gc_next_undisplay_ts = next_undisplay_ts;
+    track->n_events = start;
+}
```
TheOneric, line 1414: Yep this now avoids spurious data moves, thanks. It leaks a bit too much state outside the loop and a little bit of polishing imho improves readability and reduces mutable variables outside the loop to just two:

```C
void ass_prune_events(ASS_Track *track, long long deadline)
{
    if (deadline < track->parser_priv->gc_next_undisplay_ts)
        return;

    const bool check_readorder = track->parser_priv->check_readorder;
    const ASS_Event *events = track->events;
    const int old_n_events = track->n_events;

    long long min_undisplay_remaining = LLONG_MAX;
    int n_kept = 0;

    for (int k = 0; k < old_n_events;) {
        // discardable sequence
        for (; k < old_n_events && events[k].Start + events[k].Duration < deadline; k++) {
            if (check_readorder)
                clear_read_order_bit(track, events[k].ReadOrder);
            ass_free_event(track, k);
        }

        // to-be-kept sequence
        int move_from = k;
        for(; k < old_n_events && events[k].Start + events[k].Duration >= deadline; k++)
            min_undisplay_remaining = FFMIN(min_undisplay_remaining, events[k].Start + events[k].Duration);

        // Relocate kept events
        if (move_from < k) {
            int cnt = k - move_from;
            memmove(events + n_kept, events + move_from, cnt * sizeof(*track->events));
            n_kept += cnt;
        }
    }

    track->n_events = n_kept;
    track->parser_priv->gc_next_undisplay_ts = min_end_remaining;
}
```

I was wondering why `check_readorder` was cached to a local variable eventhough it’s only checked in a single place, but this seems to slightly improve generated instructions inside the loop and we established this should be as fast as possible. So i kept it and added two more shortcuts (though `events` prob helps more with readability and conciseness rather than codegen)*

TheOneric (commented):
libtool and macro version bumps must be in the same commit as the interface change. Caching of end timestamps in a different commit is cleaner and good, though if it makes adapting easier squashing this also is imho fine too.

libass/ass.c
cubicibo: Consider the following:
```c
ass_set_check_readorder(track, 0);

//Add some packets
while (pkt = pkt->next)
    ass_process_chunk(track, pkt->data, pkt->size, pkt->pts, pkt->dur);

ass_set_check_readorder(track, 1);
ass_prune_events(track, 0); //if index is not checked, memory corruption would ensue
```

TheOneric: Thought changing `check_readorder` after init was forbidden, but apparently no such restriction exists. Makes sense then, thx

libass/ass.h
```diff
@@ -734,6 +734,24 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
  */
 void ass_set_check_readorder(ASS_Track *track, int check_readorder);
 
+/**
+ * \brief Garbage collect events that preceed deadline
+ * \param track track
+ * \param deadline cut-off timestamp in milliseconds.
+*/
+void ass_execute_gc(ASS_Track *track, long long deadline);
```
TheOneric, line 742: fwiw i liked the previous name, `ass_prune_events`, better (similar change in `\brief`)

libass/ass_render.c
```diff
@@ -3428,6 +3428,9 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
     ass_frame_unref(priv->prev_images_root);
     priv->prev_images_root = NULL;
 
+    if (track->parser_priv->gc_enabled)
+        ass_execute_gc(track, now-track->parser_priv->gc_delay);
```
TheOneric, line 3432: style nit: `now - track->parser_priv->gc_delay`

libass/ass.c
```diff
@@ -1359,6 +1374,52 @@ void ass_flush_events(ASS_Track *track)
...
+        // discardable sequence
+        for (; k < old_n_events && events[k].Start + events[k].Duration < deadline; k++) {
+            if (check_readorder)
+                clear_read_order_bit(track, events[k].ReadOrder);
+            ass_free_event(track, k);
+        }
+
+        // to-be-kept sequence
+        int move_from = k;
+        for(; k < old_n_events && events[k].Start + events[k].Duration >= deadline; k++)
```
TheOneric, line 1408: whoops my bad; style nit: `for (;`

TheOneric (commented):
looking good, afaict all that’s left is some bikeshedding over naming, style and how to handle combinations of `bool enabled, long long deadline` or whether to merge it

libass/ass.h
cubicibo: I went with the _gc prefix to associate it to the other function. I am fine with either.

astiob (approved):

libass/ass.c
```diff
@@ -190,14 +192,23 @@ static int test_and_set_read_order_bit(ASS_Track *track, int id)
 {
     if (resize_read_order_bitmap(track, id) < 0)
         return -1;
-    int index = id / 32;
-    uint32_t bit = 1u << (id % 32);
+    int index = id >> 5; //div32
+    uint32_t bit = 1u << (id & 0x1F); //mod32
```
MrSmile, line 196: I think those comments aren't really needed. Or at least they should be uniform between `test_and_set_read_order_bit` and `clear_read_order_bit`.

libass/ass.c
```diff
@@ -1359,6 +1374,52 @@ void ass_flush_events(ASS_Track *track)
...
+        for (; k < old_n_events && events[k].Start + events[k].Duration >= deadline; k++)
+            ASS_CACHE_MIN_TS(min_undisplay_remaining, events[k].Start + events[k].Duration);
+
+        // Relocate kept events
+        if (move_from < k) {
+            int cnt = k - move_from;
+            memmove(events + n_kept, events + move_from, cnt * sizeof(*track->events));
+            n_kept += cnt;
+        }
+    }
```
MrSmile, line 1417: I think straightforward approach would be simpler to understand and not that much different in performance:
```c
    int n = 0;
    for (int i = 0; i < old_n_events; i++) {
        if (events[i].Start + events[i].Duration < deadline) {
            if (check_readorder)
                clear_read_order_bit(track, events[i].ReadOrder);
            ass_free_event(track, i);
        } else {
            ASS_CACHE_MIN_TS(min_undisplay_remaining, events[i].Start + events[i].Duration);
            events[n] = events[i];
            n++;
        }
    }
```

Also I think it's better to move `if (check_readorder)` check inside `clear_read_order_bit()`.

libass/ass.h
```diff
@@ -734,6 +734,24 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
  */
 void ass_set_check_readorder(ASS_Track *track, int check_readorder);
 
+/**
+ * \brief Garbage collect events that preceed deadline.
```
MrSmile, line 738: I think "garbage collection" is incorrect term here. It's deterministic and unambiguous deletion of events, so simple "Prune events" or "Delete events" would be better.

libass/ass_priv.h
```diff
@@ -65,6 +65,10 @@ struct parser_priv {
     uint32_t header_flags;
 
     uint32_t feature_flags;
+
+    int gc_enabled;
+    long long gc_delay;
+    long long gc_next_undisplay_ts;
```
MrSmile, line 71: I think `gc_enabled` and `gc_delay` should be merged. Also if we're moving form GC to prune, it should be renamed too:
```c
int prune_enabled;
long long prune_delay;
long long prune_next_ts;
```

libass/ass.c
```diff
@@ -59,6 +59,8 @@ static const char *const ssa_event_format =
 
 #define ASS_STYLES_ALLOC 20
 
+#define ASS_CACHE_MIN_TS(CACHE_TS, NEW_TS) CACHE_TS = FFMIN(CACHE_TS, NEW_TS)
```
MrSmile, line 62: This looks excessive for me. At least I prefer function over macro:
```c
static inline update_prune_ts(long long *ts, const ASS_Event *events)
{
    *ts = FFMIN(*ts, event->Start + event->Duration);
}
```
Or maybe even
```c
static inline update_prune_ts(ASS_Track *track, const ASS_Event *events)
{
    track->parser_priv->gc_next_undisplay_ts =
        FFMIN(track->parser_priv->gc_next_undisplay_ts, event->Start + event->Duration);
}
```

MrSmile (commented):
I think `gc` should be replaced with `prune` for consistency.
Also I vote for merging `enabled` and `delay`, as separate variant introduces unnecessary complexity and edge cases.

Sadly, our API is ill-suited for removing events from the start of the array. I think we really should move towards deprecation of direct access.

libass/ass.c
cubicibo: > Also I think it's better to move `if (check_readorder)` check inside `clear_read_order_bit()`.

That would be an inconsistency between `test_and_set_read_order_bit` and `clear_read_order_bit`. I would suggest to not do anything, or inline both with the check moved inside.

MrSmile: `memmove` can be faster than simple loop, but in this case it's simple scan + `memmove`. I think one simple loop would be faster, especially in the case where continuous bunch of events is larger than L1 cache. But yes, it should be benchmarked.

cubicibo: I did some very basic artificial benchmarking with `__rdtsc()`, 50,000 runs each. No particular gcc optimization.

Given a list of 1000 events:
1. Element-wise interleaving: every other event is to be pruned.
```
    memmove: 11330.00888   1-1copy: 6737.40456
```
2. Block-wise interleaving: every other chunk of 25 events is to be pruned:<br/>
```
    memmove: 5906.99984    1-1copy: 6829.36928
```
3. Ordered list: only one event the middle is to be pruned.
```
    memmove: 4810.76426    1-1copy: 8892.79756
```

MrSmile: > No particular gcc optimization.

Do you mean the default `-O2`? There's no sense in benchmarking `-O0` builtds.

Anyway, I suspect that it's not due to `memmove`, but due to `if (move_from < k)` check. Can you test one event at the start, instead of the middle? Similar optimization can be added to direct copy approach too, but given that discarded events usually would be at the start, I think it's not worth it.

cubicibo: `-O0` had been used, but the tendency is similar with `-O2`.
Only the first event is discarded:
```
    memmove: 1342.11    1-1copy: 2912.24
```

With `-O2`, equivalence is at a chunk size of 7-8.

TheOneric (approved):
Seems like benchmark favour the current version and otherwise it’s all good too afaict sans a minor doc nit

libass/ass.h
```diff
@@ -734,6 +734,24 @@ void ass_process_chunk(ASS_Track *track, const char *data, int size,
...
+*/
+void ass_prune_events(ASS_Track *track, long long deadline);
+
+/**
+ * \brief Configure automatic pruning of events.
+ * \param track track
+ * \param delay delay from "now" (ass_render_frame) in milliseconds.
+ * After every succesful render, events whose undisplay timestamp
+ * predate "now - delay" will be deleted.
+ * Negative delays disable pruning. A delay of 0 prunes aggressively.
```
TheOneric, line 750: *(for some reason this disappeared from the initial approving review)*

>  Negative delays disable pruning.

To avoid confusion abouth whether this affects manual `ass_prune_events` calls, i think *“Negative delays disable **automatic pruning.”* would be better

libass/ass.c
TheOneric: remark obsoleted by merging `enabled` and `delay`

TheOneric:
Thanks again!

Issue #736, opened 2024-03-08
Rendering: performance issue while rendering running rainbow animated subtitles


NBruderman:
### Description of the issue

I tried to watch the video while using mpc-hc with libass implementation on, and came across weird stuttering while
a specific animation of running rainbow colors subtitles ran on screen.
The issue also occurs while using the latest mpv build.
Below there is a small sample of the relevant part.

[Sample](https://mega.nz/file/iFdiSSyQ#7K3Yt0o3B24txrt7VnjnIqoMoeoBovLKCt7AUzHHyik)

*The issue does not occur when using the internal subtitle renderer of mpc-hc with libass turned off

### libass version

0.17.1

### Special Fonts

The issue doesn't depend on a specific font

### Additional info

I opened this issue under rendering as it seems to be the closest option to the issue

arch1t3cht:
While I can reproduce the lag on mpc-hc+libass, this sample drops 0 frames for me on mpv. So this seems to be specific to mpc-hc's usage of libass.

This file *is* however slightly peculiar because the strings inside the BlockGroups of the muxed subtitle track all contain null terminators, which I at least haven't seen in an mkv file before. The sample's mediainfo claims it was muxed using Lavf, but from testing and skimming the source code it doesn't look like libavformat includes null terminators in these subtitle packets.

I noticed this because these subtitles would fail to render properly on libass when demuxed with Aegisub (which currently does not account for such packets). Now, theoretically this could be seen as a libass/VSFilter performance differential since VSFilter *does* render a file like this (that is, a file where every Dialogue event ends with a null byte) normally while libass seems to see any null byte as terminating the entire file, but this could also be intentional.

NBruderman:
Actually, on mpv I got much less video lag, however, I got audio lags, which didn't occur on mpc-hc.
I understand my sample doesn't contain audio, so I'm adding here the same sample, but with audio intact.

[Sample with audio](https://mega.nz/file/HNlS3Awa#riTtnTINEuT-VVHkejtlASMAcm7kGhMPZ0FBEl1PsqY)

astiob:
The ASS consists of the exact same event repeated 596 times with the only differences being the primary colour (but not alpha) and rectangular clip, and with the clips animated (plus a few additional easy events). The event text is plain English. We should be caching all of this, but…

I’m not very good with profilers, but when I try Valgrind’s Callgrind on this, these items seem to stand out:

Accumulated/inclusive times:

```
[...]

Issue #782, opened 2024-06-04
Discuss: -ffast-math and other math optimization flags


rcombs:
This was brought up in https://github.com/libass/libass/pull/780#issuecomment-2148145960.

`-ffast-math` is shorthand for several flags:
- `-fno-math-errno` (we decided in #780 that we definitely want this)
- `-funsafe-math-optimizations`, which is shorthand for:
  - `-fno-signed-zeros` (probably fine?)
  - `-fno-trapping-math` (unclear what this would mean for us?)
  - `-fassociative-math` (unsure)
  - `-freciprocal-math` (unsure)
- `-ffinite-math-only` (we have a couple `isnan`s; not sure if this is safe wrt those?)
- `-fno-rounding-math` (this is actually the default anyway?)
- `-fno-signaling-nans` (this is also the default)
- `-fcx-limited-range` (we don't use complex numbers)
- `-fexcess-precision=fast` (this is the default)

So this leaves `-fno-signed-zeros`, `-fno-trapping-math`, `-fassociative-math`, `-freciprocal-math`, and `-ffinite-math-only` up for discussion. One simple test would be to build with and without each flag, and see what code changes and what perf impact they have.

I think if we have cases where `-fassociative-math` or `-freciprocal-math` help (and are safe), we should probably be explicitly adjusting our floating-point code to allow the compiler to produce optimal output even without those flags.

kasper93:
I think most if not all flags would be safe to use. For `-ffinite-math-only` it mostly means we have to be careful about dividing by zero, which is most common case of producing nans. Either way, library like libass hardly needs a IEEE compliant floats. Even `-ffast-math` will probably work in most of the cases, though from experience it is generally better to avoid it, unless we can with confidence say that libass code is good for this. Better start little by little and enable only safe(r) flags. `-ffast-math` can be deceptive, because it gives nice performance boost and things look fine at first sight, but after a while you start hitting corner cases.

`-fno-signed-zeros` is safe to enable
`-fno-trapping-math` probably also safe, I'm not familiar with libass codebase, but I imagine it is not needed.

`-fassociative-math` and `-freciprocal-math` will increase error and start producing floats that are bit off in some cases. Questions is if it is a problem. I would be careful with those, but still probably nothing catastrophic will not happen.

MrSmile:
`-funsafe-math-optimizations` looks safe to me. We don't use float hacks, only treat them as approximation for real numbers.

[...]

PR #806, opened 2024-07-24
utils: implement our own lrint() for Visual C++

xenu:
Commit cd35a4dd added the -fno-math-errno flag which allows lrint() to be inlined on GCC and Clang.

Unfortunately, Visual C++ can't inline lrint() without -fp:fast. That flag has some nasty implications, so I think it's better to just write our own lrint.

This change helps a bit with the video sample from #736. It halves the number of dropped frames in MPC-HC on my machine (from ~180 to ~80).

### DIFF
--- a/libass/ass_utils.h
+++ b/libass/ass_utils.h
@@ -30,6 +30,10 @@
 #include <limits.h>
 #include <math.h>
 
+#if defined(_MSC_VER) && !defined(__clang__)
+#include <intrin.h>
+#endif
+
 #include "ass.h"
 
 #ifndef SIZE_MAX
@@ -159,6 +163,17 @@ static inline uint32_t ass_bswap32(uint32_t x)
 #endif
 }
 
+static inline long ass_lrint(double x)
+{
+#if defined(_MSC_VER) && !defined(__clang__) && !defined(_M_ARM64EC) && (defined(_M_X64) || _M_IX86_FP == 2)
+    // Unlike GCC and Clang, Visual C++ can't inline lrint() without -fp:fast, so
+    // we provide our own implementation
+    return _mm_cvtsd_si32(_mm_set_sd(x));
+#else
+    return lrint(x);
+#endif
+}
+
 static inline int d6_to_int(int x)
 {
     return (x + 32) >> 6;
@@ -189,7 +204,7 @@ static inline double d6_to_double(int x)
 }
 static inline int double_to_d6(double x)
 {
-    return lrint(x * 64);
+    return ass_lrint(x * 64);
 }
 static inline double d16_to_double(int x)
 {
@@ -197,7 +212,7 @@ static inline double d16_to_double(int x)
 }
 static inline int double_to_d16(double x)
 {
-    return lrint(x * 0x10000);
+    return ass_lrint(x * 0x10000);
 }
 static inline double d22_to_double(int x)
 {
@@ -205,7 +220,7 @@ static inline double d22_to_double(int x)
 }
 static inline int double_to_d22(double x)
 {
-    return lrint(x * 0x400000);
+    return ass_lrint(x * 0x400000);
 }
 
 static inline int32_t lshiftwrapi(int32_t i, int32_t shift)
--- a/libass/ass_outline.c
+++ b/libass/ass_outline.c
@@ -418,8 +418,8 @@ bool ass_outline_transform_2d(ASS_Outline *outline, const ASS_Outline *source,
             ass_outline_free(outline);
             return false;
         }
-        outline->points[i].x = lrint(v[0]);
-        outline->points[i].y = lrint(v[1]);
+        outline->points[i].x = ass_lrint(v[0]);
+        outline->points[i].y = ass_lrint(v[1]);
     }
     memcpy(outline->segments, source->segments, source->n_segments);
     outline->n_points = source->n_points;
@@ -457,8 +457,8 @@ bool ass_outline_transform_3d(ASS_Outline *outline, const ASS_Outline *source,
             ass_outline_free(outline);
             return false;
         }
-        outline->points[i].x = lrint(v[0]);
-        outline->points[i].y = lrint(v[1]);
+        outline->points[i].x = ass_lrint(v[0]);
+        outline->points[i].y = ass_lrint(v[1]);
     }
     memcpy(outline->segments, source->segments, source->n_segments);
     outline->n_points = source->n_points;
@@ -478,7 +478,7 @@ void ass_outline_update_min_transformed_x(const ASS_Outline *outline,
         double x = (m[0][0] * pt[i].x + m[0][1] * pt[i].y + m[0][2]) / FFMAX(z, 0.1);
         if (isnan(x))
             continue;
-        int32_t ix = lrint(FFMINMAX(x, -OUTLINE_MAX, OUTLINE_MAX));
+        int32_t ix = ass_lrint(FFMINMAX(x, -OUTLINE_MAX, OUTLINE_MAX));
         *min_x = FFMIN(*min_x, ix);
     }
 }
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -1096,7 +1096,7 @@ void ass_process_karaoke_effects(RenderContext *state)
                     info->c[1] = tmp;
                 }
             }
-            x = x_start + lrint((x_end - x_start) * dt);
+            x = x_start + ass_lrint((x_end - x_start) * dt);
         }
 
         for (GlyphInfo *info = start; info < end; info++) {
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -542,7 +542,7 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
         center[i] -= delta[i];
         if (!(fabs(center[i]) < max_val))
             return false;
-        qr[i] = lrint(center[i]);
+        qr[i] = ass_lrint(center[i]);
     }
 
     // Minimal bounding box z coordinate
@@ -602,7 +602,7 @@ static bool quantize_transform(double m[3][3], ASS_Vector *pos,
             double val = m[i][j] * mul[j];
             if (!(fabs(val) < max_val))
                 return false;
-            qm[i][j] = lrint(val);
+            qm[i][j] = ass_lrint(val);
         }
 
     // x_lim = |m_xx| * dx + |m_xy| * dy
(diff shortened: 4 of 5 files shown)


xenu:
Compiler Explorer proof that the generated machine code of ass_lrint() is identical to how GCC inlines lrint(): https://godbolt.org/z/aPWcGqjrE

NBruderman:
@astiob is looks like this quite help with my issue. Do you see any downsides to this?

libass/ass_utils.h
```diff
@@ -159,6 +163,17 @@ static inline uint32_t ass_bswap32(uint32_t x)
...
+static inline long ass_lrint(double x)
+{
+#if defined(_MSC_VER) && !defined(_M_ARM64EC) && (defined(_M_X64) || _M_IX86_FP == 2)
+    // Unlike GCC and Clang, Visual C++ can't inline lrint() without -fp:fast, so
+    // we provide our own implementation
+    return _mm_cvtsd_si32(_mm_set_sd(x));
+#else
+    return lrint(x);
+#endif
+}
```
TheOneric, line 175: This adds a function call indirection on all platforms which we previously got rid off with `-fno-errno-math`. This is not acceptable. Instead `ass_lrint` should be a macro expanding to `lrint` on most platforms, or perhaps directly override `lrint` with a macro on MSVC

TheOneric (dismissed):
Did you/someone ask Microsoft for a fine grained `no-errno-math` toggle yet?  

In general I’m not too fond of having too much platform-specific optimisations in the code and would much prefer a simple compiler toggle in configuration logic. This particular case may be relevant enough to add a code variant anyway, but on the other hand many more projects might profit from MSVC having a `no-errno-math` equivalent

libass/ass_utils.h
rcombs: It's a `static inline`, and no function call is emitted in practice (though I wouldn't object to `#define ass_lrint lrint` on non-MSVC): https://godbolt.org/z/Pcn7TMddo

Defining macros over standard library symbols already provided by the libc is generally UB.

libass/ass_utils.h
```diff
@@ -159,6 +163,17 @@ static inline uint32_t ass_bswap32(uint32_t x)
 #endif
 }
 
+static inline long ass_lrint(double x)
+{
+#if defined(_MSC_VER) && !defined(_M_ARM64EC) && (defined(_M_X64) || _M_IX86_FP == 2)
```
rcombs, line 168: Why the explicit exclusion of arm64 here? Shouldn't the X64/X86 conditions not pass on ARM?

libass/ass_utils.h
TheOneric: oh, it’s in the header not `.c`. Yeah, it’s unlikely to end up as a function call then

rcombs (commented):
Out of curiosity, why are you building with MSVC? (I'm trying to understand the use-cases behind this better.)

libass/ass_utils.h
xenu: When compiling for the ARM64EC ABI the `_M_X64` macro is also defined, as [documented here](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170).

The longer explanation is that there are two ABIs on ARM64 Windows. "ARM64" is the normal ABI and it behaves as one would expect. "ARM64EC" is a special ABI that tries to be as X64 compatible as possible to allow linking with emulated X64 binaries.

xenu:
>Out of curiosity, why are you building with MSVC? (I'm trying to understand the use-cases behind this better.)

Because that's what MPC-HC does.

xenu:
>In general I’m not too fond of having too much platform-specific optimisations in the code and would much prefer a simple compiler toggle in configuration logic. This particular case may be relevant enough to add a code variant anyway, but on the other hand many more projects might profit from MSVC having a no-errno-math equivalent

I generally agree with your concern, but I would argue that this particular case is not *that* invasive, and the performance benefit is very noticeable.

libass/ass_utils.h
rcombs: Huh! Bizarre, but I suppose it's harmless. I see that MSVC ends up emitting a call to `_CVTSD2SI32` here, and I wonder how that differs from `lrint`.

MrSmile:
I feel that `/fp:fast` is a much better workaround for this. Should be fairly safe especially in MSVC, as it's not that aggressive at optimizing as GCC and clang.

xenu:
In both `-fp:fast` and `-ffast-math`, I'm always uncomfortable with associative math optimizations, as they have the potential to change the results. Of course, it's very likely that it doesn't affect libass. I'm not really familiar with this code, certainly not as much as you guys.

BTW, while researching for this PR, I found [this ticket](https://developercommunity.visualstudio.com/t/MSVCs-sincos-implementation-is-incorrec/10582378) about a nasty bug in a combined sin + cos optimization in `/fp:fast`. I found this because I noticed this pattern in libass and I wondered if MSVC has sincos (it doesn't). That said, it seems this bug mostly affects very large numbers so I think it shouldn't cause problems in practice.

xenu:
PS. I've built unmodified libass with `/fp:fast` and the number of dropped frames is about the same as with my patch. It seems that inlining lrint() is the only FP optimization that significantly affects performance in that sample video.

TheOneric:
> (In my experience Microsoft tends to ignore feature requests.)

I didn’t find anything explicitly requesting an `-fno-errno-math` flag on the page you linked earlier. But there’s https://developercommunity.visualstudio.com/t/trivial-math-functions-as-rintf-generate/10469611 complaining about `rintf` etc not compiling to a simple instructions and it was marked as fixed recently, supposedly available in the preview build since yesterday.
Iiuc `rintf` should also sometimes set `errno`, e.g. when rounding leads to `HUGE_VAL`, but perhaps they are able to determine whether errno get read or not.

Could you check if this already helps? If not we probably want to carry this code-tweak (for some while) due to its large impact, but i still think it’d be wort to at least try to file a feature request

xenu:
>Could you check if this already helps? 

I've installed the latest preview (17.11 Preview 5) and none of the examples provided by the author of that ticket are inlined, with or without `/fp:fast`. I don't understand this.

There's no change with `lrint`, it's inlined only with `/fp:fast`.

xenu:
>This affects clang code generation for msvc target, so please explicitly exclude clang via __clang__.

Done (although it's a bit pedantic as the generated code should be the same).

xenu:
>but i still think it’d be wort to at least try to file a feature request

https://developercommunity.visualstudio.com/t/Add-a-flag-similar-to--fno-errno-math-th/10710444

astiob:
> This affects clang code generation for msvc target, so please explicitly exclude clang via `__clang__`.

If Clang is being used in MSVC emulation mode, doesn’t it _not_ get `-fno-errno-math`, therefore doesn’t it fail to inline `lrint`, and therefore don’t we _want_ to include it in this optimization?

xenu:
It's... complicated. Clang can be configured to use the MinGW toolchain or to use the Visual C++ toolchain.

It also has two drivers, both are present in either configuration.

`clang.exe` sets `_MSC_VER` if it's configured to use the VC++ toolchain. But it takes standard flags regardless of the configuration, so `-fno-math-errno` works there.

`clang-cl.exe` always sets `_MSC_VER` regardless of the configuration and since its commandline is compatible with VC++, it doesn't understand `-fno-math-errno`.

It's a mess.

I imagine Meson users will typically use the `clang.exe` frontend.

astiob:
> so I used clang trunk and found that intrinsics generate suboptimal code.

Out of curiosity, what was suboptimal about the code? Should we be worried and search for alternatives without the intrinsic?

xenu:
>Out of curiosity, what was suboptimal about the code? Should we be worried and search for alternatives without the intrinsic?

I have a suspicion what happened:

ass_lrint() assumes that `sizeof(long) == 4` which is [always true on Windows](https://learn.microsoft.com/en-us/cpp/cpp/data-type-ranges?view=msvc-170) on all architectures.

When you switch to Clang in the Compiler Explorer, you switch to Linux, so the size of `long` changes to 8. Because of that, Clang emits additional `cdqe` instruction (convert doubleword to quadword). It disappears if you switch the function's return type to `int` or switch the intrinsic to the 64-bit version.

@Andarwinux is this what happened?

TheOneric (approved):

NBruderman:
Any news about this merge? Or is this on hold until it's time for the next release?

kasper93:
`lrint` is not the only one affected function. Generally I think mid-term goal should be to adjust the code to run safely with `/fp:fast` as this is most likely fine for libass needs. Only few NaN checks would need to be handled differently, but not biggie.

Reimlementing standard library functions is not feasible approach long-term.

xenu:
> Only few NaN checks would need to be handled differently, but not biggie.

NaN behaviour is far from the only difference between `/fp:strict` (the default) and `/fp:fast`.

Anyway, I'm abandoning this PR because this is getting tiring.

NBruderman:
> > Only few NaN checks would need to be handled differently, but not biggie.
> 
> NaN behaviour is far from the only difference between `/fp:strict` (the default) and `/fp:fast`.
> 
> Anyway, I'm abandoning this PR because this is getting tiring.

Why closing this? Just keep it open until the maintainers will decide what approach to take. You don't need to work on this anymore, just keep it open

kasper93:
> Anyway, I'm abandoning this PR because this is getting tiring.

Opened last week... Approved last week... `+41 −26`... tiring. Ok, buddy. 

Back on topic.

> NaN behaviour is far from the only difference between /fp:strict (the default) and /fp:fast.

Of course, but I argue that only NaN is affecting current libass code base in a meaningful way, because there is two checks that would be broken by that. I've analyzed it last time, with similar discussion about `-fno-math-errno` and siblings. The rest of code does not require IEEE compliant floating point operations.

xenu:
>PR here can take quite a while to be merged.

Waiting isn't an issue, I've had PRs that waited for years in other projects.

But I'm not invested in this project enough to continue arguing in this thread, especially with people like this guy above. This is the part that is tiring.

NBruderman:
Maybe you can open this PR just so people remember it exists, while the discussion can take place here?
https://github.com/libass/libass/issues/782
It feels like just too important to be just left out. It's basically 2.2x performance gain.

astiob:
Merged as https://github.com/libass/libass/commit/b13f5d107237abb15487dc031f83c11e369a4be9. Thanks! We can improve on this later if/when we decide on a better way to handle this (in #782 or otherwise).

(I can’t say that I, personally, fully understand why this was closed and deleted, but everyone here is a volunteer. If the discussion seemed to become unpleasant, I feel sorry; but I can respect the wish to abandon it. Thanks for your contribution!)

PR #807, opened 2024-07-25
fuzz: avoid Coverity warning about umask-less mkstemp

TheOneric:
Coverity is content with this, but tbh it seems _almost_ useless considering *(a)* this is no confidential data and files for an explicit output path are created with default umask (without Coverity complaining) and *(b)* POSIX Issue 7 (2008) already requires `600` perms anyway. The only benefit, other than pleasing Coverity, is consistency with across Issue ≤6 POSIX and ≥7

Opening PR for comments and if we dismiss the issue instead of pushing this, I’m not sure if “false positive” or “intentional” is a better fit

### DIFF
--- a/fuzz/writeout.c
+++ b/fuzz/writeout.c
@@ -22,6 +22,7 @@
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/stat.h>
 
 #include "writeout.h"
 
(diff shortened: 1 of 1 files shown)


rcombs (commented):
Is there even a reason for this to use mkstemp, rather than e.g. defaulting to `parsedSubs.ass` in the current working directory?

TheOneric:
The benefit of `mkstemp` is avoiding conflict with potentially already existing files and as a bonus utomatic cleanup of files no longern (at least) the next reboot. *(And `mkstemp` avoids some race conditions etc of `mktemp` or `tmpname`)*.  

I guess we could also use a fixed name and error out if it already exists (with a race condition, but this may be acceptable here). For consistency explicitly specified output path probably also should error instead of overwrite then.

TheOneric:
fwiw i marked this as a “false positive” now in Coverity

TheOneric:
i think we can change this when we eventually move to a C11 baseline, but until then the current approach seems fine and the Coverity report is already closed

PR #808, opened 2024-07-26
meson: Add support for android, simplify some nasm detection logic

nirbheek:
1. Redefining the meaning of `cpu_family` is a bit confusing, and unnecessary
2. Now setting `-Dasm=enabled` will error out early, so people know exactly what is wrong
3. When erroring out due to platform support, print the platform name

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -293,7 +293,7 @@ if not asm_option.disabled()
                 error(
                     'Please contact libass upstream to figure out if ASM support ' +
                     'for your platform @0@ can be added. '.format(host_system) +
-                    'In the meantime you will need to use -Dasm=disabled.'
+                    'In the meantime you will need to use -Dasm=disabled.',
                 )
             endif
 
--- a/meson.build
+++ b/meson.build
@@ -237,15 +237,15 @@ asm_args = []
 
 # ASM architecture variables
 asm_option = get_option('asm')
-cpu_subfamily = host_machine.cpu_family()
-if cpu_subfamily.startswith('x86')
-    cpu_family = 'x86'
+cpu_family = host_machine.cpu_family()
+if cpu_family.startswith('x86')
+    generic_cpu_family = 'x86'
 else
-    cpu_family = cpu_subfamily
+    generic_cpu_family = cpu_family
 endif
 
 if not asm_option.disabled()
-    if cpu_family == 'x86'
+    if generic_cpu_family == 'x86'
         asm_is_nasm = add_languages(
             'nasm',
             required: false,
@@ -272,7 +272,7 @@ if not asm_option.disabled()
             conf.set('ARCH_X86', 1)
             nasm_args = ['-Dprivate_prefix=ass', '-DPIC=1']
 
-            if cpu_subfamily == 'x86_64'
+            if cpu_family == 'x86_64'
                 conf.set('ARCH_X86_64', 1)
                 nasm_args += '-DARCH_X86_64=1'
             else
@@ -280,7 +280,7 @@ if not asm_option.disabled()
             endif
 
             if host_system in ['windows', 'cygwin']
-                if cpu_subfamily == 'x86'
+                if cpu_family == 'x86'
                     nasm_args += '-DPREFIX'
                 endif
             elif host_system == 'darwin'
@@ -299,7 +299,7 @@ if not asm_option.disabled()
 
             add_project_arguments(nasm_args, language: 'nasm')
         endif
-    elif cpu_subfamily == 'aarch64'
+    elif generic_cpu_family == 'aarch64'
         enable_asm = true
         conf.set('ARCH_AARCH64', 1)
         if host_system == 'darwin'
@@ -308,7 +308,7 @@ if not asm_option.disabled()
     else
         warning(
             'Assembly optimizations are not yet supported for the "@0@" architecture; disabling.'.format(
-                cpu_subfamily,
+                cpu_family,
             ),
         )
     endif
--- a/libass/meson.build
+++ b/libass/meson.build
@@ -83,9 +83,9 @@ endif
 
 if enable_asm
     asm_sources = []
-    if cpu_family == 'x86'
+    if generic_cpu_family == 'x86'
         asm_sources = src_x86
-    elif cpu_subfamily == 'aarch64'
+    elif generic_cpu_family == 'aarch64'
         asm_sources = src_aarch64
     endif
 


astiob:
If I’m not mistaken, we support assembly also on x32, which presumably has a different `cpu_family()` value. Hence the prefix check in the existing code.

nirbheek:
Meson does not have a separate `cpu_family()` value for x32: https://mesonbuild.com/Reference-tables.html#cpu-families. cpu_family() does not detect or report kernel/userland combinations. It just tells you what the processor is (or what mode it is running in).

TheOneric:
x32 is one of the things the meson build just doesn’t support at all atm. Originally due to difficulties detecting x32 in meson *(before meson gained built-in NASM support)* and now because meson misconfigures NASM on x32 systems *(using built-in support)*.  
That said, while `_family == 'x86` or `arch in ['x86', 'x86_64']` doesn’t make too much of a difference now, once meson finally (hopefully) acts on https://github.com/mesonbuild/meson/issues/6361 it might change

> It just tells you what the processor is (or what mode it is running in).

I believe this is not correct. **muon** currently uses `uname` to detect arch and thus fails e.g. in 32bit containers without a prior “personality” syscall (e.g. via `linux32`), but meson (additionally/solely?) checks compiler defines: https://github.com/mesonbuild/meson/blob/6258f0dc86188570e7d00316557bfb85948ed290/mesonbuild/environment.py#L370
*(Which in theory makes it easy to detect x32 and atl east fixing built-in NASM, but there #s no good way to communicate this for custom configuration atm)*

TheOneric:
re changes, the last commit for Android ASM support seems good, thanks. Looking at the host OS table it fortunately seems like we didn’t miss any other value. I’m not sure about the prior two commits though and will take a closer look at a later date

nirbheek:
> meson (additionally?) checks compiler defines: https://github.com/mesonbuild/meson/blob/6258f0dc86188570e7d00316557bfb85948ed290/mesonbuild/environment.py#L370

The implementation is not the intention :)

What I have stated is the intention behind the property. For example, see the lengths meson goes to to get the actual processor architecture on windows: https://github.com/mesonbuild/meson/blob/6258f0dc86188570e7d00316557bfb85948ed290/mesonbuild/meson

meson.build
```diff
@@ -281,16 +281,14 @@ if not asm_option.disabled()
...
                 nasm_args += []
             else
-                error(
-                    'Please contact libass upstream to figure out if ASM support ' +
-                    'for your platform can be added. In the meantime you will need ' +
-                    'to use -Dasm=disabled.',
-                )
+                error('Please contact libass upstream to figure out if ASM support ' +
+                    'for your platform @0@ can be added. In the '.format(host_system) +
+                    'meantime you will need to use -Dasm=disabled.')
```
TheOneric, line 291: cosmetic nit: the formatting is inconsistent with other long log messages and imho it’d be nicer to add `.format()` just after ` added.` when the relevant sentence ends

meson.build
```diff
@@ -232,16 +232,14 @@ asm_is_nasm = false
...
-if cpu_subfamily.startswith('x86')
-    cpu_family = 'x86'
-else
-    cpu_family = cpu_subfamily
-endif
+cpu_family = host_machine.cpu_family()
+asm_option = get_option('asm').require(
+    cpu_family in ['x86', 'x86_64', 'aarch64'],
+    error_message: 'ASM support is not available on ' + cpu_family,
+)
```
TheOneric, line 239: i’m not fond of this; it requires updating this list _and_ the configuration logic for all future additions (or if it ever happens: removals)

If ASM was explicitly requested on an unsupported arch the warning informing about it should be immediately before the error anyway iinm

meson.build
```diff
@@ -244,7 +244,7 @@ if not asm_option.disabled()
     if cpu_family == 'x86'
         asm_is_nasm = add_languages(
             'nasm',
-            required: false,
+            required: asm_option,
```
TheOneric, line 245: hmm.. won’t this make it _harder_ to figure out where the problem is or put another way: what paths forward there are for `-Dasm=enabled` setups?  
Previously warnings and errors clearly indicate NASM is only required for assembly and build failed because assembly was explicitly requested but couldn’t be set up. Now it just fails due to not finding NASM without any context and no indication `-Dasm=disabled` can work too *(as  much as assembly-les builds shouldn’t be used for production, this might be fine/preferable for a quick test of some non-perf change)*

TheOneric:
> What I have stated is the intention behind the property. For example, see the lengths meson goes to to get the actual processor architecture on windows: 

The link is broken, but e.g.: [`mesonbuild/environment.py@detect_windows_arch`](https://github.com/mesonbuild/meson/blob/6258f0dc86188570e7d00316557bfb85948ed290/mesonbuild/environment.py#L276) explicitly _avoids_ returning the actual processor arch, but queries the compiler target arch.  
Anyway, i don’t really mind whether `cpu_family` returns the true underlying native processor arch, currently configured processor arch mode or the userland/compiler arch as long as there’s some other easily queryable property for the latter. Afaik there isn’t though and the incompleteness of currently exposed properties affects e.g. x32 and everything were manual setup for hardfloat/softfloat is required.

TheOneric (commented):
The third commit adding `android` as a Linux alias is good apart from a minor cosmetic nit; thanks!

CHanges to `cpu_(sub)family` in the second commit are mixed. Using the same `cpu_family` name as meson’s function but with a different menaing is indeed confusing. On the other hand needing to duplicate lists of all subarches everywhere isn’t great either. How about renaming the properties to `arch_family` and `arch_full` *(maybe with a `cpu_` prefix)* to distinguish it from meson’s properties?

I’m not convinced the first commit brings any benefit; see comment.

meson.build
nirbheek: Sorry, missed that, fixed now.

meson.build
nirbheek: IME people will hardly ever read error messages, let alone warnings above it. I found this issue specifically because other people did not read those warnings, and when we moved from autotools to meson, asm support had been silently disabled (because the old build defaulted to `--enable-asm` but now the default is `auto`, although that's a separate issue).

I think it's also an easy way to know at a glance what the arch support is. Unfortunately, meson's syntax doesn't (yet?) support a way to make these things declarative / DRY.

meson.build
nirbheek: Hm the error message from meson indeed leaves something to be desired:

```
meson.build:243:22: ERROR: Unknown compiler(s): [['nasm'], ['yasm']]
The following exception(s) were encountered:
Running `nasm --version` gave "[Errno 2] No such file or directory: 'nasm'"
Running `yasm --version` gave "[Errno 2] No such file or directory: 'yasm'"
```

Does libass even support yasm? The version check seems to suggest no. I have added more error messages, please check.

nirbheek:
> On the other hand needing to duplicate lists of all subarches everywhere isn’t great either. How about renaming the properties to arch_family and arch_full (maybe with a cpu_ prefix) to distinguish it from meson’s properties?

I think the best name for that would be `arch_is_x86_or_64` which is longer than `['x86', 'x86_64']` ;)

I don't think DRY is applicable here, I think it's clearer to duplicate it.

meson.build
TheOneric: > IME people will hardly ever read error messages, let alone warnings above it

If someone doesn’t read error messages after configuration failed, the changes to the unread message here won’t help, but our logic gets more complicated and by duplication more prone to future bugs. If the error message is read, the existing message already explicitly points to prior warnings which will be placed immediately above the error so it’d be hard to miss. Sorry, I don’t see a benefit

> because the old build defaulted to --enable-asm but now the default is auto, although that's a separate issue

Was this an issue in the downstream setup, e.g. omitting `-Dasm=enabled` where `--enable-asm` was set previously, or did some behaviour difference between autotools and the meson sneak in?

meson.build
TheOneric: > Does libass even support yasm?

If it works then only by chance. Originally libass supported only yasm, because NASM was less maintained at the time, then roles reversed and we switched to only NASM in 2017. I see yasm has some activity still albeit far less than NASM, so mayhaps it’s now able to provide all features we use, but it’s entirely untested.

TheOneric:
> I think the best name for that would be arch_is_x86_or_64 which is longer than ['x86', 'x86_64'] ;)

This is supposed to group _any_ related architectures, not just x86 specifically; e.g. `riscv32` and `riscv64` would probably also end up with common sources but are split in meson’s `*.cpu_family()`. WebAssembly was also discussed before, but idk how similar or not `wasm32` and `wasm64` are.  
Having a generic architecture family scales better with future additions.

meson.build
nirbheek: The issue was that we weren't passing `-Dasm=enabled`.

The difference from autotools was that with meson's toolchain support for nasm, `nasm` needs to be specified in the cross-file to be picked up for cross-builds, unlike a regular program  (this will be fixed in the next meson release).

meson.build
nirbheek: The question is, do you want to support people who want to use yasm? If so, the version check should be conditional on a compiler-id check.

meson.build
nirbheek: I have removed the `.require()` now.

nirbheek:
> Having a generic architecture family scales better with future additions.

I have changed the code to use `generic_cpu_family`. The only outstanding question now is whether you want to support `yasm` or not.

TheOneric:
Trying to build with yasm 1.3.0, the current latest release, fails so it isn’t suitable

TheOneric:
Apologies, i completely forgot about this; i merged the third and fourth commit *(handling Android and rejecting yasm)* as b26ea492b183c57eee7534df7c154ca7b0ab18b2 and 1774117b8048d6296c79425cfbb1086bcd2ac2fa to not needlessly stall those part any longer; thanks again!

Even with the simplicifcations, the first commit still seems like a lot of extra branching and duplication for a minor cosmetic output gain to me. Similarly for the
 ```
if cond
  error(...)
else
  warn(...)
end
```
pattern in the second commit.

Could you reduce this to just `cpu_family` renaming?

**Comment by TheOneric** [comment]
for reference, runnng `meson.get_compiler('nasm')` if NASM wasn’t found can uses an error for crossbuilds, but not native builds:

```
Has header "windows.h" : NO 
Compiler for language nasm for the host machine not found.

meson.build:254:30: ERROR: Tried to access compiler for language "nasm", not specified for host machine.

A full log can be found at /media/werkbank/FOSS/libass/build_meson/meson-logs/meson-log.txt
```

This happens with meson < 1.5.0 when nasm isn’t specified in the crossfile _or_ isn’t available and for meson >= 1.5.0 when it was explicitly specified in the crossfile but isn’t available.
I _think_ the `< 1.5.0` behaviour was a meson bug and `>= 1.5.0` behaviour seems reasonable enough, though since 1.5.0 is only about a month old i guess many might run into this when building libass with distro-provided meson without nasm or on non-x86 :\

Is there a more elegenat way of avoiding this problem other than repeating `meson.get_compiler('nasm')` everywhere?

TheOneric:
Caching `meson.get_compiler('nasm')` was reverted in 5c00ca266dd49dd7cf93a79f09a661ab1ff66439 since even with current meson master this still caused issues for native Windows builds without nasm installed; see https://github.com/mpv-player/mpv/actions/runs/10418042994/job/28853424386

TheOneric:
Rebased and merged the renaming too now; thanks again!

When/if the conditional warning-or-error syntax from https://github.com/mesonbuild/meson/pull/13482 gets merged and we’re able to bump our minimum meson version sufficiently high, early erroring out can be revisited

PR #811, opened 2024-08-02
build: add -fno-math-errno for clang-cl

Andarwinux:


astiob (approved):

meson.build
```diff
@@ -25,8 +25,10 @@ cc = meson.get_compiler('c')
 cc_warnings = []
 cc_features = []
 
-if cc.get_id() != 'msvc'
+if cc.get_id() not in ['msvc', 'clang-cl']
```
kasper93, line 28: I think it would be more readable to do 
``` meson
if cc.get_id() == 'clang-cl'
...
elif if cc.get_id() != 'msvc'
...
endif
```

configure.ac
```diff
@@ -52,6 +52,7 @@ FUZZ_CPPFLAGS="${FUZZ_CPPFLAGS:--DASS_FUZZMODE=0}"
 
 # Checks for flags support
 AX_APPEND_COMPILE_FLAGS([-fno-math-errno])
+AX_APPEND_COMPILE_FLAGS([-clang:-fno-math-errno])
```
kasper93, line 55: nitpick, but `clang-cl` uses nominally cl.exe style arguments, so for consistency it should be `/clang:-fno-math-errno`. I believe both should work, but just point this out.

TheOneric (approved):

NBruderman:
Does this have any downsides? Or is this good to go?

PR #819, opened 2024-09-05
blend_bitmaps: allow compiler to optimize C variants

kasper93:
There are hand-optimized version, but it is good to have optimized baseline too. The difference is not big, but marking those pointers with restrict give compiler more freedom.

### DIFF
--- a/libass/c/c_blend_bitmaps.c
+++ b/libass/c/c_blend_bitmaps.c
@@ -29,8 +29,8 @@
  * \brief Add two bitmaps together at a given position
  * Uses additive blending, clipped to [0,255]. Pure C implementation.
  */
-void ass_add_bitmaps_c(uint8_t *dst, ptrdiff_t dst_stride,
-                       const uint8_t *src, ptrdiff_t src_stride,
+void ass_add_bitmaps_c(uint8_t *restrict dst, ptrdiff_t dst_stride,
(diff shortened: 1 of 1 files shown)


rcombs (approved):

TheOneric:
Note: this is a subset of the changes from #749

Out of interest, when/if you measured, how big was the perf difference (just for those functions or/and for a complete rendering) with/out `restrict`

astiob (approved):
Could you add the qualifiers to `ass_bitmap_engine.h` too? It probably won’t make a difference in the code generation, but that would be technically more correct and also more self-documenting.

rcombs:
Should I just split out the first 3 commits of #749 for merge, actually?

kasper93:
> Should I just split out the first 3 commits of https://github.com/libass/libass/pull/749 for merge, actually?

I wasn't aware, there are changes already. Feel free to merge it from your branch.

> Could you add the qualifiers to ass_bitmap_engine.h too?

I think @rcombs already have those.

> Out of interest, when/if you measured, how big was the perf difference (just for those functions or/and for a complete rendering) with/out restrict

Nothing, significant. I did it more for correctness and little nicer codegen, not really performance reasons.

kasper93:
Superseded by #820

PR #822, opened 2024-09-06
ASSUME: tighten assumptions even more

MrSmile:
I think explicitly specified pointer alignment can be beneficial too.

Also we should definitely replace `ASSUME` with `assert` for CI builds. I.e.
```c
#ifdef TESTING
    #define ASSUME(x) assert(x)
#elif __has_builtin(__builtin_assume)
    #define ASSUME(x) __builtin_assume(x)
#elif defined(_MSC_VER)
    #define ASSUME(x) __assume(x)
#elif __has_builtin(__builtin_unreachable)
    #define ASSUME(x) if (!(x)) __builtin_unreachable()
#else
    #define ASSUME(x) (void) 0
#endif
```
We can reuse the same define later for threading cache asserts too.

### DIFF
--- a/libass/ass_bitmap_engine.h
+++ b/libass/ass_bitmap_engine.h
@@ -26,7 +26,9 @@
  * All of these routines require some basic preconditions about their args:
  * - Widths and heights must be > 0
  * - For be_blur, width and height must be > 1
- * - All strides must be multiples of the engine alignment (16 or a greater power of 2)
+ * - All strides must be multiples of the engine alignment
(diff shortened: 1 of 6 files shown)


rcombs (approved):
Not sure if this actually gets used effectively by current compilers, but sure, doesn't hurt to have.

MrSmile:
Dunno about exact performance numbers, but at least clang replaces `movdqu` with `movdqa`.

TheOneric (approved):

PR #820, opened 2024-09-05
Restrict-qualification + ASSUME macro usage in bitmap engine

rcombs:
First 3 commits of #749, rebased, as brought up by #819. Changes:
- `restrict`-qualify pointer args to all our C bitmap-engine functions
  - These never alias, and letting the compiler know that allows it to skip emitting some unnecessary defensive code
- Add an `ASSUME` macro
  - This tells the compiler that it can assume a particular expression will always evaluate to true, which can help it optimize more effectively (especially when autovectorizing)
- Use `ASSUME` to signal some trivial invariants in bitmap-engine routines
  - Stride is always 16-byte-aligned, width/height are never zero (or 1, for be_blur), etc

PR #824, opened 2024-09-06
Verify assumed call requirements in CI builds

TheOneric:
See prior discussion in #820 and #822

Feel free to suggest other names for the macro

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,17 @@ Windows. It is suited for static-only builds on any platform well supported by
 Meson and as a Meson subproject.
 Notably, Meson supports MSVC and generation of VS project files.
 
+Macro defines
+-------------
+
+Unless developing libass there’s usually no need to manually define macros
+and there are no stability guarantees for these manual defines.
+
+- `DEBUG_LEVEL=0..3`
+  - `0` use the default set of asserts; implied if macro is not defined at all
+  - `1`, `2` *unused*
+  - `3` additionally assert assumptions usually pledged to the compiler for optimization purposes
+
 Information about the ASS format:
 =================================
 - [ASS format overview](https://github.com/libass/libass/wiki/ASS-File-Format-Guide)
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -215,6 +215,7 @@ jobs:
       - name: configure
         run: |
           export CC="${{ matrix.cc }}\
+            -DDEBUG_LEVEL=3\
             ${{ matrix.extra_cflags && ' ' }}${{ matrix.extra_cflags }}\
             ${{ steps.sanitizer.outputs.flags && ' ' }}${{ steps.sanitizer.outputs.flags }}"
           export LD="${{ matrix.ld }}"
--- a/libass/ass_utils.h
+++ b/libass/ass_utils.h
@@ -59,7 +59,9 @@
     #define __has_builtin(x) 0
 #endif
 
-#if __has_builtin(__builtin_assume)
+#if DEBUG_LEVEL >= 3
+    #define ASSUME(x) assert(x)
+#elif __has_builtin(__builtin_assume)
     #define ASSUME(x) __builtin_assume(x)
 #elif defined(_MSC_VER)
     #define ASSUME(x) __assume(x)


rcombs (commented):
For reference, ffmpeg has `ASSERT_LEVEL`, where 0 enables only always-on asserts, 1 enables asserts that aren't in performance-critical code but aren't enabled by default, and 2 enables asserts in performance-critical code. We wanted something similar in the threading PR; maybe "convert ASSUMEs to asserts" could be ASSERT_LEVEL 3?

TheOneric:
sounds good; adopted `ASSERT_LEVEL` and lacking a better place documented it in our readme

rcombs (approved):

MrSmile:
Not sure about `ASSERT_LEVEL` without actual influence on `assert` (contrary to `NDEBUG`). I think it should be paired with corresponding `ass_assert` macro with level parameter (or `ass_assert_0`..`ass_assert_4`). Actually, if we're creating our own macro, we can merge `assert` and `ASSUME` into one.

TheOneric:
> Not sure about `ASSERT_LEVEL` without actual influence on `assert` . I think it should be paired with corresponding `ass_assert` macro with level parameter (or `ass_assert_0`..`ass_assert_4`).

It affects assert insofar as it changes how many asserts are placed.
v1 of this patch just used `EXTRADEBUG`, but conditionally omitting some asserts came up before and in anticipation of future usage imho it makes sense to use a common knob for `assume → aassert` and assert omission.  
On the other hand, if we expect non-omission and assume-to-assert transforms to always appear together having 4 different levels would be needlessly complex and one singe `ifdef` condition work too. @rcombs how and why does FFmpeg use its three different assert levels?

Adding just the assume-to-assert capability now makes sense to me as we’re currently adding assumes and this can help catch bogus assumes early. I don’t want to couple this with reclassifying our existing asserts or adding new asserts for higher levels; this can happen later.

> Actually, if we're creating our own macro, we can merge assert and ASSUME into one.

In the default build configuration `ASSUME` and `ass_assert_{1,2,,3}` act differently enough to warrant a clear distinction imho

rcombs:
ffmpeg's levels are:
- 0 is always on, and is used to catch conditions that should absolutely never be possible, where continuing would be unsafe, and crashing would be preferable even in a production build
- 1 is similar to a traditional C `assert`: it's expected to be on for most development builds, but off in most production builds; it's used for states that the program can probably safely recover from, but that probably indicate a problem that a developer would want to trap into a debugger to investigate
- 2 is further restricted to only trigger in builds where performance isn't a concern (usually this means e.g. `-g -O0` builds, not `-g -O3` builds that might be used for performance testing)

You could imagine tying these to different build configurations, like 0 for `release`, 1 for `reldebug`, and 2 for `debug`.

I think libass would benefit from having either this same set of 0-2, or perhaps 0-3 (where ASSUME is only converted to an assertion at level 3). Currently ffmpeg doesn't have an `ASSUME` equivalent (and perhaps it should).

MrSmile:
> where ASSUME is only converted to an assertion at level 3

I think that assert should always convert to ASSUME when its level is disabled, separate macros do not make much sense. So usual assert = ass_assert_1, ASSUME = ass_assert_2 or 3.

TheOneric:
I don’t mind much how the toggle and assert/assume macros are named; my primary concern is ensuring `ASSUME`s get some testing. I’m not sure what will be the best for planned, not yet existing optional asserts etc. If everyone agrees on one way to set it up now, that’ſ great and i can adopt this. But if the best way isn’t clear yet i suggest moving forward with just the `ASSUME` testing here and defer details of other assert levels to when they’ll actually get used.

rcombs:
IMO we shouldn't automatically turn asserts into assumes, and if something's intended as an assume but also benefits from an assert in lower ASSERT_LEVELs, the developer should explicitly mark that. Some asserts may perform memory accesses or function calls that we don't want to ever happen in release builds, and the `ASSUME` may allow those accesses/calls to be emitted in some cases; it's also possible that we might have some asserts (at eg level 1 or 2) that we don't want to automatically become UB, but that we do want to trap into a debugger to diagnose on in development builds.

MrSmile:
> But if the best way isn’t clear yet i suggest moving forward with just the `ASSUME` testing here and defer details of other assert levels to when they’ll actually get used.

In that case, I think it would be better to rename `ASSERT_LEVEL` to `DEBUG_LEVEL`.

> Some asserts may perform memory accesses or function calls that we don't want to ever happen in release builds

Fair point. But as I understand, both `__builtin_assume` and `__assume` do not actually evaluate expression, only `__builtin_unreachable` variant is a problem.

MrSmile (approved):

README.md
```diff
@@ -30,6 +30,17 @@ Windows. It is suited for static-only builds on any platform well supported by
...
+Macro defines
+-------------
+
+Unless developing libass there’s usually no need to manually define macros
+and there are no stability guarantees for these manual defines.
+
+- `DEBUG_LEVEL=0..3`
+  - `0` use the default set of asserts; implied if macro is not defined at all
+  - `1`, `2` *unused*
+  - `3` additionally assert assumptions usually pledged to the compiler for optimsation purposes
```
rcombs, line 42: typo: optimization

rcombs (approved):

README.md
TheOneric: fixed thx

PR #827, opened 2024-09-15
libass/ass_render: add ASS_OVERRIDE_BIT_BLUR to ASS_OverrideBits 

llyyr:


llyyr:
This is broken and wrong, and I should've thought more about why this wasn't set before

astiob:
Hm. Would it perhaps make more sense to put this under `ASS_OVERRIDE_BIT_BORDER` if you want to reuse one of the existing bits? After all, what exactly gets blurred is determined by the other border overrides, not by font size or spacing.

llyyr:
> Hm. Would it perhaps make more sense to put this under `ASS_OVERRIDE_BIT_BORDER` if you want to reuse one of the existing bits? After all, what exactly gets blurred is determined by the other border overrides, not by font size or spacing. (Quick recap: blur applies to outline-border if it exists, which is decided by outline width and borderstyle.)

I just followed what the header says https://github.com/libass/libass/blob/master/libass/ass.h#L157

TheOneric:
> I just followed what the header says https://github.com/libass/libass/blob/master/libass/ass.h#L157

Looking at the [introducing commit](https://github.com/libass/libass/commit/51a93b5571acf51d3c7fe841d3e1e34720524c23) this just rescales any existing blur to the new overridden font size to preserve the `blur / font_size` ratio. Which makes much more sense to me than tying user blur overrides to a `FONT_SIZE` override. *(We should clarify the docs though)*  
The introduction of [`Blur` as a style API field](https://github.com/libass/libass/commit/794fa0f2cb8c3b3b5fc9e26dbca489fae1b5d1ab) predates the above commit by i guess it was forgotten about when splitting up override bits.

Imho blur fits `ASS_OVERRIDE_BIT_ATTRIBUTES` *(currently handling)* the best, but i’m also slightly concerned whether any API users rely on this **not** including blur. Then again, does anyone but mpv even use this API?

Currently blur can already and also only be overridden by `ASS_OVERRIDE_FULL_STYLE`. Looking at the motivating mpv issue, wouldn’t `FULL_STYLE` be a good match for `override=force`?

I haven't tested whether blur and font size overrides with `FULL_STYLE` mean blur gets both overridden and then the overridden value again scaled together with the new font size.

llyyr:
> Currently blur can already and also only be overridden by `ASS_OVERRIDE_FULL_STYLE`. Looking at the motivating mpv issue, wouldn’t `FULL_STYLE` be a good match for `override=force`?

No, we don't want `ASS_OVERRIDE_BIT_ALIGNMENT` and we also don't want `ASS_OVERRIDE_BIT_ATTRIBUTES` because they'll just set those values to mpv default.

> Imho blur fits ASS_OVERRIDE_BIT_ATTRIBUTES (currently handling)

What do you mean "currently handling"? Blur isn't handled by anything currently, and this PR adds it to ASS_OVERRIDE_BIT_BORDER.

TheOneric:
> > Imho blur fits ASS_OVERRIDE_BIT_ATTRIBUTES (currently handling)
>
> What do you mean "currently handling"?

Sorry, forgot to add the list; this should’ve been: “(currently handling Bold, Italic, Underline, StrikeOut)”

llyyr:
Hmm, I think ASS_OVERRIDE_BIT_BORDER makes most sense for mpv. FONT_SIZE_FIELDS is also fine. I'd prefer it not be part of ASS_OVERRIDE_BIT_ATTRIBUTES though.

TheOneric:
> I'd prefer it not be part of ASS_OVERRIDE_BIT_ATTRIBUTES though

Can you elaborate on why not `ATTRIBUTES` ad why `BORDER` works for mpv, so the usecase is better understood?

llyyr:
> Can you elaborate on why not `ATTRIBUTES` ad why `BORDER` works for mpv, so the usecase is better understood?

When you set `--sub-ass-override=force` with `ASS_OVERRIDE_BIT_ATTRIBUTES`, mpv will apply every single `--sub-*` option currently. This includes options like `--sub-bold`, `--sub-italics` etc.

The goal of `sub-ass-override` isn't to be destructive, but to allow users to set custom styling while being as destructive to the rest of the file as possible, so effectively stripping away bold/italics from the original ASS file shouldn't happen because it's too destructive.

Similarly for `ASS_OVERRIDE_BIT_ALIGNMENT`, if mpv used this flag every subtitle would be rendered at the bottom because that's the default value of `--sub-pos`.

mpv needs to be smarter about not passing through mpv defaults when `--sub-ass-override=force` is set, but I haven't looked into what that might entail so it'd be nicer if this wasn't part of `ASS_OVERRIDE_BIT_ATTRIBUTES` because I'd have to disable it on mpv's side then to not strip away bold/italics from ASS files.

Also I think `ASS_OVERRIDE_BIT_BORDER` fits better since this only affects borders.

astiob:
> Looking at the [introducing commit](https://github.com/libass/libass/commit/51a93b5571acf51d3c7fe841d3e1e34720524c23) this just rescales any existing blur to the new overridden font size

TBH I don’t see any link between blur and the `FONT_SIZE_FIELDS` bit in that commit at all. In any case, this must’ve been an oversight of some sort.

`ATTRIBUTES` is font attributes/styles. `BORDER` still seems like a better fit to me (or an entire new bit), because blur is effectively a property of the border, not of the font (except with Outline=0, but that _itself_ is a border property).

It does seem odd that we have a `FULL_STYLE` option if mpv doesn’t actually use it, but the rationale for not using it seems fair.

TheOneric:
> > Looking at the [introducing commit](https://github.com/libass/libass/commit/51a93b5571acf51d3c7fe841d3e1e34720524c23) this just rescales any existing blur to the new overridden font size
> 
> TBH I don’t see any link between blur and the FONT_SIZE_FIELDS bit in that commit at all.

You’re right, this is triggered by a different `FONT` override bit; I didn’t read it carefully enough:  
At the end of the diff for `ass_render.c` there’s a conditional to multiply various scale values by a `font_size_coeff` which was controlled by an explicit check for `OVERRIDE_BIT_FONT_SIZE` before and `apply_font_scale` after the commit.
`apply_font_scale` is initialised further up in line 774 of post-commit `ass_render.c` based on, among otherfactors, `ASS_OVERRIDE_BIT_SELECTIVE_FONT_SCALE`.

> `ATTRIBUTES` is font attributes/styles. `BORDER` still seems like a better fit to me (or an entire new bit), because blur is effectively a property of the border, not of the font (except with Outline=0, but that itself is a border property).

Hmm, i’m thinking of blur more as an unrelated thing which happens to be affected by value of `Outline`, so as an API user I’d find it surprising if border overrides can end up blurring text *(which might not have a border)*. So i’m leaning towards a new bit.

We also seem to currently miss a fine-grained toggle for `Angle` which actually exists in ASS files and not just our API. How about combining `Blur` and `Angle` into `ASS_OVERRIDE_BIT_EFFECTS` or similar?

astiob:
> We also seem to currently miss a fine-grained toggle for `Angle` which actually exists in ASS files and not just our API. How about combining `Blur` and `Angle` into `ASS_OVERRIDE_BIT_EFFECTS` or similar?

I wouldn’t expect anything good to come out of users overriding `Angle` tbh. And combining it with blur seems weird.

So, do we want a brand new bit or do we merge this as is?

astiob:
To be brutally honest, I don’t see a lot of value in these bits at all (and by implication, in splitting out a new bit): the API user can just not set overrides they don’t want in the first place. In the comments above, it’s clarified that mpv currently sets all user styles indiscriminately as overrides and uses these bits for damage control, and it’s agreed that it probably shouldn’t but this works well enough at the moment.

So the pragmatic question is: if this is grouped together with `BIT_BORDER`, does that work well enough for mpv? I assume the answer is yes, although FWIW it’s not obvious to me as a user why `--sub-bold` and `--sub-blur` should work any differently from each other.

llyyr:
> if this is grouped together with BIT_BORDER, does that work well enough for mpv?

It being part of BIT_BORDER or its own flag, either way works for mpv.

> FWIW it’s not obvious to me as a user why --sub-bold and --sub-blur should work any differently from each other.

I believe what makes --sub-ass-override useful is that it allows applying custom styles in a reasonably non destructive way. This is just my personal opinion but I believe applying `ASS_OVERRIDE_BIT_ATTRIBUTES` is too destructive considering Bold/Italics/Underline aren't so much _styling_ as they are integral components of the text itself. Blur not being overridden appears to just be an oversight on wm4's part, considering the comment says Blur is part of `ASS_OVERRIDE_BIT_FONT_SIZE_FIELDS`, even though it's not actually the case, of course we're free to decide now which bit should actually toggle it.

TheOneric:
> To be brutally honest, I don’t see a lot of value in these bits at all (and by implication, in splitting out a new bit): the API user can just not set overrides they don’t want in the first place.

I agree, but alas the bitflag API is already present for quite some time and we gotta work with it *(or actually mark it as deprecated and officially discourage any further use in documentation and the next release announcement)*.

While mpv is the only user _we know of_ and here squeezing blur into an existing flag might work out ok, i feel this doesn't quite fit and will just make this API even odder possibly creating problems in the future for users and/or us *(and maybe current users we don't yet know about)*, thus I prefer a new bit for this. Also adding a toggle for Angle would complete the set, but if you feel it’s too unlikely to be used we can omit it until, if at ever, we get a request for it.

llyyr:
> squeezing blur into an existing flag might work out ok

I don't agree with this interpretation. I believe my PR is more akin to a bug fix than introducing new functionality to an existing flag or a new flag altogether. According to the comment, Blur was _always_ intended to be part of it and it not being coded as such is a bug. However, since it was never part of any flags, we can choose a more appropriate flag without it being a backwards incompatible change.

> maybe current users we don't yet know about

Current users should already assume it's part of `FONT_SIZE_FIELDS`, because that's what the comment says.

If adding this to `BIT_BORDER` is going to make it seem like a new addition, then I propose just making this a part of `FONT_SIZE_FIELDS` as originally intended.

astiob:
I disagree with the idea that documentation trumps actual behaviour for backwards compatibility purposes. Any change in behaviour is a visible change. The bug was in the documentation, and the way to fix the _bug_ is to remove the mention of blur from that comment. It’s not like we accidentally removed blur for a release or two; it never was there. So on this front, I think I’m siding with Oneric.

Indeed, the linked mpv issue tackles the same issue, and your stance _there_ seems to be that when the actual behaviour makes sense but differs from the documentation—which still says that all `--sub-*` options are applied—then the behaviour stays and it’s the documentation that may need editing. I don’t think blur makes any sense in a group of size-related properties.

By the way, this documentation means `--sub-bold` and `--sub-italics` should _also_ be applied.

And really, maybe they should. Weight in particular is typically a font’s property in subtitles: most fonts are _always_ bolded, and weight changes are not used for emphasis or at all. If the user picks a font that naturally looks bold but has a low nominal weight, they may prefer to avoid it getting faux-bolded. Meanwhile, italics are used for emphasis or stylistic marking, but maybe the user hates them and wants to see all text in upright font?

---

Separately, it’s dawned upon me that we may have some misunderstanding about what the override bits actually do at all. I certainly misunderstood this, at least:

> the API user can just not set overrides they don’t want in the first place

No, they can’t. The API user sets _a whole style,_ and these bits control which parts of this style are applied. For example, the API user can’t configure this style to “not have a border width”: there _always_ is a border width, even if zero, and the bitmask is precisely what allows it to be ignored.

These bits _do not_ affect explicit user overrides like `--sub-ass-style-overrides=Outline=5`. Those are always applied regardless of this bitmask.

From this perspective, I’m more inclined to give blur a new bit, and frankly, I’d welcome a separate flag for _each_ field. I still don’t mind blur being added to the border bit, though.

astiob:
Just to spell this out explicitly: putting philosophy aside, the practical difference between adding a new bit for blur and adding blur to the existing border bit is what happens when different versions of libass and mpv are used together:

If we add it to the existing bit, dropping in a newer libass build will automatically fix mpv’s `--sub-blur`.

If we introduce a new bit, it will require a newer libass _and_ a newer mpv.

llyyr:
> If we introduce a new bit, it will require a newer libass _and_ a newer mpv.

This is fine. I'm not opposed to any of the ideas above, as long as it's possible to control whether user overridden `Blur` is ignored or not. I don't think mpv needs the ability to control each field separately, and afaik I don't believe there are any other users of this API.

TheOneric:
If I’m reading this correctly everyone is content with this proceeding with a new bit for blur

-----

> Separately, it’s dawned upon me that we may have some misunderstanding about what the override bits actually do at all. I certainly misunderstood this

ooohh, thanks for pointing this out; I forgot we also have this override API. Looking at git history, it seems the string list overrides existed from the beginning while this “set an `ASS_STyle` plus selection bits” API was added later for 0.12.0.
I’m not sure why the API ended up in this shape rather than a selectively-applied string list counterpart to the existing always-applied API, but this is off-topic for the change here

libass/ass.h
```diff
@@ -194,6 +194,10 @@ typedef enum {
      * On dialogue events override: Justify
      */
     ASS_OVERRIDE_BIT_JUSTIFY = 1 << 10,
+    /**
+     * On dialogue events override: Blur
+     */
+    ASS_OVERRIDE_BIT_BLUR = 1 << 11,
```
TheOneric, line 200: To allow users to guard usage of the new enum, `LIBASS_VERSION` should be bumped too

astiob:
> I’m not sure why the API ended up in this shape rather than a selectively-applied string list counterpart to the existing always-applied API

Oh, I think this is actually pretty clear. There’d be no point in a selectively-applied string API: just don’t pass strings you don’t need. But this selective API allows an alternative when you don’t want to build strings: for example, mpv has `--sub-blur` etc. and it just builds an `ASS_Style` out of them. When rendering unstyled subtitles, it uses this style directly to render the subs. When rendering styled subs, it passes the same style instance but selectively disables some of its fields.

TheOneric:
> Oh, I think this is actually pretty clear. There’d be no point in a selectively-applied string API: just don’t pass strings you don’t need.

Yes, but to my understanding the “selective” distinguishing `ass_set_selective_style_override*` from `ass_set_style_overrides` is the latter being _always_ applied to all (matching) styles and events, while (unless using `ASS_OVERRIDE_FULL_STYLE`) the former apparently only acts on events heuristically classified as dialogue.
It just seems to me like it might’ve been better to also use a string list for `ass_set_selective_style_override` instead of `ASS_Style` and separately a bunch of override bits

llyyr:
> actually mark it as deprecated and officially discourage any further use

This would need to introduce a new API to pass like `ass_set_style_overrides`, but it would take in ASS_Style instead of clients needing to build a string.

TheOneric:
> >    actually mark it as deprecated and officially discourage any further use
>
> This would need to introduce a new API

After we cleared up the misunderstanding which override API the bits affect, deprecation isn’t considered anymore since as you say, it would remove the ability for context-sensitive (dialogue/typeset) overrides requiring a replacement.

TheOneric (approved):

PR #839, opened 2024-10-15
build: properly spell "Fontconfig", "Core Text", "macOS", "DirectWrite"

astiob:


configure.ac
```diff
@@ -28,11 +28,11 @@ AC_ARG_ENABLE([profile], AS_HELP_STRING([--enable-profile],
 AC_ARG_ENABLE([fuzz], AS_HELP_STRING([--enable-fuzz],
     [enable fuzzing consumer @<:@default=no@:>@]))
 AC_ARG_ENABLE([fontconfig], AS_HELP_STRING([--disable-fontconfig],
-    [disable fontconfig support @<:@default=check@:>@]))
+    [disable Fontconfig support @<:@default=check@:>@]))
 AC_ARG_ENABLE([directwrite], AS_HELP_STRING([--disable-directwrite],
     [disable DirectWrite support (Windows only) @<:@default=check@:>@]))
 AC_ARG_ENABLE([coretext], AS_HELP_STRING([--disable-coretext],
-    [disable CoreText support (OSX only) @<:@default=check@:>@]))
+    [disable Core Text support (macOS only) @<:@default=check@:>@]))
```
astiob, line 35: In fact, this isn’t actually correct: Core Text is also available on iOS and probably on Apple’s other platforms (watchOS, tvOS…) if libass can be built there at all. Should we change this to “macOS/iOS”? Or something else?

rcombs: I usually just say "Apple" (ie "(Apple only)" here).

astiob: Added a commit that switches to “Apple”.

configure.ac
```diff
@@ -238,7 +238,7 @@ AS_IF([test "x$enable_require_system_font_provider" != xno  dnl
         && test "x$directwrite" != xtrue                    dnl
         && test "x$coretext" != xtrue                       ], [
     AC_MSG_ERROR(m4_text_wrap(m4_normalize([
-            Either DirectWrite (on Windows), CoreText (on OSX), or Fontconfig
+            Either DirectWrite (on Windows), Core Text (Apple), or Fontconfig
             (Linux, other) is required. If you really want to compile without
             a system font provider, add --disable-require-system-font-provider]),
```
TheOneric, line 243: While we’re at it: this is technically not quite correct or at least misleading: Fontconfig is available on all platforms not just “Linux, other” and providing just Fontconfig also sufficies on Windows and Apple platforms

TheOneric: Maybe: “At least one of Core Text (Apple exclusive), DirectWrite (Windows exclusive) or Fontconfig is required.”?

astiob: Added a commit that switches to (almost) your suggested wording.

TheOneric (approved):

Issue #642, opened 2022-08-21
[Feature Request] ass.h - Add method ass_get_fonts


moi15moi:
### Feature Request
Would it be conceivable to add the method: ``ass_get_fonts(ASS_Track *track)``?

### What would it do
This method would return all the font path that libass would use to render one .ass file.
In brief, it would be an FontCollector like Aegisub, but directly in libass.

astiob:
My thoughts (not necessarily a rejection; discussion welcome):

libass uses fonts from a variety of sources, most prominently fonts already embedded in the video file and in the subtitle file, which have no paths.

And libass’s goal when it searches for fonts is very much not what a font collector’s goal should be. libass’s choice can be suboptimal (but during playback, any choice is better than nothing), and even if you were to grab all of the currently selected fonts, put them in an MKV or ASS file and play it back using the same libass on a different computer, you might see some of the attached fonts ignored and different fonts from the system picked instead.

For a good subtitle release, you ideally want the subtitles to never use font fallback and to always select fonts exactly, with no room for ambiguity. In my opinion, a font collector should strive to verify these properties and complain to the user if their subtitles are ambiguous. libass currently has no way of doing this, although something like an “authoring mode” has been proposed in the past and might make this viable.

moi15moi:
**Reason why I think an FontCollector would need to be part of libass.**

Libass has a algorithm to make a choice when it pick font.
Any FontCollector need to mimic this algorith to know which font libass need.

Even if the algorithm is not supposed to change, it can (see #613).
So, a FontCollector could be obsolete. If libass would propose a FontCollector, it would always be up to date.

**Currently, how libass pick font?**
```
1. ass_render_frame (ass_render)
2. ass_render_event (ass_render)
3. ass_shaper_find_runs (ass_shaper)
4. ass_font_get_index (ass_font)
5. add_face (ass_font)
6. ass_font_select (ass_fontselect)
```

We could add ``fontpath`` to the struct ``ASS_FontDesc``. 
The method ``add_face`` would set the fontpath.

**Implementation problematic**
- What do we do with font added with ass_add_font (font embedded in the video)? Do we ignore them?
- How do we ignore the "default" font?
- How to reuse the code from ``ass_render_frame`` to ``ass_shaper_find_runs``? Some part of the code are needed.

> although something like an “authoring mode” has been proposed...

Is that discussion is on github? I would like to get more information about this possibility.

astiob:
[...]

PR #841, opened 2024-10-22
Grab font path on Windows too (as is done on Linux and Mac already)

p0358:
Currently on Linux it's like:
```
libass: fontselect: (Times New Roman, 700, 100) -> /usr/share/fonts/truetype/msttcorefonts/Times_New_Roman_Bold_Italic.ttf, 0, TimesNewRomanPS-BoldItalicMT
```

While on Windows it's like (since path is empty, it falls back to just displaying postscript name there):
```
libass: fontselect: (Times New Roman, 700, 100) -> TimesNewRomanPS-BoldItalicMT, 0, TimesNewRomanPS-BoldItalicMT
```

After this PR:
```
libass: fontselect: (Times New Roman, 700, 100) -> C:\WINDOWS\FONTS\TIMESBI.TTF, 0, TimesNewRomanPS-BoldItalicMT
```

For the new function `get_utf8_path` I mimicked the code style of `get_data` and `get_utf8_name` funcs.

Yes, interface `IDWriteLocalFontFileLoader` inherits from `IDWriteFontFileLoader`, despite that not being documented properly in Microsoft Docs (they claim it inherits just from IUnknown), had to copy-paste its method into the vtable above. If you remove that from there, it will break...

Notable caveat: it will flip the code-path here:
https://github.com/libass/libass/blob/6a759836e5e76bb7b69b0ac244eea76b0d290512/libass/ass_fontselect.c#L396-L411


p0358:
(force-pushed to remove a buggy line that wasn't needed anyways, I confirmed they write the null terminator on call to `IDWriteLocalFontFileLoader_GetFilePathFromKey`)

moi15moi:
When you call ``ass_font_provider_add_font``, it calls ``ass_face_open`` which use the [``FT_New_Face``](https://github.com/libass/libass/blob/6a759836e5e76bb7b69b0ac244eea76b0d290512/libass/ass_font.c#L317).
FT_New_Face has a issue on the freetype gitlab: https://gitlab.freedesktop.org/freetype/freetype/-/issues/1098
Because of that, we cannot use the file path.

p0358:
Well, my original goal was actually to be able to extract the font paths corresponding to given styles, but only later I realized that the font paths are actually completely discarded after being logged and not preserved in some struct like ASS_Font... (apart from ASS_FontInfo struct, but it's only internally used for already looked up fonts, but not in font_cache) But it looks like the APIs really wasn't designed for arbitrary font pulling like that. For now I ended up re-coding a copy of `ass_update_font` to copy values from ASS_Style* to ASS_FontDesc* instead and then I can pretty much call `ass_font_select` with a dummy ASS_Font object and dummy renderer (after exporting some extra symbols and wrapping it all in custom Cache class). And it looks like at least without fontconfig, the fonts have to be loaded into freetype for extra metadata, so there's no going around that.

But generally not sure if libass would want to refactor some code to allow such a use-case (enlisting fonts used in styles [and with usage in events] and preserving font path to read it out later [within font_cache too]) and I might have to fork the repo anyways to include C# bindings, so I might just be fine keeping the changes in a fork then maybe (the intended use-case is for Jellyfin media server to figure out what available fonts are needed for given ass file and then send them to client for its libass to use, with the reasoning that it'd be best to apply the exact same font matching logic on both sides faithfully).

So with that recognized the PR is probably a bit less useful than I originally thought, but it would still be nice to get it in for the sake of completionness I guess and to at least allow that log, after figuring out some way to fix the aforementioned code path that causes font to be read from file instead of stream...

astiob:
This PR has merit in any case, but it sounds to me like you want to use moi15moi’s [FontCollector](https://github.com/moi15moi/FontCollector) instead of hacking libass.

> When you call `ass_font_provider_add_font`, it calls `ass_face_open` which use the [`FT_New_Face`](https://github.com/libass/libass/blob/6a759836e5e76bb7b69b0ac244eea76b0d290512/libass/ass_font.c#L317).
> FT_New_Face has a issue on the freetype gitlab: https://gitlab.freedesktop.org/freetype/freetype/-/issues/1098
> Because of that, we cannot use the file path to open the font.

The DirectWrite provider uses `ass_face_stream` instead of `ass_face_open`, and filling in the path [doesn’t change this](https://github.com/libass/libass/blob/6a759836e5e76bb7b69b0ac244eea76b0d290512/libass/ass_font.c#L447).

p0358:
> This PR has merit in any case, but it sounds to me like you want to use moi15moi’s [FontCollector](https://github.com/moi15moi/FontCollector) instead of hacking libass.

Ah, wasn't aware of this project, looks useful overall. But probably won't be able to use it for this particular case since it's in Python. But it's interesting it has strategies for both libass and VSFilter...

> The DirectWrite provider uses ass_face_stream instead of ass_face_open, and filling in the path [doesn’t change this](https://github.com/libass/libass/blob/6a759836e5e76bb7b69b0ac244eea76b0d290512/libass/ass_font.c#L447).

Well, if I make the get_utf8_path do:
```cpp
path = malloc(15);
strcpy(path, "C:/GARBAGE!!!");
return path;
```
then I get a whole bunch of these in logs:
```
libass: Error opening font: 'C:/GARBAGE!!!', 0
```

astiob (commented):
Conceptually, I like this as it makes the log more useful, but it does add an extra function to maintain without improving any core functionality, so I’m curious to hear what the other maintainers think.

Regarding the implementation:

Does `loader` not need to be released as well?

The docs don’t say explicitly, but the default is for output parameters (as in `GetLoader`) to require releasing by the caller. You should be able to determine this empirically using this hack:

```c
IDWriteFontFileLoader_AddRef(loader);
printf("Current reference count: %lu\n", IDWriteFontFileLoader_Release(loader));
IDWriteFontFile_GetLoader(file, &loader);
IDWriteFontFileLoader_AddRef(loader);
printf("Reference count after another GetLoader call: %lu\n", IDWriteFontFileLoader_Release(loader));
```

astiob:
Ah, right, I wasn’t looking at the right conditional, I guess:
https://github.com/libass/libass/blob/6a759836e5e76bb7b69b0ac244eea76b0d290512/libass/ass_fontselect.c#L396

We should probably just change this to `if (provider->funcs.get_data)`.

p0358:
> Does `loader` not need to be released as well?
You might be right, since I see others use CComPtr for it online. But that would mean that the other existing usage of `IDWriteFontFile_GetLoader` in libass is buggy and leaks ref too then probably. I remember I had an itch that something might be amiss here, but that's why I kept it like that.

Regarding your code, I can't really do AddRef before GetLoader. But I decided to just go with:
```cpp
IDWriteFontFile_GetLoader(file, &loader);
IDWriteFontFileLoader_AddRef(loader);
printf("Reference count after another GetLoader call: %lu\n", IDWriteFontFileLoader_Release(loader));
```

And you're right:
```
Reference count after another GetLoader call: 4
Reference count after another GetLoader call: 5
Reference count after another GetLoader call: 6
Reference count after another GetLoader call: 7
Reference count after another GetLoader call: 8
Reference count after another GetLoader call: 9
Reference count after another GetLoader call: 10
Reference count after another GetLoader call: 12
Reference count after another GetLoader call: 13
Reference count after another GetLoader call: 14
Reference count after another GetLoader call: 15
Reference count after another GetLoader call: 16
Reference count after another GetLoader call: 17
Reference count after another GetLoader call: 18
Reference count after another GetLoader call: 20
Reference count after another GetLoader call: 21
Reference count after another GetLoader call: 22
Reference count after another GetLoader call: 23
Reference count after another GetLoader call: 24
Reference count after another GetLoader call: 25
Reference count after another GetLoader call: 26
Reference count after another GetLoader call: 28
Reference count after another GetLoader call: 29
Reference count after another GetLoader call: 30
Reference count after another GetLoader call: 31
Reference count after another GetLoader call: 32
Reference count after another GetLoader call: 33
Reference count after another GetLoader call: 34
```

After making it release:
```
Reference count after another GetLoader call: 4
Reference count after another GetLoader call: 4
Reference count after another GetLoader call: 4
Reference count after another GetLoader call: 4
Reference count after another GetLoader call: 4
Reference count after another GetLoader call: 4
Reference count after another GetLoader call: 4
Reference count after another GetLoader call: 5
Reference count after another GetLoader call: 5
Reference count after another GetLoader call: 5
Reference count after another GetLoader call: 5
Reference count after another GetLoader call: 5
Reference count after another GetLoader call: 5
Reference count after another GetLoader call: 5
Reference count after another GetLoader call: 6
Reference count after another GetLoader call: 6
Reference count after another GetLoader call: 6
Reference count after another GetLoader call: 6
Reference count after another GetLoader call: 6
Reference count after another GetLoader call: 6
Reference count after another GetLoader call: 6
Reference count after another GetLoader call: 7
Reference count after another GetLoader call: 7
Reference count after another GetLoader call: 7
Reference count after another GetLoader call: 7
Reference count after another GetLoader call: 7
Reference count after another GetLoader call: 7
Reference count after another GetLoader call: 7
```
I also fixed it in init_font_private_stream locally and it still seems to raise like this, but more slowly. Welp, something else might be internally raising it somewhere, but certainly a bug will be fixed here.

libass/ass_directwrite.c
```diff
@@ -828,7 +925,12 @@ static void add_font(IDWriteFont *font, IDWriteFontFamily *fontFamily,
     font_priv->font = font;
     font = NULL;
 
-    ass_font_provider_add_font(provider, &meta, NULL, 0, font_priv);
+    char *path = get_utf8_path(font_priv);
+
+    ass_font_provider_add_font(provider, &meta, path, 0, font_priv);
```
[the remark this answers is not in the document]
p0358: I think that's the case too. And I agree it makes sense, since it seems that's gonna always be called later anyways in Windows's case...

astiob:
> I also fixed it in init_font_private_stream

Oh, good catch!

> Maybe I should put a fix in a separate commit since it'd fix two places, or no?

I guess the optimal commit split would be to fix `init_font_private_stream` in a separate commit that comes first, and then to have the necessary releases from the get-go in the path extraction commit.

p0358:
> The `add_font` path is the one.
Oh, I've misread `font_priv->font` and `font_priv->face` there. Yeah that makes sense now.

> I guess the optimal commit split would be to fix `init_font_private_stream` in a separate commit that comes first, and then to have the necessary releases from the get-go in the path extraction commit.

Yeah, that's actually the most reasonable way to go probably, time to do some rebasing then.

> By the way: if `init_font_private_stream` and `get_utf8_path` share about half their code, perhaps you want to extract it to another new function?

I can try, but then also in a separate last commit to avoid turning the original one into a royal mess (and so that it's easier to follow the changes)

libass/ass_directwrite.c
```diff
@@ -254,24 +254,41 @@ static void init_FallbackLogTextRenderer(FallbackLogTextRenderer *r,
...
-    return true;
+    hr = IDWriteFontFile_GetLoader(file, loader);
+    if (FAILED(hr) || !*loader)
+        goto cleanup;
+
+    ret = true;
+
+cleanup:
+    if (file)
+        IDWriteFontFile_Release(file);
```
astiob, line 289: Ah, I’m afraid this isn’t as easy as it seemed.

You can’t release the file here on the success path, because this potentially invalidates the reference key that you’re returning: its lifetime is tied to the file’s lifetime. So then you have to move this release back into the callers, which means you must return the file itself via another output argument, which reduces the gains from factoring this out in the first place. But at least the three API calls above are shared, so perhaps it’s still beneficial?

I also wanted to suggest returning the loader directly instead of a Boolean, but if you need to return both the loader and the file anyway, then the Boolean might be cleaner.

libass/ass_directwrite.c
```diff
@@ -604,7 +606,7 @@ static void add_font_face(IDWriteFontFace *face, ASS_FontProvider *provider,
 {
     ASS_FontProviderMetaData meta = {0};
 
-    IDWriteFontFace3 *face3;
+    IDWriteFontFace3 *face3 = NULL;
```
astiob, line 661: FWIW we trust that if the call has `SUCCEEDED`, then it has filled this variable with some well-defined value per its contract. But this is fine too.

p0358: Technically it should, but it's probably still a bit safer and more consistent with the rest of the file where it was generally assigned to NULL in advance. I can picture some edge case situations like stub code in Wine and such, which might return a fake success response but not do anything with the params (they often have a bunch of stubs for incomplete compatibility layers just to get a portion of apps running at all). So that's why I'd consider it a worthwhile good practice

libass/ass_directwrite.c
p0358: Hmm fair. The fact that it works probably means that IDWriteFontFace keeps references to file alive, but we probably shouldn't rely on that at all.

I think the boolean is cleaner because it saves us from potentially checking values of the other params if it returns true (otherwise someone would look at the invocation and think the coder forgot to check them if it returned just the object instead)

libass/ass_fontselect.c
```diff
@@ -405,10 +405,11 @@ ass_font_provider_add_font(ASS_FontProvider *provider,
                 meta->postscript_name : meta->extended_family;
             face = ass_face_stream(selector->library, selector->ftlibrary,
                                    name, &stream, index);
-        } else {
+        } else if (path) {
             face = ass_face_open(selector->library, selector->ftlibrary,
                                  path, meta->postscript_name, index);
-        }
+        } else
+            goto error;
```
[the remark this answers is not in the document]
p0358: Ah, you're completely right, it was my C++ habits getting in place. After all FT_Face is just a pointer in the end (`typedef struct FT_FaceRec_*  FT_Face;`), so it can be just initialized to NULL (if that was a C++ object with constructor, then that wouldn't work out)

libass/ass_directwrite.c
p0358: This leaves the question on what to do in case of failure in terms of cleanup. I think I will just leave the cleanup to the callers to avoid code duplication there...

p0358:
Okay both changes implemented

libass/ass_directwrite.c
astiob: You _might_ want to explicitly initialize `*file` to `NULL` inside the function for extra clarity/safety, although, of course, we do currently null-initialize it in the callers anyway.

astiob: (This is turning into bikeshedding, and the code is good for merge as it is, but:)

Hm, I don’t think it’s useful to initialize the key reference variables here, nor for that matter the loader, because if the function fails, those aren’t supposed to be used by the caller in any way. But the file specifically, by your contract, must be freed by the caller even if this function fails—so my logic was that it should always have a valid value even if this function fails.

p0358: > But the file specifically, by your contract, must be freed by the caller even if this function fails

Well in case of failure it's only file, but in case of success both file and loader. In the parent funcs we implement it by simply null-checking for both and then freeing them, and in this situation it's probably better to NULL-initialize them both (otherwise we risk someone using the func in some 3rd place down the line, without NULL-setting the input vars, and they'd have a bug unbeknownst to them where they'd if-check for non-NULL-initialized argument and try to Release it, if they copy-pasted the `cleanup` section).

And then when we set 2 of 4 params to null, we may as well just set all 4 for consistency xD I think out of nulling 0/4 or 4/4, nulling only 2/4 would be the least intuitive

p0358:
Fixed typo in WinRT codepath that CI detected (I couldn't test WinRT one due to no proper environment)

libass/ass_directwrite.c
astiob: Uhh, not quite sure I follow: upon success, the loader _is_ filled—regardless of the added null-initialization.

p0358: Yeah, but on early failure it could remain untouched without being set to null in helper func.

Then if some kind of hypothetical new parent func passed arg of `IDWriteFontFileLoader * loader;` without NULL-initializing it itself, and then did `if (loader) IDWriteFontFileLoader_Release(loader);`, then they could potentially be trying to Release uninitialized garbage and get a crash (but they'd miss the bug, because normally the func suceeds).

And with that I believe it's beneficial to also null it in the helper as well, just like you said to do to file, if we're doing that already now anyways

astiob: To me, there’s a clear difference in the API contract:

If the function call fails, it doesn’t produce a `loader`, so it makes no promises about its value, so the caller mustn’t just `if (loader)` unless the caller itself has zero-init it. Just like in our calls to DirectWrite.

Whereas for `file`, this function _requires_ the caller to handle it both on success and on failure, so it is this function’s own duty to ensure that it actually is able to be handled by the caller in every case. It returns a `loader` only upon success, but it returns a `file` _always._

Of course, it can also promise that it always fills the `loader`, as your code does now, and that’s a valid choice—just not one I would personally take in this context. But like I said, this doesn’t even matter all that much.

astiob (approved):
Looks good to me. Thanks!

moi15moi:
I don't know if it really matter, but since Directwrite always give a file path in uppercase, you may want to normalize it like this: https://github.com/arch1t3cht/Aegisub/blob/74ed5a6917c99e99962156eda2788255fde1b5b4/src/font_file_lister_gdi.cpp#L29-L64

p0358:
Hmm, so right now it seems that all the added font files will be opened by FreeType anyways to get extended font info on Windows (unlike fontconfig where n_family should already be populated from its cache). Otherwise I would protest that it's not worth it on a case-insensitive system if it requires CreateFile, which for cold files (not recently opened and in FS cache) could be a very expensive operation.

But since it will still open these files later, it probably shouldn't change all that much in practice in terms of performance impact? That is unless the behavior is changed in the future. But then again the path is only used for logging currently and it doesn't technically break anything on a case-insensitive system in theory.

(I'm so cautious of CreateFile since I know opening 1000 file handles can take 10+ seconds for cold files on NTFS, even if you're on NVMe... now granted most people probably won't have exactly 1000 fonts on Windows desktop probably but still (I have 1000+ on a server though))

p0358:
> If you want to give it a try, I don’t mind waiting. But I also don’t want to leave you hanging, so if you’d prefer that this PR be merged as is (and perhaps add path normalization in a separate PR or not at all), I’m happy to do that, too.

Yeah I'd say let's just merge it for now. Path normalization could land in a separate PR if someone would want to do it (not necessarily me and not necessarily now)

libass/ass_directwrite.c
```diff
@@ -326,6 +326,78 @@ static bool init_font_private_stream(FontPrivate *priv)
...
 
+static char* get_utf8_path(FontPrivate *priv)
+{
+    HRESULT hr = S_OK;
+    IDWriteFontFile *file = NULL;
+    IDWriteFontFileLoader *loader = NULL;
+    IDWriteLocalFontFileLoader *local_loader = NULL;
+    UINT32 n_files = 1;
+    const void *refKey = NULL;
+    UINT32 keySize = 0;
```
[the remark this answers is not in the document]
astiob: Yes. FWIW these specific names also occur in `init_font_private_stream` above, and this style in general is used throughout the file for names directly reflecting Windows API parameter names, but it’s mixed with snake_case for things that don’t.

libass/ass_directwrite.c
```diff
@@ -917,7 +894,10 @@ static void add_font(IDWriteFont *font, IDWriteFontFamily *fontFamily,
 cleanup:
     free(meta.postscript_name);
     free(meta.extended_family);
+    free(font_priv);
```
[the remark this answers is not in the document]
astiob: Oh, good catch. I must’ve missed it when this got added. Why `free` this?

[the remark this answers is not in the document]
p0358: Oh shoot, I messed up some logic here. Okay so the reasoning is that previously either calloc would fail or it'd be assumed that `ass_font_provider_add_font` call always succeeds, no problem there. But after adding:
```cpp
    HRESULT hr = IDWriteFont_CreateFontFace(font_priv->font, &face);
    if (FAILED(hr) || !face)
        goto cleanup;
```
it now can fail here in between. So if we just went to cleanup like before, it'd leak the memory allocated to font_priv.

What I didn't account for is that the cleanup is of course also ran on successful call and that will cause a problem. Good catch for sure, will correct this

libass/ass_directwrite.c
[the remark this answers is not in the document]
p0358: This exactly, I just mimicked it from existing `init_font_private_stream`. Changing it would require changing it throughout the whole file consistently, as we have names like `fontSet`, `faceRef` etc etc.

Generally the majority of things in this file that are `ID*` objects have names in this convention, so it'd require a separate commit to consistentify it if anything, but I think it basically follows the naming convention style of Microsoft docs. (there are some exceptions like `IDWriteFontCollection *system_font_coll`).

I don't think it's necessarily a bad thing either, because it makes it clear what things belong to/originate from WinAPI, and which things are something internal to libass's code.

p0358:
> Edit 24-10-2024:
> Since C++17, you can simply do std::filesystem::canonical(path);
> MSVC implemented it like [this](https://github.com/microsoft/STL/blob/27973ad9e661c0edc8c1d42068a17ed83a67e049/stl/inc/filesystem#L3050-L3101) which is basically the same thing that I provided, so it doesn't seems there is a way to avoid calling [CreateFileW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew).

Fair. Although worth nothing that Microsoft's implementation of std::filesystem::path aren't exactly too performance-conscious. I used `std::filesystem::relative` for game mod's filesystem at one point, it had a big impact.

I had to change
```cpp
auto filepath = std::filesystem::relative(entry.path(), scanpath).lexically_normal();
```
to
```cpp
if (entry.path().native().length() <= scanpath.native().length() + 1) [[unlikely]]
{
    m_logger->error("[RefreshReplacementsCache] invalid path \"{}\"", Util::Narrow(entry.path().native()));
    continue;
}
std::wstring_view filepath{ entry.path().native().c_str() + scanpath.native().length() + 1 }; // no copy
```

And now in the hindsight guess why: https://github.com/microsoft/STL/blob/27973ad9e661c0edc8c1d42068a17ed83a67e049/stl/inc/filesystem#L4225

In Microsoft's implementation, weakly_canonical still attempts to do normal _Canonical first and falls back to weakly canonicalizing a theoretical non-existent path there only as a fallback. (and I should've used ::lexically_relative now that I look into docs again, welp)

And the worst offender was doing `if (cached_path.compare(path) == 0)` for some reason, it caused the game to take 5+ minutes to load instead of say 20 seconds or something like that somehow...

So yeah overall this caused me to be very conscious about filesystem operations in Windows and STL. They can be quite absurdly performance-impactful compared to the Linux equivalents... How much it'd matter in practice given current usage and given theoretical usage if loading fonts by FT for extra metadata was removed: this'd have to be benchmarked first imo, to avoid introducing a potentially big performance regression for little benefit practically.

astiob:
> They can be quite absurdly performance-impactful compared to the Linux equivalents...

Just a passing comment: I believe Linux `stat` is also quite slow, which is why higher-level languages/frameworks try to minimize and cache their `stat` calls. It would make sense if `CreateFile` was comparable to `stat`.

p0358:
- fixed commit style (capped at common max of 72)
- fixed wrong space style around that `*`
- fixed the wrong free, now the two handles are also only transferred at once to font_face once we know we won't be freeing it anymore and nulling local vars, otherwise they'll be freed in cleanup upon failure if needed, I think the logic is nailed now

> Just a passing comment: I believe Linux stat is also quite slow, which is why higher-level languages/frameworks try to minimize and cache their stat calls. It would make sense if CreateFile was comparable to stat.

Yes generally filesystems don't like many operation on small files and there's a reason pack files in programs like games are common, but `CreateFile` can only be worse than `stat`. Well for one it returns a file handle instead of statbuf. And one very particular and concrete reason that's slow on Windows is the prevalence of global filesystem hooks, such as the beloved Windows Defender :) And for example something like PyCharm prompts on first launch to consider adding its python dirs to exclusion list to dramatically speed up operations...

libass/ass_directwrite.c
```diff
@@ -253,79 +253,92 @@ static void init_FallbackLogTextRenderer(FallbackLogTextRenderer *r,
...
+ * A helper function for init_font_private_stream and get_utf8_path.
+ * Remember to release the file and loader after obtaining it, even if the function fails.
+ */
+static bool get_font_file_loader_and_reference_key(
+    FontPrivate *priv,
+    const void **refKey,
+    UINT32* keySize,
+    IDWriteFontFile **file,
+    IDWriteFontFileLoader **loader
+    )
```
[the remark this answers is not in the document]
astiob: That too, but I suspect MrSmile means the way the whole declaration is split across lines. Usually, we do it like this:

```c
static bool get_font_file_loader_and_reference_key(FontPrivate *priv,
                                                   ...,
                                                   IDWriteFontFileLoader **loader)
{
```

But once again, this file already breaks with this tradition for some of its functions, and you’ve merely copied that second style here. It looks like it does this for functions with super long names, where fitting even one parameter on the same line as the function name makes the line arguably too long, e. g.:

https://github.com/libass/libass/blob/6a759836e5e76bb7b69b0ac244eea76b0d290512/libass/ass_directwrite.c#L115-L120

Personally, I don’t have a strong opinion on this, but perhaps try using our usual style, and if it you find the lines to be too long, try shortening the function name? I do feel it’s pretty verbose. You could perhaps do with `get_loader_and_reference` or `get_font_reference` or `get_loader_reference` or `get_font_loader` or maybe even something completely different.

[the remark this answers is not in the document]
p0358: Yeah between lines 96 and 217 all the funcs are in this style and I must have mimicked it from there.

I fixed it up to match up the style. As for names I have a few more ideas, but then again the function is pretty specific and does all these things grabbing these objects at once, so perhaps it's not that bad.

It could've been worse like `get_font_file_and_font_file_loader_and_reference_key` or something xD

But currently it's named for the two things we actually *need* from the function, being the font file loader and reference key. The third item font file we grab just to properly dereference it later, so it's not mentioned as the intermediary there. I think it kinda makes sense.

At most it could be shortened to `get_font_loader_and_reference_key`, since there can't be non-file loader, but then again emphasizing "file" makes sense if one of the funcs has to obtain the file's path. So yeah I'd just keep it...

astiob:
By the way: @p0358, could you elaborate on your use case for collecting system fonts? You want the server to gather all fonts necessary to display a subtitle stream and send them to a client, right? But is there good reason to believe the server has better system fonts than the client does? Would it perhaps suffice—or be even better—to send all fonts attached to the MKV file and then let the client fall back to its own system fonts? Is this meant for clients on platforms without proper font stores or poorly supported by libass? or to allow users to add specific extra fonts on the server in case someone forgets to attach them to the MKV?

p0358:
> But is there good reason to believe the server has better system fonts than the client does?

Yes: that the server admin explicitly installs the required fonts there for this exact purpose while setting up the media. When you have fansub releases of animes, they often some with .ass files and then a folder of subtitles. So just extract all these subtitles to a custom folder, mount that in Docker under appropriate dir, and you're good to go.

Once it's landed, the docs can recommend people to do that more directly.

> Would it perhaps suffice—or be even better—to send all fonts attached to the MKV file and then let the client fall back to its own system fonts? Is this meant for clients on platforms without proper font stores or poorly supported by libass? or to allow users to add specific extra fonts on the server in case someone forgets to attach them to the MKV?

As above, the primary intended purpose is to stream fonts for loose .ass files, without having to bake a new MKV file and without other workarounds. This is the only remaining option and I believe it will be the most ideal for users both for ease of use and performance. Given you cannot install new fonts on any other clients than PC and maybe Android pretty much (web with Subtitle Octopus also has no access to client fonts by default!! the Web API for that isn't out of experimental stage yet). So all these reasons come down together pointing at this approach.

MKV attachments are already covered in Jellyfin and sent out properly. Currently the prototype/mvp PR skips getting subtitles for MKV subtitles, but it might come later, it needs more work on deduplicating fonts with what's already in MKV.

Also, Jellyfin already uses server-side fonts if someone transcodes with subtitle burn-in type of scenario. So installing the fonts on the server is beneficial anyways, the planned feature will make it a consistent single place to install the fonts and have them always work for ASS, hopefully in more clients than just web too. An in a way where I can hopefully have my 1000+ fonts without waiting 10 seconds for subtitles to show up (how I ended up there was that fansub groups were missing fonts in per-series subfolders, so I decided to brute-force it and just extract them all to a single dir xD).

You can read the whole description here for full background: https://github.com/jellyfin/jellyfin/pull/12877

But generally I thought that if we have to parse ASS on server, I may as well try to use the exact same library to aid me for best consistency and maybe least work (okay that latter one was naive)

astiob:
Just FYI, if you’re thinking of a fansub release that comes with its own pack of fonts, ideally you should use those fonts and no others—as opposed to having all fonts from all releases of all shows available together, because other fonts can unintentionally provide closer matches and result in broken rendering. And in that case, you could just send the whole folder attached to the current video without attempting to figure out which fonts are actually referenced and which aren’t.

p0358:
That's a fair point, I was considering it, I know there's a lot of duplicates and sometimes there could be a broken version of some fonts. But on the contrary often the fonts that come with single fansub release are wrong and missing too (missing or extra unused fonts). I've come to find that neither source is ideal.

But hmm, some option of pinning certain fonts for given show might not be a bad idea, at least at some point, if someone is sure they're not bad.

But I think this could be set back to rethink after minimal implementation later to limit the initial score, because I'm afraid that this would require much more debating in Jellyfin, creating some conventions in Server Guide for adding in the library layout fonts like there are for external subtitles or extras for shows etc. And things like whether to use *only* the provided fonts, or treat them with some kind of bigger or smaller precedence, while allowing fallbacks at all or not etc. The normal `fonts_dir` wouldn't be good as it'd require reinitializing everything and throwing away font cache + I think fonts dir is added by default as equal source to font system provider.

But I think for most simple default behavior it then could be something like:
- if no fonts in dir, use system provider as normally
- if there are any fonts, match with just them first
  - if *any* font match for given query is found, use that (even if something else in system store might potentially match it better)
  - if no matches are found at all, fall back to system store lookup

There's no globals as far as I see, so it should be fine to have both global ASS_Library instance for system and then some fonts_dir-only transient ones with system provider disabled for the second point. This could be the best compromise if such idea was to be implemented... Actually now that I think about it, the very similar logic could be used for MKV attachment fonts as well!

libass/ass_directwrite.c
```diff
@@ -253,79 +253,90 @@ static void init_FallbackLogTextRenderer(FallbackLogTextRenderer *r,
...
+ */
+static bool get_font_file_loader_and_reference_key(FontPrivate *priv,
+                                                   const void **refKey,
+                                                   UINT32 *keySize,
+                                                   IDWriteFontFile **file,
+                                                   IDWriteFontFileLoader **loader)
+{
+    HRESULT hr = S_OK;
+    UINT32 n_files = 1;
+    *refKey = NULL, *file = NULL, *loader = NULL, *keySize = 0;
```
MrSmile, line 268: I think it's better to not abuse operator `,` in this context. Also I think it's more logically correct to move this initialization at the very start of the function, as `hr` and `n_files` are directly related to `IDWriteFontFace_GetFiles` call.

MrSmile (approved):

TheOneric:
> Conceptually, I like this as it makes the log more useful, but it does add an extra function to maintain without improving any core functionality, so I’m curious to hear what the other maintainers think.

I’m sceptical about the motivating use case of collecting used font files; see e.g. https://github.com/libass/libass/issues/642 where the pitfalls with this were discussed before and the change is not trivial. On the other hand, this might be useful for debugging incompatible or just in general suboptimal font selections.
But also, I’m not not an expert on Win32 API nor using or working on the DirectWrite font selector and almost all changes 
are within the Windows specific code. In conclusion: I don’t know, but it seems all who actually work with the DirectWrite provide are fine with this change so it seems good?

p0358:
> I’m sceptical about the motivating use case of collecting used font files; see e.g. https://github.com/libass/libass/issues/642 where the pitfalls with this were discussed before and the change is not trivial.

Yeah, I realize the pitfalls and my actual use-case will probably be implemented in a fork, not sure if I'll want to try upstreaming it or not. I avoid some of the pitfalls there, since my use-case isn't permanent/ideal font-picking for the purpose of creating an MKV release. But rather to find a list of fonts available "right here, right now" and show that to the client, instead of falling back to default otherwise. So I don't need idealism, the 99% of good enough is just fine. But I don't necessarily agree that an official API like proposed in that Feature Request even with the pitfalls considered, it could still be useful for example for unit tests where behavior of different font picking programs could be compared etc. But whether it's something worth upstreaming in the end in some form would have to be debated later after I finish up the fork.

But motivation aside, the paths should just come up as useful in logs, since the other OSs also have that.

libass/ass_directwrite.c
```diff
@@ -314,11 +314,13 @@ static bool init_font_private_stream(FontPrivate *priv)
...
     hr = IDWriteFontFileLoader_CreateStreamFromKey(loader, refKey, keySize, &stream);
     if (FAILED(hr) || !stream) {
+        IDWriteFontFileLoader_Release(loader);
         IDWriteFontFile_Release(file);
         return false;
     }
 
     priv->stream = stream;
+    IDWriteFontFileLoader_Release(loader);
     IDWriteFontFile_Release(file);
```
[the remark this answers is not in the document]
p0358: Ah this is an outdated commit (as GitHub says too). In final version there's a single cleanup section at the bottom, after outlining the common code to a separate function (and then they need to be if-checked, since the helper function may not initialize all of them successfully)

libass/ass_directwrite.c
p0358: I could move it to the start. I used the operator `,` because I didn't really want to make it too verbose by occupying four lines and then a newline for better visual clarity... (cause we weren't really sure if that nulling is needed to begin with, it was done just to have the function output behavior be cleaner no matter what comes in the args in.

PR #847, opened 2024-11-05
Consistent C standard and feature flags

TheOneric:
If someone figures out how to sidestep the array type incompatibility in C99 without additional copies, we can also use `c99` everywhere instead.


TheOneric:
Merged in simplified form since the CI setup justifying the complexity didn’t come to be yet, so at least the meson-built CoreText gets unborked

PR #849, opened 2024-11-12
aarch64: Set the right intended alignment for constants

mstorsjo:
The align parameter to these macros is passed to the gas directive .align. This directive is architecture dependent; on some architectures, the given alignment is an alignment in bytes, while on others, its the power of two to align to. On aarch64, .align is for power of two alignment, see [1] for details; i.e. it behaves the same as .p2align.

This means that when the macros are invoked with align=16, we actually requested 64 KB alignment, not 16 byte alignment. Fix this, setting the alignment to the intended 16 byte alignment.

This fixes building for aarch64-windows targets with assert enabled versions of Clang; such versions of Clang error out with "unsupported section alignment" on this case, while regular release versions of Clang silently let this issue pass.

[1] https://sourceware.org/binutils/docs/as/Align.html

### DIFF
--- a/libass/aarch64/blend_bitmaps.S
+++ b/libass/aarch64/blend_bitmaps.S
@@ -18,7 +18,7 @@
 
 #include "asm.S"
 
-const edge_mask, align=16
+const edge_mask, align=4
     .dcb.b 16, 0xFF
     .dcb.b 16, 0x00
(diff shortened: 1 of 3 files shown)


TheOneric:
Thanks!

PR #850, opened 2024-11-21
various: use ass_lrint for rounding consistently

kasper93:


libass/ass_blur.c
```diff
@@ -162,7 +162,7 @@ static void find_best_method(BlurMethod *blur, double r2)
         calc_coeff(mu, blur->radius, r2, mul);
     }
     for (int i = 0; i < blur->radius; i++)
-        blur->coeff[i] = (int) (0x10000 * mu[i] + 0.5);
+        blur->coeff[i] = ass_lrint(0x10000 * mu[i]);
```
astiob, line 165: cc @MrSmile

(btw would I be right to assume `mu[i]` are always non-negative?)

libass/ass_fontconfig.c
```diff
@@ -138,10 +138,10 @@ static bool scan_fonts(FcConfig *config, ASS_FontProvider *provider)
         // from typographical weight. we're using truetype weights, so
         // convert appropriately.
 #if FC_VERSION >= 21292
-        meta.weight = FcWeightToOpenTypeDouble(weight) + 0.5;
+        meta.weight = ass_lrint(FcWeightToOpenTypeDouble(weight));
 #elif FC_VERSION >= 21191
         // Versions prior to 2.12.92 only had integer precision.
-        meta.weight = FcWeightToOpenType(weight + 0.5) + 0.5;
+        meta.weight = FcWeightToOpenType(ass_lrint(weight));
```
astiob, line 144: cc @rcombs

This is probably fine?

astiob (commented):
`lrint` behaves differently from `lround` and `(int) (x + 0.5)`: it rounds half to even, whereas the latter round half up (assuming positive arguments). Half-to-even is not desirable everywhere. In at least some of these cases, the choice is intentional:

For example, an upscaling factor of 1.5 is common (e. g. 720p video on 1080p screen), and for whole-pixel rounding in clipping during animations or within gradients, the rounding produces a staircase effect, and I think it’s more desirable to have the steps alternate 1-2-1-2-1-2-1-2 (`lround`) than 1-2-2-1-1-2-2-1 (`lrint`), which looks less uniform. The same argument can apply to karaoke, although exact half-integer steps seem less likely there.

kasper93:
> `lrint` behaves differently from `lround` and `(int) (x + 0.5)`: it rounds half to even, whereas the latter round half up (assuming positive arguments). Half-to-even is not desirable everywhere. In at least some of these cases, the choice is intentional:

I understand the difference. I wanted to start discussion about it.

> For example, an upscaling factor of 1.5 is common (e. g. 720p video on 1080p screen), and for whole-pixel rounding in clipping during animations or within gradients, the rounding produces a staircase effect, and I think it’s more desirable to have the steps alternate 1-2-1-2-1-2-1-2 (lround) than 1-2-2-1-1-2-2-1 (lrint), which looks less uniform. The same argument can apply to karaoke, although exact half-integer steps seem less likely there.

This is good argument and this is precisely why I wanted to discuss this topic. Because it looks like currently libass is not consistent when rounding. If believe that it is more uniform to use `lround`, why is `lrint` used in some places that are equally affected by this logic?

astiob:
We tend to use `lrint` for intermediate, high(er)-precision (fixed-point 26.6) computation results, where it’s presumed desirable for the same error averaging properties that make it the usual choice in floating point.

For more visible-to-the-naked-eye whole-pixel rounding, I have personally, at least, tried to always use `lround`. It’s still plausible that we may have some inconsistency, so if you spot specific examples of where my rationale seems to apply but we’re currently using `lrint`, I’d appreciate direct links.

I imagine that it may often not be clear whether either option is better than the other, so we’re just using whatever the author of a patch preferred at the time.

I initially assumed that this PR was motivated by performance concerns. Is this correct?

> Also note that lrint rounding depends on current rounding mode

I think I’ve brought this up before and the other maintainers said something to the effect that our entire code (e. g. our floating-point operations) essentially assumes half-to-even rounding and that this is standard practice.

libass/ass_blur.c
MrSmile: Yes, it's explicitly enforced [here](https://github.com/libass/libass/blob/master/libass/ass_blur.c#L139). I've used `(int)` as it generates optimal code more consistently than `lrint`, but it probably doesn't matter much here.

libass/ass_render.c
```diff
@@ -1020,14 +1020,14 @@ ASS_Vector ass_layout_res(ASS_Renderer *render_priv)
...
-            FFMAX(1, lround(track->PlayResY * render_priv->frame_content_width
+            FFMAX(1, ass_lrint(track->PlayResY * render_priv->frame_content_width
                     / render_priv->frame_content_height / settings->par)),
             track->PlayResY
         };
     else
         return (ASS_Vector) {
             track->PlayResX,
-            FFMAX(1, lround(track->PlayResX * render_priv->frame_content_height
+            FFMAX(1, ass_lrint(track->PlayResX * render_priv->frame_content_height
```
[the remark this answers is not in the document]
TheOneric: VSFilter always has `layout_res == frame_res` and thus doesn’t need a fallback, xy-SubFilter afaik requires `original_video_size` to be set *(though my knowledge about non-CSRI interfaces is quite small)* thus also no need and therefore lack of such code.  
While we don’t require it to render anything at all, we always required layout res to be set for _correct_ rendering. I think/hope we’re free to tweak the fallback calculation relatively freely.

However, VSFilter/xy_SubFilter’s `layout_res` is a `CSize` which is a pair of integers, thus later scaling operations might rely on integer arithmetic and diverge if we were to change ours to floating point. While we can tweak the roundng, I think we should keep the final result as integers. “Closest aspect ratio” sounds good to me *(provided we exclude `0` from “closest” match for extreme ratios)*

PR #852, opened 2024-11-29
ass.h: clarify comments on struct ass_image

rcombs:
The existing comment described the bitmap as being "1bpp", which is commonly understood to mean "1 bit per pixel", when in reality the bitmaps are 1 *byte* per pixel (8 bits).

The description of the bitmap's size was also somewhat unclear.

### DIFF
--- a/libass/ass.h
+++ b/libass/ass.h
@@ -62,9 +62,12 @@ extern "C" {
 typedef struct ass_image {
     int w, h;                   // Bitmap width/height
     int stride;                 // Bitmap stride
-    unsigned char *bitmap;      // 1bpp stride*h alpha buffer
+    unsigned char *bitmap;      // 1-byte-per-pixel alpha buffer
                                 // Note: the last row may not be padded to
                                 // bitmap stride!
+                                // The guaranteed allocated size is
+                                // `(stride * (h - 1)) + w`, and bytes past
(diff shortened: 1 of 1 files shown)


libass/ass.h
```diff
@@ -62,9 +62,11 @@ extern "C" {
 typedef struct ass_image {
     int w, h;                   // Bitmap width/height
     int stride;                 // Bitmap stride
-    unsigned char *bitmap;      // 1bpp stride*h alpha buffer
+    unsigned char *bitmap;      // 1-byte/pixel alpha buffer
```
astiob, line 65: `byte-per-pixel` would be more unambiguously fool-proof (this can be parsed as “byte or pixel” and leave the reader scratching their head)

libass/ass.h
```diff
@@ -62,9 +62,11 @@ extern "C" {
...
                                 // Note: the last row may not be padded to
                                 // bitmap stride!
+                                // The guaranteed allocated size is
+                                // `(stride * (h - 1)) + w`
```
TheOneric, line 69: Perhaps also mention anything outside the actual `w`×`h` area may be uninitialised even if allocated?

TheOneric (approved):

astiob (approved):

PR #853, opened 2024-12-08
ci/gha/meson: build both static and dynamic lib on Windows

TheOneric:
As brought up in the threading PR #793 

@astiob, do you know which flags we need with MSVC to enable similar errors as for GCC-like compilers?

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -22,7 +22,7 @@ jobs:
             msystem: MINGW64,
             shell: 'powershell',
             buildtype: release,
-            args: '-Ddefault_library=static'
+            args: '-Ddefault_library=both'
           }
           - {
             name: Windows MinGW Release,
@@ -32,7 +32,7 @@ jobs:
             msystem: MINGW64,
             shell: 'msys2 {0}',
             buildtype: debugoptimized,
-            args: ''
+            args: '-Ddefault_library=both'
           }
           - {
             name: Ubuntu Debug,
--- a/meson.build
+++ b/meson.build
@@ -46,6 +46,10 @@ if cc.get_argument_syntax() == 'gcc'
 elif cc.get_id() == 'msvc'
     cc_warnings += [
         '/we4013', # implicit function definition
+        '/we4775', # non-standard format strings
+        '/we4317', # not enough arguments for format string
+        '/we4473', # same as above
+        '/we4474', # too many arguments for format string
     ]
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -43,6 +43,10 @@ if cc.get_argument_syntax() == 'gcc'
         '-Wformat-non-iso',
         '-Werror=format-non-iso',
     ]
+elif cc.get_id() == 'msvc'
+    cc_warnings += [
+        '/we4013', # implicit function definition
+    ]
 endif
 
 add_project_arguments(


TheOneric:
i suspect the linking error in the MSVC job might result from the pacman dep reuse introduced in #774, cc @rcombs

rcombs:
Looks like the gcc-built harfbuzz C++ objects aren't compatible with the MSVC linker, yeah. I think we have 3 reasonable options:
1. Link against dynamic libs instead of static (assuming the pacman packages provide them?)
2. Go back to using wrapdb, but cache the output
3. Get the deps from choco or some other package manager that provides MSVC-built versions of them

astiob:
> do you know which flags we need with MSVC to enable similar errors as for GCC-like compilers?

We can try [setting the C standard version](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170) if we don’t do it yet: `/std:c17` or `/std:c11`. The docs say the only real difference is the value of `__STDC_VERSION__`. No, stdatomic isn’t present in either.

If that doesn’t help, there’s an option to [treat a warning as an error](https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170): `/we4013`

TheOneric:
> We can try [setting the C standard version](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170) if we don’t do it yet: /std:c17 or /std:c11

Hmm, neither matches our current standard target C99, but the linked document states C99 was omitted on purpose and i guess since we don't/can't set C99 whatever we get atm differs too.
Currently meson only maps `cXX` to `/std:cXX` (if existing) though and just ignores `gnuXX`. Any chance specifying `/std:cXX` disables some Microsoft extensions we’d like to keep around?

> If that doesn’t help, there’s an option to [treat a warning as an error](https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170): /we4013

This perfectly matches our `-Werror-` for implicit function definition in GCC, great. The only other warning-as-error we use atm is clang’s error on non-ISO format strings; but I suspect MSVC doesn't have an equivalent for this?

astiob:
> Any chance specifying `/std:cXX` disables some Microsoft extensions we’d like to keep around?

Doesn’t seem like it from the documentations I saw, but I haven’t tested this. (There’s a separate flag for disabling Microsoft extensions, although its description is quite unclear as to what exactly it disables.)

astiob:
> The only other warning-as-error we use atm is clang’s error on non-ISO format strings; but I suspect MSVC doesn't have an equivalent for this?

Might be 4775, see below. It appears to have a whole bunch of potentially useful warnings for format strings. I can’t tell the difference between some of them, but here’s the full list of warnings [in Microsoft’s list](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-c4000-c5999?view=msvc-170) that mention “format” and seem to refer to format strings:

Enabled by default (`/W1`):
* [4313: '<var>function</var>' : '%<var>format_specifier</var>' in format string conflicts with argument <var>argument_number</var> of type '<var>type</var>'](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4313?view=msvc-170)
* 4317: '<var>printf_family</var>' : not enough arguments passed for format string
* [4473: '<var>function</var>' : not enough arguments passed for format string](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c4473?view=msvc-170)
* [4477: '<var>function</var>' : format string '<var>string</var>' requires an argument of type '<var>type</var>', but variadic argument <var>number</var> has type '<var>type</var>'](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c4477?view=msvc-170)
* 4478:	'<var>function</var>' : positional and non-positional placeholders cannot be mixed in the same format string
* 4776:	'%<var>character</var>' is not allowed in the format string of function '<var>function</var>'

Disabled by default (why, lol) (normally included in `/W3`, but I suspect that turning them individually into errors also force-enables them; if not, they can be individually moved to the enabled-by-default warning level 1 using `/w1XXXX`, e. g. `/w14474`):
* 4474: '<var>function</var>' : too many arguments passed for format string
* 4475: '<var>function</var>' : length modifier '<var>modifier</var>' cannot be used with type field character '<var>character</var>' in format specifier
* 4476: '<var>function</var>' : unknown type field character '<var>character</var>' in format specifier
* 4775:	nonstandard extension used in format string '<var>string</var>' of function '<var>function</var>'
* 4778:	'<var>function</var>' : unterminated format string '<var>string</var>'

Disabled by default (same deal as before but this time `/W4`):
* 4774: '<var>string</var>' : format string expected in argument <var>number</var> is not a string literal
* 4777: '<var>function</var>' : format string '<var>string</var>' requires an argument of type '<var>type1</var>', but variadic argument number has type '<var>type2</var>'

TheOneric:
Thanks for the list of format string warnings! I picked out some which seemed fairly safe to promote as errors and it doesn't seem to cause any issues in CI

PR #854, opened 2024-12-08
fuzz: optionally reject too large files and OSS-Fuzz observations

TheOneric:
In https://github.com/google/oss-fuzz/pull/12687 we’ve found out OSS_Fuzz currently reports pretty bad coverage for our corpus appearing to not even covering some basic override tags listed in the dictionary.  

To speed up the fuzzer and avoid the unhelpful time-out reports popping up occasionally (which due to CIFuzz deciding to abort on the first time out, also make the former unusable for us), just setting a size limit for testcases seems like the easiest option (apart from figuring out a good cutoff). However, while AFL and libFuzzer support CLI flags to limit the size of generated test cases, OSS Fuzz supports more engines and it’s `option` file apparently also only contains engine-specific settings  meaning we’d have to set it for all current and future engines.  
[OSS-Fuzz docs recommend](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#input-size) to just return early on too large input and libFuzzer documents a return value of `-1` to prevent addition to the corpus.

Even with this setting added here though, OSS-Fuzz won't use it until someone who signed Google’s CLA sends a patch.

---

As it turns out though, there is more to this than just time-out induced inefficiency:

Downloading the current public corpus, a simple `grep` shows purportedly non-covered tags being present inside override blocks.
Creating coverage reports manually shows _all_ override tags are in fact covered! *(i did not test with OSS-Fuzz infra)*

OSS Fuzz is suffering from a missing fontconfig config file (and presumably usable fonts), which we should also address, but this shouldn’t affect early parsing stages i think?

---

Analysing the current corpus, file sizes are distributed approximately like this *(approximated via kernel density estimation to reduce noise; logarithmic size scale)*:
2024-12-08_libass_public-corpus__file-size-dist-kde

Some (exact) reference points:
- the minimum file size is $1$ byte
- the maximum file size is $1,047,597$ byte
- 99.0% are <= $2^{16}$ bytes
- 98.4% are <= $2^{14}$
- 97.0% are <= $2^{12}$
- 92.4% are <= $2^{10}$
- 81.0% are <= $2^{8}$
- 45.7% are <= $2^{6}$

The currently open time-out report’s testcase takes up 384K

Introducing a limit affects the total coverage like this:

|  Limit  |Regions  |Missed Regions  |Cover  |Functions |  Missed Functions  |Executed  |Lines  |Missed Lines  |Cover  |Branches  |Missed Branches  |
|--------|--------|--------|--------|--------|--------|--------|--------|--------|--------|--------|--------|
|  2^04  |9183  |8492  |7.52%  |529  |451  |14.74%  |10940  |9934  |9.20%  |5956  |5645  |
|  2^05  |9183  |7728  |15.84%  |529  |423  |20.04%  |10940  |9262  |15.34%  |5956  |5251  |
|  2^06  |9183  |3789  |58.74%  |529  |163  |69.19%  |10940  |4281  |60.87%  |5956  |2889  |
|  2^07  |9183  |3492  |61.97%  |529  |152  |71.27%  |10940  |3921  |64.16%  |5956  |2616  |
|  2^08  |9183  |3345  |63.57%  |529  |145  |72.59%  |10940  |3715  |66.04%  |5956  |2514  |
|  2^09  |9183  |3288  |64.19%  |529  |144  |72.78%  |10940  |3664  |66.51%  |5956  |2461  |
|  2^10  |9183  |3183  |65.34%  |529  |143  |72.97%  |10940  |3551  |67.54%  |5956  |2385  |
|  2^12  |9183  |3165  |65.53%  |529  |143  |72.97%  |10940  |3530  |67.73%  |5956  |2364  |
|  2^13  |9183  |2981  |67.54%  |529  |139  |73.72%  |10940  |3207  |70.69%  |5956  |2231  |
|  2^14  |9183  |2981  |67.54%  |529  |139  |73.72%  |10940  |3209  |70.67%  |5956  |2227  |
|  2^15  |9183  |2980  |67.55%  |529  |139  |73.72%  |10940  |3208  |70.68%  |5956  |2223  |
|  2^16  |9183  |2980  |67.55%  |529  |139  |73.72%  |10940  |3208  |70.68%  |5956  |2224  |
|  none  |9183  |2958  |67.79%  |529  |139  |73.72%  |10940  |3203  |70.72%  |5954  |2214  |

Up to $2^{13}$ coverage barely decreases making it seem pretty safe as a cutoff. 

If someone wants to investigate which parts loose coverage above/below $2^{13}$ I can share the full line-by-line coverage report; perhaps we could go even lower then if we manually add a few files to fill in the missing spots or just wait for the fuzzer to rediscover them.

This will obsolete 220 existing corpus entries; i guess we’ll need to ask OSS-Fuzz maintainers to drop them.

### DIFF
--- a/fuzz/fuzz.c
+++ b/fuzz/fuzz.c
@@ -34,6 +34,15 @@
     #define ASS_FUZZMODE FUZZMODE_STANDALONE
 #endif
 
+// Limit maximal processed size in continuous fuzzing;
+// to be used when the fuzzing setup doesn't expose its own max size control.
+// Has no effect in standalone builds
+#ifdef ASSFUZZ_MAX_LEN
+    #define LEN_IN_RANGE(s) ((s) <= (ASSFUZZ_MAX_LEN))
+#else
+    #define LEN_IN_RANGE(s) true
+#endif
+
 // MSAN: will trigger MSAN if any pixel in bitmap not written to (costly)
 #ifndef ASSFUZZ_HASH_WHOLEBITMAP
     #define ASSFUZZ_HASH_WHOLEBITMAP 0
@@ -347,6 +356,8 @@ int main(int argc, char *argv[])
     buf = __AFL_FUZZ_TESTCASE_BUF;
     while (__AFL_LOOP(100000)) {
         len = __AFL_FUZZ_TESTCASE_LEN;
+        if (!LEN_IN_RANGE(len))
+            continue;
 
         if (!init_renderer()) {
             printf("Failing renderer init, skipping a sample!\n");
@@ -372,9 +383,14 @@ int main(int argc, char *argv[])
 #elif ASS_FUZZMODE == FUZZMODE_LIBFUZZER
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
+    // OSS Fuzz docs recommend just returning 0 on too large input
+    // But libFuzzer docs tell us to use -1 to prevent addition to corpus
+    if (!LEN_IN_RANGE(size))
+        return -1;
+
     ASS_Track *track = NULL;
 
-    // All return values but zero are reserved
+    // All return values but zero and -1 are reserved
     if (!init())
         return 0;
 


rcombs (approved):

fuzz/fuzz.c
```diff
@@ -346,7 +355,9 @@ int main(int argc, char *argv[])
     __AFL_INIT();
     buf = __AFL_FUZZ_TESTCASE_BUF;
     while (__AFL_LOOP(100000)) {
-        len = __AFL_FUZZ_TESTCASE_LEN;
+        len = LIMIT_LEN(__AFL_FUZZ_TESTCASE_LEN);
```
astiob, line 358: Is this macro defined somewhere in fuzzer headers? The other macros seem to start with double underscores, so this seems suspicious.

TheOneric: It’s not defined anywhere, sorry. This is a leftover from an earlier draft of the patch which cut off excess bytes instead of skipping the testcase; apparently I forgot to remove it here and didn't build in AFL mode since.

astiob (approved):

PR #855, opened 2024-12-09
Tests in meson

rcombs:
Adds tests for meson builds in CI.

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -155,7 +155,7 @@ jobs:
       - name: configure
         run: |
           muon version
-          muon setup -Dtest=true build
+          muon setup -Dtest=enabled build
 
       - name: build
         run: |
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -51,10 +51,19 @@ jobs:
       run:
         shell: ${{ matrix.config.shell || 'bash' }}
 
+    env:
+      ART_SAMPLES: ${{ github.workspace }}/ext_art-samples
+
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: '0'
+
+      - name: download test samples
+        uses: actions/checkout@v4
+        with:
+          repository: libass/libass-tests
+          path: ${{ env.ART_SAMPLES }}
 
       - name: Setup MSys2
         if: startsWith(matrix.config.os, 'windows')
@@ -124,14 +133,25 @@ jobs:
           meson wrap update-db
 
       - name: Configure
-        run: meson setup build ${{ matrix.config.args }} -Dbuildtype=${{ matrix.config.buildtype }}
+        run: meson setup build ${{ matrix.config.args }} -Dbuildtype=${{ matrix.config.buildtype }} -Dtest=enabled -Dcompare=enabled -Dprofile=enabled -Dfuzz=enabled '-Dart-samples=${{ env.ART_SAMPLES }}'
 
       - name: Build
         run: meson compile -C build
 
+      - name: Run tests
+        id: run-tests
+        if: matrix.config.skip_tests != 'yes'
+        env:
+          ART_REG_SKIP: ${{ matrix.config.art_reg_skip }}
+        run: meson test -C build
+
       - name: Dump Meson Debug Info
         if: failure()
         run: cat build/meson-logs/meson-log.txt
+
+      - name: Dump Meson Test Debug Info
+        if: ${{ failure() && steps.run-tests.conclusion == 'failure' }}
+        run: cat build/meson-logs/testlog.txt
 
   build-muon:
     runs-on: ubuntu-latest
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -86,6 +86,13 @@ jobs:
               $pre-fribidi $pre-freetype $pre-harfbuzz \
               $pre-fontconfig $pre-libpng
             echo "PKG_CONFIG_PATH=$(cygpath -w "$PWD/pacman-root/mingw64/lib/pkgconfig")" >> "$GITHUB_ENV"
+            cygpath -w "$PWD/pacman-root/mingw64/bin" >> "$GITHUB_PATH"
+            cp "$PWD/pacman-root/mingw64/lib/libharfbuzz.dll.a" "$PWD/pacman-root/mingw64/lib/libharfbuzz.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libfreetype.dll.a" "$PWD/pacman-root/mingw64/lib/libfreetype.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libfribidi.dll.a" "$PWD/pacman-root/mingw64/lib/libfribidi.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libfontconfig.dll.a" "$PWD/pacman-root/mingw64/lib/libfontconfig.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libpng.dll.a" "$PWD/pacman-root/mingw64/lib/libpng.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libpng16.dll.a" "$PWD/pacman-root/mingw64/lib/libpng16.lib"
 
       - name: Install dependencies (not MSVC)
         if: ${{ !matrix.config.msvc }}
--- a/checkasm/meson.build
+++ b/checkasm/meson.build
@@ -1,0 +1,34 @@
+checkasm_src = files(
+    'checkasm.c',
+    'rasterizer.c',
+    'blend_bitmaps.c',
+    'be_blur.c',
+    'blur.c',
+)
+
+checkasm_src_x86 = files(
+    'x86/checkasm.asm',
+)
+checkasm_src_aarch64 = files(
+    'arm/checkasm_64.S',
+)
+
+checkasm_asm_sources = []
+if generic_cpu_family == 'x86'
+    checkasm_src += checkasm_src_x86
+elif generic_cpu_family == 'aarch64'
+    checkasm_src += checkasm_src_aarch64
+endif
+
+libass_checkasm = executable(
+    'checkasm',
+    checkasm_src + config_h,
+    install: false,
+    include_directories: incs,
+    dependencies: deps,
+    link_with: libass_for_tools,
+    c_args: asm_args,
+    build_by_default: false,
+)
+
+test('checkasm', libass_checkasm)
--- a/compare/compare.c
+++ b/compare/compare.c
@@ -18,9 +18,9 @@
 
 #include "image.h"
 #include "../libass/ass.h"
+#include "../libass/ass_filesystem.h"
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <dirent.h>
 #include <string.h>
 
 #if defined(_WIN32) && !defined(__CYGWIN__)
@@ -485,14 +485,14 @@ static bool add_img_item(ItemList *list, const char *dir, const char *file, size
 
 static bool process_input(ItemList *list, const char *path, ASS_Library *lib)
 {
-    DIR *dir = opendir(path);
-    if (!dir) {
+    ASS_Dir dir;
+    if (!ass_open_dir(&dir, path)) {
         printf("Cannot open input directory '%s'!\n", path);
         return false;
     }
-    struct dirent *file;
-    while ((file = readdir(dir))) {
-        const char *name = file->d_name;
+
+    const char *name;
+    while ((name = ass_read_dir(&dir))) {
         if (name[0] == '.')
             continue;
         const char *ext = strrchr(name + 1, '.');
@@ -527,10 +527,10 @@ static bool process_input(ItemList *list, const char *path, ASS_Library *lib)
         } else {
             continue;
         }
-        closedir(dir);
+        ass_close_dir(&dir);
         return false;
     }
-    closedir(dir);
+    ass_close_dir(&dir);
     return true;
 }
 
--- a/compare/image.c
+++ b/compare/image.c
@@ -25,6 +25,15 @@ static inline bool is_little_endian(void)
     return *(char *) &(uint16_t) {1};
 }
 
+static void read_callback(png_structp png_ptr, png_bytep data, size_t length)
+{
+    if (png_ptr == NULL)
+        return;
+
+    if (fread(data, 1, length, png_get_io_ptr(png_ptr)) < length)
+        png_error(png_ptr, "Read Error");
+}
+
 bool read_png(const char *path, Image16 *img)
 {
     FILE *fp = fopen(path, "rb");
@@ -55,6 +64,11 @@ bool read_png(const char *path, Image16 *img)
         return false;
     }
 
+    // Explicitly set a read callback instead of using the default.
+    // This ensures all accesses to the FILE* happen within the program that created it,
+    // rather than in the libpng library, which may be linked against a different libc
+    // (particularly on win32).
+    png_set_read_fn(png, fp, &read_callback);
     png_init_io(png, fp);
     png_read_info(png, info);
 
--- a/compare/meson.build
+++ b/compare/meson.build
@@ -1,0 +1,19 @@
+compare_src = files(
+    'compare.c',
+    'image.c',
+)
+
+libass_compare = executable(
+    'compare',
+    compare_src,
+    install: false,
+    include_directories: incs,
+    dependencies: deps + png_deps,
+    link_with: libass_for_tools,
+)
+
+art_samples = get_option('art-samples')
+if art_samples != ''
+    dir = join_paths(art_samples, 'regression')
+    test('art-compare', files(join_paths(dir, 'run-all.sh'))[0], workdir: dir, args: [libass_compare.full_path()], depends: [libass_compare])
+endif
--- a/fuzz/fuzz.c
+++ b/fuzz/fuzz.c
@@ -23,7 +23,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef _WIN32
+#include <io.h>
+#else
 #include <unistd.h>
+#endif
 #include "ass.h"
 #include "ass_types.h"
 
@@ -173,6 +177,11 @@ struct settings {
     const char *output; // path or NULL for tmp file
 };
 
+#ifdef _WIN32
+#define READ_RET int
+#else
+#define READ_RET ssize_t
+#endif
 
 static ASS_Track *read_track_from_stdin(void)
 {
@@ -181,11 +190,11 @@ static ASS_Track *read_track_from_stdin(void)
     if (!buf)
         goto error;
     size_t s = 0;
-    ssize_t read_b = 0;
+    READ_RET read_b = 0;
     do {
         // AFL++ docs recommend using raw file descriptors
         // to avoid buffering issues with stdin
-        read_b = read(STDIN_FILENO, buf + s, smax - s);
+        read_b = read(fileno(stdin), buf + s, smax - s);
         s += read_b > 0 ? read_b : 0;
         if (s == smax) {
             size_t new_smax = smax > SIZE_MAX / 2 ? SIZE_MAX : smax * 2;
--- a/fuzz/writeout.c
+++ b/fuzz/writeout.c
@@ -23,6 +23,12 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#ifdef _WIN32
+#include <io.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#endif
+
 #include "writeout.h"
 
 
@@ -177,7 +183,12 @@ void write_out_track(ASS_Track *track, const char *outpath)
         printf("Parsed File will be written to:  %s\n", outpath);
     } else {
         char filename[] = "/tmp/parsedSubs_XXXXXX";
+#ifdef _WIN32
+        _mktemp(filename);
+        int fd = _open(filename, O_CREAT | _O_EXCL | _O_WRONLY, _S_IREAD | _S_IWRITE);
+#else
         int fd = mkstemp(filename);
+#endif
         if (fd == -1) {
             printf("Failed to acquire temporary file!\n");
             return;
--- a/meson.build
+++ b/meson.build
@@ -14,7 +14,7 @@ project(
 
 conf = configuration_data()
 deps = []
-test_deps = []
+png_deps = []
 
 host_system = host_machine.system()
 
@@ -120,8 +120,12 @@ if libunibreak_dep.found()
     conf.set('CONFIG_UNIBREAK', 1)
 endif
 
-if get_option('test')
-    test_deps += dependency('libpng', version: '>= 1.2.0')
+png_dep = dependency(
+    'libpng',
+    version: '>= 1.2.0',
+    required: get_option('test').enabled() or get_option('compare').enabled())
+if png_dep.found()
+    png_deps += png_dep
     conf.set('CONFIG_LIBPNG', 1)
 endif
 
@@ -373,11 +377,20 @@ else
     libass_for_tools = libass
 endif
 
-if get_option('test')
+if get_option('test').enabled()
     subdir('test')
 endif
-if get_option('profile')
+if get_option('compare').require(default_library != 'shared').enabled()
+    subdir('compare')
+endif
+if get_option('profile').enabled()
     subdir('profile')
+endif
+if get_option('fuzz').enabled()
+    subdir('fuzz')
+endif
+if get_option('checkasm').require(enable_asm and default_library != 'shared').allowed()
+    subdir('checkasm')
 endif
 
 # libass.pc
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,6 +18,8 @@ EXTRA_DIST = libass.pc.in Changelog MAINTAINERS RELEASEVERSION ltnasm.sh
 # Meson build setup
 EXTRA_DIST += gen_defs.py meson_options.txt meson.build \
               libass/meson.build libass/ass/meson.build \
+              fuzz/meson.build checkasm/meson.build \
+              compare/meson.build \
               profile/meson.build test/meson.build
 
 pkgconfigdir = $(libdir)/pkgconfig
--- a/Makefile_util.am
+++ b/Makefile_util.am
@@ -16,7 +16,7 @@ if ENABLE_COMPARE
 noinst_PROGRAMS += compare/compare
 endif
 compare_compare_SOURCES = compare/image.h  compare/image.c  compare/compare.c
-compare_compare_LDADD = libass/libass.la
+compare_compare_LDADD = libass/libass_internal.la
 compare_compare_LDFLAGS = $(AM_LDFLAGS) $(LIBPNG_LIBS) -static
 EXTRA_DIST += compare/README.md
 
--- a/checkasm/checkasm.h
+++ b/checkasm/checkasm.h
@@ -35,6 +35,10 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+#ifdef _WIN32
+#include <windows.h>
+#endif
+
 #if !ARCH_X86_64 || !defined(_WIN32)
 #include <setjmp.h>
 #define checkasm_context jmp_buf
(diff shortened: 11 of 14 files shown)


.github/workflows/meson.yml
```diff
@@ -86,6 +95,13 @@ jobs:
               $pre-fribidi $pre-freetype $pre-harfbuzz \
               $pre-fontconfig $pre-libpng
             echo "PKG_CONFIG_PATH=$(cygpath -w "$PWD/pacman-root/mingw64/lib/pkgconfig")" >> "$GITHUB_ENV"
+            cygpath -w "$PWD/pacman-root/mingw64/bin" >> "$GITHUB_PATH"
+            cp "$PWD/pacman-root/mingw64/lib/libharfbuzz.dll.a" "$PWD/pacman-root/mingw64/lib/libharfbuzz.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libfreetype.dll.a" "$PWD/pacman-root/mingw64/lib/libfreetype.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libfribidi.dll.a" "$PWD/pacman-root/mingw64/lib/libfribidi.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libfontconfig.dll.a" "$PWD/pacman-root/mingw64/lib/libfontconfig.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libpng.dll.a" "$PWD/pacman-root/mingw64/lib/libpng.lib"
+            cp "$PWD/pacman-root/mingw64/lib/libpng16.dll.a" "$PWD/pacman-root/mingw64/lib/libpng16.lib"
```
kasper93, line 104: Mixing mingw build with msvc one defeats the purpose, no? (and is generally unsafe, if public headers are not carefully vetted to work) All those libraries are available in meson's wrap db which make it very easy to build when needed.

compare/meson.build
```diff
@@ -0,0 +1,19 @@
...
+    install: false,
+    include_directories: incs,
+    dependencies: deps + test_deps,
+    link_with: libass_for_tools,
+)
+
+art_samples = get_option('art-samples')
+if art_samples != ''
+    dir = join_paths(art_samples, 'regression')
+    test('art-compare', files(join_paths(dir, 'run-all.sh'))[0], workdir: dir, args: [libass_compare.full_path()], depends: [libass_compare])
```
kasper93, line 18: I would suggest using native binaries to run tests. It is more portable. As it is defined currently, meson doesn't know that the binary is executed, only path is passed to the script. Therefore it won't add varaibles to environment like PATH, ASAN_OPTIONS and most imprtantly when cross compiling it wouldn't know to use exe_wrapper (like wine or qemu) if needed. All this makes running tests portable a nightmare, while it would be already handled by Meson.

The easiest and canonical way would be to loop the files in meson and declare test for each of them, instead of using `run_all.sh` helper.

fuzz/fuzz.c
```diff
@@ -164,6 +168,11 @@ struct settings {
     const char *output; // path or NULL for tmp file
 };
 
+#ifdef _WIN32
+#define READ_RET int
+#else
+#define READ_RET ssize_t
+#endif
```
kasper93, line 175: Could it be in some common header? Also `<io.h>` and `<unistd.h>` could be included from common header.

Also, that's not correct for mingw build. This should be done only for MSVC. And using larger type is fine, makes easier for downstream compatibility.

```suggestion
#ifdef _MSC_VER
typedef long long ssize_t;
#endif
```

fuzz/fuzz.c
```diff
@@ -172,11 +181,11 @@ static ASS_Track *read_track_from_stdin(void)
     if (!buf)
         goto error;
     size_t s = 0;
-    ssize_t read_b = 0;
+    READ_RET read_b = 0;
     do {
         // AFL++ docs recommend using raw file descriptors
         // to avoid buffering issues with stdin
-        read_b = read(STDIN_FILENO, buf + s, smax - s);
+        read_b = read(fileno(stdin), buf + s, smax - s);
```
kasper93, line 188: You can easily defined those helper nacros in MSVC case in some common header.

``` c
#ifdef _MSC_VER
#ifndef STDIN_FILENO
#define STDIN_FILENO 0
#endif
#ifndef STDOUT_FILENO
#define STDOUT_FILENO 1
#endif
#ifndef STDERR_FILENO
#define STDERR_FILENO 2
#endif
#endif
```

fuzz/writeout.c
```diff
@@ -177,7 +183,12 @@ void write_out_track(ASS_Track *track, const char *outpath)
         printf("Parsed File will be written to:  %s\n", outpath);
     } else {
         char filename[] = "/tmp/parsedSubs_XXXXXX";
+#ifdef _WIN32
+        _mktemp(filename);
+        int fd = _open(filename, O_CREAT | _O_EXCL | _O_WRONLY, _S_IREAD | _S_IWRITE);
+#else
         int fd = mkstemp(filename);
```
kasper93, line 190: Unrelated note, but for fuzzing on Linux it is better to use `memfd_create` in such cases. See for example https://github.com/mpv-player/mpv/blob/baf528069a584c04686262ef6f76e2c9232adba3/fuzzers/fuzzer_load.c#L45-L58

fuzz/fuzz.c
rcombs: Win32 defines `ssize_t` as an unsigned type.

compare/meson.build
```diff
@@ -0,0 +1,19 @@
+compare_src = files(
+    'compare.c',
+    'image.c',
+)
+
+libass_compare = executable(
+    'libass_compare',
```
TheOneric, line 7: `libass_test` is only namespaced because meson reserves the `test` name; does the same apply here?

checkasm/meson.build
```diff
@@ -0,0 +1,33 @@
...
+    checkasm_src += checkasm_src_aarch64
+endif
+
+libass_checkasm = executable(
+    'checkasm',
+    checkasm_src + config_h,
+    install: false,
+    include_directories: incs,
+    dependencies: deps,
+    link_with: libass_for_tools,
```
TheOneric, line 29: incorrect linkage; this will break on dynamic-only builds on Windows since checkasm relies on accessing private libass functions *(only on Windows because meson still lacks cross-platform symbol export control)*

compare/compare.c
```diff
@@ -18,9 +18,9 @@
 
 #include "image.h"
 #include "../libass/ass.h"
+#include "../libass/ass_filesystem.h"
 #include <sys/types.h>
 #include <sys/stat.h>
-#include <dirent.h>
 #include <string.h>
```
TheOneric, line 24: Using libass internal functions means this can no longer be built with dynamic libass

Either avoid this by just not supporting `compare` with MSVC, or force static linkage like for `checkasm` i both autotools and meson

meson.build
```diff
@@ -120,7 +120,7 @@ if libunibreak_dep.found()
     conf.set('CONFIG_UNIBREAK', 1)
 endif
 
-if get_option('test')
+if get_option('test') or get_option('compare')
     test_deps += dependency('libpng', version: '>= 1.2.0')
```
TheOneric, line 124: since it’s no longer used exclusively by `test`, this should be renamed

fuzz/writeout.c
TheOneric: This code is not even compiled in for continuous fuzzing builds and `memfd_create` defeats the purpose of the writeout functionality

checkasm/meson.build
rcombs: Should I just disable checkasm and compare when building shared-only? (`compare` is already off by default)

compare/compare.c
astiob: > by just not supporting `compare` with MSVC

Don’t we need it for tests? Isn’t this (it being unsupported with MSVC) the current status quo that this PR is aiming to change?

TheOneric: We need it for full test coverage, yes. If the goal of the PR is to get the Windows threading implementation tested in CI, MinGW is sufficient if i’m not mistaken about it also using Win32-native API. If the goal is to get specifically MSVC tested in CI, then yeah, omitting `compare` support goes against that

checkasm/meson.build
TheOneric: In autotools we split the actual libass lib build into two stages *(building internal static-only lib; pushing this through the linker to get the final lib)* to be able to always provide a static lib for checkasm, so if you want to keep `ass_filesystem` in `compare`, you can use this too.
Given meson’s position as a secondary build system with limited support, just erroring out if a tool needing internal API is requested in a dynamic-only build is probably fine too.

meson.build
```diff
@@ -376,9 +376,18 @@ endif
 if get_option('test')
     subdir('test')
 endif
+if get_option('compare') and default_library != 'shared'
```
TheOneric, line 379: just like every other option in the autotools and meson config setups, this should error out if it was explicitly requested but is unavailable instead of silently omitting it

meson.build
```diff
@@ -376,9 +376,18 @@ endif
...
+    subdir('compare')
+endif
 if get_option('profile')
     subdir('profile')
 endif
+if get_option('fuzz')
+    subdir('fuzz')
+endif
+if get_option('checkasm') and default_library != 'shared'
```
TheOneric, line 388: same as with `compare`

compare/compare.c
astiob: Not sure about threads, but at the very least, MSVC builds use different atomics from MinGW.

rcombs:
All build issues are now resolved; this now succeeds on all targets. I believe the meson feature-management issues are also resolved. This should be ready for merge once approved. I think #793 should be blocked on this, since running the full test suite with threading enabled on the MSVC build will be very useful.

I might also add a test pass to the muon build, but I think that can be scoped outside of this PR. The muon build might end up being our "threading-disabled" CI build in the medium-term (as it has bugs that prevent threading from playing nicely with asm).

meson.build
```diff
@@ -373,12 +377,21 @@ else
     libass_for_tools = libass
 endif
 
-if get_option('test')
+if get_option('test').enabled()
```
kasper93, line 380: What's up with `.enabled()`? If it is intended to be a boolean option instead of feature, it should be defined as such in `meson_options.txt`.

meson.build
```diff
@@ -373,12 +377,21 @@ else
...
+endif
+if get_option('profile').enabled()
     subdir('profile')
 endif
+if get_option('fuzz').enabled()
+    subdir('fuzz')
+endif
+if get_option('checkasm').require(enable_asm and default_library != 'shared').allowed()
+    subdir('checkasm')
+endif
```
kasper93, line 394: So to enable libass tests, user has to enable 4 different flags? Why? Could there be single a single flag? It's unclear to me, the user, how to test libass before deployment.

meson.build
```diff
@@ -373,12 +377,21 @@ else
     libass_for_tools = libass
 endif
 
-if get_option('test')
+if get_option('test').enabled()
     subdir('test')
 endif
-if get_option('profile')
+if get_option('compare').require(default_library != 'shared').enabled()
```
kasper93, line 383: Why this is disabled in shared build?

compare/image.c
```diff
@@ -25,6 +25,15 @@ static inline bool is_little_endian(void)
...
 
+static void read_callback(png_structp png_ptr, png_bytep data, size_t length)
+{
+  if (png_ptr == NULL)
+    return;
+
+  if (fread(data, 1, length, png_get_io_ptr(png_ptr)) < length)
+    png_error(png_ptr, "Read Error");
+}
+
```
kasper93, line 36: Is this expected that indentation is different than in the rest of the file?

.github/workflows/meson.yml
```diff
@@ -51,11 +51,20 @@ jobs:
       run:
         shell: ${{ matrix.config.shell || 'bash' }}
 
+    env:
+      ART_SAMPLES: ${{ format('{0}/ext_art-samples', github.workspace) }}
```
kasper93, line 55:
```suggestion
      ART_SAMPLES: ${{ github.workspace }}/ext_art-samples
```

.github/workflows/meson.yml
TheOneric: Putting aside whether this mixing is in general safe or not, the build setup is more complex yet slower than wrapdb+sccache from #857. The latter brought down MSVC build times into the range of other Windows jobs, while the mixed setup here sticks out taking ~60% more time than the next slowest CI job.

Perhaps the argument can be made for there being some merit in having this setup in git history as a point of reference for what it takes to get a mixed setup working; but for further use sccache+wrapdb just seems more beneficial to me

meson.build
TheOneric: Because it _requires_ access to libass-internal symbols which is not possible with a proper dynamic library.
The autotools build has a way to deal with this which can eventually be adapted into the meson setup, but for now just getting tests at all and restricting this is fine imo

kasper93: If a tool or test depends on internal symbols, indeed it shouldn't like with libass directly. Meson has easy way to handle this case, without splitting into static libraries, `libass.extract_all_objects(recursive: true)`  or `libass.extract_objects(...)` for more fine-grained objects to link.

meson.build
rcombs: This is targeting parity with the autoconf build; simplifying the flags would be reasonable, but is out-of-scope for this PR.

rcombs:
I think this is in a good position to merge at this point. There are some suggestions to adjust the deps situation on windows, but I don't think they need to block this.

meson.build
kasper93: Ok, clear. So how should I test libass? Is `-Dtest=enabled` good enough as integration tests or should I enable all other flags too?

compare/image.c
```diff
@@ -55,6 +64,7 @@ bool read_png(const char *path, Image16 *img)
         return false;
     }
 
+    png_set_read_fn(png, fp, &read_callback);
```
astiob, line 67: If I understand correctly, this is a workaround for the MinGW-originating libpng using a different libc (and thus a different `FILE`) from libass on Windows?

Firstly, this deserves a comment, because this code looks like it’s completely redundant. Secondly, try changing the `msystem` in the GitHub CI matrix to `UCRT64`: assuming a new-enough MSVC, I think this should get you the right libc, making this code unnecessary. (Thirdly, of course, this is just another argument for not sourcing dependencies for an MSVC build from MSYS2.)

compare/meson.build
```diff
@@ -0,0 +1,19 @@
...
+    install: false,
+    include_directories: incs,
+    dependencies: deps + png_deps,
+    link_with: libass_for_tools,
+)
+
+art_samples = get_option('art-samples')
+if art_samples != ''
+    dir = join_paths(art_samples, 'regression')
+    test('art-compare', files(join_paths(dir, 'run-all.sh'))[0], workdir: dir, args: [libass_compare.full_path()], depends: [libass_compare])
```
astiob, line 18: How does this work anyway? AFAICT the shell is set to PowerShell, but the executable is a Bourne shell script. Is MSYS2 automatically registered as a handler for `sh` files?

meson.build
astiob: This repository doesn’t contain any integration tests; only the infrastructure.

These flags control which utilities are built. `checkasm` can be run to verify that our assembler code matches our C code and is run by `meson test`/`make check`. The others are special-purpose utilities that can be run by hand; in particular, `test` just takes an ASS file, renders it onto a blank canvas and saves it as a PNG.

Actual integration tests are in the dedicated libass-tests repository, which contains a shell script to run them using these utilities: `compare` for regression tests and `fuzz` for crash tests.

So if you want to run all tests, you must check out the libass-tests repo, build libass with `checkasm`, `compare` and `fuzz` (but not necessarily `test`) enabled, and either
* set `-Dart-samples=path/to/libass-tests` in Meson and run `meson test` (which, at the current state of this PR, will run the tests but only give a summary verdict for all of them together)
* or run `meson test` without defining `art-samples` (which will only run `checkasm`) and run libass-tests separately (which will let you see individual test results).

compare/meson.build
astiob: Further, looking at the CI output, it seems that this causes Meson to swallow the output of our test scripts and just print a single “OK” for the whole lot of tests. Which is certainly better than nothing—and at the end of the day we can run the tests manually to find the failing test—but not ideal. Although maybe it prints the output if the tests should fail?

Actually, does it even properly detect failures?

But anyway, on the other hand, the suggested solution seems to amount to a full rewrite of our test infrastructure in Meson, which also isn’t ideal. On the third hand, this might be easier than it sounds, because some of that code is too generalized/configurable for basic CI and could perhaps be simplified during this rewrite.

arch1t3cht: By default meson will redirect the output into `meson-logs`, but this can be configured with e.g. `-v`

kasper93: > Although maybe it prints the output if the tests should fail?

It logs everything in test-log and can be enable to stdout with verbose flag. Though if it passes shorter output generally make it easier to ready, when there is no need for more info.

> But anyway, on the other hand, the suggested solution seems to amount to a full rewrite of our test infrastructure in Meson

I don't see it that way, I had a glance at run_all.sh script. And it is essentiality aggregator of testcases. So what already Meson does and sure it would be duplicating some logic, so running script like that is fine, but as you noticed it bundles all tests into one, which may not be ideal. Though it is not critical to change it, I feel like it would simplify the flow for Meson.

meson.build
kasper93: Thank you for explanation, I will see if this is feasible for me to integrate on each build.

compare/image.c
rcombs: This has little to do with mingw vs MSVC deps and everything to do with dynamic deps vs static deps. See [this Microsoft docs article](https://learn.microsoft.com/en-us/cpp/c-runtime-library/potential-errors-passing-crt-objects-across-dll-boundaries?view=msvc-170): on Windows, it should be presumed unsafe to pass libc structures across DLL boundaries, or to e.g. `malloc()` an object in one DLL and `free()` it in another. Dynamically-linked deps aren't uncommon by any means, and IMO are a case worth having tests for in our CI.

There are two related potential issues here:
- Different libraries being linked against different *versions* of the libc code (mingw vs MSVC, but also eg MSVC 2013 vs 2015, or UCRT vs MSVCRT)
- Different libraries calling into different *copies* of the same libc code (e.g. either [or both] are statically linked against UCRT)

In either of these cases (which are very normal, esp since statically linking the libc is common on windows, and indeed is enabled by default in some situations), the DLLs have separate heaps, and it is not safe to pass libc/CRT objects (including `FILE*`) across the DLL boundary. This is why we have #775, and why it's so common for libraries to have similar routines (ffmpeg, openssl, etc). This is a common error, and seems extremely worth having tests for!

compare/meson.build
rcombs: Meson runs `sh` with `run-all.sh` as the first arg.

compare/meson.build
rcombs: I already have a step that prints the full log if tests fail.

rcombs: But anyway, as far as I'm aware, meson doesn't support globbing or other methods of listing the contents of a directory, so this suggestion is not actually possible without hardcoding the list of tests (from `libass/libass-tests`) within `libass/libass`. I suppose it might be possible to `run_command('ls')` and then parse the output? But that'd mean pulling in a list of tests at configure-time, which would be brittle (i.e. if you added a test case, it wouldn't get pulled in until you re-ran meson), and I don't think it'd provide any actual *benefit* here.

compare/image.c
kasper93: > Different libraries being linked against different versions of the libc code (mingw vs MSVC, but also eg MSVC 2013 vs 2015, or UCRT vs MSVCRT)

That's historically true, and for that reason many binary releases are/were shipping multiple .dll versions built with each of the CRT. But Microsoft resolved those issue with UCRT. They no longer break compatibility on each version.

It's even noted in the article linked

> Because the DLL version of the Universal CRT library used by Visual Studio 2015 and later is now a centrally deployed Windows component (ucrtbase.dll), it's the same for apps built with Visual Studio 2015 and later versions. 

I strongly believe that in present year, there is no need to care about MSVC 2013 (EOL Apr 9, 2019) or MSVC 2015 (EOL Oct 13, 2020) or about MSVCRT at all. Everyone should be using UCRT which is available since Windows Vista SP2 (with an update). And if anyone needs to deploy MSVCRT they should likely built their dependencies to avoid any issues.

MinGW switched by default to UCRT, which in turn is compatible with MSVC builds.

(the only reason MSVCRT is still popular is because prebuilt toolchains in linux distros are outdated)

DLLs on windows are self-contained, unlike in POSIX world symbols are not loaded into global namespaces and by design thy can use different libc or have it duplicated.

> This is a common error, and seems extremely worth having tests for!

Sure. For me this is Frankenstein build, it doesn't reflect real scenario. And if we want to test corner cases, we should also have nominal cases tested, otherwise what is the point?

compare/meson.build
kasper93: > But anyway, as far as I'm aware, meson doesn't support globbing or other methods of listing the contents of a directory

And that's a good thing. But you can list dir or whatever you need with a script. https://mesonbuild.com/FAQ.html#but-i-really-want-to-use-wildcards

compare/image.c
astiob: I didn’t expect that this affects static builds even when using the same libc, but it makes sense.

Anyway, I’d appreciate a short code comment here explaining this is done to ensure the same stdio is called to open the file and to read it regardless of what stdio libpng is linked to. (Technically, this also affects non-Windows; it’s just more rare.)

> MinGW switched by default to UCRT, which in turn is compatible with MSVC builds.

Not sure what this means. MSYS2 has multiple environments with different C libraries (all of them with MinGW-w64 compilers), and you choose one of them explicitly. This is what I suggested changing in the CI configuration.

compare/meson.build
astiob: Interesting.

[This doc page](https://mesonbuild.com/External-commands.html) (while not specifically about `test`) says “Meson will autodetect scripts with a shebang line and run them with the executable/interpreter specified in it both on Windows and on Unixes”. I’m gonna assume tests do the same, and that on Windows it discards the explicit path from the shebang and tries to run the executable by basename from PATH etc.

compare/image.c
kasper93: > Not sure what this means. MSYS2 has multiple environments with different C libraries (all of them with MinGW-w64 compilers), and you choose one of them explicitly. This is what I suggested changing in the CI configuration.

MSYS2 is not MinGW. MSYS2 is package manager. They provide various environments. See, https://www.msys2.org/docs/environments/ especially MSVCRT vs UCRT section.

Anyway, what I meant is that [MinGW](https://www.mingw-w64.org/) starting from v12.0.0 is using UCRT by default. Of course you can build MSVCRT version (like MSYS2 provides), but we are at the point where we should be migrating everything to UCRT.

See also https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-doc/howto-build/ucrt-vs-msvcrt.txt

> This is what I suggested changing in the CI configuration.

This is good suggestion, also worth testing is CLANG64 environment.

compare/meson.build
kasper93: Also my point is mostly, that if MSVC Windows support is something that is a real target it should work without third-party tools like sh.exe (above other things) from MSYS2. I know it is currently not feasible, but this is generally something to keep in mind.

meson.build
rcombs: A couple of these have actual dependency interactions, and it didn't make sense for some of the utils to be features while others were booleans.

compare/image.c
rcombs: Added comment clarifying intent.

astiob (approved):

.github/workflows/meson.yml
```diff
@@ -155,7 +182,7 @@ jobs:
       - name: configure
         run: |
           muon version
-          muon setup -Dtest=true build
+          muon setup -Dtest=enabled build
```
TheOneric, line 185: noteworthy: like shown here, the option changes affect the build config interface, which should be properly announcement in an eventual release text. But with how the commits are named and not mentioning this in their desription this will now likely be forgotten
Also, can you elaborate on why it was changed?

checkasm/meson.build
```diff
@@ -0,0 +1,34 @@
...
+checkasm_src_aarch64 = files(
+    'arm/checkasm_64.S',
+)
+
+checkasm_asm_sources = []
+if generic_cpu_family == 'x86'
+    checkasm_src += checkasm_src_x86
+elif generic_cpu_family == 'aarch64'
+    checkasm_src += checkasm_src_aarch64
+endif
```
TheOneric, line 21: Unlike automake, meson does **not** properly separate compile flags for assembly and C cources (see also `c_args: asm_args` below. That’s why for the main lib non-NASM assembly is split off into an internal static lib, built with `asm_args` while C sources use the project default `c_args`.

As it stands, *all* checkasm sources, even on NASM platforms, do not use the usual C compiler arguments at all iiuc, but only the much smaller and different set meant for non-NASM assembly files (potentially subtly breaking it in some configs but in any event disabling our warnings etc)*

meson.build
```diff
@@ -120,8 +120,12 @@ if libunibreak_dep.found()
...
 endif
 
-if get_option('test')
-    test_deps += dependency('libpng', version: '>= 1.2.0')
+png_dep = dependency(
+    'libpng',
+    version: '>= 1.2.0',
+    required: get_option('test').enabled() or get_option('compare').enabled())
+if png_dep.found()
+    png_deps += png_dep
```
TheOneric, line 128: This is suboptimal, it will also try to look up libpng and show messages about it to the builder even when no enabled build target uses it. Instead the lookup should be dependend on users being enabled like before, and the dep then being required *(the default iirc)*.

compare/image.c
TheOneric: reminder: `test` writes pngs and so does `compare`. If we consider this a supported usecase this needs fixing too and if not this is change can be removed once we stop mixing in CI

rcombs: Yes, that should be added; I considered it out-of-scope for this PR (esp since we don't have any tests that use those, so it'd be hard to validate).

.github/workflows/meson.yml
rcombs: > A couple of these have actual dependency interactions, and it didn't make sense for some of the utils to be features while others were booleans.

compare/image.c
TheOneric: > esp since we don't have any tests that use those, so it'd be hard to validate

The muon CI job actually builds and executes the `test` utility atm though it ofc doesn't use MSVC. With the PR here it now makes more sense to run the usual tests instead

checkasm/meson.build
TheOneric: Does MSVC accept `-D` macro defines? If not this might actually already break for MSVC Windows ARM builds

rcombs: MSVC does accept `-D` args.

As far as I can tell, the full set of C args *is* passed here; an example command from a local build:
`[40/49] cc -Icheckasm/checkasm.p -Icheckasm -I../checkasm -I. -I.. -Ilibass -I../libass -I/opt/homebrew/opt/freetype/include/freetype2 -I/opt/homebrew/opt/libpng/include/libpng16 -I/opt/homebrew/Cellar/fribidi/1.0.15/include/fribidi -I/opt/homebrew/Cellar/harfbuzz/9.0.0/include/harfbuzz -I/opt/homebrew/Cellar/glib/2.80.4/include/glib-2.0 -I/opt/homebrew/Cellar/glib/2.80.4/lib/glib-2.0/include -I/opt/homebrew/opt/gettext/include -I/opt/homebrew/Cellar/pcre2/10.44/include -I/opt/homebrew/Cellar/graphite2/1.3.14/include -I/opt/homebrew/Cellar/libunibreak/6.1/include -I/opt/homebrew/Cellar/fontconfig/2.14.2/include -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -std=gnu11 -O2 -g -Wno-sign-compare -Wno-unused-parameter -Werror-implicit-function-declaration -Wstrict-prototypes -Wpointer-arith -Wredundant-decls -Wno-missing-field-initializers -Wformat-non-iso -Werror=format-non-iso -Watomic-implicit-seq-cst -D_GNU_SOURCE -D_XPLATFORM_SOURCE -fno-math-errno -DPREFIX -MD -MQ checkasm/checkasm.p/checkasm.c.o -MF checkasm/checkasm.p/checkasm.c.o.d -o checkasm/checkasm.p/checkasm.c.o -c ../checkasm/checkasm.c`

So it gets everything we passed to `add_project_arguments`, plus `asm_args` (which is just `-DPREFIX`, which is harmless to apply to our C sources as well).

That said, we should really add a ARM builds to the CI (both on Windows and Linux).

meson.build
rcombs: Does this really matter? It's a one-time cached configure check (and tbh we might end up turning the test/compare/etc programs on by default behind a libpng-availability gate at some point anyway).

Issue #825, opened 2024-09-12
[RFC/Discussion] Text to shape / Text metrics API


arch1t3cht:
Adding some form of text metrics API to libass has been in discussion for a long time (see #87 and #348), but (as far as I can see) never with any too concrete proposals. I've been planning to draft out a more detailed proposal for a long time now, but never really had the time for it. I have a bit more time now and was reminded of this issue by a discussion today, so I thought I could start by listing my thoughts so far and see if anyone else has relevant ideas or concerns. Apologies in advance for the long issue.

### Use Cases
Maybe it's best to start with the API user's perspective, to see what kinds of metrics would be needed in such an API.
There are two big classes of requests:
- **Text metrics**: These are, of course, needed to lay out text, especially when splitting it by character or by syllable. Most importantly, they are used in Aegisub's karaoke templater to split a line into syllables or characters and position them correctly. In my fork of Aegisub, they are also used in the [perspective tool](https://github.com/arch1t3cht/Aegisub/releases/tag/feature_09) to determine how much a line will be shifted by `\fax`. Similarly, they're used by any Lua scripts that deal with perspective.

  In Lua scripts (including the karaoke templater), these metrics are obtained from the [`aegisub.text_extents`](https://aegisub.org/docs/latest/automation/lua/miscellaneous_apis/#aegisubtext_extents) function:
  ```
  width, height, descent, ext_lead = aegisub.text_extents(style, text)
  ```
  On Windows, this function uses GDI (implemented [here](https://github.com/Aegisub/Aegisub/blob/6f546951b4f004da16ce19ba638bf3eedefb9f31/src/auto4_base.cpp#L72-L120)) - `width` and `height` are obtained using `GetTextExtentPoint32`, while `descent` and `ext_lead` are obtained from `GetTextMetrics` (and are independent of the given text), i.e. it returns the same metrics that VSFilter uses for layouting.

  However, on Linux, where GDI is not available, it has to use wxWidgets, which (usually) internally uses gtk, which in turn uses pangocairo to compute the metrics. The resulting metrics are accurate some of the time (though iinm even then there are differences in rounding resulting in shifts on the order of one PlayResY unit), but at other times they can be very wrong. This can make advanced typesetting on Linux very hard (and some typesetters use strange workarounds like running Aegisub through wine, which has a better, albeit not perfect, emulation of GDI)
[...]

PR #856, opened 2024-12-09
[WIP/POC] Add API to obtain metrics and shape data

arch1t3cht:
This is an initial proof-of-concept attempt to implement an API along the lines of #825. It's not yet complete, but before spending more time on this I'd first like to ask if you would be open to adding such a metrics API at all, and if so, if you agree with the overall structure used in this PR. If that is the case, I can polish the PR and discuss some of the finer details (like what fields should be exposed, what data structures to use, and other implementation details).

A couple of comments.
- I went with the general structure outlined in #825, i.e. adding an `ass_get_metrics` function that internally does all the same operations as `ass_render_frame` (even including rasterization), but in the end returns shape data rather than bitmaps. The idea being that this is fairly easy to implement from libass's side and provides a very powerful API to the user. (To clarify, I'm imagining the user either creating a dedicated `ASS_Track` to pass to `ass_get_metrics`, or to add an extra event they want to measure to an existing track (which they might need if they care about collision detection) and picking out the `ASS_Metrics` struct pointing to that added event.) A downside may be that the user will need to create an entire `ASS_Track` with headers and styles and everything in order to obtain metrics, but since this API is mainly aimed at authoring tools, most programs that would use such an API should already have the machinery for this.
- I added the API in a separate header to not confuse users that just want to render subtitles, and so that the metrics API could theoretically be versioned separately (i.e. more finely) from the main API.
- My main aim when implementing the API was to impact the normal rendering code as little as possible, i.e. to not add any big separate code paths and to instead just write some of the intermediate data into structs. In particular, `ass_get_metrics` still runs the full rasterization code even though the result is not used (though this could also be avoided with more work). This being an authoring API, performance should not be as critical, and I feel like this way it should have better maintainability.

Again, the main feedback I'm interested in right now is whether you would consider adding this kind of API at all. If so, I can flesh this out further.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -29,11 +29,13 @@
 #endif
 
 #include "ass.h"
+#include "ass_metrics.h"
 #include "ass_outline.h"
 #include "ass_render.h"
 #include "ass_parse.h"
 #include "ass_priv.h"
 #include "ass_shaper.h"
+#include "ass_utils.h"
 
 #define MAX_GLYPHS_INITIAL 1024
 #define MAX_LINES_INITIAL 64
@@ -177,6 +179,7 @@ void ass_renderer_done(ASS_Renderer *render_priv)
     if (!render_priv)
         return;
 
+    ass_free_metrics(render_priv);
     ass_frame_unref(render_priv->images_root);
     ass_frame_unref(render_priv->prev_images_root);
 
@@ -1118,6 +1121,10 @@ init_render_context(RenderContext *state, ASS_Event *event)
 {
     ASS_Renderer *render_priv = state->renderer;
 
+    state->collect_metrics = false;
+    state->run_metrics = NULL;
+    state->current_run_metrics = NULL;
+
     state->event = event;
     state->parsed_tags = 0;
     state->evt_type = EVENT_NORMAL;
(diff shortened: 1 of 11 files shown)


rcombs:
This sounds very good to me! Given the design of this PR, I wonder if anyone would have a use-case for exposing both the images and the metrics from a single render pass (i.e. a way to retrieve the metrics from the just-rendered frame).

TheOneric (commented):
API-wise this seems mostly good to me, some remarks:

Metrics and rendering sharing the same internal process mean they’ll affect each others caches etc and the `ass_get_metrics`already mentions the lifetime of metrics thus being affected by render calls. But the reverse is also true and will need to be documented in `ass_render_frame`; same for `changed`.

I know you’re not looking for code comments yet, but since it might get relevant in testing: the new API needs to be added to the list of exported symbols, else it only usable from static libs *(or broken meson builds)*.

In prior discussion the distinction between logical and visual metrics came up regularly. Both the issue and this implemented/seem focussed on logical metrics.
Which makes sense since visual extents can already be extracted from rendered bitmaps. I guess providing access to visual and logical metrics in the same call in the future is another reason to keep the internal pipelines for metrics and rendering identical?

libass/ass_render.c
```diff
@@ -3393,7 +3486,8 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
                     realloc(priv->eimg,
                             priv->eimg_size * sizeof(EventImages));
             }
-            if (ass_render_event(&priv->state, event, priv->eimg + cnt))
+            (priv->eimg + cnt)->metrics.runs = NULL;
```
filip-hejsek: This only initializes the `EventImages` partialy, which is insufficient.

libass/ass_render.c
filip-hejsek, line 3092: This initialization is skipped if this function returns early.

libass/ass_render.c
```diff
@@ -3393,7 +3486,8 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
...
+            if (ass_render_event(&priv->state, event, priv->eimg + cnt, collect_metrics))
```
filip-hejsek, line 3490: This only initializes the metrics partialy, which is insufficient.

libass/ass_render.c
```diff
@@ -3427,14 +3521,95 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
...
+    ass_free_metrics_outlines(next);
+}
+
+/**
+ * \brief Free all memory that was allocated for metrics.
+ */
+void ass_free_metrics(ASS_Renderer *priv)
+{
+    if (!priv->eimg_size)
+        return;
```
filip-hejsek, line 3597: This only checks if `EventImages` have been allocated, not initialized.

libass/ass_render.c
```diff
@@ -3427,14 +3521,95 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
...
+
+    // The ASS_Metrics structs themselves are part of EventImages and
+    // are hence allocated and freed by ass_render_frame_internal.
+    // All contained pointers are owned by the ASS_Metrics struct
+    // and need to be freed here.
+    for (ASS_Metrics *metrics = &priv->eimg[0].metrics; metrics; metrics = metrics->next) {
```
filip-hejsek, line 3603: Here we crash because `metrics->next` is uninitialized.

filip-hejsek (commented):
I have tried using the POC to implement Aegisub's text metrics API and encountered crashes. Since I have already debugged them and found the cause, I figured I might as well post my analysis here as a review.

PR #857, opened 2024-12-09
ci/gha: add sccache caching for MSVC build

kasper93:


.github/workflows/meson.yml
```diff
@@ -112,7 +115,7 @@ jobs:
           esac
 
       - name: Set up WrapDB
-        if: matrix.config.pkgmanager == 'choco' || matrix.config.os == 'macos-latest'
+        if: matrix.config.pkgmanager == 'choco' || matrix.config.os == 'macos-latest' || matrix.config.msvc
```
TheOneric, line 118: The `matrix.config.pkgmanager == 'choco'` part is actually a leftover from before the MSys dep reuse and nothing sets this now. Either MSVC jobs should now again declare their pkgmanager to be `choco` and there’s no need to add `|| matrix.config.msvc`, or the `choco` condition removed here

rcombs:
I still don't find wrapdb usage to be particularly useful here (it's still third-party forks of our dependencies, so it's only reasonably usable for local personal builds, and won't be supported for production builds, which are what we want to mimic in CI), but if we're going to have a workflow that uses them, we should also cache the download step; `meson configure` is currently taking over a minute on the MSVC build (it's under 10 seconds everywhere else).

kasper93:
> (it's still third-party forks of our dependencies, so it's only reasonably usable for local personal builds, and won't be supported for production builds, which are what we want to mimic in CI)

This mimics what everyone is doing on Windows, just wrapped in Meson. Vcpkg may be a larger package manager, but having a Meson-wrapped build system also works. It's quite rare to build everything by hand.

For your information, FreeType2, FriBidi, and HarfBuzz are all official release tarballs since they provide first-party meson.build files. So, there are no forks here. LibPNG and zlib are wrapped, but these are tiny libraries where it hardly matters how they are built. The code itself is not modified in any way (official release tarballs are used), except for the addition of meson.build files.

I also find it amusing to mention "production builds" when the current build is side-loading some MinGW libraries from third-party binary blob repository. This is further from production than you might think. Also, if you're already using MinGW/MSYS, why even bother with MSVC? It makes no sense. No one would do that in "production". If MinGW is in their pipeline, they'll build libass with MinGW. If it’s not in their pipeline, they won't build dependencies with MinGW either.

Either way, I made this PR as a response to request made on IRC. I myself don't understand the issue of build taking 2, 4, 6 minutes. How long on average it takes to merge a PR? Is Windows build time the bottle-neck in this process? Of course if you build only dozen of libass files (like on Linux) it will be faster than downloading and building dependencies.

Like I said in the other discussion, letting Meson handle dependencies is convenient way to make it work. Without additional complexity, renaming libraries and so on. I'm trying to help here and keep it simple, not adding hacks on top of hacks to make it work. Feel free to close this PR if this is not what is wanted.

> we should also cache the download step

This can be easily done by caching `subproject` directory, but it requires little maintenance, because to update subprojects it would likely have to be cleared manually. (by updating cache version)

.github/workflows/meson.yml
```diff
@@ -67,25 +70,27 @@ jobs:
...
+          "timestamp=$((Get-Date).Ticks)" >> $env:GITHUB_OUTPUT
 
-      - name: Install more MSVC dependecies with msys
+      - name: Restore cache
+        if: matrix.config.msvc
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            ${{ env.SCCACHE_DIR }}
+            ${{ github.workspace }}/subprojects
```
TheOneric, line 85: Does `meson wrap update-db`  update existing project downloads?
While the meson PR was being wroked on we had some wrap files directly in the repo and back then meson just always kept existing project sources around leading to issues everytime a dep was upgraded.

kasper93: No, it does not update. wrap db is just a json file with list of available projects, but once they are initialized in subproject directory they will not be updated until manually changed. Things in subproject directory always takes precedence.

> meson just always kept existing project sources around leading to issues everytime a dep was upgraded.

Yes, this is the case. As far as I know, the only solution is to clear subprojects driectory manually.

TheOneric: If there’s a straightforward way of retrieving the availbe version on wrapdb, they could be included in the cache key. Otherwise you could limit the lifetime of the subproject download cache like done here to avoid us getting stuck on the same dep versions forever: https://github.com/libass/libass-tests/blob/cee2df5fa840df02dda1a33f75a99844b5cfd61e/.github/workflows/art.yml#L124

(Note: the linked setup also includes an additional `rev` number to allow easily forcing a cache rebuild if necessary, but since GHA now allows deleting caches from the web UI this shouldn't be necessary anymore)

kasper93: After taking a look into it, `--reset` will remove stale sources, so all we have to do is `meson subprojects update --reset`.

TheOneric: hmm.. it’s not documented at [the usual place]() and the code looks like it indiscriminately deletes even up-to-date subdirs and but reuse already downloaded tarballs. Those tarball names are versioned, so that’s fine. However:
- the cache will slowly grow since old tarballs are never cleaned up
- `meson subprojects update` doesn't recognize extracted project dirs anymore if the wrap file version changed (since the versioned dir name changed); so they too never get cleaned up:

    ```
    Updating fribidi-1.0.15...
      -> Cannot update subproject with no wrap file
    Using fribidi source from cache.
    Updating fribidi...
      -> New version extracted
    ```
  
- neither `update` nor `wrap update-db` actually seem to modify an existing `*.wrap` file. To upgrade fribidi i had to first delete the existing wrap

How did you get this working on your end?

kasper93: > hmm.. it’s not documented at [the usual place](https://github.com/libass/libass/pull/857) and the code looks like it indiscriminately deletes even up-to-date subdirs and but reuse already downloaded tarballs.

I think your link is not correct, but it is documented here https://mesonbuild.com/Subprojects.html#update-subprojects And yes, what it does it to remove source dir and update it from .wrap file, which would otherwise never done.

> the cache will slowly grow since old tarballs are never cleaned up

Typical cache issue, sccache will also grow with old object files that are not used. Up to limit.

I don't think the amount and frequency of updates is that big to warrant added complexity to manage that. Even if it grows with a "few" of older version at some point, it can be purged manually, from GHA cache.

> doesn't recognize extracted project dirs anymore if the wrap file version changed (since the versioned dir name changed); so they too never get cleaned up:

Yes, indeed. And I tried the [`purge`](https://mesonbuild.com/Release-notes-for-0-58-0.html#purge-subprojects-folder) option seems to remove new version instead of old one. Things are strange. 

(Also it is really annoying that Meson document functions in release notes only and don't put it in the normal documentation...)

Anyway, I changed it now to cache only `subprojects/packagecache`, so tarballs and patches. So we avoid any wrap/source dir issues, because it clearly is not designed to be managed in automatic way.

> neither update nor wrap update-db actually seem to modify an existing *.wrap file. To upgrade fribidi i had to first delete the existing wrap

To update wrap, you can use `meson wrap update`, which is NOT the same as `meson subproject update`. But not important anymore, wrap will be recreated each time and only cache will be reused.

kasper93:
Updated harfbuzz to newer version too, because freetype2 has pretty old one defined as wrap. Newer version is little bit slower to build, but what can you do. Those Windows VM are slooow.

.github/workflows/meson.yml
TheOneric: > I think your link is not correct, 

Yeah, I forgot to actually copy the link and GH interprets an empty `[link text]()` as a link to the current page, sorry.
This is the link I meant to reference *(`purge` is documented, but `--reset` is absent; annoying indeed)*: https://mesonbuild.com/Commands.html#subprojects

> Typical cache issue, sccache will also grow with old object files that are not used. Up to limit.

**“Up to limit”** being the important distinction here. For sccache *(and other sane caches)* we basically allocate a chunk of our cahce budget to this (the max limit)*, but as implemented the subprojects cache will grow indefinitely. It growing slowly just means it’s less likely someone keeps an eye on it or immediately remembers what to do once problems show up, or jobs slow down from down/uploading bloated cache entries.

Given we have a known set of project names to worry about, properly managing amounts to just grouping all file entries matching up to their first digit, sorting the tail part starting from the first digit numerically and deleting all but the (something like: `sort -nr | tail -n +2 | xargs -n 1 printf '%s%s\n' "$project" | xargs rm`)

>

kasper93: I've added a cache clearing step. I still think it is not strictly necessary though. For two reasons. First even few years worth of tarballs for our dependencies would produce any real issue with the cache itself. And second and more importantly GitHub cache lifetime is 7 days, so unless libass will keep the cache up for literally years by bumping it every few days, it would be naturally cleaned. But sure there is theoretical possibility that it will grow...

TheOneric (commented):
I’m just trusting you on the powershell and VSCode setup changes, otherwise this looks good apart from a couple remarks. Speedup is also good bringing time even a bit below (completely uncahced) MinGW autotools builds. Thanks!

---

I’ve seen it succeed before, but now the most recent GHA run failed for dynamically linked MSVC. Logs just show both `fuzz` and `compare`  exiting with error 127 everywhere *(also meson apparently also runs libpng’s tests and enables sanitizers but no sanitizer output is shown on the errors)*. Debugging this in GHA will be annoying, can you look into this locally?

.github/workflows/meson.yml
```diff
@@ -75,13 +77,25 @@ jobs:
...
+
       - name: Install MSVC dependecies with choco
         if: matrix.config.msvc
         run: |
-          # for whatever reason choco insist on a reboot if it installs meson
-          choco install ninja nasm python pkgconfiglite
-          python -m pip install --upgrade pip
-          pip install meson
+          python -m pip install meson
+          choco install sccache nasm
```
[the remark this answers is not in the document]
kasper93: The current comment is not true. I've added updated one. It's simply easier to manage PATH, when installed through pip.

.github/workflows/meson.yml
```diff
@@ -15,15 +15,25 @@ jobs:
...
-          - name: Windows MinGW
+          - name: Windows (CLANG64)
+            os: windows-latest
+            msystem: CLANG64
+            shell: 'msys2 {0}'
+
+          - name: Windows (UCRT64)
+            os: windows-latest
+            msystem: UCRT64
+            shell: 'msys2 {0}'
```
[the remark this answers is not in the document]
kasper93: I don't remember how many runners are available, but there are certainly enough to run all configurations in parallel as we do in mpv. The added value is that we don't have to think what conf should be tested and just run them all, we can afford that.

.github/workflows/meson.yml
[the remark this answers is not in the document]
TheOneric: It definitely used to be true, but good to know it no longer does that.  
Can you elaborate on what issues there are with `PATH`? `nasm` is also installed via `choco` and seemingly shows up in the `PATH` without further action required.

.github/workflows/meson.yml
```diff
@@ -44,6 +44,7 @@ jobs:
 
           - name: Ubuntu
             os: ubuntu-latest
+            args: '-Db_sanitize=address,undefined'
```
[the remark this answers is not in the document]
TheOneric: How does this meson option deal with clang’s UBSAN either never setting a non-zero exit code, or not printing a report about _what_ went wrong?

.github/workflows/meson.yml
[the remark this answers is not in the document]
TheOneric: The limit appears to be [20 concurrent runners, but at most 5 macOS ones](https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits), atm. Meaning (unless overlapping with libass-test) this still is within bounds but already close to the limit

I guess we ca still drop the superfluopus WIndows runners later if it becomes relevant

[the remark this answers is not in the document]
moi15moi: In this [example](https://github.com/moi15moi/FindSystemFontsFilename/actions/runs/12448903177), you can see that I have 6 ``macos`` runners, so I feel that the information isn't correct.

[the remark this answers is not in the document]
TheOneric: The limit is supposed to apply to how many jobs can run concurrently; you can still have more than that, but some will only start to run after the others finished.

In the example workflow run you linked, each job only takes a couple seconds so but their end time is only displayed with minute precision so i can't tell whether they ran concurrent or partially sequential.

profile/meson.build
```diff
@@ -6,5 +6,5 @@ libass_profile = executable(
     install: false,
     include_directories: incs,
     dependencies: deps,
-    link_with: libass_for_tools,
+    objects: libass.extract_all_objects(recursive: true),
```
TheOneric, line 9: Only checkasm and compare need a static lib, the rest can work with either (thus they can also be used to test dynamic linking works)

fuzz/meson.build
```diff
@@ -9,11 +9,11 @@ libass_fuzz = executable(
     install: false,
     include_directories: incs,
     dependencies: deps,
-    link_with: libass_for_tools,
+    objects: libass.extract_all_objects(recursive: true),
```
TheOneric, line 12: Here too `libass_for_tools` should be kept

test/meson.build
```diff
@@ -6,5 +6,5 @@ libass_test = executable(
     install: false,
     include_directories: incs,
     dependencies: deps + png_deps,
-    link_with: libass_for_tools,
+    objects: libass.extract_all_objects(recursive: true),
```
TheOneric, line 9: same here

.github/workflows/meson.yml
```diff
@@ -15,105 +15,102 @@ jobs:
...
+            os: windows-latest
+            msystem: CLANG64
+            shell: 'msys2 {0}'
+            args: '-Ddefault_library=both -Db_sanitize=address,undefined'
+
+          - name: Windows (UCRT64)
+            os: windows-latest
+            msystem: UCRT64
+            shell: 'msys2 {0}'
             args: '-Ddefault_library=both'
```
TheOneric, line 37: Might be good to use `-Ddefault_library=shared` for one Windows runner to ensure shared-only functions too

.github/workflows/meson.yml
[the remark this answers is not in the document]
kasper93: meson can find nasm in standard location, so it does not have to be in PATH. On the other hand, we call meson directly, so we have to have it in PATH. And since this jobs steps are shared, we would need to add some logic to add to path conditionally or call meson with absolute path. pip just seems clear to me

Also since meson is purely Python project, I find use of pip appropriate for the job, instead of some installers through.

.github/workflows/meson.yml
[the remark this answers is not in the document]
kasper93: > Actually this decreases build diversity since all Ubuntu-based autotools runners use ASAN and UBSAN, but this seems hardly problematic

Fair. But still ASAN is generally good to run either way. I mean if it works with ASAN it will work without it too.

> How does this meson option deal with clang’s UBSAN either never setting a non-zero exit code, or not printing a report about what went wrong? (ref 

Meson sets UBSAN_OPTIONS and ASAN_OPTIONS that are appropriate for running tests. Specifically it sets `halt_on_error=1`.

Probably ghci.yml could also set `UBSAN_OPTIONS=halt_on_error=1` this will alow the print of the report.

profile/meson.build
kasper93: In this case it should link to libass directly.

fuzz/meson.build
kasper93: Updated.

test/meson.build
kasper93: Updated.

.github/workflows/meson.yml
kasper93: Sure, I can make each of this build different.

profile/meson.build
TheOneric: No it shouldn’t. Linking to libass directly prefers dynamic, while we want to prefer but not require static

.github/workflows/meson.yml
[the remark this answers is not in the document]
TheOneric: Sure, if LLVM/CLang now actually has a way to do this that will be preferable (we already set `-fno-sanitize-recover` anyway, but in the past clang’s UBSAN used to ignore all exitcode runtime flags)

profile/meson.build
kasper93: Why? As I understand the libass_for_tools were added only to make those tests work in 'both' build.

Why do you want to prefer static? I'm asking because it has to be documented.

TheOneric: > As I understand the libass_for_tools were added only to make those tests work in 'both' build.

It wasn’t and if it were, would only have been added to the relevant utilities. `libass_for_tools` exists since the original meson implementation predating any tests being hooked up. No changes or comments wrt to `libass_for_tools`are needed.
 
- Statically linked, uninstalled binaries are resilient to directory renaming, can be copied around and e.g. kept as a reference during bisects
- preferring static matches the autotools builds *(there a simple libtool link flag suffices to prefer static when both are available)*

kasper93: I understand why static tools may be preferable. Sorry, it's my fault for not asking questions clearly. If you prefer statically linked tools, why don't you want that when building shared libass?

> preferring static matches the autotools builds (there a simple libtool link flag suffices to prefer static when both are available)

Meson has `prefer_static` flag. I don't see how autotools are relevant here?

TheOneric (approved):
Seems all good now and even went beyond just the initial caching goal; thanks!

profile/meson.build
TheOneric: Utilities are never installed and serve as convenient tools for developers and testing; therefore it makes sense to tweak the process for it and since it’ſ not installed we can be more opinionated.
The libass library on the other hand is the main part installed on systems later, here it should follow whatever the builder chose. Following/Keeping common defaults unless there’s a reason not too (like with the default library type) is also good to yield expected results.

I believe `prefer_static` used to require a too-new meson version for most of the PRs lifetime. But even now, `prefer_static` affects _all_ libraries iiuc, while we mostly just care about our own code during e.g. bisects.

Consistency with autotools matters insofar as it means there are less differences to keep in mind when switching back and forth between the build systems during testing creating less surprises. *(Or if you meant why I mentioned how the behaviour is achieved in autotools: so there’s no need to search for a non-existing `libass_for_tools` equivalent as a reference in the autotools build setup)*

PR #860, opened 2025-01-02
ass_directwrite: don't vendor custom dwrite.h

kasper93:
The required headers have been included in MinGW since version 8.0.0, released on 2020-09-18. This version is readily available on modern distributions like Ubuntu 22.04 (previous LTS) and Debian Bullseye (oldstable), which represents a reasonable cutoff point for cross-compilation support. Moreover, older platforms are not actively tested on CI, so ensuring compatibility with them cannot be guaranteed.

See: https://sourceforge.net/p/mingw-w64/mingw-w64/ci/716676b8bfbec6742618194e5c31a7bb3b5d7866


astiob:
> The required headers have been included in MinGW

But not in MSVC.

kasper93:
Why even support MSVC in this case...

PR #864, opened 2025-01-28
fontconfig: fix minor memory leak in fallback font selection

kasper93:
When the used config is completely devoid of fonts FcFontSort will attempt to allocate an empty set and return it while still setting the result to invalid.

Similar case as fixed in 6e83137cdbaf4006439d526fef902e123129707b.

### DIFF
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -243,8 +243,11 @@ static void cache_fallbacks(ProviderPrivate *fc)
 
     // If this fails, just add an empty set
     // (if it fails, cache_fallbacks will just be reattempted later)
-    if (result != FcResultMatch)
+    if (result != FcResultMatch) {
+        if (fc->fallbacks)
+            FcFontSetDestroy(fc->fallbacks);
         fc->fallbacks = FcFontSetCreate();
+    }
 
     FcPatternDestroy(pat);
 }


astiob (commented):
Can/should we just skip the `FcFontSetCreate` call in this case?

kasper93:
> Can/should we just skip the `FcFontSetCreate` call in this case?

We could do this 
``` c
    // If this fails, just add an empty set
    // (if it fails, cache_fallbacks will just be reattempted later)
    if (result != FcResultMatch && !fc->fallbacks)
        fc->fallbacks = FcFontSetCreate();
```

Though I'm not sure at what state the `fc->fallbacks` is if `result != FcResultMatch`. I'm more comfortable with creating empty set explicitly. I will let you decide though.

TheOneric:
> Though I'm not sure at what state (and if it won't change in the future fontconfig version) the fc-fallbacks is if result != FcResultMatch.

In almost all cases `result != FcResultMatch` is paired iwth a `NULL` from `FcFontSetSort` and/or `FcFontSort`. Afaict the only exception currently is a valid (fallback) config but zero sets being passed *or* all passed set referenced being either `NULL` or empty. If that happens a new empty set is returned without changing the default failure indicator *(near the start of `FcFontsSetSort`)*.

It could be argued since an empty set is a valid sorting for zero fonts, `result` should be a success, but it seems like pairing it with a failure indicator was intentional: [`fontconfig@a18ca17b6211f62fb`](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/commit/a18ca17b6211f62fbd1d893811b94b8c83db4cc0).

I’d hope fontconfig at least will never return a stale pointer, but who knows :\. I think i weakly prefer the current explicit empty set creation

PR #865, opened 2025-01-29
ci/gha/meson: use ubuntu-latest directly for tarball test

kasper93:
It has been upgraded to 24.04 which should be new enough.


astiob:
The CI run has succeeded, so I assume this is good indeed.

PR #866, opened 2025-01-30
ci: enable ARM linux builds

rcombs:
These runners are now available to free projects.

### DIFF
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -32,6 +32,11 @@ jobs:
             docker_image: alpine:latest
             shell: '/tmp/docker_shell {0}'
             art_reg_skip: 'font_nonunicode'
+          # ARM linux builds
+          - os: ubuntu-24.04-arm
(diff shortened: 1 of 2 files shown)


TheOneric (approved):

TheOneric:
note we are not exceeding the limit of 20 concurrent runners per org; after this is merged we might want to consider dropping a few low relevance jobs (e.g. we don't need to cover all Windows libc configurations in both autotools and meson imho)

astiob (approved):
> note we are not exceeding the limit of 20 concurrent runners per org

I assume this was meant to be “now”?

It seems we’re only just over the limit and the speed of our jobs varies enough that there’s a decent chance two faster jobs manage to complete in the same time one slower job does. But of course, we may have multiple CI flows running concurrently from PRs and master and whatnot, and maybe we really don’t need so many. Then again, it’s not like they’re _super_ slow, anyway; we can generally just wait for any queued jobs if we have to.

PR #871, opened 2025-03-23
utils: use unambiguous return type for ass_lrint

kasper93:
All users of this function expect a 32-bit integer. The intrinsic version already converts to a 32-bit value. The non-intrinsic version uses the standard lrint, which returns a long, whose size differs between Windows and Linux targets. Be explicit about the return type and consistent between platforms.


rcombs:
We should probably explicitly handle overflow cases here to ensure we can't end up with UB (or implementation-defined behavior), as with `dtoi32` in `ass_parse.c`. The intrinsic implementation's behavior (which I assume matches what happens in vsfilter?) is that anything out of range converts to 0x80000000 (`INT32_MIN`). Compilers seem to be pretty bad at handling both of these functions, so maybe they should both have x86 intrinsic/asm implementations (both on MSVC and gcc/clang)?

kasper93:
I think it is good assumption that `ass_lrint` is called with the values that are representable with int32_t type that is expected on the other end.

We could add `ass_lrint_checked` to check for overflows if this makes easier to handle. But in most cases it should already be handled before "final" to integer conversion. Because it likely that we work with bogus values already.

PR #872, opened 2025-03-23
meson: add support for oss-fuzz

kasper93:
This will allow to enable MSAN and i386 builds in oss-fuzz. See: https://github.com/kasper93/oss-fuzz/commit/2709d7fed5303a0601c6e7c85792a2ba58258119

### DIFF
--- a/fuzz/meson.build
+++ b/fuzz/meson.build
@@ -9,6 +9,8 @@ libass_fuzz = executable(
     install: false,
     include_directories: incs,
     dependencies: deps,
+    link_args: get_option('fuzz-link-args'),
+    link_language: get_option('fuzz-link-language'),
     link_with: libass_for_tools,
 )
 
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -4,6 +4,8 @@ option('profile', type: 'feature', description: 'enable profiling program')
 option('fuzz', type: 'feature', description: 'enable fuzzing consumer')
 option('checkasm', type: 'feature', description: 'enable assembly unit test program')
 option('art-samples', type: 'string', description: 'Path to the root of regression testing sample repository. If set, it is used in meson test.')
+option('fuzz-link-args', type: 'string', description: 'Additional link flags for fuzz targets.')
+option('fuzz-link-language', type : 'combo', choices : ['c', 'cpp'], value : 'c', description: 'Linking language for fuzz targets.')
 
 option('fontconfig', type: 'feature', description: 'Fontconfig support')
 option('directwrite', type: 'feature', description: 'DirectWrite support (win32 only)')


fuzz/meson.build
```diff
@@ -9,6 +11,8 @@ libass_fuzz = executable(
     install: false,
     include_directories: incs,
     dependencies: deps,
+    link_args: get_option('fuzzing-engine'),
+    link_language: 'cpp',
```
TheOneric, line 15: This now always requires a C++ linker or perhaps depending on what the language declaration earlier does even a full C++ toolchain? Yet, only OSS-Fuzz actually needs C++ linkage making it unnecessarily hard to build the fuzzer for everyone else.

meson_options.txt
```diff
@@ -4,6 +4,7 @@ option('profile', type: 'feature', description: 'enable profiling program')
 option('fuzz', type: 'feature', description: 'enable fuzzing consumer')
 option('checkasm', type: 'feature', description: 'enable assembly unit test program')
 option('art-samples', type: 'string', description: 'Path to the root of regression testing sample repository. If set, it is used in meson test.')
+option('fuzzing-engine', type: 'string', description: 'Additional link flags for fuzz targets.')
```
TheOneric, line 7: I’d prefer to neutrally name it `fuzz-ldflags` or something similar, since it can also be used for other things than the fuzzing engine. On the other side of this, from the name `fuzzing-engine` I would never expect to have to specify link flags but instead would try to name the desired engine.

TheOneric (commented):
I guess this helps with MSAN due to making it easier to build all dependencies (which MSAN requires). How does this help with 32-bit though?

Unlike the autotools setup, this necessitates passing fuzz macro defines to the entire build, which sure is simpler but at the same time feels less clean to me. Admittedly this is unlikely to cause issues in practice due to the `ASSFUZZ` prefix

kasper93:
> I guess this helps with MSAN due to making it easier to build all dependencies (which MSAN requires). How does this help with 32-bit though?

It unifies the build. For 32-bit it would be possible to use prebuilt packages, but in practice for fuzzing it is always better to build whole binary instrumented, to avoid any black box parts from fuzzer point of view. It helps drilling into the code paths, with smaller input changes. Granted it not always matters, but its good nonetheless. 

> Unlike the autotools setup, this necessitates passing fuzz macro defines to the entire build, which sure is simpler but at the same time feels less clean to me. Admittedly this is unlikely to cause issues in practice due to the `ASSFUZZ` prefix

True, but I wouldn't expect any issues with that.

fuzz/meson.build
kasper93: > You must use $CXX as a linker, even if your project is written in pure C.

This is requirement from oss-fuzz. Because some "engines" link static libraries that depends on C++ runtime. I don't recall which one exactly needs it, but it's good to be consistent here.

> Yet, only OSS-Fuzz actually needs C++ linkage making it unnecessarily hard to build the fuzzer for everyone else.

Is this a problem? I doesn't think it is common to have C only toolchain setup, especially on fuzzing machines.

I would prefer to keep it simply. Meaning not fragment it to much. I could make it cpp only if oss or add an option. But the question is, does the added complexity is needed to solve any real use-case?

meson_options.txt
kasper93: Fair, I can change that. It's kinda unfortunate we need this at all, but some meson sanity checks are generting main function, so they cannot be linked with fuzzer, which already defined it too.

Granted if we were to only support libfuzzer it could be set implicitly, but I think an option is clean and connect to oss-fuzz nicely.

fuzz/meson.build
TheOneric: C-only toolchains certainly exist and there’s no point in requiring them here for standalone or any other native variants of the utility. Keep in mind that this is also used in testing and to conveniently reproduce crashes.

TheOneric:
The current commit appears to have a typo leading to nearly all meson jobs failing

> See: https://github.com/kasper93/oss-fuzz/commit/2709d7fed5303a0601c6e7c85792a2ba58258119

While not directly related to the path here, I’ve been thinking it may be best to move all OSS-Fuzz logic into this repo under `fuzz/oss-fuzz` or something. Then in the oss-fuzz repo only copy scripts etc from this subfolder after cloning. This way it wouldn’t be so much hassle to get any changes through anymore

kasper93:
> While not directly related to the path here, I’ve been thinking it may be best to move all OSS-Fuzz logic into this repo under `fuzz/oss-fuzz` or something. Then in the oss-fuzz repo only copy scripts etc from this subfolder after cloning. This way it wouldn’t be so much hassle to get any changes through anymore

That's certainly a possibility. Though I didn't do to for mpv either like that, because:
* Dockerfile still would have to be hosted upstream
* build.sh logic is not related to libass itself and more so the oss-fuzz build specifics. It's mostly unrelated code to libass repository. Also to test it, you still need to go through oss-fuzz. While the ci fuzz action is possible it does run differently than upstream tests when you merge it to oss-fuzz, but this also run different than actual build runners, so there is no winning.
* If they do build configuration changes upstream, they will update built scripts if needed, if this is something that can be done in bulk. Though it doesn't really happen often.

What I'm trying to say, build.sh logic almost fully oss-fuzz specific and generally shouldn't change on libass side. Of course this is not fully true, because any breaking changes would need to be adjusted. But this is basically a question if you want to keep this file in libass repository.

One thing I made sure when I made mpv script, is to make sure that adding new fuzzers is possible without changes to script. The rest I wasn't concerned about. Although it's true there were multiple changes needed to the build.sh script, than I initially expected.

```
meson compile -C build fuzzers
find ./build/fuzzers -maxdepth 1 -type f -name 'fuzzer_*' -exec mv {} "$OUT" \; -exec echo "{} -> $OUT" \;
```

fuzz/meson.build
```diff
@@ -3,12 +3,23 @@ fuzz_src = files(
     'writeout.c',
 )
 
+link_language = 'c'
+fuzz_ldflags = get_option('fuzz-ldflags')
+# Assume that if fuzz_ldflags is set, C++ linking is required.
+# This is necessary for OSS-Fuzz.
+if fuzz_ldflags != ''
+    add_languages('cpp', native: false)
+    link_language = 'cpp'
```
TheOneric, line 12: i can’t otoh say if there’s any setup needing `fuzz_ldflags` which _doesn't_ always come with a C++ toolchain, but at least a plain, non-OSS-Fuzz libfuzzer setup [needs this option](https://github.com/libass/libass/commit/9987eef44384b368670d962fca57b5994b216499) but is fine with just the C-interface of clang. 

I’d much prefer if such a change wouldn't just always or implicitly magically happen, but instead be an explicit toggle; whether it’s a specific "link fuzzer with C++" toggle or a special-purpose "oss-fuzz" toggle i don’t mind

TheOneric, line 12: nit: `fuzz_link_langauge` to make it clearer this only applies to the fuzzer binary

kasper93: To be honest, the "link with cpp is required" by oss-fuzz is bit hairy. But they established system like that, so I want to follow it to avoid like breakage of build when the do something stupid in env.

Unfortunately it doesn't work, now I know it will break regardless, one way or anther.

I will add an option `fuzz_link_langauge` to go in hand with `fuzz_ldflags`

TheOneric (approved):
Seems in principle fine to me now, but lets wait if or what thoughts maintainers with OSS-Fuzz access have on using meson here

TheOneric:
Well there were no further comments but it can’t hurt to add. Especially since `fuzz-ldflags` is apparently needed sometimes even outside OSS-Fuzz

PR #878, opened 2025-04-14
fuzz: allow use of portable fontconfig runtime config

kasper93:


fuzz/fuzz.c
```diff
@@ -400,6 +400,12 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     if (!LEN_IN_RANGE(size))
         return 0;
 
+#ifdef FONTCONFIG_SYSROOT
+#define STR_(x) #x
+#define STR(x) STR_(x)
+    setenv("FONTCONFIG_SYSROOT", STR(FONTCONFIG_SYSROOT), 1);
+#endif
```
TheOneric, line 407: It would be better to do this only once during `LLVMFuzzerInitialize`.
Also i think we should not encroach on fontconfigs macro namespace here and prefix this with `ASSFUZZ_`

kasper93: done

kasper93:
It works well with this branch on oss-fuzz side https://github.com/kasper93/oss-fuzz/tree/add_libass_msan

One remaining item to discuss is what fonts to include in the build. With that added, libass should be fully operational in fuzzing environment.

fuzz/fuzz.c
```diff
@@ -412,9 +412,8 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 
     ASS_Track *track = NULL;
 
-    // All return values but zero and -1 are reserved
     if (!init())
-        return 0;
+        exit(1);
```
TheOneric, line 416: Can you elaborate on what benefit this change brings?
Whether the fuzzer is restarted or continues with the next sample shouldn’t make much difference considering no state but a fontconfig env var persists across samples anyway, except for the former being slower

TheOneric:
> One remaining item to discuss is what fonts to include in the build.

To summarise from IRC: covering TrueType and Postscript outlines, including cubic splines might be a  good idea.
We aren’t sure whether it’s better to use minimal tiny fonts which are quick to parse and won’t take up much memory, or fonts with wide coverage.

Since no better fonts were proposed, we can perhaps just start out with the minimal version of FanSubBlock used as an embedded font in our regression tests: [fansubdoge_0.otf.zip](https://github.com/user-attachments/files/19840482/fansubdoge_0.otf.zip)

fuzz/fuzz.c
```diff
@@ -390,6 +390,16 @@ int main(int argc, char *argv[])
     return 0;
 }
 #elif ASS_FUZZMODE == FUZZMODE_LIBFUZZER
+int LLVMFuzzerInitialize(int *argc, char ***argv)
+{
+#ifdef ASSFUZZ_FONTCONFIG_SYSROOT
+#define STR_(x) #x
+#define STR(x) STR_(x)
+    setenv("FONTCONFIG_SYSROOT", STR(ASSFUZZ_FONTCONFIG_SYSROOT), 1);
```
TheOneric, line 398: this should also be set for `FUZZMODE_AFLXX_SHAREDMEM` before entering the sample loop

fuzz/fuzz.c
kasper93: This is critical error and should never happen. If it does, fuzz.c needs to be fixed. It does not depend on fuzzing input.

TheOneric: At least in AFL, a crash like this wouldn’t stop fuzzing either; instead it creates a bunch of bogus "crash" reports. Meaning if the fuzz environment is setup is so broken that we can't even init *(e.g. extremely restrictive memory limits)* we’d get a bunch of bogus reports.

kasper93: > At least in AFL, a crash like this wouldn’t stop fuzzing either

Good point, I've updated AFL code too.

> Meaning if the fuzz environment is setup is so broken that we can't even init (e.g. extremely restrictive memory limits) we’d get a bunch of bogus reports.

What do you mean by bogus? Those are very valid reports that fuzz environment setup is broken and cannot proceed.

Also note that any OOM would be caught earlier, because ASAN already aborts on any OOM. OSS-Fuzz is configured to 2G limit, I think by default ASAN defaults to 8G limit.

kasper93:
> I think there’s no issue per se, but i still don’t quiet get the motivation for `exit(1)` ing on init failure which almost certainly is an issue with the fuzzer setup, not libass.

Exactly, so it doesn't make sense to continue with this broken fuzzer setup, hance why we exit early to notify about the issue.

> This likely stems from different experiences how fuzz setups will handle this; could you again further elaborate on when and how you expect this to become visible?

exiting/aborting will generate the report that something is wrong. Not sure what is the question. `return 0` in current code is just hiding the issue and your fuzzing can spin in place without any progress, because of hidden failure that doesn't depend on input data. 

And especially if that failure is intermittent, you might get interesting input data generated that would be dropped, because of unrelated error.

kasper93:
> _How_ exactly will this be visible? _What kind of_ report is generated (both when running LLVM’s fuzzer standalone and as part of OSS-Fuzz)?

> Crash Type: 	Unexpected-exit

```
 	==410== ERROR: libFuzzer: fuzz target exited
	    <stack trace>
	
	SUMMARY: libFuzzer: fuzz target exited
```

https://issues.oss-fuzz.com/issues/392929029

> E.g. does the fuzzing process abort completely and if so what kind of error is indicated?

Fuzzing process aborts always on any error it founds. 

> Does it create a flood of ""buggy samples"" as fuzzing continues and nearly all samples ""fail""? Something else?

What do you mean by flood? Every similar fuzzing failure is merged into the same report, you don't get multiple reports for the same error.

kasper93:
> Is this OSS-Fuzz specific or does it also hold when running the LLVM fuzzer by itself with default arguments?

What is the question? LLVM fuzzer will abort on errors, yes.

PR #879, opened 2025-04-15
ci/gha/meson: add Windows Arm builds

kasper93:


.github/workflows/meson.yml
```diff
@@ -15,21 +15,39 @@ jobs:
...
 
+          - name: Windows (MSVC arm64)
+            os: "windows-11-arm"
+            arch: arm64
+            msvc: true
+            shell: 'pwsh'
+            # asm preprocessing is not supported by Meson, could be done manually with custom target
+            # https://github.com/mesonbuild/meson/issues/9889
+            args: '-Ddefault_library=shared -Dasm=disabled'
+            devenv: |
```
TheOneric, line 36: unless MSVC is using gasm/llvm-as, afaik we don’t intend to support MSCVC’s assembler anyway

TheOneric (commented):
do we really need both MSVC and clang for Windows aarch64 builds?
Is there anything sufficiently different between x86_64 and aarch64 MSVC to expect any difference wrt compiling C? *(Differences in Windows’ API should be uncovered by clang too)*

kasper93:
> do we really need both MSVC and clang for Windows aarch64 builds? Is there anything sufficiently different between x86_64 and aarch64 MSVC to expect any difference wrt compiling C? _(Differences in Windows’ API should be uncovered by clang too)_

In perfect world, we would need only a single build as "everything" should be compatible. But we know that there are differences in mingw vs Windows SDK. And on top of that different target architectures may have different issues. If you think it's not worth testing, we can skip it. I don't think it's that critical to test, but at the same time it was easy to add and symmetric with existing jobs.

.github/workflows/meson.yml
kasper93: Do you suggest to remove the comment? asm is disabled and probably remain as such for aarch64 build, unless someone contribute changes needed to make it work. I don't even have hardware to work on this if I wanted to.

TheOneric: As is the comment is misleading as it suggests ASM would start to work if the meson issue was resolved or worked around — but it actually will never work with this toolchain’s assembler.
If the job was staying, i would replace the comment to say *“MSVC’s assembler is not supported”*

TheOneric:
We are already over the limit for simultaneous GHA jobs and we cover both MSVC and aarch64-Windows already, so imho if there are no known additional divergences, this job isn’t worth the extra time and cache space it uses up.

> Also those build are also useful if you ever decide to share build artifacts

I expect this to be pretty unlikely. I don’t think anyone here is interested in dealing with distribution which requires knowledge of the target platform and dealing with licence requirements of also all our (then bundled) deps. I’m happy to leave distribution to downstreams with expertise in their respective target setup.

.github/workflows/meson.yml
kasper93: > As is the comment is misleading as it suggests ASM would start to work if the meson issue was resolved or worked around — but it actually will never work with this toolchain’s assembler. If the job was staying, i would replace the comment to say _“MSVC’s assembler is not supported”_

I'm not sure I follow. If the mentioned issue would be resolved, ASM code will be well supported. The only missing part is pre-processing. MSVC assembler is just converting ASM to object binary.

For example ffmpeg is using custom preprocessor for ARM where assembler doesn't do preprocessing (Apple and MSVC) https://github.com/ffmpeg/gas-preprocessor

PR #884, opened 2025-05-31
Move ART CI jobs into main repo

TheOneric:
This adapts the VM and qemu-user jobs from libass-tests, omitting configurations already covered by existing CI jobs.

We no longer need to exhaust our entire cache allowance for this and being in the main repo means tests always run automatically on each commit and even pull request. This also avoids the need to manually reenable scheduled CI runs in libass-tests  over rand over again if there were no changes to the test repo for a while.


astiob:
Hm, what has changed regarding the cache?

Issue #886, opened 2025-06-09
Adding animated transform to \pos


jeybee:
I have locally altered libass to add support for the \pos tag within \t to allow for use of the acceleration parameter. It was easy to implement, just a few lines, and allows {\move(0, 0, 200, 200, 0, 500} to be replaced with {\pos(0, 0)\t(0, 500, 1, \pos(200, 200)}. It also allows for multiple pos transitions within one line.

When trying to match video text movement, I find linear movement rarely works so acceleration is key.

Is there a reason to not support this? I'm just wondering why there's this limitation on pos when the functionality to support it is basically already there.

techguru0:
we need ALOT more tags for simpler typsetting
theres already a bunch of ones added to the modded vsfilter

PR #889, opened 2025-06-16
ASS/SSA Format: add animated transform for \pos behind header

jeybee:
Adding support for \pos in \t tags to allow use of acceleration and multiple movements within a single tag.

More discussion here: https://github.com/libass/libass/issues/886

### DIFF
--- a/libass/ass_parse.c
+++ b/libass/ass_parse.c
@@ -610,14 +610,18 @@ char *ass_parse_tags(RenderContext *state, char *p, char *end, double pwr,
                 v2 = argtod(args[1]);
             } else
                 continue;
-            if (state->evt_type & EVENT_POSITIONED) {
+
+            int allow_transform = state->renderer->track->PositionTransform;
+
+            if (state->evt_type & EVENT_POSITIONED && 
+                (!allow_transform || (allow_transform && !nested))) {
                 ass_msg(render_priv->library, MSGL_V, "Subtitle has a new \\pos "
                        "after \\move or \\pos, ignoring");
             } else {
                 state->evt_type |= EVENT_POSITIONED;
                 state->detect_collisions = 0;
-                state->pos_x = v1;
-                state->pos_y = v2;
+                state->pos_x = allow_transform ? v1 * pwr + state->pos_x * (1 - pwr) : v1;
+                state->pos_y = allow_transform ? v2 * pwr + state->pos_y * (1 - pwr) : v2;
             }
         } else if (complex_tag("fade") || complex_tag("fad")) {
             int32_t a1, a2, a3;
--- a/fuzz/ass.dict
+++ b/fuzz/ass.dict
@@ -47,6 +47,7 @@
 "YCbCr Matrix"
 "Kerning"
 "Language"
+"PositionTransform"
 ## Interesting Script Info Header Values
 "v4.00"
 "v4.00+"
--- a/fuzz/writeout.c
+++ b/fuzz/writeout.c
@@ -112,6 +112,7 @@ static void write_header(FILE *f, ASS_Track *track, const char *originalformat_n
     HEADER_BOOL(ScaledBorderAndShadow);
     HEADER_BOOL(Kerning);
     HEADER_STR(Language);
+    HEADER_BOOL(PositionTransform);
     if (track->YCbCrMatrix != YCBCR_DEFAULT) /* Or normalise this to TV.601? */
         fprintf(f, "YCbCr Matrix: %s\n", ycbcr_to_str(track->YCbCrMatrix));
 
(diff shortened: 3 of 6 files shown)


libass/ass_parse.c
```diff
@@ -610,14 +610,18 @@ char *ass_parse_tags(RenderContext *state, char *p, char *end, double pwr,
                 v2 = argtod(args[1]);
             } else
                 continue;
-            if (state->evt_type & EVENT_POSITIONED) {
+
+            int allow_transform = state->renderer->track->PositionTransform;
+
+            if (state->evt_type & EVENT_POSITIONED && 
+                (!allow_transform || (allow_transform && !nested))) {
```
astiob, line 617: I’d be inclined to align the behaviour to other transformable tags and just have repeated tags reset the value:

```suggestion
                !allow_transform) {
```

astiob (commented):
Not sure how I feel about locking this behind a specific Boolean header. (Maybe it’s time for something like `OptionalFeatures: a,b,c,d`? Or maybe a survey of existing files will reveal that repeated and wrongly nested \pos tags are actually effectively nonexistent and this can be enabled unconditionally?)

But FWIW here’s an as-is review, and thanks for volunteering to implement this!

arch1t3cht:
I agree that this may be the time to settle on some sensible format for future feature flags. I don't yet have any too strong opinions on *which* format, but before yet another header is added (that then has to be supported in all kinds of authoring tooling), it'd be nice to figure out a system that makes adding future such feature flags easier.

Systems like `OptionalFeatures: a,b,c,d` would also allow for renderers or tooling to detect when a script depends on features they aren't aware of and e.g. print a warning (while general headers are also used as a free field for various project metadata, so unknown headers may not mean much).

A few thoughts I *do* have about the format are that
a) it should be as easy as possible to opt into all currently available extensions
b) I am slightly concerned about some users deciding to explicitly opt out of certain features that somehow [break their workflow](https://xkcd.com/1172/). This is probably unavoidable to some degree, especially as some tooling may never be updated to match new extensions, but what I'm worried about specifically is that convoluted combinations of the form "opt out of feature X but opt into features Y and Z" could arise, which could become a maintenance burden. I'm not yet sure if it's really a good idea, but I feel like one should at least think about making the feature toggles a linear scale `OptionalFeaturesVersion: 5` instead. This has other potential downsides like centralizing feature additions (though this isn't *currently* a problem since all real development currently happens at libass) and introducing yet another version next to the `ScriptType` (though this is arguably the case either way), but I at least wanted to bring it up and give my reasoning.

Zahuczky:
In my head I've been toying with different ideas for a long time on how to handle new features of libass if it ever comes to that, so what about defining future levels for user facing interfaces simply as libass versions?

To handle different compatibility levels, internally it could rely on of a switch just like mentioned above,
```
1.2.3
    a,b
    
1.2.4
    a,b,c,d
```

This may make it more manageable if the number of optional features grows in the future.

I don't know too much about the politics of the different currently in-development renderers, but assuming we take libass as the de-facto ass renderer of today, it might be a reasonable expectation of others to align themselves with the feature level versioning used by libass.

This could also introduce an easy way for subtitle authors to specify version-based messages in their subtitles, to eg. notify their audience if they're using a renderer which lacks some of the optional features the subtitle used. 


Not sure if it should be a header thing, or an event thing, but for example:
Let's say in the header the feature level is set as:
`Feature level: 1.2.3`
and an event as such is used: 
`{\compat\an5}You're using a video player with an outdated renderer, pls update`
which will only render on versions lower than specified (and by nature, any renderer that doesn't support this way of specifying feature levels, eg everything up to this point)

I'm leaning towards it being an event, (maybe not with a tag, but an effect?) since multilingualism is a core principle of subtitling, so people authoring subtitles in different languages should be able to provide their own message to their own audience.
This would also allow authors to make decision based on how severely they think the compatibility level will affect their subs, and tune their warning based on that. If they believe the subtitles will be severely borked when using an old renderer, they can make a big, obtrusive warning, or if they believe at most it'll make some signs not blend in as well, it could be a small, few second long event at the start. 

Now that I think about it, if not tied to a libass version, a simple integer of "feature level" could be also be used for this, starting with 0 being today, so mostly vsfilter compatible. 

Authoring tools such as aegisub could also check for the usage of different optional features in the file, detecting and gently pushing users to provide their own compatibility message.

astiob:
Just a quick note: the reason I suggested feature flags rather than version numbers is that different features are more or less difficult to implement in different renderers. Even if one considers libass to be the only actively developed renderer at the moment (which I may not agree with), strictly linearizing the set of features might make it more difficult to develop a new renderer in the future.

PR #891, opened 2025-07-03
Scale margin values received from user override style.

X2E4VXpZKv:
These are also effected by script resolution and the override style is supposed to be resolution independent.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -991,9 +991,9 @@ static ASS_Style *handle_selective_style_overrides(RenderContext *state,
         new->Justify = user->Justify;
 
     if (requested & ASS_OVERRIDE_BIT_MARGINS) {
-        new->MarginL = user->MarginL;
-        new->MarginR = user->MarginR;
-        new->MarginV = user->MarginV;
+        new->MarginL = user->MarginL * scale;
+        new->MarginR = user->MarginR * scale;
+        new->MarginV = user->MarginV * scale;
     }
 
     if (!new->FontName)


kasper93:
I have not check this in details. But on the surface it looks iffy. First of all scale here is defined as `scale = render_priv->track->PlayResY / 288.0;`. Since those are L R margins too, this would probably make more sense to scale by `PlayResX / 384.0`, no?

In mpv we do this to fix margins, after playres change. 

https://github.com/mpv-player/mpv/blob/8cc65c838e2a688f1e9af9f12cd17792961265c0/sub/sd_ass.c#L625-L646

``` c
            // ffmpeg and mpv use a default PlayResX of 384 when it is not known,
            // this comes from VSFilter.
            double fix_margins = track->PlayResX / (double)MP_ASS_FONT_PLAYRESX;
            for (int n = 0; n < track->n_styles; n++) {
                track->styles[n].MarginL = lrint(track->styles[n].MarginL * fix_margins);
                track->styles[n].MarginR = lrint(track->styles[n].MarginR * fix_margins);
                track->styles[n].MarginV = lrint(track->styles[n].MarginV * set_font_scale);
            }
```

We don't use / expose `ASS_OVERRIDE_BIT_MARGINS` itself though.

X2E4VXpZKv:
I was trying to get overriden margins to be consistent when using the same window dimensions (independent of the script resolution/aspect ratio). I did also think about the horizontal scaling being wrong after submitting this pr if I remember correctly but scaling by playresx was not correct, or maybe i just tested it wrong not sure, it was 8 months ago.

I have also have a patch trying to fix different aspect ratios [here](https://github.com/X2E4VXpZKv/libass/commit/5a39010d75215ea202e49b8e2ad52aa2ef6ec988) which does seems to work but by the end was mostly just trial and error so my math could be even more wrong there.

Files that i tested with and a patch to mpv to enable margin overrides are [here](https://github.com/X2E4VXpZKv/mpv/tree/master/subtest) (ass is second subtitle track).

TheOneric:
> I have not check this in details. But on the surface it looks iffy. First of all scale here is defined as scale = render_priv->track->PlayResY / 288.0;. Since those are L R margins too, this would probably make more sense to scale by PlayResX / 384.0, no?

I see your point, but note this already now scales e.g. x shadow and outline width using the `scale` value based on the relative `y` extent of the canvas. And since this works on _style_ definitions there isn’t even a separate parameter for the `x` extent of these properties, meaning this cannot be changed.

But even when calculating a `x` specific scaling factor for margins, I’m not sure any arbitrarily chosen `x` base size will actually be better. Afterall the actual video might have any aspect ratio, but by chosing a pair of `x` and `y` base extent values we’ll assume a particular AR leading to odd results if it doesn’t match.  
The scale factor just ensures the same style override can be used for different video up/downscales. In practice the margins will need to be adjusted for different aspect ratio videos anyway and a single scale value at least gives more consistent behaviour imho

--------

> I have also have a patch trying to fix different aspect ratios [here](https://github.com/X2E4VXpZKv/libass/commit/5a39010d75215ea202e49b8e2ad52aa2ef6ec988) which does seems to work but by the end was mostly just trial and error so my math could be even more wrong there.

By replacing and removing the basic `*2scr*` functions, this changes/breaks behaviour of regular tags and events, not just user overrides.

TheOneric:
> In mpv we do this to fix margins, after playres change.

Which means, mpv also does it wrong atm, since it’s scaling value for margins has no relation at all to the ratio of the original and new `PlayResX`. It used to do it conceptually correctly before since https://github.com/mpv-player/mpv/commit/f862d3b6cd66abee98f2af21c36255ef7361cb80 but this was reverted it seems because of mpv sometimes puts garbage values in `PlayResX` before the scaling happens if i understand https://github.com/mpv-player/mpv/pull/15538#issuecomment-2550728511 correctly?  
Imho mpv should (as this comment suggested) make ure to get the correct, original `PlayResX` instead of wrongfully assuming some constant previous size *(see the reverted fix and my comments on its PR)*

This isn’t applicable to scaling user overrides though, since we don’t change `PlayRes*` values here and work independently of any AR.

kasper93:
@llyyr: Could you take a look at this topic?

llyyr:
Yes, the hardcoding to 384 fix was misguided and I've fixed it correctly here: https://github.com/mpv-player/mpv/pull/17846

kasper93:
> Yes, the hardcoding to 384 fix was misguided and I've fixed it correctly here: [mpv-player/mpv#17846](https://github.com/mpv-player/mpv/pull/17846)

Thank you, @TheOneric and @llyyr for making this fix happen.

kasper93:
> The scale factor just ensures the same style override can be used for different video up/downscales. In practice the margins will need to be adjusted for different aspect ratio videos anyway and a single scale value at least gives more consistent behaviour imho

Again, I've not gone into details, but intuitively it seems that by "respecting" aspect ratio it could possibly work better without manual adjustment. There is no issue obviously with up/downscale while keeping the same ratio, because both x and y scaling ratio should be the same in such case.

It's a choice. Scaling only by y scale is fine too, so long we are consistent about it. Would be good to align mpv to whatever is preferred by upstream. The extent of the issue is unclear to me, however I see the inconsistent/different behavior which I wanted to flag.

I trust you guys to figure it out, I'm just passerby complainer.

TheOneric:
> Again, I've not gone into details, but intuitively it seems that by "respecting" aspect ratio it could possibly work better without manual adjustment

The situation for mpv’s adjustemnts after changing `PlayRes*` and the selective user stale override libass API modified by this  change here is completely different.

For reference, here are the relevant API functions for "selective" style overrides *(not to be confused with text-based style addition and overrides from `ass_read_style` or the string-based partial overrides from `ass_set_style_overrides`)*:
```C
// determine _when_ to apply the override and _which parts_ of the override
void ass_set_selective_style_override_enabled(ASS_Renderer *priv, int bits);
// set the one, single style override applied on every style in the file
void ass_set_selective_style_override(ASS_Renderer *priv, ASS_Style *style);
```

These overrides operate solely on styles alone (not in-line override tags) and provide a single source style in a vacuum. We have absolutely _no idea_ what AR or other considerations the source style was designed for.
The rescaling from an arbitrarily chosen base `PlayResY` makes it possible to keep the same user-override style for different resolutions of the same aspect ratio. But that’s it and there’s nothing we can do _because we don’t know_ about mismatching ARs.  
Nor do we know what the intent of the override style author is wrt to margins. Was it to ensure there at least N pixels of padding between text and the edges? Or at least N% of the video frame padding? Or maybe it was not about the distance to the edge per se but to restrict the _dialogue text lines_ to be at most N canvas units wide. When changing the AR (and storage size, or perhaps even render frame size), the ideal rescaling of margin values differs for each of those scenarios.

While when mpv changes its `PlayRes*` values *(for auto-converted subs iirc?)*, both the original style values _and its target environment, including AR_ are known. And so is the target environment. Thus it is possible to make meaningful scaling decisions for `x` and `y` independently.  
The circumstances and goals here are different so I don’ŧ think there’s any need for the scaling behaviour between the two to match. If you’d wanted to match and it was a good fit here, you’d just be using this API afterall isntead of implementing it yourself.

If you want to discuss mpv’s behaviour more, it would be good to move this to an actual mpv issue though, to not derail this PR too much.

Issue #538, opened 2021-08-19
Rendering direction differ from Aegisub


Bayany:
libass :

![image](https://user-images.githubusercontent.com/60580365/130065251-416a441d-9da7-40bb-873c-d7b7bea4a6b4.png)

 Aegisub v3.2.2

![image](https://user-images.githubusercontent.com/60580365/130066783-9b6a3db2-ad8b-494b-8f38-0c4db5f5e577.png)

Text ⬇
{\pos(935.273,202.364)\fnSG Kara SemiBold\fs110\b1\shad0\bord1\3c&HEEF8F7&}{\fs110\c&H71FF2C&}!{\c&H2CABD4&}‫نمی‌خوره{\fs110\c&H71FF2C&} شکست{\i0\fs110\c&H2CABD4&} جاهی ساما{\i0}   

it just happens when each word has a different style

TheOneric:
Thanks, for reporting. This is already being worked on and will be fixed soon™ once #441 gets merged.

As of master a8456e673c473a081febb4c2da5ba90457574c6d and pr#441 ead32a18ffe2c049eb42bfb6369ee256799de824:
![grafik](https://user-images.githubusercontent.com/1668471/130115820-f86ef024-c35a-4026-a817-6f4edf9b57a1.png)

Bayany:
I found another compatibility issue with ordering ... it is a bit different, so I thought I should share it.
![image](https://user-images.githubusercontent.com/60580365/130333038-8c4d8dd2-e129-4b3d-af64-00387af1bffe.png)

Text:
21م آوریل 2095 - بعد از مدرسه

TheOneric:
I get the same result matching you “libass” rendering, regardless of PR#443 *(with fribidi 1.0.5-3.1+deb10u1 and harfbuzz 2.3.1)*.
Though not authorative on this matter, interestingly recent Firefox (using a differnt version of harfbuzz and fribidi) and nano agree with the libass rendering, but wxSCT and presumably native GTK3.24.5 elements (eg in Aegisub and pluma) agree with your “Aegisub” result.

However, we need more info. Aegisub can use all sorts of different VSFilters and in fact also libass, so labeling the result “Aegisub” doesn't tell us much. Look in Aegisub's options at `Advanced → Video → Subtitle-Provider` to get a rough idea of what's rendering your subtitles. If you know more, like the exact flavour and version of this subtitle-provider that would be great.
[...]

Issue #295, opened 2018-01-01
libass no longer changes base direction based on encoding


lirbar:
I was referred here from VLC (VideoLAN) [forums](https://forum.videolan.org/viewtopic.php?f=12&t=141606)

It seems libass doesn't respect ASS Encoding flag. I specifically testing Hebrew subtitles with latest nightly VLC (3.0.0 rc2) on latest macOS 10.13.2 - which have the following issue. While previous VLC versions 2.2.8 doesn't (probably since it used older libass version).

Hebrew is written from right to left, so even when VLC knows that the subtitle is in Hebrew (track's language) and ASS Encoding flag is set to 177 - it's still shows LTR punctuations which means that " . ? ! , " shows in the beginning and not in the end of the sentence.
This means that the punctuations shown in the wrong direction - like English (ASS encoding flag = 0).

Here is the info about ASS Styles Encoding flag:
https://en.wikipedia.org/wiki/SubStation_Alpha
Encoding. Useful for writing other alphabet characters in ASCII. With Latin characters, this is set to zero. With Japanese characters, it can be set to 128 (ex: ‹ó displays Sora kanji). With Hebrew characters, it can be set to 177.

As said, the issue doesn't exist (ASS encoding flag 177 leads to RTL punctuations and flag 0 leads to LTR - as it should be) on VLC 2.2.8 in the same macOS system.

anoosi:
use -1 flag and libass will detect the direction automatically

lirbar:
Thanks that works nicely, but what with older ASS files that have the flag set with the old 177 flag ?
Is it a lot of work to support also the 177 flag, as was in the past ?

anoosi:
the same with me :(
i use arabic language, it's flag was 178 rtl also
but this option change from 0.13.3 version you can see the changelog
https://github.com/libass/libass/blob/master/libass/ass_shaper.c
see the comment in the last

lirbar:
Oh, I see. Anyone from the dev team can help us fix both Hebrew and Arabic by updating the code to the following ?

/**
 * \brief Resolve a Windows font charset number to a suitable base
 * direction. Generally, use LTR for compatibility with VSFilter. The
 * special value -1, which is not a legal Windows font charset number,
 * can be used for autodetection.
 * \param enc Windows font encoding
 */
FriBidiParType resolve_base_direction(int enc)
{
    switch (enc) {
        case -1:
            return FRIBIDI_PAR_ON;
        case 177:
[...]

PR #901, opened 2025-08-01
ass_shaper: map Encoding=-2 to FRIBIDI_PAR_RTL

llyyr:
This is useful for forcing RTL rendering of text if they come from sources that sideload text layout information, such as Youtube where the text layout is set externally via CSS and isn't present in the sub file itself. It's not trivial to detect those cases, but it is trivial to provide a way to force RTL if the user chooses to.

### DIFF
--- a/libass/ass_shaper.c
+++ b/libass/ass_shaper.c
@@ -1130,13 +1130,15 @@ FriBidiStrIndex *ass_shaper_get_reorder_map(ASS_Shaper *shaper)
 /**
  * \brief Resolve a Windows font charset number to a suitable base
  * direction. Generally, use LTR for compatibility with VSFilter. The
- * special value -1, which is not a legal Windows font charset number,
- * can be used for autodetection.
+ * special values -1 and -2, which are not legal Windows font charset numbers,
+ * can be used for autodetection and forcing RTL, respectively.
  * \param enc Windows font encoding
  */
 FriBidiParType ass_resolve_base_direction(int enc)
 {
     switch (enc) {
+        case -2:
+            return FRIBIDI_PAR_RTL;
         case -1:
             return FRIBIDI_PAR_ON;
         default:


astiob:
If this is to be used together with an `ASS_Feature` anyway, why not make this an `ASS_Feature` of its own right? Or introduce a new dedicated base-direction field in `ASS_Style`.

Personally, I dislike the use of `Encoding` for this. If it didn’t break some files that have gotten released into the wild since libass introduced this, I’d rather see `Encoding` −1 gone, too. Not only are these modes unsupported in other renderers (VSFilter variants), but they also actually _conflict_ with the existing interpretation of this field in VSFilter (although tbf, unlike −1, −2 seems to not map to any valid charset value, so uhh I guess VSFilter will always fall back to a default font? I’m not sure) and take away the ability to actually choose the desired font charset. On the other hand, nobody actually uses negative values in VSFilter in practice, and RTL base direction within runs could be fairly easily added to VSFilter too (and explicitly choosing the charset isn’t ever strictly necessary). On the third hand, using a field named `Encoding` to control base direction doesn’t make intuitive sense, anyway.

If this mode doesn’t actually flip layout order (as mentioned), then this seems only more undesirable: what use is it to introduce such a half-measure into the file format? Once added, it’ll stick with us forever, and any full-featured renderer will have to support LTR layout built from RTL blocks, and yet it will keep being surprising or useless to subtitle authors.

It’s worth noting that base RTL can currently be achieved (assuming `ASS_FEATURE_WHOLE_TEXT_LAYOUT`) by prefixing every event and `\N`-delimited line with a Unicode RTL mark, but I can see how an API toggle may be more sensible.

astiob:
> why not make this an `ASS_Feature` of its own right? Or introduce a new dedicated base-direction field in `ASS_Style`.

The latter could also be supported in user style overrides, like the `Justify` field.

Alternatively yet, −2 could implicitly enable `ASS_FEATURE_WHOLE_TEXT_LAYOUT` like −1 does, but that would further decrease the chance of it ever being supported in VSFilter.

llyyr:
> Alternatively yet, −2 could implicitly enable ASS_FEATURE_WHOLE_TEXT_LAYOUT like −1 does, but that would further decrease the chance of it ever being supported in VSFilter.

I'm fine with doing this.

> However, it increases the chances of actual _files_ rather than merely ephemeral converted data relying on a libass-exclusive feature _(which to be clear, is undesirable)_.

IMHO if we don't want to expose this for files to use, we should make Encoding={-1,-2} both ASS_Features. I'm waiting for us to reach some understanding before making changes to the PR.

astiob:
@llyyr Sorry this isn’t going fast, and thanks for your patience. I trust we’ll get there in time.

---

Personally, I dislike the very idea of using `Encoding` for any kind of extensions unrelated to font charsets, but we _could_ also make an `ASS_Feature` that enables the ability to set `Encoding=-2` (and/or indeed `-1`). Then, you’d have per-style control, but regular ASS without the feature enabled would continue to not support this.

If we’re considering an extension anyway, we could introduce a new style property dedicated to base direction. I mentioned an API field before, but it could even be a parsable column/field that one could specify in the style section’s `Format` line, since we still honour those. Again, this could be either unconditionally available or guarded by an `ASS_Feature`. (While we’re at it, we could add `Justify` there, too.)

If the main goal of the change is to enable a user override, then a new overridable style property like `Justify` seems perfect to me: it achieves the goal and has no effect on ASS portability.

If the main goal is to simplify automatic subtitle conversions, then I’m guessing an always-on format extension is indeed the simplest because format converters probably aren’t aware of `ASS_Feature` and user overrides? But perhaps it would be better to extend converters so they _are_ aware of them instead.

If per-style control is unnecessary, then another possible extension (also suggested in #295) is a new ASS header. It could have a clear meaning, obviously be an extension, avoid hijacking an existing format feature, and be settable both in text (in an automatic converter from WebVTT) and in a user override. I’d be inclined to have it enable `WHOLE_TEXT` when present (except perhaps if set to a special value that explicitly requests the legacy behaviour). Personally, I think per-style control is more useful, but this seems quite attractive.

> Reminder that extending `ASS_Style` _always_ breaks ABI necessitating a major soname bump which in the past you’ve seemed to prefer avoiding.

Sure, but if it needs to be done, then it needs to be done. I don’t think we should prefer to sacrifice common sense and stuff unrelated functionality into existing, usable fields just to avoid an ABI bump.

Technically, there is a way to do this without breaking ABI: keep a second, private array of “extended properties of styles” (“base direction” being the single extended property for now), keep its length in sync with the primary style array, and provide accessor functions to get/set these properties.

At some point, we should really consider moving away from direct array access in the API/ABI, although that in itself will, of course, involve an API/ABI break.

> it is possible to combine ["RTL base direction"] with a default, enabled and disabled state of `ASS_FEATURE_WHOLE_TEXT`. Generally it should then work correctly with _any_ `WHOLE_TEXT` state, which means it shouldn’t just ignore the layout direction issue.

Maybe I misunderstand, but this sounds backwards to me. If the new toggle sets _only_ the base direction flag, then it is indeed combinable with any `WHOLE_TEXT` state and does what it says on the tin in small print: sets the base direction of each run when `WHOLE_TEXT` is disabled or of each line/paragraph when it’s enabled.

For simple one-liners, this would actually be plenty enough, and even on lines with override tags, it would simplify RTL authors’ eternal pains about leading/trailing punctuation getting rendered at the wrong end of the run while still providing a clear and predictable workaround for override tag ordering. (Which might make them want to use it in actual releases, which is where VSFilter compatibility could be beneficial.)

llyyr:
> imho Encoding=-2 is the most realistic approach to actually realising VSFilter support and making it something RTL-language authors can actually utilise. Due to the benefits for RTL-language authors, I believe this is the ideal (realistic) outcome.

RTL language authors don't actually need Encoding=-2 at all. Encoding=-1 along with implicit directional marks should suffice already. Encoding=-2 is needed only for cases like playback of subtitles where the directional information is sideloaded through something external. (In this case, css).

llyyr:
What is actually desired here? Should I make this an ASS_Feature flag, or Encoding=-2 where it also enables WHOLE_TEXT or Encoding=-2 but without enabling WHOLE_TEXT? Currently the PR hints RTL and enables WHOLE_TEXT with Encoding=-2 set.

TheOneric:
I’d prefer to have `Encoding=-2` be independent of `WHOLE_TEXT`. To my understanding only RTL-by-default but `WHOLE_TEXT=false` semantics are feasible to implement in VSFilters and imho any (new) in-file extension ought to be *(at the very least possible to be)* implemented in VSFilters too. And the ability to set "RTL by default" does seem like it would be useful for authors of RTL languages.

It’s admittedly a bit odd to use an `Encoding` field to toggle the default base direction, but given `Encoding=-1` is already well established at this point and current VSFilters will ignore negative values, I’m personally fine with continuing to use it for an "RTL-by-default" in the form of `Encoding=-2`. If that’s really a concern, we could also introduce a new `TextDirection: [ ltr | auto | rtl ]` header. Except that as discussed before `auto` does not seem feasible to implement in VSFilters. Meaning only `ltr` (the default) and `rtl` would be portable ASS.

I remain confused about parts of @astiob’s earlier comments though and likely simply misunderstand them *(see my previous comment)*; a clarification would be appreciated.

llyyr:
I've removed the implicit WHOLE_TEXT part of the commit

astiob:
Just to clear up confusion around previous comments:

> > we could also make an ASS_Feature that enables the ability to set Encoding=-2 (and/or indeed -1). Then, you’d have per-style control,
> 
> Can you elaborate on the what interface you imagine for per-style `ASS_Feature` s? Otoh only unergonomic and/or artificially limited options come to my mind, stemming from `ASS_Feature`s taking just a single integer value.

The hypothetical ASS_Feature would be a Boolean that enables or disables the interpretation of Encoding=-2 as “RTL base direction”, which, in turn, is a per-style property. With the ASS_Feature disabled, Encoding would [aim to] be strictly compatible to older VSFilter.

If there’s something else you’re confused by, could you please quote/repeat the particulars? Thanks.

TheOneric:
> I've removed the implicit WHOLE_TEXT part of the commit

It’s still not laying out the runs themselves from right to left though *(first comment)*, right?
I believe there was more discussion about how this affects the results and why it is (not?) desirable, but I can't recall the conclusion.

To make sense as a new extension it should (as mentioned before) be sufficiently easy to implement in VSFilters, but also _make sense_. I’m not sure whether LTR-arranged runs of RTL text are sensible.

astiob:
On the main topic, _personally_ I remain hesitant about using `Encoding` for something completely encoding-unrelated and thereby removing the ability to actually specify the font charset. On the other hand, this does seem to be the most pragmatic and feasible way of adding new features with minimal disruption to parsing, since font charsets are barely useful even when supported, so I’m prepared to see us proceed with this.

If we want to take an incremental-improvement route, we _could_ do “RTL base direction within run” and “RTL layout of whole runs” as two separate and potentially independent steps. Even if we _only_ do the former, that’s still an improvement for RTL authors as it means they no longer have to work around punctuation placement quirks.

If we dream of making them available in `Encoding`, then I suppose we must either expect to make a dedicated new negative value for every plausibly useful subsequent combination of features or make the `Encoding` a bit mask (1 = bidi/shaping across runs and auto base dir by default; 2 = base RTL; perhaps 4 = RTL runs).

Two new things’ve just popped into my head:

1. We could make the `Encoding` values _positive_ but larger than 255. Put the new feature flags in higher bits, and keep the font charset in the lowest byte. This would keep the ability to set the font charset and in fact keep it compatible with older/current VSFilters, although they would, of course, get the bidi wrong.

2. `Encoding` is a per-run property that can be changed mid-line using `{\q...}` and `{\r...}`. How does this interact with positive and negative values? How _should_ it all interact? Even if `-1` already behaves badly, it might be wise to avoid repeating its mistakes. `WHOLE_TEXT` is definitely a per-event property, but base direction _within_ run can easily be per-run. One could even see it as an explicit bidi control element for simple cases.

TheOneric:
> The hypothetical ASS_Feature would be a Boolean that enables or disables the interpretation of Encoding=-2 as “RTL base direction”, which, in turn, is a per-style property

Oh ok, this makes more sense. Though if we’re going to add `Encoding=-2` to VSFilters, this seems counter-productive *(and also a reminder that [current VSFilter already treat negative values as a `1` anyway](https://github.com/pinterf/xy-VSFilter/commit/102897189902658ec68ef691eb3dd8d89f099985))*. Furthermore, we currently do not implement VSFilter-like font slection for positive encodings other than `1` anyway. Thus, at least at the moment the usefuleness of such a toggle seems questionable. Though we could figure out whatever held it back until now and implement the `encoding`-related font selection restrictions.
We don’t know of any releases actually using font encoding `255` or `-1` however.

> If there’s something else you’re confused by, could you please quote/repeat the particulars? Thanks.

For one, the following part doesn’t make sense to me, since the way I read it both what you quoted and then describe as the actual result are the same. Perhaps I’m misunderstanding either what you wrote or how the implemtnation will/should behave?

> > it is possible to combine ["RTL base direction"] with a default, enabled and disabled state of ASS_FEATURE_WHOLE_TEXT. Generally it should then work correctly with any WHOLE_TEXT state, which means it shouldn’t just ignore the layout direction issue.
> 
> Maybe I misunderstand, but this sounds backwards to me. If the new toggle sets only the base direction flag, then it is indeed combinable with any WHOLE_TEXT state and does what it says on the tin in small print: sets the base direction of each run when WHOLE_TEXT is disabled or of each line/paragraph when it’s enabled.
> 
> For simple one-liners, this would actually be plenty enough, and even on lines with override tags, it would simplify RTL authors’ eternal pains about leading/trailing punctuation getting rendered at the wrong end of the run while still providing a clear and predictable workaround for override tag ordering. (Which might make them want to use it in actual releases, which is where VSFilter compatibility could be beneficial.)

Then and also more generally, I’m unclear on what ABI and API changes yu consider acceptable for this and other features. Since several of the options you proposed break ABI or even API. I agree that they can be improved (see e.g. the wiki page), but in the past concerns about ABI stability are what stalled implementing parsing of v4++ files in (after splitting it out) #651, so this seems at odds.

----

Since we’re talking about implementability in VSFilter, it’s probably also worth discussing how exactly a theoretical patched VSFilter would behave. Otoh I believe the idea was to simply pass a flag indicating RTL base direction to GDI. I’m assuming this means RTL-arranged blocks of RTL text runs?

astiob:
> For one, the following part doesn’t make sense to me, since the way I read it both what you quoted and then describe as the actual result are the same.

We probably described the same thing in distinct words. I’m afraid I can’t comment further because your original comment (the one I quoted) doesn’t itself quote whatever it was meant to be a response to. We clearly did have a misunderstanding at some point, but it seems like we’re on the same page now.

> I’m unclear on what ABI and API changes yu consider acceptable for this and other features.

If a feature needs an ABI break to be sensible/useful, then it needs an ABI break. We should avoid ABI breaks when we can help it, but it shouldn’t force us to shoehorn features into undesirable forms that will haunt us forever. Does this make sense?

> Since we’re talking about implementability in VSFilter, it’s probably also worth discussing how exactly a theoretical patched VSFilter would behave. Otoh I believe the idea was to simply pass a flag indicating RTL base direction to GDI. I’m assuming this means RTL-arranged blocks of RTL text runs?

Simply passing a flag to GDI (the simplest possible change) means LTR-arranged blocks of RTL text runs. Which, based on your comments, is exactly how this PR behaves right now.

(I see that I initially viewed this as an undesirable half-measure, but if this is all we can muster at this time and we hope to some day™ also add RTL run layout, then perhaps it’s rather a useful starting step.)

astiob:
> Though if we’re going to add `Encoding=-2` to VSFilters, this seems counter-productive

And yes, that particular option assumed we would _not_ be adding this to VSFilters.

llyyr:
> It’s still not laying out the runs themselves from right to left though _(first comment)_, right?

Changing the runs themselves to go from right to left produce highly undesirable results for me, both with and without WHOLE_TEXT. I said in my [response last year](https://github.com/libass/libass/pull/901#issuecomment-3145321081) that I needed WHOLE_TEXT, but I can't reproduce this anymore with the same samples I was using then.

TheOneric:
> If a feature needs an ABI break to be sensible/useful, then it needs an ABI break. We should avoid ABI breaks when we can help it, but it shouldn’t force us to shoehorn features into undesirable forms that will haunt us forever. Does this make sense?

This does make perfect sense and I agree with it. *(I don’t see then though how this was an argument against the v4++ patches, but this is another topic and can be revisited later)*

> [text direction within a run vs ordering of whole runs]

Hmm. If it’s useful enough for RTL authors and the most easily achievable in VSFilter, I guess having (in the absence of `WHOLE_TEXT`) a feature for as “LTR-arranged runs of RTL text” is still sensible and if VSF patches can be prepared we can follow up with a mode for RTL in both levels.

Regarding how such modes should be toggled (`-2`, bit flags, …):  
I personally think nobody should ever use `Encoding` for font selection purposes in newly authored files anyway *(even if/when we eventually support it for compatability with existing old files)*, so I’m not concerned about it being impossible to use it as such if we continue using negative values. I’m not opposed to using higher bits to allow new extensions to coexist with it however.

As you brought up, if a future setting for the ordering between full text runs is a style property, it can change within an event, complicating things. While I think it would be possible to treat this similar to explicitly bidi marks in unicode text and actually flip directions, I doubt this will ever be useful in the context of ASS events and having the direction of text runs be an event-wide property should be fine. While it would be cleanest to then have a field in the event itself for this, this is incompatible with the existing file format *(if we want a VSF implementation)*. Ignoring contradicting additional values, similar to what happens with repeated event-wide override tags is probably fine?

TheOneric:
To clarify my thoughts on the way forward:

- If @llyyr or someone else is up to implementing RTL run ordering in the _near_ future, then I think it’d be good to implement RTL run and tesst-within-run ordering at the same time now and combine them into a single toggle
- If nobodoy is up to that in the _near_ future *(i’m not)*, then I think it’d be good to get input from actual RTL ASS authors on whether "RTL text within LTR runs" is useful to them.
    - if yes: we can add this as is and easily add support for it too to VSFilter
    - if not: i think this shuld be an API-only toggle rather than an in-file flag. Accepting an ABI break would allow us to tie this to either the style or event and also bundle some other ABI breaks which piled up.  
        Possibly even some API breaks to explicitly forbid thread--local stuff in the log callback (required for current threading draft), remove a dead font patter flag and perhaps overhaul the whole direct-modification API to forbid direct array access getting rid of the need for ABI breaks when we extend structs

TheOneric:
> I think it’d be good to get input from actual RTL ASS authors on whether "RTL text within LTR runs" is useful to them. 

cc @realfinder: do you want to provide your view on whether this would be helpful?

realfinder:
> > I think it’d be good to get input from actual RTL ASS authors on whether "RTL text within LTR runs" is useful to them.
> 
> cc @realfinder: do you want to provide your view on whether this would be helpful?

you mean by using unicode control character like **`RLE`** like how I did it with https://github.com/libass/libass/issues/538#issuecomment-1013581908 and https://sourceforge.net/projects/rle-tool/ but automatically with Encoding=-2? so that it work in both libass and vsfilter; if so then yes, I think it's better than make it work only with libass (like in old Encoding=-1 case)

realfinder:
and I think maybe its better to use **`-177`** and **`-178`** instead of **`-2`**

TheOneric:
> you mean by using unicode control character like RLE like how I did it with https://github.com/libass/libass/issues/538#issuecomment-1013581908 and https://sourceforge.net/projects/rle-tool/ but automatically with Encoding=-2?

It would have similar effects, yes. Every run will be arranged RTL by default (until overriden by control characters), **but** the runs themselves are still arranged LTR. To my understanding this should match what’s currently possible by inserting Unciode control characters at the start of each run.     
The question is whether this would actually be useful and picked up by RTL authors, or whether only a full RTL mode where the runs too are arranged from right to left *(actually providing more than inserting control characters)*. The latter is harder to implement though and might not show up in the near term.

`Encoding=-1` differs from this by automatically detecting the base direction and laying out the whole event instead of each run individually. This mode is especially hard to implement in VSFilter and thus will presumably always remain libass-only.

> and I think maybe its better to use -177 and -178 instead of -2

I don’t think we need to worry about resemblance to (positive-valued) Windows codepages. Using actual codepages in styles and overrides is strongly discouraged and error-prone anyway. A single value also seems preferable than having several to me.

realfinder:
> The question is whether this would actually be useful and picked up by RTL authors

I think if it work in both libass and vsfilter and it has an option to be used in aegisub it will be useful and it will be used for sure

TheOneric:
thank you! Yes, porting it to active xy-VSFilter is the intent

Then I think we can move this forward with `Encoding=-2` or one of the other discussed in-file mechanisms

PR #902, opened 2025-08-19
meson: Support version-script for symbol visibility

robxnano:
Hello
This is a PR for the libass Meson port to correctly export shared library symbols in more cases. Instead of using the `-fvisibility` attribute, it emulates the libtool behavior of using a version script for the linker. Any linker that supports the GNU --version-script flag should work, as well as the Apple ld64 linker.

The existing script to generate a MSVC .def file has been expanded so it can generate the other formats as well. If the MSVC and Apple formats are not supported, it runs a test to ensure that the linker recognizes the GNU format, and if not it reverts to the current behavior of warning about the lack of symbol visibility support. I've tested it with:

- Apple ld64 and LLVM lld on macOS
- GNU ld.bfd, lld, and mold on Linux
- lld in the Android SDK
- lld and mold on FreeBSD

In all these cases it correctly limited the exported symbols to the ones listed in libass.sym. Please let me know if you want it tested with any other toolchain combinations, and I'll try to make sure it works there too.

### DIFF
--- a/gen_defs.py
+++ b/gen_defs.py
@@ -13,6 +13,16 @@ def gen_mapfile(symfile_path: str, mapfile_path: str):
         with open(mapfile_path, 'w') as mapfile:
             mapfile.writelines(output)
 
+# Exported symbols file, for Apple linker
+def gen_expsym(symfile_path: str, expsymfile_path: str):
+    with open(symfile_path) as symfile:
+        lines = symfile.readlines()
+        output = ''
+        for line in lines:
+            output += f'_{line.strip()}\n'
+        with open(expsymfile_path, 'w') as expsymfile:
+            expsymfile.writelines(output)
+
 # DEF file, for Microsoft link.exe
 def gen_deffile(symfile_path: str, deffile_path: str):
     with open(symfile_path) as symfile:
@@ -23,5 +33,7 @@ def gen_deffile(symfile_path: str, deffile_path: str):
 
 if sys.argv[1] == '--mapfile':
     gen_mapfile(sys.argv[2], sys.argv[3])
+elif sys.argv[1] == '--expsym':
+    gen_expsym(sys.argv[2], sys.argv[3])
 else:
     gen_deffile(sys.argv[1], sys.argv[2])
--- a/libass/meson.build
+++ b/libass/meson.build
@@ -113,6 +113,17 @@ if host_system == 'windows'
         command: [gen_defs, '@INPUT@', '@OUTPUT@'],
     )
     kwargs = {'vs_module_defs': libass_def}
+elif cc.get_linker_id().contains('ld64')
+    # Apple macOS linker
+    have_symbol_visibility = true
+    libass_expsym = custom_target(
+        'libass.expsym',
+        input: ['libass.sym'],
+        output: ['libass.expsym'],
+        command: [gen_defs, '--expsym', '@INPUT@', '@OUTPUT@'],
+    )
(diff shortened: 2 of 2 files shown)


eli-schwartz:
Hello. Speaking with my meson maintainer hat on, this is absolutely something I want to see merged. It is even on my sprawling to-do list of things I'm interested in working on personally -- I just never seem to get around to finding the time to do it. :(

I'm very interested in seeing this completed. I have some ideas about how it should probably work, which I'm very happy to share with anyone who has time to take a stab at it.

- we should definitely allow a flat list of symbols, and generate the correct linker mapping file for each linker we know about (not just GNU, there's a number of different linkers that support the same thing but with a different file syntax and command line argument)
- we should really support manually specified map files probably as an argument accepting a dictionary of `{'gnu': 'foo.map', 'msvc': 'foo.def'}` etc, because lots of people do some fiddly things beyond simply hiding symbols, e.g. this should be flexible enough to support GNU style binding symbols to version nodes
- we should probably support enhanced GNU version maps, with the flat list transformed into a msvc defs / apple symbol hiding / various others as a fallback when an explicit file doesn't exist. So, passing both at the same time?
- do we want to have a compiler method for discovering the supported syntax so that custom tricks can be used? Is it enough to support `{'gnu': custom_target(...)}`? 
- we probably do want to support pattern matching for symbols by scanning objects files after they are built, before linking, to dynamically generate the mapping file based on what symbols the library actually exposes within some form of prefix. This is tricky to get right without proper meson built-in support because the `library()` call defines both the object files and the link command, and the resulting library is immutable -- you can't run extract_all_objects, process it with a custom target, then pass the custom target as another link args. So, a compelling reason to support it as a built-in meson feature! :) The sticking point is that libtool can do regexes here but the only regex engine Meson has access to is the one in the python standard library, which is hardly standard. So, for portability reasons I don't e.g. want to force Muon to implement a bug for bug compatible python regex engine. I'd rather not expose a regex language in Meson at all. But I think that `glob()` can work here and honestly I feel like every actual usage of libtool regexes was always anyway `xxx_.*` so hopefully nobody will actually find this to be an annoying limitation.
- open questions: how should we handle linkers that don't support anything at all? How should we handle people specifying a GNU binutils mapping file only, if they a) don't support other systems at all, b) support other systems but don't care about controlling symbols there? 

Mostly this just requires someone to actually do the legwork and "make a decision" while implementing it. A working prototype to play with is a good way to answer open questions. :)

eli-schwartz:
> we probably do want to support pattern matching for symbols by scanning objects files after they are built, 

This bit in particular would be good for a well rounded feature worth having in Meson's core, since there are quite a number of people who prefer the Unix style "export all symbols" and wish to provide it on Windows -- and have asked after this feature. (Just like there are many that wish Unix linkers behaved like Windows by default and forced libraries to be careful about providing exact exports).

And if you want to blanket export everything on Windows you really do need to scan objects and export `*` via collected names...

robxnano:
@eli-schwartz Glad to hear you're interested in this feature. Your ideas on how it should work are quite similar to what I was considering. I hadn't considered using a dictionary to pass different types of script, but it does make sense in order to support libraries that want to use all the features of GNU version scripts. I'll try and make a prototype which supports at least the basic features like a flat symbol list, then we can go from there.

robxnano:
I've got a working prototype in Meson now, but it needs to be cleaned up before I can make a pull request. In the meantime, I've opened an [issue](https://github.com/mesonbuild/meson/issues/14980) to discuss how to implement the feature.

Issue #532, opened 2021-07-19
Short-circuit rendering on alphaFF


rcombs:
When glyphs are alphaFF (or all individual layers are), we should be able to skip rasterizing them altogether; this can be a useful optimization for glyphs used for spacing, particularly on very large text, esp. if it flows outside the window. A precursor to skipping rendering outside the window bounds.

astiob:
This is trivial to implement, but it will affect output bitmap sizes, which some applications currently use for extent measurement.

rcombs:
We could provide a flag to not do it, or more practically, provide an actual text-extents API rather than just expecting people to rely on bitmap size (which is a poor proxy anyway for a load of reasons).

arch1t3cht:
~~Note that depending on how this is implemented it could break composition caches in certain cases. Specifically, I've been working with lines like~~
```
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(960,540)\an5\fnArial\fs200\bord30\3a&H80&\shad0\clip(800,400,960,700)}A{\2a1\alpha&HFF&}A
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\pos(960,540)\an5\fnArial\fs200\bord30\3a&H80&\shad0\clip(960,400,1200,700)\alpha&HFF&}A{\alpha\3a&H80&\2a1}A
```
(on a 1920x1080 PlayRes)

where rect clips are used to avoid overlapping transparent shapes in situations where a run break cannot be avoided.

~~At the moment, this AFAICT benefits from composition caching, so that only one copy of the line needs to be rasterized and (if necessary) blurred, and skipping rasterization of fully transparent bitmaps could break this.~~

~~I don't know whether this can be worked around by accounting for this in the cache keys and/or whether this is a sensible trade-off to make, but I wanted to mention it here.~~

Meanwhile, from a short look it seems that many applications using libass do not skip fully transparent bitmaps when blending, which could maybe be improved downstream (which I might do at some point, but I'm noting it here for future reference).

astiob:
If I understand you correctly, then your case shouldn’t be (negatively) affected: the composite bitmap cache has an entry _per run,_ not per event. Fully transparent runs would just not be rendered and not be cached, but the non-transparent entries would be cached in exactly the same way they are now. Since transparency is a property of the whole run (i. e., any change in transparency is a run break), a composite bitmap could only ever be skipped as a whole but not internally changed.

[...]

PR #914, opened 2025-12-30
Do not output fully transparent bitmaps

arch1t3cht:
Work towards #532.

With this, fully transparent bitmaps are still fully rasterized, blurred, shifted, etc., but are at least not returned to the calling application to be blended.

---

The main objection I can see to this is the following comment in #532:

> it will affect output bitmap sizes, which some applications currently use for extent measurement.

I personally don't see how this would break such extent measurements, since applications can just not use `\alphaFF` in these cases, but if this *is* a problem this check could also just be implemented in all relevant downstream applications instead (with simpler code, too). But all else being equal it's nicer for this to be implemented upstream.

### DIFF
--- a/libass/ass_render.c
+++ b/libass/ass_render.c
@@ -3366,6 +3366,25 @@ static int ass_detect_change(ASS_Renderer *priv)
 }
 
 /**
+ * \brief free a single image.
+ * \param img image as returned by ass_render_frame().
+ *        Only img will be freed, not img->next.
+ * \return img->next
+ * Should not be called outside of ass_frame_unref
+ * once the image list's refcounting is set up.
+ */
+static ASS_Image *ass_free_image(ASS_Image *img) {
+    ASS_Image *next = img->next;
+
+    ASS_ImagePriv *priv = (ASS_ImagePriv *) img;
+    ass_cache_dec_ref(priv->source);
+    ass_aligned_free(priv->buffer);
+    free(priv);
+
+    return next;
+}
+
+/**
  * \brief render a frame
  * \param priv library handle
  * \param track track
@@ -3415,16 +3434,25 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
     if (cnt > 0)
         fix_collisions(priv, last, priv->eimg + cnt - last);
 
-    // concat lists
+    // concat lists, removing fully transparent bitmaps
     ASS_Image **tail = &priv->images_root;
     for (int i = 0; i < cnt; i++) {
         ASS_Image *cur = priv->eimg[i].imgs;
         while (cur) {
+            if (_a(cur->color) == 0xFF) {
+                cur = ass_free_image(cur);
+                continue;
+            }
+
             *tail = cur;
             tail = &cur->next;
             cur = cur->next;
         }
     }
+
+    // If the last image was skipped in the above loop, *tail may not be NULL and needs to be set to NULL.
+    *tail = NULL;
+
     ass_frame_ref(priv->images_root);
 
     if (detect_change)
@@ -3460,10 +3488,6 @@ void ass_frame_unref(ASS_Image *img)
     if (!img || --((ASS_ImagePriv *) img)->ref_count)
         return;
     do {
-        ASS_ImagePriv *priv = (ASS_ImagePriv *) img;
-        img = img->next;
-        ass_cache_dec_ref(priv->source);
-        ass_aligned_free(priv->buffer);
-        free(priv);
+        img = ass_free_image(img);
     } while (img);
 }


rcombs (commented):
I suppose in theory this would help long-term for cases where the `detect_collisions` flag is set (which would require us to still produce a bitmap of the correct size)? But I wouldn't expect that to be true in the vast majority of performance-relevant cases, so this feels like a lower-priority stage to target.

arch1t3cht:
Yes, this is a mostly minor improvement; a full "fix" would skip rasterizing and blurring of fully transparent bitmaps entirely. But I was in the middle of implementing the same logic in Aegisub and a PR to mpv and realized that ideally this should just implemented upstream.

libass/ass_render.c
```diff
@@ -3415,11 +3429,20 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
     if (cnt > 0)
         fix_collisions(priv, last, priv->eimg + cnt - last);
 
-    // concat lists
+    // concat lists, removing fully transparent bitmaps
     ASS_Image **tail = &priv->images_root;
     for (int i = 0; i < cnt; i++) {
         ASS_Image *cur = priv->eimg[i].imgs;
         while (cur) {
+            if (_a(cur->color) == 0xFF) {
```
cubicibo, line 3442: Maybe adapt that to an if-else to drop the `continue`, and to move that unlikely codepath in the else statement?

libass/ass_render.c
```diff
@@ -3415,11 +3429,20 @@ ASS_Image *ass_render_frame(ASS_Renderer *priv, ASS_Track *track,
...
+                ASS_Image *next = cur->next;
+                ass_free_image(cur);
+
+                *tail = NULL;
```
astiob: I assume you’re performing this assignment for safety, out of abundance of caution? Personally, I think it’s more confusing than helpful. I think I find it easier to reason that `*tail` is initially empty and is (re)assigned exactly once, and the single assignment always has a valid non-empty value.

libass/ass_render.c
```diff
@@ -3460,10 +3483,8 @@ void ass_frame_unref(ASS_Image *img)
...
         return;
     do {
-        ASS_ImagePriv *priv = (ASS_ImagePriv *) img;
-        img = img->next;
-        ass_cache_dec_ref(priv->source);
-        ass_aligned_free(priv->buffer);
-        free(priv);
+        ASS_Image *next = img->next;
+        ass_free_image(img);
+        img = next;
```
astiob: Consider making `ass_free_image` return `img->next` and simplifying this pattern (here and above) to:
```suggestion
        img = ass_free_image(img);
```

arch1t3cht: That's a neat pattern; done.

libass/ass_render.c
arch1t3cht: `*tail = NULL` is necessary here because the images are already partially linked beforehand (so it's not actually initially empty). If the image list of some event is `img1 -> img2 -> img3` where `img2` is fully transparent, the first iteration will result in `tail` being `&img1->next` and `*tail` being `img2`. If `img2` is to be removed from the list, `*tail` needs to be cleared too. (It will be overwritten in a later loop iteration if there *is* one, but if `img2` is the last image in the event's list it will not.)

I moved `*tail = NULL` into the main loop body now, though. That should hopefully make this easier to reason about.

libass/ass_render.c
arch1t3cht: Personally I prefer the current structure with the unlikely path in another `if` and the generic path in the main loop body, but if the maintainers prefer a different structure I can change it.

(The unlikely path also got much shorter after addressing the review comments.)

arch1t3cht:
Sorry, I somehow missed the notifications for the review. Replied now.

libass/ass_render.c
TheOneric: iiuc the `NULL` assignment is at most needed for the very last, not-deleted image. Imho moving the assignment outside and after the loop would be the easiest to understand (and slightly more efficient but not sure if this one assigment matters here). To avoid any potential for confusion, perhaps with a short comment like “necessary if trailing images trimmed”.  
If it needs to be inside the loop, clearing the affected pointer immediately after the problematic free seems clearer to me than reassigning it in the "normal-case" block

libass/ass_render.c
```diff
@@ -3365,6 +3365,25 @@ static int ass_detect_change(ASS_Renderer *priv)
     return diff;
 }
 
+/**
+ * \brief Free a single image.
+ * \param img image as returned by ass_render_frame().
+ *        Only img will be freed, not img->next.
+ * \return img->next
+ * Should not be called outside of ass_frame_unref once
+ * the image list's refcounting is set up.
```
astiob: Nit: move `once` to the second line?

libass/ass_render.c
arch1t3cht: Right, that makes the loop even simpler. I pushed this now.

FWIW, what I like more about keeping the assignment inside the loop is that it prevents `*tail` from pointing to now-invalid `ASS_Image`s at the end of a loop iteration. But I also see the benefit of keeping the loop as simple and understandable as possible.

After all of this I don't really have any strong preference between all three of these options, so I'm happy to choose whichever the others prefer.

astiob (approved):

libass/ass_render.c
arch1t3cht: Done.

MrSmile (approved):

TheOneric (approved):

TheOneric:
I’m not sure why this wasn’t merged yet; it has three approvals and still seems fine to me so we probably just forgot. Sorry for the delay, merged it now.

Issue #381, opened 2020-03-17
support color emoji


0ct0cat:
i want render ass subtitle with emoji.but the  font of text can't support emoji.

so i use NotoColorEmoji.ttf for color emoji, but the libass support only vector font and gray text.

the freetype2 can raster the text with color emoji use NotoColorEmoji.ttf normal.
so the libass raster text algorithm by self ?

vgnorth:
We are seeing similar problem in our platform. It would be great to have this support.

superbonaci:
If it can't support color emoji, at least have a fallback to display it as black/white or grayscale, because as it is, mpv doesn't display the emoji at all. It shows this:

```
□□CAMELBACK□
```

but the metadata is:

```
🐪🐪CAMELBACK🐪
```

https://github.com/mpv-player/mpv/issues/15746

superbonaci:
I've tried but doesn't work, same error:

```
$ mpv --sub-font="Symbola" movie.mp4
$ mpv --sub-font="ttf-ymbola" movie.mp4
```

The zip was downloaded from here: https://fontlibrary.org/en/font/symbola
And installed correctly:

```
$ fc-list | grep -i "Symbola"
/Users/memetix_/Library/Fonts/Symbola.ttf: Symbola:style=Regular
```

TheOneric:
This is not a *verbose* log and while you’re checking Fontconfig’s `fc-list` here, information from mpv’s tracker show you’re actually using Core Text at runtime.

Either way, this is off-topic on this issue tracking progress and obstacles with supporting colour emoji. Please [file a new issue](https://github.com/libass/libass/issues/new?template=wrong-rendering-report.yml) and provide all information the template asks for *(you may skip comparing to VSFilter here; we already know VSFilter doesn't support colour emoji either)*

TheOneric:
Summarising the major challenges with supporting colour emoji (both bitmap and vector-based formats) from recent IRC discussion for future reference:

- bitmaps and anything with non-axis-aligned gradients will require RGBA outputs
  - if monochrome and coloured `ASS_Image`s coexists, possibly with the latter containing a monochrome fallback for API compatibility, this will effectively force a run break between emoji and surrounding text. This has implications on RTL and BiDi text and they are likely undesirable.
      Even if we should later switch to RGBA everywhere, we’ll have to keep this run-break behaviour forever anyway to keep the format semantics stable.  
[...]

Issue #73, opened 2014-04-03
Proposal for new RGBA output API


ghost:
There are a few reasons why we'd want this:
1. Downstreams often to this manually, e.g. grouping ASS_Images in order to upload them at once in one step, and similar.
2. We might switch to RGBA compositing internally, and the current API can't output that.
3. We might need it for similar reasons if we want to add fancy features like real gradients.

It might make a nice simplification too, because ASS_Image is really a bit obscure.

Here's a suggestion what the API could look like. It would (for now) exist in parallel to ass_render_frame().

``` C
typedef enum {
    /**
     * Each pixel is an uint32_t. Each color component takes 8 bits.
     * Bits 0-7: blue
     * Bits 8-15: green
     * Bits 16-23: red
     * Bits 24-31: alpha (0 means transparent, 255 opaque)
     * The alpha is premultiplied, which means the RGB values have already
     * been multiplied with the alpha. This means the condition
     * r <= alpha && g <= alpha && b <= alpha holds true.
     */
    ASS_RGBA_PREMULTIPLIED = 1,
} ASS_FrameFormat;

typedef struct ASS_Rect {
    int x, y, w, h;
};

typedef struct ASS_Frame {
    ASS_FrameFormat format;     // Format of the pixels referenced by bitmap
    void *bitmap;               // Points to the top/left pixel
    size_t stride;              // Line size in bytes (xxx: maybe explain better)
    int w, h;                   // same as ass_set_frame_size()
    int change_detect;          // same as ass_render_frame() change flag
    // A list of rectangles on the bitmap that contain non-transparent
    // pixel data. The rectangles never overlap, which means you can use this
    // to accelerate rendering or texture upload. The number of rectangles
    // is limited by ass_set_max_dirty_rects(). If the number is too small,
    // some rectangles will be coalesced to reach the limit.
    ASS_Rect *dirty_rects;
    int num_dirty_rects;
} ASS_Frame;

/**
 * Set the format of the frame that should be rendered with ass_render_frame2().
 */
int ass_set_frame_format(ASS_Renderer *priv, ASS_FrameFormat format);

/**
 * Set the maximum number of dirty rectangles ASS_Frame should have set.
 */
int ass_set_max_dirty_rects(ASS_Renderer *priv, int max);

/**
 * Set whether regions not covered by dirty rects in ASS_Frame should be cleared.
 * (You can switch it off as a minor possible optimization.)
 */
void ass_set_clear_dirty(ASS_Renderer *priv, int clear);

[...]

PR #915, opened 2026-01-06
feat: Add color emoji support (SBIX/CBDT bitmap fonts)

tonycpsu:
## Summary

This PR adds support for rendering color emoji from SBIX (Apple) and CBDT/CBLC (Google/Noto) bitmap fonts.

- Detects color glyphs via FreeType's `FT_HAS_COLOR` and `FT_LOAD_COLOR`
- Extends `ASS_Image` with `flags` and `color_bitmap` fields for RGBA data
- Backward compatible: apps unaware of color emoji continue working (alpha fallback)
- Compile-time detection via `#ifdef ASS_IMAGE_FLAG_COLOR`
- Full blur support (`\blur`, `\be`) using pre-multiplied alpha
- Shadow support for color glyphs
- Tested on macOS (Apple Color Emoji) and Linux (Noto Color Emoji)

## API Changes

```c
typedef struct ass_image {
    // ... existing fields unchanged ...
    
    // NEW: Color emoji support (at end of struct for ABI compat)
    uint32_t flags;              // ASS_IMAGE_FLAG_COLOR if RGBA
    unsigned char *color_bitmap; // RGBA data, NULL for traditional glyphs
} ASS_Image;

#define ASS_IMAGE_FLAG_COLOR 0x01
```

## Addresses

- Issue #381 (color emoji support discussion)
- Related to #73 (RGBA API discussion)

## Design Decisions

**Per-glyph RGBA vs full-frame RGBA**: This implementation provides RGBA data per-glyph rather than a pre-composited full-frame buffer. This was chosen because:
1. Backward compatible with existing linked-list API
2. Allows GPU-accelerated compositing in players
3. Preserves `detect_change` optimization semantics
4. All major players (mpv, VLC, ffmpeg) already use the linked-list approach

A full-frame RGBA API could be added as an optional parallel API if desired.

**Blur implementation**: Uses pre-multiplied alpha approach - extracts to 4 channels, blurs each with existing SIMD code, recombines. ~4x slower than monochrome but requires no new SIMD code and blur on color emoji is rare.

## Test Plan

- [x] macOS: Apple Color Emoji (SBIX format)
- [x] Linux: Noto Color Emoji (CBDT/CBLC format)  
- [x] Blur effects on color emoji
- [x] Shadow rendering
- [x] Regression tests added (`compare/test/emoji/`)
- [x] Verified with mpv integration


CounterPillow:
Hello clanker, please fetch your fleshie and make them explain why the outputs in the tests are the way they are. Why do things just shift position around and overlap in strange ways?

astiob (commented):
Besides the strange shift in the blur test image and the seemingly inadequate spacing/scaling in all test images, this code also seems to make no attempt to tackle borders, 3D transforms, or layer order issues. Of course, 3D transforms (and to a lesser extent layer ordering) are a pretty fundamental issue that likely prevents bitmap emoji from _ever_ being allowed in ASS.

This also seems to (undesirably) reject non-colour emoji glyphs even in explicitly requested fonts… but only when using Core Text, with no equivalent code changes included for any of the other font providers at all.

(There are also some random stylistic edits in areas that aren’t otherwise modified in this PR.)

But if by some miracle we _should_ eventually get support for colour emoji, I like the idea of putting them in dedicated RGBA bitmaps with fallback monochrome alpha maps for backwards compatibility.

PR #916, opened 2026-01-13
x86inc.asm: add support for separate .text sections

Andarwinux:
this allows linkers to discard unused asm code

Separated from #909

### DIFF
--- a/libass/x86/x86inc.asm
+++ b/libass/x86/x86inc.asm
@@ -84,6 +84,10 @@
     %define FORCE_VEX_ENCODING 0
 %endif
 
+%ifndef FUNCTION_SECTIONS
+    %define FUNCTION_SECTIONS 0
+%endif
+
 %macro SECTION_RODATA 0-1 16
     %ifidn __OUTPUT_FORMAT__,win32
         SECTION .rdata align=%1
@@ -828,12 +832,19 @@ BRANCH_INSTR jz, je, jnz, jne, jl, jle, jnl, jnle, jg, jge, jng, jnge, ja, jae, 
     %xdefine current_function_section __SECT__
     %if FORMAT_ELF
         %if %1
+            %if FUNCTION_SECTIONS
+                section .text.%2 progbits alloc exec nowrite
+                %define last_branch_adr $$
+            %endif
             global %2:function hidden
         %else
             global %2:function
         %endif
     %elif FORMAT_MACHO && HAVE_PRIVATE_EXTERN && %1
         global %2:private_extern
+    %elif WIN64 && %1 && FUNCTION_SECTIONS
+        section .text$%2 comdat=1:%2
+        %define last_branch_adr $$
     %else
         global %2
     %endif
--- a/configure.ac
+++ b/configure.ac
@@ -41,6 +41,8 @@ AC_ARG_ENABLE([asm], AS_HELP_STRING([--disable-asm],
     [disable compiling with ASM @<:@default=check@:>@]))
 AC_ARG_ENABLE([large-tiles], AS_HELP_STRING([--enable-large-tiles],
     [use larger tiles in the rasterizer (better performance, slightly worse quality) @<:@default=disabled@:>@]))
+AC_ARG_ENABLE([trim-asm], AS_HELP_STRING([--enable-trim-asm],
+    [eliminate redundant asm functions where possible @<:@default=disabled@:>@]))
 
 AC_ARG_VAR([ART_SAMPLES],
     [Path to the root of libass' regression testing sample repository. If set, it is used in make check.])
@@ -332,6 +334,9 @@ AS_IF([test "x$enable_asm" != xno], [
                     ASFLAGS="$ASFLAGS -DFORCE_VEX_ENCODING=1"
                 ])
 
+                AS_IF([test "x$enable_trim_asm" = xyes], [
+                    ASFLAGS="$ASFLAGS -DFUNCTION_SECTIONS=1"
+                ])
                 AC_MSG_CHECKING([if $AS supports vpmovzxwd])
                 echo "vpmovzxwd ymm0, xmm0" > conftest.asm
                 AS_IF([$AS conftest.asm $ASFLAGS -o conftest.o >conftest.log 2>&1], [
@@ -384,6 +389,7 @@ AS_IF([test "x$enable_shared" != xno], [
 
 ## Setup conditionals for use in Makefiles
 AM_CONDITIONAL([ASM], [test "x$can_asm" = xtrue])
+AM_CONDITIONAL([TRIM_ASM], [test "x$enable_trim_asm" = xyes])
 AM_CONDITIONAL([X86], [test "x$cpu_family" = xx86])
 AM_CONDITIONAL([X86_64], [test "x$cpu_family" = xx86 && test "x$BITS" = x64])
 AM_CONDITIONAL([AARCH64], [test "x$cpu_family" = xaarch64])
@@ -404,6 +410,9 @@ AM_CONDITIONAL([DIRECTWRITE], [test "x$directwrite" = xtrue])
 ## Define C Macros not relating to libraries
 AM_COND_IF([ASM], [
     AC_DEFINE(CONFIG_ASM, 1, [ASM enabled])
+    AM_COND_IF([TRIM_ASM], [
+        AC_DEFINE(CONFIG_ASM_DCE, 1, [TRIM ASM enabled])
+    ])
     AM_COND_IF([X86], [
         AC_DEFINE(ARCH_X86, 1, [targeting a 32- or 64-bit x86 host architecture])
     ])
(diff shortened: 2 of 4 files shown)


libass/x86/x86inc.asm
```diff
@@ -763,9 +763,9 @@ DECLARE_ARG 7, 8, 9, 10, 11, 12, 13, 14
 
 %define last_branch_adr $$
 %macro AUTO_REP_RET 0
-    %if notcpuflag(ssse3)
-        times ((last_branch_adr-$)>>31)+1 rep ; times 1 iff $ == last_branch_adr.
-    %endif
+    ;%if notcpuflag(ssse3)
+    ;    times ((last_branch_adr-$)>>31)+1 rep ; times 1 iff $ == last_branch_adr.
+    ;%endif
```
rcombs: Why the removal here?

rcombs (commented):
Can we try to send this upstream?
I forget what we currently pull our copy of x86inc from, but https://code.videolan.org/videolan/x86inc.asm exists as a standalone repo now. I'd generally prefer to avoid maintaining our own ad-hoc patched copy, so we can pull from an upstream easily in the future.

libass/x86/x86inc.asm
Andarwinux:
```
FAILED: [code=1] libass/libass.a.p/x86_rasterizer.asm.o
nasm -Ilibass/libass.a.p -f elf64 -DELF -D__x86_64__ -Ox -Dprivate_prefix=ass -DPIC=1 -DARCH_X86_64=1 -DSTACK_ALIGNMENT=16 -I/usr/include/freetype2 -I/usr/include/libpng16 -DWITH_GZFILEOP -I/usr/include/harfbuzz -I/usr/include/freetype2 -I/usr/include/libpng16 -DWITH_GZFILEOP -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -I/usr/include/sysprof-6 -I/usr/include/fribidi -I/usr/include/freetype2 -I/usr/include/libpng16 -DWITH_GZFILEOP -I. -I../libass -Ilibass -I.. -I. -I../libass -Ilibass -Ilibass/libass.a.p -MD libass/libass.a.p/x86_rasterizer.asm.o.d -MQ libass/libass.a.p/x86_rasterizer.asm.o -o libass/libass.a.p/x86_rasterizer.asm.o ../libass/x86/rasterizer.asm
../libass/x86/rasterizer.asm:894: error: shift operator may only be applied to scalar values
../libass/x86/rasterizer.asm:890: ... from macro `MERGE_TILE' defined here
../libass/x86/x86inc.asm:665: ... from macro `RET' defined here
../libass/x86/x86inc.asm:767: ... from macro `AUTO_REP_RET' defined here
../libass/x86/rasterizer.asm:894: error: non-constant argument supplied to TIMES
../libass/x86/rasterizer.asm:890: ... from macro `MERGE_TILE' defined here
../libass/x86/x86inc.asm:665: ... from macro `RET' defined here
../libass/x86/x86inc.asm:767: ... from macro `AUTO_REP_RET' defined here
../libass/x86/rasterizer.asm:895: error: shift operator may only be applied to scalar values
../libass/x86/rasterizer.asm:890: ... from macro `MERGE_TILE' defined here
../libass/x86/x86inc.asm:665: ... from macro `RET' defined here
../libass/x86/x86inc.asm:767: ... from macro `AUTO_REP_RET' defined here
../libass/x86/rasterizer.asm:895: error: non-constant argument supplied to TIMES
../libass/x86/rasterizer.asm:890: ... from macro `MERGE_TILE' defined here
../libass/x86/x86inc.asm:665: ... from macro `RET' defined here
../libass/x86/x86inc.asm:767: ... from macro `AUTO_REP_RET' defined here
[33/34] Compiling C object libass/libass.a.p/c_c_blur.c.o
```

Andarwinux:
> Can we try to send this upstream?

To do this, I had to make sure that the change was fully compatible with [ffmpeg](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg+%22times+%28%22+language%3AAssembly&type=code), [dav1d](https://github.com/search?q=repo%3Avideolan%2Fdav1d+%22times+%28%22&type=code), x264, etc. I tested it and found that ffmpeg and dav1d had problems like `error: non-constant argument supplied to TIMES` with this change. Unlike the case of libass, they are used similarly outside of x86inc.asm so it's hard to fix.

libass/x86/x86inc.asm
rcombs: Please figure out what's actually going wrong here instead of just breaking this functionality.

Andarwinux: The problem is this feature is fundamentally incompatible with non-unique .text.

MrSmile: `%define last_branch_adr $$` immediately after `.section` solves it.

libass/x86/x86inc.asm
```diff
@@ -828,12 +828,25 @@ BRANCH_INSTR jz, je, jnz, jne, jl, jle, jnl, jnle, jg, jge, jng, jnge, ja, jae,
...
             global %2:function
         %endif
-    %elif FORMAT_MACHO && HAVE_PRIVATE_EXTERN && %1
-        global %2:private_extern
+    %elif FORMAT_MACHO
+        %if HAVE_PRIVATE_EXTERN && %1
+            global %2:private_extern
+        %else
+            global %2
+        %endif
```
rcombs: Isn't this change redundant? Since it's the same as the top-level `%else` path anyway.

Andarwinux:
> > Can we try to send this upstream?
> 
> To do this, I had to make sure that the change was fully compatible with [ffmpeg](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg+%22times+%28%22+language%3AAssembly&type=code), [dav1d](https://github.com/search?q=repo%3Avideolan%2Fdav1d+%22times+%28%22&type=code), x264, etc. I tested it and found that ffmpeg and dav1d had problems like `error: non-constant argument supplied to TIMES` with this change. Unlike the case of libass, they are used similarly outside of x86inc.asm so it's hard to fix.

I think it would be possible to add a macro to guard DCE feature so that it won't break ffmpeg.

> Can we try to send this upstream?

I'm trying to do that, but code.videolan.org doesn't seem to allow new users to register.

libass/x86/x86inc.asm
```diff
@@ -828,12 +832,19 @@ BRANCH_INSTR jz, je, jnz, jne, jl, jle, jnl, jnle, jg, jge, jng, jnge, ja, jae,
...
+                %define last_branch_adr $$
+            %endif
             global %2:function hidden
         %else
             global %2:function
         %endif
     %elif FORMAT_MACHO && HAVE_PRIVATE_EXTERN && %1
         global %2:private_extern
+    %elif WIN64 && %1 && ENABLE_DCE
+        section .text$%2 comdat=1:%2
```
rcombs, line 846: Doesn't this also need `global %2`? Or does the `section .text$` somehow do that implicitly?

libass/x86/x86inc.asm
```diff
@@ -84,6 +84,10 @@
     %define FORCE_VEX_ENCODING 0
 %endif
 
+%ifndef ENABLE_DCE
+    %define ENABLE_DCE 1
```
rcombs: Feels like a weirdly specific name (there are other reasons to want function sections, and it's not like the assembler does any DCE itself; also, x86inc doesn't use any other ENABLE_ macros); I'd suggest eg `FUNCTION_SECTIONS`. Also, are there cases where this is undesirable?

libass/x86/x86inc.asm
Andarwinux: comdat already did the same thing. If global is kept, it will cause duplicate symbols.

https://github.com/libass/libass/actions/runs/20396468689/job/58612663761

libass/x86/x86inc.asm
Andarwinux: > I'd suggest eg `FUNCTION_SECTIONS`

good idea.

> Also, are there cases where this is undesirable?

https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg+%22times+%28%22+language%3AAssembly&type=code
https://github.com/search?q=repo%3Avideolan%2Fdav1d+%22times+%28%22&type=code

I don't think these cases can be fixed transparently within x86inc.asm alone. After changes are committed to upstream x86inc.asm, ffmpeg and dav1d can define FUNCTION_SECTIONS=0 for those asm to fix the build.

rcombs: Hmmm, if this breaks all of those cases, then I'd be inclined to have FUNCTION_SECTIONS be off by default (so, matching the behavior of common C compilers).

Andarwinux: done.

rcombs: I'm not entirely clear on the reason to not just pass `-DFUNCTION_SECTIONS=1` unconditionally in libass?

Andarwinux: I thought you meant libass should also disable FUNCTION_SECTIONS by default. However, disabled by default simplifies the implementation of #909, as it eliminates the need for ass_bitmap_engine_init_checkasm.

PR #921, opened 2026-04-28
ci/gha: convert UWP job to regular desktop UCRT

TheOneric:
The job failed for over four months now as the result of toolchain changes and UWP is deprecated making a future fix seems unlikely. Nobody stepped up to fix the CI setup either.
A permanently failing job brings no benefit, so let’s not.

### DIFF
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -98,10 +98,10 @@ jobs:
             output_stream.write(f"count={num_cpus}\n")
 
       - name: checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: download test samples
-        uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


rcombs (approved):

astiob (approved):
Last I heard, VLC still had UWP builds? So it’d be great to have them on our end too, but this is certainly better than failing.

TheOneric:
> Last I heard, VLC still had UWP builds?

Indeed, and they keep actively fixing their build setup for it, like most recently [here](https://code.videolan.org/videolan/vlc/-/commit/efe0e7527da52ad9971e1c8f2aa470eddde6da69). However, it helps noone to keep a permanently failing to build CI job around since it tests nothing at all and just pollutes the CI result. If a way is found to make UWP CI actually work, it can be brought back.

PR #922, opened 2026-04-29
ci: add illumos job back

TheOneric:
cross-platform actions now supports OmniOS. Unlike in OpenIndiana we don’t get our deps as "normal" system packages, but they are still available via the extra "ooce" *(OmniOS Community Edition)* repo just requiring an extra pkg-config opt-in

### DIFF
--- a/.github/workflows/vm.yml
+++ b/.github/workflows/vm.yml
@@ -65,14 +65,14 @@ jobs:
             # no libubsan available, thus just SIGILL
             CC: 'gcc -fsanitize=undefined,float-cast-overflow -fno-sanitize-recover=all -fsanitize-undefined-trap-on-error'
 
-          # No illumos in cross-platform-actions/action yet, see
-          #   https://github.com/cross-platform-actions/action/issues/20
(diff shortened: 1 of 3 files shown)


.github/workflows/vm.yml
```diff
@@ -132,36 +152,24 @@ jobs:
...
-                    developer/build/pkg-config \
-                    system/library/freetype-2 \
-                    library/fribidi \
-                    library/c++/harfbuzz \
-                    system/library/fontconfig \
-                    image/library/libpng
+              omnios)
+                # "ooce" extra packages are not in regular search paths.
+                # Activate via /opt/ooce/lib/64/pkgconfig etc.
+                # Harfbuzz and fridi are bundled with pango (no standalone package).
```
astiob: typo in “fribidi”?

astiob (approved):

.github/workflows/vm.yml
TheOneric: yep. fixed now, thx

PR #924, opened 2026-05-03
ass_fontconfig: release libfontconfig global state on destroy paths

llyyr:
FcConfigDestroy frees the FcConfig object but doesn't release libfontconfig's global state, which remains populated when parsing the font config XML. This causes LSAN to report leaks originating from ass_set_fonts when the fontconfig provider is selected.

Add FcFini in the regular destroy path, and in the error path of ass_fontconfig_add_provider where FcInitLoadConfig may have been called before bailing out.

### DIFF
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -85,6 +85,7 @@ static void destroy(void *priv)
     if (fc->fallbacks)
         FcFontSetDestroy(fc->fallbacks);
     FcConfigDestroy(fc->config);
+    FcFini();
     free(fc);
 }
 
@@ -380,6 +381,7 @@ ass_fontconfig_add_provider(ASS_Library *lib, ASS_FontSelector *selector,
         ass_msg(lib, MSGL_ERR,
                 "No valid fontconfig configuration found!");
         FcConfigDestroy(fc->config);
+        FcFini();
         free(fc);
         return NULL;
     }


rcombs:
The data freed by `FcFini` is not refcounted, so this is an unconditional global destroy of data that may be being used from different libraries or on other threads. This is not safe to call from within library code. If this is an issue for you, please either request either a library-safe refcounting global-destructor API or a global-state-free API from fontconfig, or suppress this allocation in your lsan tooling (it is not a leak; the allocations remain live and reachable at all times).

llyyr:
https://github.com/libass/libass/pull/925

arch1t3cht:
I looked into this myself a while ago. `FcFini` *is* actually supposed to be reference counted on new versions, see https://gitlab.freedesktop.org/fontconfig/fontconfig/-/work_items/448 and https://gitlab.freedesktop.org/fontconfig/fontconfig/-/merge_requests/410 .

But, at least back when I checked, fontconfig wasn't very explicit about which of the fontconfig init functions increase the reference count in this way, and the code looked quite hard to reason about too. In particular it didn't look to me like `FcFini` was actually really thread-safe, even if the code claims it is. So I didn't end up making a PR.

rcombs:
Yeah, as far as I can tell, `FcFini` remains non-thread-safe as of the most recent versions (which have made some attempts to make it thread-safe, but have not fully succeeded), eg because:
- `free_lock` races against other threads' calls to `lock_config` and `lock_cache`; that part might be fine if they stopped heap-allocating the mutexes for no discernible reason (maybe they just didn't want to implement a `once` routine?)
- `FcCacheFini` also performs nonatomic reads of `fcCacheChains` outside the lock, so that's non-threadsafe on the face of it
- Agreed; the global config refcount in newer versions *also* looks to be full of fairly obvious race conditions and I wouldn't trust it as far as I can throw it

Trying to destroy heap-allocated global state is hell at the best of times, and serves very little actual purpose beyond silencing overzealous runtime sanitizers; I see no reason to subject libass consumers to the risks involved. Happy to accept any PRs that remove usage of global state (a la https://github.com/libass/libass/pull/925) though.

PR #925, opened 2026-05-03
ass_fontconfig: use FcConfigSetDefaultSubstitute when available

llyyr:
FcDefaultSubstitute allocates state into global context that is not freed by FcConfigDestroy. FcConfigSetDefaultSubstitute introduced in 2.17.0 by https://gitlab.freedesktop.org/fontconfig/fontconfig/-/commit/90a84c477c4006b5539973a5487f28b3ed09711d scopes the allocation to the provided FcConfig instance, allowing FcConfigDestroy to fully clean up all state and eliminating the LSAN reported leaks

### DIFF
--- a/libass/ass_fontconfig.c
+++ b/libass/ass_fontconfig.c
@@ -233,7 +233,7 @@ static void cache_fallbacks(ProviderPrivate *fc)
         return;
     FcPatternAddString(pat, FC_FAMILY, (FcChar8 *)"sans-serif");
     FcPatternAddBool(pat, FC_OUTLINE, FcTrue);
-    FcConfigSubstitute (fc->config, pat, FcMatchPattern);
+    FcConfigSubstitute(fc->config, pat, FcMatchPattern);
 #if FC_VERSION >= 21700
     FcConfigSetDefaultSubstitute(fc->config, pat);
 #else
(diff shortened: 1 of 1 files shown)


rcombs (approved):

libass/ass_fontconfig.c
```diff
@@ -100,7 +100,11 @@ static bool scan_fonts(FcConfig *config, ASS_FontProvider *provider)
     if (!pat)
         return false;
 
-    FcDefaultSubstitute(pat);
+#if FC_VERSION >= 21700
+    FcConfigSetDefaultSubstitute(config, pat);
+#else
+    FcDefaultSubstitute (pat);
```
TheOneric: you probably aligned the spacing here with the `FcDefaultSubstitute` (and `FcConfigSubstitute`) call(s) below, but i don’t know why those have an extra space in the first place; probably just a mishap which slipped through. We don’t want to have space between function name and parameter list.

TheOneric (approved):
The linked fontconfig commit indicates the versions without config parameter will be deprecated in favour of the new ones, so this change is good and necessary even regardless of sanitiser complaints

libass/ass_fontconfig.c
TheOneric: looks like those two calls are the only instances of fonctofig functions with an extra space atm according to `grep -rnE 'Fc[a-zA-Z]* \(' libass`

PR #927, opened 2026-05-10
Support legacy arabic charmap

### LINKED ISSUE #292: Rendering Arabic fonts
Hi, I just switched to Libass recently and I came to realize that it doesn't render a whole lot of Arabic fonts.
Example:
Xy-Vs

Libass

This is just a drop in the bucket. Is there any way to solve this?
Link for this exact font: [Here](http://www.arfonts.net/download.php?url=mutawar-symbol,hesham-nagham-normal-traditional).

moi15moi:
As specified in https://learn.microsoft.com/en-us/typography/legacy/legacy_arabic_fonts, there are 2 types of legacy arabic charmap (ARABIC_CHARSET_SIMPLIFIED and ARABIC_CHARSET_TRADITIONAL). To properly support these mapping, harfbuzz added support to them in https://github.com/harfbuzz/harfbuzz/pull/3063

But, from our side, we still needed to map the unicode codepoint to the legacy "encoding".

To do so, let's simply use the mapping harfbuzz generated (https://github.com/harfbuzz/harfbuzz/blob/da577bc545b1282f003b579189a6dcfb5a1eb6b1/src/ArabicPUASimplified.txt + https://github.com/harfbuzz/harfbuzz/blob/da577bc545b1282f003b579189a6dcfb5a1eb6b1/src/ArabicPUATraditional.txt).

Fix: https://github.com/libass/libass/issues/292


astiob (approved):
Thanks!

Just to be sure: have you retested https://github.com/libass/libass/issues/292 and made sure that it works with this?

moi15moi:
I didnt found the original sample (since the link is dead), so i "just" tested it with the 2 samples fonts in my libass-tests PR.

realfinder:
> I didnt found the original sample (since the link is dead), so i "just" tested it with the 2 samples fonts in my libass-tests PR.

similar test cases listed in https://github.com/libass/libass/issues/292#issuecomment-4373037194

Bilal2453:
Hey, thank you for your work!

I've tested this with a few of the famous fonts that I've verified are Traditional/Simplified fonts, and this seems to be working perfectly.

(using mpv with the custom built shared libass binary)

SKR-HEAD1:
image

Khalid King:
image

AGA-Aladdin:
image

Edit: and this is the original test case
image

(and a few others)

Bilal2453:
While this fix is probably going to be fine for a long time (LGTM for now), wouldn't you want to use harfbuzz directly for the codepoint mapping for a long term solution? Otherwise I am afraid you will have to maintain them not just for Arabic, but other languages too where the Windows 3.1 curse might've hit (Farsi/Hebrew I believe), plus something in those Arabic mappings might change at one point to fix something.
Although, I haven't looked too much at the code, so not sure how easy it is to do that.

moi15moi:
> While this fix is probably going to be fine for a long time (LGTM for now), wouldn't you want to use harfbuzz directly for the codepoint mapping for a long term solution?

No, we want to properly emulate gdi, so we need to have the control.

For example, to fix https://github.com/libass/libass/issues/319, we will need to do it from our side. Harfbuzz cant really do anything for these.

MrSmile:
Not sure about autogenerated sources — we already rejected similar approach for parsing. Also compilers generate abysmal assembly for that switch, is it on hot path? If we adopt generation then it would be better to generate something more performant (perfect hash, for example).

moi15moi:
> is it on hot path?

Yes it is (for symbol arabic font).

> If we adopt generation then it would be better to generate something more performant (perfect hash, for example).

I don't see how we could avoid using generation to solve this problem.
Do you have a suggestion?
I do see how I could use a binary search for this, but it wouldnt as fast has a perfect hash

MrSmile:
I mean to use more sophisticated generation, something like [this](https://gist.github.com/MrSmile/e420bf4e5a45c5a621dcddcdabd67458).
Also, maybe it's better to have local copies of `ArabicPUA*.txt` too.

astiob:
Each of the two charsets is single-byte. Just put them in an array, no hashing required.

astiob:
Looks like they cover only a few narrow ranges, so maybe a few simple lookup tables would suffice?

TheOneric:
For both the unicode ranges it is mapping from seem to be:
- `0x0020–0x005D` via bit-or with a prefix
- `0x00AB`, `0x00BB`, `0x00D7`, `0x00F7` individually
- `0x060C–0x066C` with some holes
- `0x200C–0x200F` via simple bitmask
- for traditional only: 
    - `0x201C` and `0x201D`
    - `0xFC08–0xFEFC` with some holes and the mapping method doesn’t seem uniform at first glance    
- for simple only:
    - `0x2018` and `0x2019`
    - `0xFE81–0xFEFC` mapping method also not uniform *(at least not obviously so)*

> Also, maybe it's better to have local copies of ArabicPUA*.txt too.

I agree. If we have (pre-)generated source files in git, we should at leat also have the sources for an exact reproduction in git too. The web sources from the current Harfbuzz repo might change or disappear / move to another location.


> No, we want to properly emulate gdi, so we need to have the control.

That is correct for the overall font processing and rendering, but just to make sure: this means there’s no Harfbuzz API we can use _just_ for the symbol mapping here *(and only when detecting an affected legacy encoding)*, right? Because if such an API existed it should be safe to hand the mapping details off to HarfBuzz, given it too just reimplements GDI’s mapping.

astiob:
> If we have (pre-)generated source files in git, we should at leat also have the sources for an exact reproduction in git too. The web sources from the current Harfbuzz repo might change or disappear / move to another location.

I thought that was part of the point: we would treat HarfBuzz’s list as the master source, and if HarfBuzz later updates their list, we could just rerun the script to generate an updated file on our end.

TheOneric:
> if HarfBuzz later updates their list, we could just rerun the script to generate an updated file on our end.

Yes, but if ti changed (or unlikely: fully disappears), we won’t be able to tell whether the generated source was modified (possibly before being committed) from what the contemporary source created, if the genreator was modified without updating the generated code, or if HarfBuzz simply updated the mapping files.  
The generator script can still refetch the files from HarfBuzz, just now it will be obvious if something changed or not.

moi15moi:
> this means there’s no Harfbuzz API we can use _just_ for the symbol mapping here _(and only when detecting an affected legacy encoding)_, right?

I just verified. To answer to your question, yes, there isn't an api for that. ``hb_font_get_nominal_glyph`` does exist, but it return the glyph id, not the cmap entry. It would require to rewrite a lot of thing.

Also, I don't know why, but harfbuzz just properly decode those symbol arabic font if their os/2 table version is 0.
See [this](https://github.com/harfbuzz/harfbuzz/blob/af541c22796405f54fd80ab066b86a4220d9a58a/src/hb-ft.cc#L402) + [this](https://github.com/harfbuzz/harfbuzz/blob/af541c22796405f54fd80ab066b86a4220d9a58a/src/hb-ot-os2-table.hh#L345).

Bilal2453:
> Also, I don't know why, but harfbuzz just properly decode those symbol arabic font if their os/2 table version is 0.

Because that is the only valid version for them. Those fonts are now largely documented by Microsoft, see [this](https://learn.microsoft.com/en-us/typography/legacy/legacy_arabic_fonts).

It is not possible to accurately identify these fonts otherwise, while you could technically do checks like ensuring the font has a GSUB table, or even looking the PUA up, it still won't be guaranteed what that you are looking at. So this is the best thing to do.

moi15moi:
GDI still use this legacy encoding if the OS/2 table version isn't 0.
GDI is the source of truth (over documentation)

Bilal2453:
Do we have a good understanding of how GDI is detecting them? does it just always assume the PUA is a valid place for the glyphs and look them up as a fallback?

Edit: on a good note, none of the fonts I've tested had their os/2 version set to anything higher than 0.

moi15moi:
> Do we have a good understanding of how GDI is detecting them?

Yes. If the cmap is a microsoft symbol cmap, it just look at the fsSelection and look at the charset (like the current code).

> does it just always assume the PUA is a valid place for the glyphs and look them up as a fallback?

It doesnt use the PUA as the glyph id. It use the pua for the cmap (those are 2 differents things).
It isn't a fallback. Its just how it map the unicode character to the cmap entry.

moi15moi:
I added the map files to the repos and I used @MrSmile ``gen_arabic_charmap.py`` (with a small change for the typing).
Note that I used the latest traditional map file (since there was an error in it) https://github.com/harfbuzz/harfbuzz/pull/5985

legacy_arabic_cmap/gen_arabic_charmap.py
```diff
@@ -0,0 +1,203 @@
...
+def main():
+    SOURCES = [
+        {
+            "file": Path(__file__).parent.joinpath("ArabicPUASimplified.txt"),
+            "function": "ass_font_charmap_arabic_simplified",
+        },
+        {
+            "file": Path(__file__).parent.joinpath("ArabicPUATraditional.txt"),
+            "function": "ass_font_charmap_arabic_traditional",
+        },
```
TheOneric, line 167: This doesn’t seem right. As committed the mapping files should be in the same directory as the generator

TheOneric (commented):
I’m wondering whether it would make sense to include the mapping files and generator script into tarballs via `EXTRA_DIST` or not.
At the one hand it’s not at all required for building the tarball, nor even can be automatically be used if present. On the other hand it _is_ the actual source of `ass_arabic_charmap.{c,h}` and explains/shows how we got to it.

*(Settling this is in no way a blocker for this patch though)*

legacy_arabic_cmap/gen_arabic_charmap.py
moi15moi: I don't understand your comment.
`Path(__file__).parent` mean the directory of `gen_arabic_charmap.py`.

I did it that way so if you execute the script from a different location than `libass/legacy_arabic_cmap`, it would still find the mapping files AND generate `ass_arabic_charmap.{c,h}` at the right location.

TheOneric: sorry, i misunderstood what this does and checking again the generated code files use a double `.parent.parent`. Yes, should be good

TheOneric (approved):

MrSmile:
My `gen_arabic_charmap.py` was proof of concept, for production code both tables should be compacted to `uint8_t` (it's enough to check only high byte of hash value and high byte of the result table is fixed). But to do that reliably, there should be integrated range checks for robustness.

MrSmile:
[New version](https://gist.github.com/MrSmile/e420bf4e5a45c5a621dcddcdabd67458) with `uint8_t` tables and range checks. Less generic, but I think such tables shouldn't change much.

moi15moi:
Thanks @MrSmile 
I updated the generator with your sample.

Note that with this sample, I dont get the same result with libass and vsfilter
```
[Script Info]
; Script generated by Aegisub 9841-feature-2d190fef7
; http://www.aegisub.org/
Title: [DAC] Kujira no Kora wa Sajou ni Utau - 07
ScriptType: v4.00+
WrapStyle: 0
ScaledBorderAndShadow: yes
YCbCr Matrix: TV.709
PlayResX: 1280
PlayResY: 720
Original Script: Asenshi
Original Translation: AlterDAC
Original Editing: Congroo
Update Details: http://www.dacsubs.com/

[Aegisub Project Garbage]
Last Style Storage: Default
Video File: ?dummy:24000/1001:40000:1920:1080:255:255:255:
Video AR Value: 1.777778
Video Zoom Percent: 1.000000
Scroll Position: 39

[V4+ Styles]
Style: Insert,Hesham Nagham,100,&H00221D18,&H000000FF,&H00545D85,&H00000000,-1,0,0,0,100,100,0,0,1,2.2,0,2,7,7,33,1

[Events]
Dialogue: 0,0:00:00.00,0:00:06.17,Insert,,0,0,0,,aّمال
```

vsfilter
image

libass
test

TheOneric:
hmm, this is presumably related to the arabic font not having an `a` glyph but i can’t find any prior mention of VSFilter or GDI, e.g. splitting runs, on fallback font changes.

The output here is broken in other ways (not sure if my copy of the font is broken or the harfbuzz build missing some bits), but HarfBuzz appears to favour adding it to the `a` glyph:

hb-view_render_2026-06-05

moi15moi:
I deciced to open an issue about this since it isn't related to this PR.
See: https://github.com/libass/libass/issues/932

TheOneric (approved):

PR #929, opened 2026-05-15
build: do not treat signed integer over- or underflow as UB

TheOneric:
We long struggled with how to deal with signed overflows during rendering, where overflow detection/prevention is unacceptably costly. At the same time, as long as the overflow itself doesn’t blow up, regardless of what value we end up with in those  rendering computations we never end up with something _unsafe_, only a nonsensical output image (sometimes we even _want_ to overflow with wrap around to emulate VSFilter).

This patch is not necessarily the best final solution *(for one, it only works on some compilers; additionally we now also won’t get warnings on places where overflow is actually unsafe)*, but intended to (re)start the discussion about how to tackle this. Alternatively we could explicitly cast operands to unsigned types in all such "safe" calculations at the cost of more unwieldy code.  
Ideally imho, we had different types for safely wrapping signed integers, signed integers where overflow is considered unsafe and also the same for _unsigned_ types, as we typically do _not_ want to ever over- or underflow `size_t` values used as pointer offsets or to track the size of a memory region.  
Maybe if [`__attribute__((overflow_behavior(…)))`](https://clang.llvm.org/docs/OverflowBehaviorTypes.html) stabilises and becomes more wide-spread we can do that, but for now, I don’t think this is possible.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,6 +55,11 @@ FUZZ_CPPFLAGS="${FUZZ_CPPFLAGS:--DASS_FUZZMODE=0}"
 AX_APPEND_COMPILE_FLAGS([-fno-math-errno])
 AX_APPEND_COMPILE_FLAGS([/clang:-fno-math-errno])
 
+## No UB on signed integer overflow
+## (this also enforces complement-of-two wrap around, but we shouldn’ŧ rely on that)
+AX_APPEND_COMPILE_FLAGS([-fwrapv])
+AX_APPEND_COMPILE_FLAGS([/clang:-fwrapv])
+
 ## Warnings
 AX_APPEND_COMPILE_FLAGS([ \
     -Wall -Wextra -Wno-sign-compare -Wno-unused-parameter -Wstrict-prototypes \
--- a/meson.build
+++ b/meson.build
@@ -26,9 +26,9 @@ cc_warnings = []
 cc_features = []
 
 if cc.get_id() == 'clang-cl'
-    cc_features += '/clang:-fno-math-errno'
+    cc_features += ['/clang:-fno-math-errno', '/clang:-fwrapv']
 elif cc.get_id() != 'msvc'
-    cc_features += ['-D_GNU_SOURCE', '-D_XPLATFORM_SOURCE', '-fno-math-errno']
+    cc_features += ['-D_GNU_SOURCE', '-D_XPLATFORM_SOURCE', '-fno-math-errno', '-fwrapv']
 endif
 
 if cc.get_argument_syntax() == 'gcc'


rcombs:
Might be worth conditionalizing this, and having debug/test builds use either `-ftrapv` or leave it UB (eg in ubsan builds)?

TheOneric:
What method of detecting or classifying sth as a "debug build" do you have in mind?

Note though, even now the UBSAN in recent-enough clang will still report signed-integer overflows anyway _if_ explicitly enabling the (also implicitly enabled by `-fsanitize=undefined`) `-fsanitize=signed-integer-overflow` as noted in the commit message. With gcc and in general regardless of UBSAN it should already be possible to override the default flag with `-ftrapv` in the user-defined `CFLAGS` with `-ftrapv`

rcombs:
Mostly our own CI builds (esp. eg fuzzing); if our existing settings will result in this being caught in CI, then I'm 👍 on this change.

TheOneric:
OSS Fuzz uses clang and seems to explicitly set `signed-integer-overflow` if I’m interpreting this right ([20.04 base image](https://github.com/google/oss-fuzz/blob/755a8154abbe5649e1500744f02830dff7225980/infra/base-images/base-builder/ubuntu-20-04.Dockerfile#L77), [24.04 base image](https://github.com/google/oss-fuzz/blob/755a8154abbe5649e1500744f02830dff7225980/infra/base-images/base-builder/ubuntu-24-04.Dockerfile#L77)), so it should still complain. *(Notably it also appears to complain about _unsigned_ overflows even though they are well defined)*

Our own CI and libass-test mostly use gcc due to it not suffering from clang’s bug of ignoring the `exit-code` for one of the sanitisers *(meaning, even if issues are discovered, it will just exit with `0` and we can’t reliably detect it in CI. Can only be solved by foregoing error details and telling it to `SIGILL` on any errors instead)*.
But we could easily add `-ftrapv` to any builds using sanitisers.

kasper93:
> (Notably it also appears to complain about unsigned overflows even though they are well defined)

It does not by default. Unsigned overflows are not reported with default UBSAN options.

> clang’s bug of ignoring the `exit-code` for one of the sanitisers _(meaning, even if issues are discovered, it will just exit with `0` and we can’t reliably detect it in CI. Can only be solved by foregoing error details and telling it to `SIGILL` on any errors instead)_. But we could easily add `-ftrapv` to any builds using sanitisers.

This is not true. When UBSAN runtime library is available, clang will show full diagnostics, with `UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1` or `-fno-sanitize-recover=undefined` at compile time. The only case where it throws `SIGILL` is when running in runtime-less mode, which should never be the case for testing/ci environment.

EDIT: You can see that for example in this build https://fate.ffmpeg.org/report.cgi?slot=amd64-clang-asan-asm-short-enums&time=20260523235542 where you get all information needed.

TheOneric:
> > (Notably it also appears to complain about unsigned overflows even though they are well defined)
>
> It does not by default. Unsigned overflows are not reported with default UBSAN options.

hmm, `-fsanitize=…,unsigned-integer-overflow,…` literally appears in the quoted configuration line though directly next to `signed-integer-overflow`. If the former does not actually take effect, the latter might not either in which case it would be unable to detect signed int overflows with `-fwrapv`.

The FFmpeg report you linked might use a customised setup. Do you have a better source on what oss fuzz’s exact sanitiser flags for the "undefined" mode are if it’s not what i linked?

>  >  clang’s bug of ignoring the exit-code for one of the sanitisers […]
>
> This is not true.

It very much was true when the CI was set up and took a while to figure out why it wasn’ŧ failing correctly. Though reading your response you might be thinking of something else.  
See https://reviews.llvm.org/D35085  (i couldn’t find the migrated version of this in their new GitHub PR and issue tracker)

kasper93:
> > > (Notably it also appears to complain about unsigned overflows even though they are well defined)
> > 
> > 
> > It does not by default. Unsigned overflows are not reported with default UBSAN options.
> 
> hmm, `-fsanitize=…,unsigned-integer-overflow,…` literally appears in the quoted configuration line though directly next to `signed-integer-overflow`. If the former does not actually take effect, the latter might not either in which case it would be unable to detect signed int overflows with `-fwrapv`.
> 
> The FFmpeg report you linked might use a customised setup. Do you have a better source on what oss fuzz’s exact sanitiser flags for the "undefined" mode are if it’s not what i linked?

The unsigned overflow checks are only inserted to guide coverage, they are not reported in any way. They are explicitly suppressed. See: [ubuntu-24-04.Dockerfile#L91-L93](https://github.com/google/oss-fuzz/blob/755a8154abbe5649e1500744f02830dff7225980/infra/base-images/base-builder/ubuntu-24-04.Dockerfile#L91-L93) 

> > > clang’s bug of ignoring the exit-code for one of the sanitisers […]
> > 
> > 
> > This is not true.
> 
> It very much was true when the CI was set up and took a while to figure out why it wasn’ŧ failing correctly. Though reading your response you might be thinking of something else. See https://reviews.llvm.org/D35085 (i couldn’t find the migrated version of this in their new GitHub PR and issue tracker)

Nothing changed. UBSAN is not faulting, you can run your application and it's not perturbed by the presence of UBSAN, except that it will print-out the reports. For "non-interactive" mode, it's important to configure it in a way to fault and report errors when they are detected. So, it's true that exit code is not set in default configuration, but with proper options it reports errors as the arrive and aborts, `SIGILL` is only visible on misconfigured environments without UBSAN runtime. Again, see ffmpeg report.

kasper93:
> We long struggled with how to deal with signed overflows during rendering, where overflow detection/prevention is unacceptably costly.

This could be re-evaluated by using checked arithmetic like https://en.cppreference.com/c/numeric/ckd_add or https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

Instead of doing expensive checks before doing arithmetic operations, you let the compiler figure this out for you. And allow access to overflow flag that your cpu already set anyway. If you ignore return value, this compiles to the same code as normal arthmetic op would. (depends on arch and so on, but you get the point). Making this very cheap way to handle signed overflows in defined way.

As a bonus it might be supported on MSVC one day, unlike the `-fwrapv` flag. And makes it more fine-grained overall, instead of making all signed overflows defined and possibly letting some unwanted ones in.

Note that if you use intrinsics or asm, the overflows are defined and depending on arch you might have access to saturating variants too.

PR #930, opened 2026-05-24
GDI-compatible rotation handling, extended.

Masaiki:
This is based on #702 and adds the remaining changes needed to make the #701 vertical layout case match VSFilter/GDI more closely.

Additional fixes on top of #702:

1. Make DirectWrite fallback selection locale-aware.

   Without this, when the requested subtitle font is not installed, DirectWrite may fall back to Arial for CJK/common characters. That still leaves the #701 no-font sample rendered incorrectly, because the selected fallback
   does not match the font GDI/VSFilter would use for the same text.

   This patch passes the requested base family and the user locale to `CreateTextFormat`, which makes DirectWrite fallback selection closer to Windows/GDI behavior.

2. Fix vertical rule-like punctuation in `@font` layout.

   PR #702 treated the whole General Punctuation block as fullwidth. In the issue701 sample, that makes characters such as U+2014 get rotated as fullwidth glyphs, so horizontal rule-like text turns into vertical bars or broken line fragments.

   This patch excludes the whole General Punctuation block from  `ass_codepoint_is_fullwidth()` and handles the affected Common/Inherited script glyphs separately. `needs_vertical_common_substitutions()` shapes these runs in top-to-bottom direction only for vertical substitutions, then copies the substituted glyph IDs back into the original run when clusters still match.

3. Fix vertical punctuation/symbol positioning.

   Some Common/Inherited-script glyphs need vertical substitution behavior, but their placement should not be identical to ordinary rotated fullwidth glyphs.
   Without the extra positioning step, punctuation and symbols in vertical text are visibly offset compared with VSFilter/GDI.

   This patch adds `needs_vertical_common_centering()` and tracks this case with `DECO_VERT_SUB`, allowing the outline rotation path to apply the additional centering needed for vertically substituted common glyphs.

libass with fix:
issue701-localfont-libass-1323

vsfilter:
issue701-localfont-vsfilter-1323


rcombs (commented):
I am extremely against making any of this hinge on the user-default locale except as a last-order fallback. See eg #607.

Masaiki:
> I am extremely against making any of this hinge on the user-default locale except as a last-order fallback. See eg #607.

@rcombs I removed the font selection related changes. Since there's already a more comprehensive PR addressing that part, it should be handled by that corresponding PR. This current PR only focuses on the vertical layout issue.

TheOneric:
Just looking at the images:
It’s clearly better tan now, but some differences with VSFilter remain. For reference and potential future patch sets, do you know or have any guesses about where these differences come from?  
Notably the placement of the roated `「` appears to be _worse_ than in https://github.com/libass/libass/pull/702#issuecomment-1871058772

Masaiki:
> Just looking at the images: It’s clearly better tan now, but some differences with VSFilter remain. For reference and potential future patch sets, do you know or have any guesses about where these differences come from? Notably the placement of the roated `「` appears to be _worse_ than in [#702 (comment)](https://github.com/libass/libass/pull/702#issuecomment-1871058772)

I looked into the `「` regression and the `、。` positioning issue again.

The root problem seems to be that the vertical positioning correction was effectively applied twice on the same visual axis: once through `GlyphInfo.offset.x` in shaping/layout, and again through `offs.x` during outline rotation via `DECO_VERT_SUB`.

That second outline-level correction does not have shaping context anymore, so it can overcorrect glyphs whose vertical alternate is already positioned well, such as `「」`. Keeping the adjustment in the shaper still helps punctuation like `、。`, while removing `DECO_VERT_SUB` avoids pushing already-correct glyphs too far.

So the updated patch keeps Common vertical glyph positioning in `ass_shaper.c`, but removes the extra outline-level correction from `ass_font.c`.

Updated render result with libass:
issue701-1323000

PR #934, opened 2026-06-11
ci: add read-only permissions and pin action SHAs

Alb3e3:
## Summary

CI workflows were missing top-level `permissions` declarations and referenced actions by mutable tags/branch names.

### Changes

- Add explicit workflow-level `permissions: contents: read` to the affected workflows
- Pin third-party action references to full commit SHAs

The workflows only need repository read access for checkout/build/test execution. Pinning action references to commit SHAs also ensures the audited action revisions are used instead of mutable tags.

### Verification

```bash
python3 - <<'PY'
import pathlib, yaml
for p in pathlib.Path('.github/workflows').glob('*.yml'):
    yaml.safe_load(p.read_text())
for p in pathlib.Path('.github/workflows').glob('*.yaml'):
    yaml.safe_load(p.read_text())
print('yaml ok')
PY
```

Result: `yaml ok`.

### DIFF
--- a/.github/workflows/ghci.yml
+++ b/.github/workflows/ghci.yml
@@ -5,6 +5,9 @@ on:
     branches: [master, ci, coverity_scan]
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ${{ matrix.os }}
@@ -98,10 +101,10 @@ jobs:
             output_stream.write(f"count={num_cpus}\n")
 
       - name: checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
 
       - name: download test samples
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
         with:
           repository: libass/libass-tests
           path: ${{ env.ART_SAMPLES }}
@@ -135,7 +138,7 @@ jobs:
           chmod a+x /tmp/docker_shell
 
       - name: Setup MSys2
-        uses: msys2/setup-msys2@v2
+        uses: msys2/setup-msys2@e9898307ac31d1a803454791be09ab9973336e1c  # v2
         if: matrix.msystem
         with:
           msystem: ${{ matrix.msystem }}
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     name: ${{ matrix.config.name }}
@@ -75,19 +78,19 @@ jobs:
       SCCACHE_MAXSIZE: 200M
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
         with:
           fetch-depth: '0'
 
       - name: Download test samples
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
         with:
           repository: libass/libass-tests
           path: ${{ env.ART_SAMPLES }}
 
       - name: Setup MSYS2
         if: matrix.config.msystem
-        uses: msys2/setup-msys2@v2
+        uses: msys2/setup-msys2@e9898307ac31d1a803454791be09ab9973336e1c  # v2
         with:
           msystem: ${{ matrix.config.msystem }}
           pacboy: >-
@@ -109,7 +112,7 @@ jobs:
 
       - name: Restore cache
         if: matrix.config.msvc
-        uses: actions/cache/restore@v5
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5
         with:
           path: ${{ env.SCCACHE_DIR }}
           key: windows-msvc-${{ matrix.config.arch }}-${{ steps.get_time.outputs.timestamp }}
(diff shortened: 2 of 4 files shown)


.github/workflows/ghci.yml
```diff
@@ -5,6 +5,8 @@ on:
     branches: [master, ci, coverity_scan]
   pull_request:
 
+permissions: read-all
```
kasper93: I think it's better to set repository with limited permissions in settings. And here only list what's needed. Note that `read-all` is certainly not needed.

TheOneric:
I changed the default perms to "only read repo content and packages" for the entire organisation.



To my reading of the setting description, write perms for _anything_ can still be obtained however, if the workflow just explicitly specifies it.  
The same will apply to the changes made in this PR; if a PR is made by someone who previously contributed some change, then they can just change the permissions and contents of the workflow however they want and it will run automatically. Pinning and restricting perms in-file does not help with this.  

Imho, read-only perms ought to be the default, PR workflows never write to the target repo and any extension of perms require approval regardless of prior contributions.

Regarding pinning hashes, this makes it harder to keep track of and greatly increases churn. We generally _want_ minor and patch updates for actions. The majority of actions is also directly from GitHub itself (`actions/*`) and if GitHub employees are compromised, then the libass repo can be compromised otherwise too. Only MSys2 and cross-platform-actions are third-party actions.  
But, since those actions will to my understanding only have read access to the git repo and (non-existent) published packages, which is already all public, can a compromised action even cause any harm other than CI failures?

kasper93:
> To my reading of the setting description, write perms for _anything_ can still be obtained however, if the workflow just explicitly specifies it. The same will apply to the changes made in this PR; if a PR is made by someone who previously contributed some change, then they can just change the permissions and contents of the workflow however they want and it will run automatically. Pinning and restricting perms in-file does not help with this.

Workflow jobs are scoped to the calling repository, unless they are run in the context of parent repository, in which case it is not possible to edit them in PR, because the definition of the job is used from the master brach.

For easier understanding, let's consider example. There are two triggers
`pull_request`: this is the most standard one, it runs in the context of the **fork**, any secrets and permissions apply to the fork. You cannot publish or read anything private from main repository. Regardless of any `permissions`.
`pull_request_target`: this is the same as the above, it triggers on the pull requests, but it runs in the context of `parent` repository, and the job definition is run from `master` branch. Any edit to this workflow in pull request will not have an effect until it is merged to master.

> Imho, read-only perms ought to be the default, PR workflows never write to the target repo and any extension of perms require approval regardless of prior contributions.

Correct. Historically though, it is more permissive by default.

> Regarding pinning hashes, this makes it harder to keep track of and greatly increases churn. We generally _want_ minor and patch updates for actions. The majority of actions is also directly from GitHub itself (`actions/*`) and if GitHub employees are compromised, then the libass repo can be compromised otherwise too.

I agree on this point. Supply chain attacks are real, but at the point when (`actions/*`) are compromised, just steal my secrets it's over anyway...

> Only MSys2 and cross-platform-actions are third-party actions. But, since those actions will to my understanding only have read access to the git repo and (non-existent) published packages, which is already all public, can a compromised action even cause any harm other than CI failures?

Depends on the job they run it, for example if you have workflow to publish (test) release that downloads artifact from build job using `actions/download-artifact`, and publishes this, naturally you would have permissions there to write content. Again, though if `actions/` are compromised it's over anyway. For smaller 3rd-party actions it indeed better to limit scope they can access.

TheOneric:
> For easier understanding, let's consider example. There are two triggers
> `pull_request` […] `pull_request_target`

Right, thank you for the reminder and update on the current situation, much appreciated. It’s been a while since I looked at GHA workflow perms.

> Depends on the job they run it, for example if you have workflow to publish (test) release that downloads artifact from build job using actions/download-artifact, and publishes this, naturally you would have permissions there to write content.

Indeed, thanks for making this explicit.

In libass though, we only use the `pull_request` trigger and never use GHA to publish releases, packages *(except in the out-of org `oldlibs` container build)* or even workflow artefacts. As I understand, this means even a compromised third-party action cannot actually do anything other than causing spurious failures. Am I’m getting this right?  
We do have one secret token for submitting builds to Coverity for scanning. But this is passed as a env variable only to the in-lined Coverity step alone. Can compromised actions query repo secrets they have not explicitly been granted access to on their own?

arch1t3cht:
> But, since those actions will to my understanding only have read access to the git repo and (non-existent) published packages, which is already all public, can a compromised action even cause any harm other than CI failures?

One notable gotcha here is that, afaik, read-only workflows can still write to caches, so a compromised action could run a cache poisoning attack to then compromise another workflow that *does* have write access. (See [here](https://tanstack.com/blog/npm-supply-chain-compromise-postmortem) for an example of how this was recently exploited in the wild.)

This is not an issue for libass specifically, since that does not have any other workflows with write access, but it's something to keep in mind in case such workflows are ever added in the future.

Alb3e3:
I checked this against the current libass workflows in this PR.

For the repository as it exists today, your reading is mostly right: the changed workflows use `pull_request`, `push`, and `workflow_dispatch`; none use `pull_request_target`, and this PR does not introduce a write-token publish workflow. A compromised third-party action in ordinary PR CI would be able to execute code in that job and tamper with that job's workspace/output, but it should not be able to write to `libass/libass` through `GITHUB_TOKEN`.

On secrets: GitHub does not give actions a way to enumerate or fetch arbitrary repository secrets. A secret becomes available only when the workflow maps it into a job or step. The Coverity token is only mapped on the Coverity step, and that step is gated with `github.repository == 'libass/libass' && github.event_name != 'pull_request'`, so fork PRs should not receive it. The caveat is same-job persistence: anything that runs earlier in a trusted, secret-bearing job can modify the workspace or runner state and affect later steps. In the current Coverity matrix, I only see GitHub-owned checkout actions before the Coverity step; the pinned MSYS2/QEMU third-party actions are not on that Coverity path.

On caches: `contents: read` does not make cache writes impossible. That does not provide direct repository write access, and libass currently has no write-token publish workflow for cache poisoning to pivot into, but cache poisoning can still affect future CI execution and would matter more if a write-token workflow is added later.

So for libass specifically, the immediate practical risk reduction is modest after setting the org/repo default to read-only. The explicit `contents: read` declarations still make least privilege local to these workflow files, and the SHA pins mainly prevent unreviewed tag movement or release changes in third-party actions. If the maintenance cost of SHA pins is not worth that tradeoff for this project, I can narrow this PR to permissions only.

PR #935, opened 2026-06-15
ci/gha/meson: use vswhere to find Visual Studio

kasper93:
Instead of assuming instalation path and version.

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -70,7 +70,6 @@ jobs:
         shell: ${{ matrix.config.shell || 'bash' }}
 
     env:
-      VS: "C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise"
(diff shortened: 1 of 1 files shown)


TheOneric:
Thank you!

Curiously, it seems GH didn’t _upgrade_, but _downgrade_ VS in the job environment. Instead of the previously hardcoded and now failing `22`, `vswhere ` now locates: ` VS=C:\Program Files\Microsoft Visual Studio\18\Enterprise`

kasper93:
> Thank you!
> 
> Curiously, it seems GH didn’t _upgrade_, but _downgrade_ VS in the job environment. Instead of the previously hardcoded and now failing `22`, `vswhere ` now locates: ` VS=C:\Program Files\Microsoft Visual Studio\18\Enterprise`

It's just Microsoft changing their naming. Old VS2022 would derive directory name from marketing name of the product. New VS2026 derive it from the actual product version, which is currently 18.

PR #936, opened 2026-06-17
ass_rasterizer: fix ilog2 impl on MSVC

kasper93:


kasper93:
Fixes compilation on clang with msvc sdk. See:  https://github.com/mpv-player/mpv/actions/runs/27693171495/job/81909291366

astiob (commented):
Thanks!

Any particular reason to prefer `__builtin_clz` on MSVC-mode Clang?

kasper93:
> Any particular reason to prefer `__builtin_clz` on MSVC-mode Clang?

I prefer using consistently using clang features if available, this is also common across other projects. In this particular cases, clang will optimize both cases to the same code anyway, but that's not always the case.

MrSmile (approved):
Why `int` worked before though? Was MSVC less strict or the pointer type changed at some point?

kasper93:
> Why `int` worked before though? Was MSVC less strict or the pointer type changed at some point?

It was less strict. And the values that it can write would not exceed int size so it kinda worked, but was aliasing violation.

TheOneric (approved):

Issue #937, opened 2026-06-18
Crash: AddressSanitizer: SEGV on unknown address


Fieash:
### libass version

libass 0.17.4-32-g35ba2ca
commit 35ba2caf5edc8f69cf6e4101c5f60d3a6853984

### Is it a regression?

Unknown. I have not tested this against older releases.

### ASS Sample

This reproduces through the Matroska streaming packet path using ass_process_chunk(), not as a standalone .ass file.

codec_private.ass
```text
[Script Info]
ScriptType: v4.00+
PlayResX: 384
PlayResY: 288

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL,
MarginR, MarginV, Encoding
Style: Default,Arial,18,&H00FFFFFF,&H00FFFF00,&H00000000,&H80000000,0,0,0,0,100,100,0,0,1,2,3,2,20,20,20,1

[Events]
Format: Layer, Start, End, Style, Name, MarginL, MarginR, MarginV, Effect, Text
```

Packet payload passed to ass_process_chunk():

packet.txt
```text
-1,0,Default,,0,0,0,,negative ReadOrder crash demo
```

Minimal reproducer:
```c
//minimal_poc.c
#include <ass/ass.h>

  int main(void) {
      ASS_Library *lib = ass_library_init();
      ASS_Renderer *renderer = ass_renderer_init(lib);

      ass_set_frame_size(renderer, 640, 480);
      ass_set_fonts(renderer, NULL, "sans-serif", ASS_FONTPROVIDER_NONE, NULL, 1);

      ASS_Track *track = ass_new_track(lib);

      const char priv[] =
          "[Script Info]\n"
          "ScriptType: v4.00+\n"
          "PlayResX: 384\n"
          "PlayResY: 288\n"
          "[V4+ Styles]\n"
          "Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, "
          "OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, "
          "ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, "
          "Alignment, MarginL, MarginR, MarginV, Encoding\n"
[...]

PR #938, opened 2026-06-22
ass: bounds-check negative index in clear_read_order_bit

kali834x:
clear_read_order_bit derives index from an event ReadOrder via id >> 5 and only checks the upper bound against read_order_elems before touching read_order_bitmap. ReadOrder comes from atoi() on a chunk field, so a negative value yields a negative index that slips past the check and reads/writes before the allocation when ass_prune_events sweeps an expired event. test_and_set_read_order_bit avoids this because resize_read_order_bitmap rejects negative ids up front; the clear path has no equivalent guard, so add the index >= 0 lower bound to match. reachable from ass_process_chunk with a crafted ReadOrder when an app toggles ass_set_check_readorder around the insert (the negative entry then stays in the bitmap-tracked set until prune). asan on the prune confirms a heap-buffer-overflow 4 bytes before the bitmap, gone after the guard.


TheOneric:
This was already reported in #937 and was now fixed in 23da65a130bb8c5b2cf0c7df0dd7d424607f98f1. The patch proposed here unfortunately has issues on platforms with a logical instead of arithmetic right shift by still having a chance of causing spurious bit clears.

EDIT: or well, given no valid index should be as large as a logically right-shiften negative value (in complement of two representation), it might have been safe and logically sound by the `< elements` check alone. But this is much less clear code and more prone to be forgotten and accidentally broken should we ever refactor how the bitmap works

PR #939, opened 2026-06-25
render: avoid signed overflow in vector clip scale shift

kali834x:
blend_vector_clip builds the clip scale_base with lshiftwrapi(1, clip_drawing_scale - 1). clip_drawing_scale comes straight from the \clip(scale, drawing) tag via parse_vector_clip, which unlike the \p handler in ass_parse_tags never clamps it, so an INT32_MIN scale overflows the clip_drawing_scale - 1 subtraction before lshiftwrapi runs, defeating the very helper meant to keep this shift well defined. doing the subtraction in unsigned leaves the wrapped result unchanged for every input while removing the overflow. ubsan flags it at ass_render.c:712 on {\clip(-2147483648,m 0 0 l 100 0 100 100 0 100)}.


kali834x:
gentle ping

TheOneric:
xy- and guliverkli(1)-VSFilter also [prevent negative values on `\clip(scale, …)` inline overrides](https://github.com/pinterf/xy-VSFilter/blob/4911cacb102bc0cbd41449793e54473265b04e62/src/subtitles/RTS.cpp#L2340).
We should adopt this instead as negative values don't make much sense here in the first place, it is cleaner and improves compat.
Note, due to the `& 31` added in d1140ac30240c0490da861d6ed8c76cc931584db negative values were still technically safe, just nonsensical.

Fixed to match VSFilter via 4af3320969f1a4a63ee37b72002f39278151f8fa

But even if VSFilter accepted negative values, the suggested patch here just trades a “technically undefined but in practice likely safe” *(also see discussion about adding `-fwrapv` to compile flags to make it explicitly safe)* for “technically implementation-defined result but in practice likely still the same as before”  since `lshiftwrapi` takes signed arguments. Which isn’t really much of an improvement imho. A proper fix needs to actually _fix_ the result to a unambiguous output, matching VSFilter where relevant.  
Though on the other hand, we do admittedly already rely on assigns of unrepresentable unsigned alues to signed types just re-using the bit content just now interpreted as complement-of-two. For the return value of the very same `lshiftwrapi` function even.

PR #940, opened 2026-06-25
Alternative approach to bitmap shifting

MrSmile:
This is my variant of PR #749, complete with SSE2/AVX2 and NEON assembly.
Main difference is the absence of additional allocations and uses of a scratch array.
Should be useful for direct performance comparisons on target hardware.

### DIFF
--- a/Makefile_util.am
+++ b/Makefile_util.am
@@ -52,6 +52,7 @@ endif
 checkasm_checkasm_SOURCES = \
     checkasm/rasterizer.c \
     checkasm/blend_bitmaps.c \
+    checkasm/shift.c \
     checkasm/be_blur.c \
     checkasm/blur.c \
     checkasm/checkasm.h checkasm/checkasm.c \
--- a/checkasm/checkasm.c
+++ b/checkasm/checkasm.c
@@ -59,6 +59,7 @@ static const struct {
 } tests[] = {
     { "rasterizer", checkasm_check_rasterizer },
     { "blend_bitmaps", checkasm_check_blend_bitmaps },
+    { "shift", checkasm_check_shift },
     { "be_blur", checkasm_check_be_blur },
     { "blur", checkasm_check_blur },
     { 0 }
--- a/checkasm/checkasm.h
+++ b/checkasm/checkasm.h
@@ -61,6 +61,7 @@ int xor128_rand(void);
 
 void checkasm_check_rasterizer(unsigned cpu_flag);
 void checkasm_check_blend_bitmaps(unsigned cpu_flag);
+void checkasm_check_shift(unsigned cpu_flag);
 void checkasm_check_be_blur(unsigned cpu_flag);
 void checkasm_check_blur(unsigned cpu_flag);
 
--- a/checkasm/meson.build
+++ b/checkasm/meson.build
@@ -2,6 +2,7 @@ checkasm_src = files(
     'checkasm.c',
     'rasterizer.c',
     'blend_bitmaps.c',
+    'shift.c',
     'be_blur.c',
     'blur.c',
 )
--- a/checkasm/shift.c
+++ b/checkasm/shift.c
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2026 libass contributors
+ *
+ * This file is part of libass.
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include "ass_compat.h"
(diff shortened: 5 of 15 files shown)


TheOneric:
For x86 this runs into the similar performance characteristics discrepancy between the CPU micro-architectures MrSmile and I run on our systems, leading to `ssse3` versions being again faster for me on Zen+. Since it’ſ still significantly faster than the C version and we let this slip before, this is not a blocker though imho.  

Since `checkasm --function="*shift*" --bench-c` doesn’t actually run benchmarlks for `nop` and the C version, I don’t have a renewed, now more direct comparison of the C implementation between this and the other approach.

I was trying to also bench this on aarch64, but all aarch64 systems i have access to just fail with *“checkasm: unable to access cycle counter”* even after setting (when available) `/proc/sys/kernel/perf_user_access` to `1`. Is there anything else I must or can do to get checkasm benches to work on ARM Linux?

x86 results on Zen+:
```
# BENCH_RUNS  (1 << 19)
# C version with gcc -g -O2 -fvect-cost-model=dynamic
# gcc 14.2.0

nop: 59.2
shift_c: 793.5
shift_sse2: 383.4
shift_ssse3: 315.8
shift_avx2: 374.1
```

rcombs:
On an M4 machine, both branches give values within measurement-error range for me:
```
checkasm: using random seed 2732232029
NEON:
 - shift.shift                      [OK]
checkasm: all 1 tests passed
nop: 0.0
shift_c: 5.5
shift_neon: 0.9
```

The `shift_c` values vary a bit (I've seen values ranging from 4.9 to 5.7), but `shift_neon` is highly consistent at 0.9 in every test I ran across both branches. So, at least on Apple cores, I have no particular preference as to which PR wins.

TheOneric:
After some modifications unrelated to the shifting logic, here are now some results for other aarch64 CPUs.

Changes are: tweaks proposed by rcombs to get it working without cycle counter access on Linux, increased bench run count and notably a change to get #749 to compile with gcc 14 (it is also failing in our gcc 13 CI, haven’t tested newer gcc builds)

I am not at all familiar with the aarch64 NEON assembly syntax and the latter is just a result of comparing with the instructions used in this PR and compiler suggestions. It still seems to pass checkasm teests, so it either should be fine, or checkasm tests are lacking

<details>
<summary>Patch fixing gcc build for #749</summary>

```diff
diff --git a/libass/aarch64/shift_bitmap.S b/libass/aarch64/shift_bitmap.S
index 7ac28f7..988c389 100644
--- a/libass/aarch64/shift_bitmap.S
+++ b/libass/aarch64/shift_bitmap.S
@@ -72,7 +72,7 @@ function shift_bitmap_neon, export=1

     str        q4, [x0, x7] // buf[x] = px

-    mov        v2.8h, v17.8h // prev = b_x
+    mov        v2.16b, v17.16b // prev = b_x

     add        x7, x7, #0x10 // x++
     cmp        x7, x2 // x < w
```
</details>


<details>
<summary>Patch for usermode timing and bench runs</summary>

```diff
diff --git a/checkasm/checkasm.h b/checkasm/checkasm.h
index ac66e55..a09ed17 100644
--- a/checkasm/checkasm.h
+++ b/checkasm/checkasm.h
@@ -81,7 +81,7 @@ int float_near_abs_eps_array(const float *a, const float *b, float eps,
 int float_near_abs_eps_array_ulp(const float *a, const float *b, float eps,
                                  unsigned max_ulp, int len);

-#define BENCH_RUNS (1 << 16) /* Trade-off between accuracy and speed */
+#define BENCH_RUNS (1 << 19) /* Trade-off between accuracy and speed */

 /* Decide whether or not the specified function needs to be tested */
 #define check_func(func, ...)\
@@ -128,6 +128,33 @@ static inline uint64_t readtime(void) {
 #ifdef _MSC_VER
 #include <windows.h>
 #define readtime() (_InstructionSynchronizationBarrier(), ReadTimeStampCounter())
+#elif !defined(AARCH64_USER_CYCLE_COUNTER)
+#include <time.h>
+static inline uint64_t readtime(void) {
+  #ifdef _WIN32
+    static LARGE_INTEGER freq;
+    LARGE_INTEGER        ts;
+    if (!freq.QuadPart) {
+        if (!QueryPerformanceFrequency(&freq))
+            return -1;
+    }
+    if (!QueryPerformanceCounter(&ts))
+        return -1;
+    return UINT64_C(1000000000) * ts.QuadPart / freq.QuadPart;
+  #else
+    struct timespec ts;
+    clockid_t id;
+    #ifdef CLOCK_MONOTONIC_RAW
+    id = CLOCK_MONOTONIC_RAW;
+    #else
+    id = CLOCK_MONOTONIC;
+    #endif
+    if (clock_gettime(id, &ts) < 0)
+        return -1;
+    return UINT64_C(1000000000) * ts.tv_sec + ts.tv_nsec;
+  #endif
+}
+#define readtime readtime
 #else
 static inline uint64_t readtime(void) {
     uint64_t cycle_counter;
```
</details>

The gist is, we can see the same heavy-compiler dependence for the C implementation as previously discussed in the other PR and x86, with neither being always clearly better. While for ASM, on the oldest CPU 749 is a tad faster, while on newer CPU designs its the other way around.

(Neoverse N1 is similar to but not quite the same as Cortex A76 design)

For reference, tests use commit f4a4706ddccf1d1cbce2fa122914b5895765455c for #940 and 6c67487b17248cf662f22af1a9b3966cc5b47460 for #749.  
Used gcc version is 14.2.0 and clang 21.1.8 everywhere.
I ran each full bench at least twice and timings changed only insignificantly between runs unless a range is shown.

**Assembly implementation**
| CPU | avg nop time | 940 | 749 |
|--|--|--|--|
| BCM2837 (ARM Cortex-A53) @ 1.4GHz | 66.8 | 1,399.5 | 1,122.4 |
| SDM632 (Cortex-A73 + Cortex-A53) @1.8 GHz | 52.0 | 495.5 | 587.3 |
| Neoverse-N1 | 37.1 |160.3 | 218.7 |

**C implementation**

| CPU | compiler | 940 | 749 |
|--|--|--|--|
| BCM2837 (ARM Cortex-A53) @ 1.4GHz | gcc | 2,563.1 | 11,082.1 | 
| BCM2837 (ARM Cortex-A53) @ 1.4GHz | clang | 11,656.1 – 11,958.8 | 1,701.3 | 
| SDM632 (Cortex-A73 + Cortex-A53) @1.8 GHz | clang | 4,421.3 | 993.7 |
| Neoverse-N1 | gcc | 388.7 | 1,493.3 |
| Neoverse-N1 | clang | 1,359.0 | 360.0 |

PR #941, opened 2026-07-03
outline: avoid abs() UB in coordinate range checks

kali834x:
ass_outline_add_point guards coordinates with abs(pt.x) > OUTLINE_MAX, but pt.x/pt.y are int32_t and abs() is undefined for INT32_MIN, where it yields a negative value so the check lets an out-of-range point through. a drawing coordinate of -33554432 reaches this from ass_drawing_parse (double_to_d6 rounds it to INT32_MIN), so the point lands in the outline in violation of the [-OUTLINE_MAX, OUTLINE_MAX] invariant and then overflows the int32 math in ass_outline_construct during rendering. replace the abs()-based range test with an explicit two-sided comparison here and in the sibling checks in ass_outline_rotate_90 and ass_outline_scale_pow2; bit-identical for valid inputs, well-defined for INT32_MIN.


rcombs (approved):
For reference, in practice compilers seem to implement `abs` such that `abs(INT_MIN)` returns `INT_MIN`, which means the check passes in that case. Is there any upshot beyond potentially allocating an excessive amount of memory? (Seems like a fine change either way; I'm just curious).

There are a few other cases in the code where we compare `abs(v)` to `OUTLINE_MAX`, though mostly in debug assertions, so these are probably the only cases that can have actual impact on release builds.

kali834x:
Right, abs(INT32_MIN) coming back as INT32_MIN is exactly the case that slips through: it stays negative so it never trips `> OUTLINE_MAX` and the point gets added.

The upshot is more than a big allocation though. INT32_MIN is a poison value for the int32 math downstream. With the repro drawing (`{\p1}m 0 0 l -33554432 0 l 0 100{\p0}`, which double_to_d6 rounds to INT32_MIN) the point lands in the outline, and then in ass_outline_construct the bbox subtraction `v->advance = bbox.x_max - bbox.x_min` evaluates `0 - -2147483648` and overflows int32. UBSan flags that, plus the negation back in ass_outline_add_point, so it's real UB on the release path rather than just excessive memory.

Agreed on the rest: the other abs()/OUTLINE_MAX comparisons are in asserts, so they only bite in debug builds. I left those as-is to keep this to the three checks that actually run in release.

rcombs:
I would've appreciated disclosure on the LLM usage instead of attributing comments directly to the user, but 🤷‍♀️ the patch itself is still fine. (The out-of-range `abs` is already "real UB", so saying that the exact same underflow happening downstream is "real UB" is missing the point pretty spectacularly.)

MrSmile:
Sadly, there's no sane `abs` with unsigned result...
Asserts should be fixed too, also `valid_point()` function.
To reduce copy-paste maybe it's worth to introduce something like
```c
static inline bool vec_in_range(ASS_Vector pt, int32_t lim)
{
    return pt.x >= -lim && pt.x <= lim && pt.y >= -lim && pt.y <= lim;
}
```

TheOneric:
yes, i too think this should be macroised or static-inlined and applied consistently. Either it’s not serious enough to bother at all, or serious and properly treated everywhere.

In theory, `-positive_value` is _also_ not guaranteed to be representable. But in practice, with C23 now mandating complement-of-two representation, i think this should be fairly safe _(even if we don't yet use/require C23)_

kali834x:
Pushed the helper as suggested, in ass_outline.h so the rasterizer assert can use it too. Converted the debug asserts and valid_point() (its labs() on FT_Pos had the same edge case). scale_pow2 keeps an open-coded check since its limits differ per axis and the single-lim helper doesn't fit there. Repro is still clean under UBSan and checkasm passes.

@rcombs fair, the abs() call itself is the UB and the downstream overflow is just where the bad value ends up, sloppy wording on my part. And yes, I do use an LLM a bit for research and sanity-checking my reasoning, should have said so.

MrSmile (approved):
LGTM, but commits should be squashed before merge.

kali834x:
Squashed into a single commit. No code changes, the tree is identical to the reviewed head.

kali834x:
gentle ping

kali834x:
gentle ping

PR #944, opened 2026-07-20
Security reporting and contributing guidelines

TheOneric:
Inspired by recent private (per-email to a single maintainer) and public security reports, AI slop PRs and such. And often fuzzing or security reports reïnventing the existing `fuzz` utilities *(only once so far actually using different API)*, making reproduction a bit more tedious and also initially more effort for the reports. Hopefully, with this we’ll get things reported via the proper channels monitored by all maintainers.

The layout of both documents was inspired by mpv’s existing counterparts, but most of the text content was rewritten from scratch or at least modified in some way. The least changed parts are commit and copyright guidelines

I believe `SECURITY.md` just sums up what we converged on in prior discussion and thus should already match existing consesn, except for the alternative reporting approaches for those without a GitHub account. In particular, whether everyone affected is actually fine with the email-based reporting option and will notice incoming reports.  
Notably, the rendered version of this file will show up in GitHub’s “Security and quality” tab too.

`CONTRIBUTING.md` too describes an alternative patch submission way, but furthermore also has a short Code of Conduct and AI policy statement.  

The former is intentionally kept brief and thus also rather general. There’s also merit to being more detailed and explicit.
On the other hand, one can also argue, that in this case we might not need a CoC statement at all since everything worked well without.  
I chose to go with a brief and general statment for the following reasons:
- I suspect barely anyone will actually read a lengthy document before interacting
- being too specific and trying to explicitly enumerate each possible individual case is prone to forgetting sth or creating loopholes.
- nonetheless, having the general idea stated explicitly is nice imo and it can be a reassuring sign for potential contributors  

It is also true though, that no form of CoC helps if the actual culture doesn’t live up to it; fortunately imo we’re good on that rn? *(But if there’s anything currently bothering you I’m oblivious too, please do speak up)*  
In any event, I’m sure the wording could be tweaked a bit and if there’s a shared preference for taking another CoC approach, I’m open to discuss this too.

The latter is basically [Gentoo’s policy](https://wiki.gentoo.org/wiki/Project:Council/AI_policy) with a slight tweak and two additional sentences, to be explicit about vendored/copied code and to not _technically_ disallow assistive input and output methods like screenreaders and such. You were all likely already exposed to ""AI"" contribution and general ethic discussions over the past year(s), so I’ll spare you a recap of arguments for a ban *(but a non-exhaustive list can be found with the linked Gentoo policy)*. If you were lucky enough to avoid this or there’s disagreement about it, an elaboration can still follow.  
On a purely personal note apart from more general rationales though, I find the prospect and dealing with AI slop submission extremely demotivating.

One tangential observation made while writing: meson doesn’t accept path’s relative to the project root for `ART_SAMPLES` unlike autotools which is a bit inconvenient. But we recommend autotools for development anyway in our `README.md`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,165 @@
+# Contributing to libass
+
+## General
+
+If you’re not sure how to start, just getting an unpatched git HEAD
+building as per our `README.md` and checking out the available
+config options might be a good idea.
+
+If your patch is quite large and/or requires significant changes to the internal
+code organisation or interfaces or affects public API in any way, it is
+**strongly** recommended to discuss the concept with us first.
+This saves you time by not investing effort into drafting patches based on an
+unviable approach, and is better for us too as we don’t have to review and
+figure out how to salvage a large patch set based on a flawed concept.
+
+Try to match the existing coding style of the individual file(s) you modify.
+
+For discussion or to ask questions about the code base or your patch,
+you can reach us via IRC *(see `README.md`)*.
+For more lengthy or asynchronous discussions about patch concepts,
+RFC issues or GitHub discussions can be used too.
+
+## Code of Conduct
+
+Aim to stay courteous and respectful.
+We want to be welcoming space for all sorts of diverse participants.
+
+Naturally, to remain a generally welcoming and friendly space
+we must not be welcoming nor tolerant of anything or anyone
+inherently detrimental to this. Thus bigotry, \*ism, \*phobia
+and other such intolerance have no place here.
+
+## Testing
+
+Make sure your changes compile in all supported build systems.
+
+Beyond that, we have a simple *(unfortunately rather incomplete)*
+test suite at [libass-tests](https://github.com/libass/libass-tests).
+To use it, clone this repository and during configure time enable the
+`compare` and `fuzz` utilities and finally also indicate where the
+cloned repo is located.
+
+For autotools, this can be done with:
+
+```
+./configure --enable-compare --enable-fuzz ART_SAMPLES=../libass-tests
+make -j "$(nproc)"
+make check
+```
+
+For meson:
+
+```
+meson setup -Dcompare=enabled -Dfuzz=enabled -Dart-samples="$HOME"/code/libass-test build_meson
+meson compile -C build_meson
+meson test -C build_meson test
+```
+
+On some platforms certain tests relying on system libraries
+might need to be skipped. You can use our CI setup for reference
+and check documentation in the test repo for further info about
+how to tweak the test suite.
+
+## Commit guidelines
+
+### Split changes into multiple commits
+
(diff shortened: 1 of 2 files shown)


SECURITY.md
```diff
@@ -0,0 +1,73 @@
...
+If you find something affecting a git state after the most recent release
+but no longer current git HEAD, then if the fix seems intentional this too
+will no longer be relevant.  
+If however, it seems like a mere coincident with the exploit fixed unintentionally
+by chance it might still be worth to bring to our attention.
+
+## What qualifies as a security bug
+
+Security bugs should be in the core libass library.
+Our extra utilities like `test`, `compare`, etc are not build by default,
```
rcombs: *built

SECURITY.md
```diff
@@ -0,0 +1,73 @@
...
+Only the most recent release and the unreleased git HEAD state are considered supported.
+Testing and analysis are strongly encouraged to be performed on one or both of these.
+
+If you find something affecting an older but not the most recent release,
+we most likely not consider this relevant.
+
+If you find something affecting a git state after the most recent release
+but no longer current git HEAD, then if the fix seems intentional this too
+will no longer be relevant.  
+If however, it seems like a mere coincident with the exploit fixed unintentionally
```
rcombs: *coincidence

SECURITY.md
```diff
@@ -0,0 +1,73 @@
...
+by chance it might still be worth to bring to our attention.
```
rcombs: *worth bringing to

TheOneric: fixed this and the other two; thanks!

CONTRIBUTING.md
```diff
@@ -0,0 +1,167 @@
...
+### Always squash fixup commits when making changes to pull requests
+
+- If you make fixup commits to your pull request, you should generally squash
+  them with `git rebase -i`, `git history fixup` or similar.
+  We prefer to have pull requests in a merge-ready state suitable
+  to be applied as a fast-forward to our development branch.
+
+### Write good commit messages
+
+
```
MrSmile: Extra line.

MrSmile (commented):
Not sure about such strict prohibition of LLM use. I fear that in the not so distant future it would be hard to find any code without some form of AI involvement. I think using LLM as a tool would be fine, given that submitter know what they're doing and remain responsible for submitted code.

rcombs:
I'll note I have mixed feelings on the ML section, but as I'm currently at Nvidia, I have a conflict of interest and think it'd be appropriate to recuse myself from that debate.

CONTRIBUTING.md
TheOneric: fixed, ty

TheOneric:
Sorry, i’m currently short on time and this doesn’t seem too urgetn yet. I’ll take a while longer to respond regarding the "AI"/LLM/ML policy comments

PR #946, opened 2026-08-10
ass_font: Better support for symbol cmaps

moi15moi:
The previous code unconditionally mapped every MS Symbol-encoded font to the 0xF000 range (`return 0xF000 | symbol`). Even though it works for most symbol fonts, it doesn't always work (for example, it doesn't work with the font "Baoloc").

Now we replicate how GDI actually handles symbol fonts.

Note that https://github.com/libass/libass/issues/319 remains unresolved/relevant, because it still doesn't use the system code page to map the symbol to the correct glyph index.

### DIFF
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -261,6 +261,7 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
         switch (face->charmap->encoding) {
         case FT_ENCODING_MS_SYMBOL: {
             TT_OS2 *os2 = FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
+
             if (os2) {
                 // See: https://learn.microsoft.com/en-us/typography/legacy/legacy_arabic_fonts
                 int ARABIC_CHARSET_SIMPLIFIED  = 178;
@@ -271,7 +272,48 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
                 if (charset == ARABIC_CHARSET_TRADITIONAL)
                     return ass_font_charmap_arabic_traditional(symbol);
             }
-            return 0xF000 | symbol;
+
+            // GDI only supports cmap format 4 for symbol fonts
+            // GDI reads the cmap's first segment startCount and
+            // uses it to know what is the first character.
+            // FT_Get_First_Char() is NOT exactly equivalent,
+            // but for well-formatted cmap, it is.
+            FT_UInt gindex;
+            FT_ULong first_char = FT_Get_First_Char(face, &gindex);
+
+            uint32_t offset;
+            switch (first_char & 0xFF00) {
+                case 0xF000:
+                    offset = 0xF000;
+                    break;
+
+                case 0xE000: // EUDC fonts
+                case 0x0000:
+                    offset = 0;
+                    break;
+
+                default:
+                    offset = (uint32_t) first_char - 0x20;
+                    break;
+            }
+
+            bool is_symbol_glyph_set = (offset) || (os2 &&
+                            os2->panose[0] == 5 && /* PAN_FAMILY_PICTORIAL */ 
+                            (os2->fsSelection & 0xFF) == 0 /* ANSI_CHARSET */);
+
+            if (!is_symbol_glyph_set)
+                return symbol;
+
+            if (symbol <= 0xFF)
+                // TODO: Use the codepage of the language track: https://github.com/libass/libass/issues/319#issuecomment-1457223424
+                // GDI use the codepage of the system to get the glyph id.
+                // There is an exception to that.
+                // If the system code page is 932, 936, 949, or 950, it will actually use 1252.
+                return offset + symbol;
+            else if (symbol >= 0xF020 && symbol <= 0xF0FF)
+                return offset + (symbol & 0xFF);
+
+            return 0;
         }
         case FT_ENCODING_MS_SJIS:
         case FT_ENCODING_MS_GB2312:
--- a/libass/ass_font.c
+++ b/libass/ass_font.c
@@ -247,7 +247,7 @@ void ass_charmap_magic(ASS_Library *library, FT_Face face)
     }
 }
 
-static uint32_t convert_unicode_to_ms_symbol(FT_Face face, uint32_t symbol)
+static bool convert_unicode_to_legacy_symbol(FT_Face face, uint32_t symbol, uint32_t *legacy_symbol)
 {
     TT_OS2 *os2 = FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
 
@@ -256,11 +256,24 @@ static uint32_t convert_unicode_to_ms_symbol(FT_Face face, uint32_t symbol)
         int ARABIC_CHARSET_SIMPLIFIED  = 178;
         int ARABIC_CHARSET_TRADITIONAL = 179;
         uint8_t charset = (os2->fsSelection >> 8) & 0xFF;
-        if (charset == ARABIC_CHARSET_SIMPLIFIED)
-            return ass_font_charmap_arabic_simplified(symbol);
-        if (charset == ARABIC_CHARSET_TRADITIONAL)
-            return ass_font_charmap_arabic_traditional(symbol);
+        if (charset == ARABIC_CHARSET_SIMPLIFIED) {
+            *legacy_symbol = ass_font_charmap_arabic_simplified(symbol);
+            return true;
+        }
+        if (charset == ARABIC_CHARSET_TRADITIONAL) {
+            *legacy_symbol = ass_font_charmap_arabic_traditional(symbol);
+            return true;
+        }
     }
+
+    return false;
+}
+
+static uint32_t convert_unicode_to_ms_symbol(FT_Face face, uint32_t symbol)
+{
+    uint32_t legacy_symbol;
+    if (convert_unicode_to_legacy_symbol(face, symbol, &legacy_symbol))
+        return legacy_symbol;
 
     // GDI only supports cmap format 4 for symbol fonts
     // GDI reads the cmap's first segment startCount and
@@ -286,6 +299,7 @@ static uint32_t convert_unicode_to_ms_symbol(FT_Face face, uint32_t symbol)
             break;
     }
 
+    TT_OS2 *os2 = FT_Get_Sfnt_Table(face, FT_SFNT_OS2);
     bool is_symbol_glyph_set = (offset) || (os2 &&
                     os2->panose[0] == 5 && /* PAN_FAMILY_PICTORIAL */ 
                     (os2->fsSelection & 0xFF) == 0 /* ANSI_CHARSET */);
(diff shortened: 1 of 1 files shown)


libass/ass_font.c
```diff
@@ -271,7 +272,48 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
...
+            FT_UInt gindex;
+            FT_ULong first_char = FT_Get_First_Char(face, &gindex);
+
+            uint32_t offset;
+            switch (first_char & 0xFF00) {
+                case 0xF000:
+                    offset = 0xF000;
+                    break;
+
+                case 0xE000: // EUDC fonts
```
moi15moi: I am still trying to find a EUDC fonts to confirm this part.
That's why the PR is currently in draft.

astiob (approved):

libass/ass_font.c
```diff
@@ -271,7 +272,48 @@ uint32_t ass_font_index_magic(FT_Face face, uint32_t symbol)
...
             }
-            return 0xF000 | symbol;
+
+            // GDI only supports cmap format 4 for symbol fonts
+            // GDI reads the cmap's first segment startCount and
+            // uses it to know what is the first character.
+            // FT_Get_First_Char() is NOT exactly equivalent,
+            // but for well-formatted cmap, it is.
+            FT_UInt gindex;
+            FT_ULong first_char = FT_Get_First_Char(face, &gindex);
```
TheOneric: I’m not sure how relevant this will be in practice, but i’m slightly concerned about this matching GDI hinging on cmaps being “well-formatted” as we’ve seen real-word fonts doing nonsensical things over and over. In particular relevant since the old and most-commonly intended `0xF000` offset now isn’t the default fallback anymore.
Can you elaborate on this? Does “well-formatted” here mean “not breaking spec” or “font authors chose to do the most intuitively sensible thing” or something else?

TheOneric (commented):
I didn’t do any tests or other research to confirm this offset calculation and masking matches GDI, but will choose to just believe it does.
Afaict this differnt base offsetting wasn’t discussed yet in #319.

Code seems fine to me, though given the `FT_ENCODING_MS_SYMBOL` case now accumulated considerable complexity, imho it is best to split its logic out into a new static function for readability.

libass/ass_font.c
moi15moi: I confirmed that EUDC font works like 0x0000.
I used this font to confirm it. [JKR - MORTAL ORNAMENT.zip](https://github.com/user-attachments/files/30917254/JKR.-.MORTAL.ORNAMENT.zip).

I also tried to edit this font and set the panose.bFamilyType to 5 + set the fsSelection to a ainsi charset to be sure that this is indeed correct:
```c
bool is_symbol_glyph_set = (offset) || (os2 &&
                os2->panose[0] == 5 && /* PAN_FAMILY_PICTORIAL */ 
                (os2->fsSelection & 0xFF) == 0 /* ANSI_CHARSET */);
```

libass/ass_font.c
moi15moi: By "well-formatted" I mean spec-compliant.
Specifically, the truetype specification requires that in a format 4 cmap subtable, segments (startCode/endCode pairs) are sorted in ascending order by startCode (see [this](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#format-4-segment-mapping-to-delta-values)).

I tested every symbol fonts in @seproDev (thanks!) database (a bit over 43,000 fonts). There was only one that had a cmap where the offset **category** (0xF000 vs 0xE000 vs 0x0000 vs other) of startCount[0] was different with what FT_Get_First_Char reported. Even in that one case, the cmap had only two segments total and both were unreachable in practice (their code point ranges couldn't actually be hit).

Note that segment ordering in the cmap is a hard spec requirement. It is not a convention. I'd expect this kind of mismatch to stay extremely rare in the wild.

Also, note that we could parse the cmap table manually with something like this, but the problem I found was that I couldn't find how to be sure the cmap subtable I parse correspond to the actual face->charmap selected :/ Also, of course, it's is a bit heavy in cpu cycle compared to ``FT_Get_First_Char`` for basically no benefit.
```c
static inline uint16_t read_u16be(const FT_Byte *buf)
{
    return ((uint16_t) buf[0] << 8) | buf[1];
}

static inline uint32_t read_u32be(const FT_Byte *buf)
{
    return ((uint32_t) buf[0] << 24) | ((uint32_t) buf[1] << 16) |
           ((uint32_t) buf[2] << 8) | buf[3];
}

/**
 * Get the startCode of the first segment of the active cmap format 4
 * subtable, mimicking what GDI reads to determine a symbol font's
 * first character. FT_Get_First_Char() is not exactly equivalent for
 * malformed cmaps.
 */
static bool get_cmap4_first_char(FT_Face face, uint32_t *first_char)
{
    if (FT_Get_CMap_Format(face->charmap) != 4)
        return false;

    FT_ULong len = 0;
    if (FT_Load_Sfnt_Table(face, TTAG_cmap, 0, NULL, &len) != FT_Err_Ok ||
        len < 4)
        return false;

    FT_Byte *buf = malloc(len);
    if (!buf)
        return false;

    bool ok = false;
    if (FT_Load_Sfnt_Table(face, TTAG_cmap, 0, buf, &len) == FT_Err_Ok) {
        uint16_t num_tables = read_u16be(buf + 2);
        for (uint16_t i = 0; i < num_tables; i++) {
            size_t record = 4 + (size_t) i * 8;
            if (record + 8 > len)
                break;

            uint16_t platform_id = read_u16be(buf + record);
            uint16_t encoding_id = read_u16be(buf + record + 2);
            if (platform_id != face->charmap->platform_id ||
                encoding_id != face->charmap->encoding_id)
                continue;

            uint32_t subtable = read_u32be(buf + record + 4);
            if (subtable + 8 > len || read_u16be(buf + subtable) != 4)
                continue;

            uint16_t seg_count_x2 = read_u16be(buf + subtable + 6);
            uint32_t start_code_offset = subtable + 14 + seg_count_x2 + 2;
            if (start_code_offset + 2 > len)
                continue;

            *first_char = read_u16be(buf + start_code_offset);
            ok = true;
            break;
        }
    }

    free(buf);
    return ok;
}
```

If you need a font that will behave differently due to this limitation (startCount[0] vs FT_Get_First_Char), you can use this one that I edited:
[AGMaturaC Edited - first startCount != FT_Get_First_Char.zip](https://github.com/user-attachments/files/31010242/AGMaturaC.Edited.-.first.startCount.FT_Get_First_Char.zip)

astiob:
Just to be clear, I have likewise not done any additional research or testing of my own, and I _trust_ that this pull request is in fact the product of thorough research and testing. Thanks!

libass/ass_font.c
TheOneric: Thank you for the detailed response!
Indeed, if this is a spec requirement, I’m hopeful it actually remains almost always true in practice and we can rely on it. You went the extra mile and even evaluated a large real-world collection, much appreciated! Given the results, i’m not just hopeful but quite optimistic this will suffice and we do not need to result to costly manual font parsing.

TheOneric (approved):

moi15moi:
Oh, I just realized something!
First of all, there are also hebrew special cmap handling needed (like for legacy arabic: https://github.com/libass/libass/pull/927).
Also, legacy hebrew/arabic font can be platformID 3 and encodingID 0 or 1. It doesn't just impact symbol fonts!

I opened an issue on harfbuzz repos to ask if they could add support for legacy hebrew font: https://github.com/harfbuzz/harfbuzz/issues/6159#issuecomment-5287193043

I will just edit this PR to also support legacy fonts where the encodingID is 1.
For hebrew, I will do that in another PR.

moi15moi:
Let me know if there is anything else needed.

TheOneric (approved):
Generally still seems good to me; thanks for the fixes!

Just one question: is the change from `encoding`/`FT_ENCODING_*` to `encoding_id`/`TT_*`, just a cosmetic change to be closer at TrueType spec and MS reference docs, or is it necessary to avoid some mapping issue?

moi15moi:
It does more then that. See how I added a case for `TT_MS_ID_UNICODE_CS`. It allow to properly handle legacy arabic font when it is in a microsoft cmap + encoding_id 1.

I needed to change from `encoding`/`FT_ENCODING_*` to `encoding_id`/`TT_*` because the encoding `FT_ENCODING_UNICODE` would include `TT_MS_ID_UNICODE_CS` + `TT_MS_ID_UCS_4`. GDI only handle legacy arabic font for symbol (encoding_id = 0) + unicode (encoding_id = 1) cmap.

TheOneric:
For reference, here are some comments from IRC after you left the channel

> **\<MrSmile>** Individual commits don't look like clean logical steps, maybe better to squash before merge
> **\<Oneric>** to me the splitting sems fine'ish. First commit fixes the offset for symbol fonts. Second is a cosmetic refactor. Last one deals with legacy arabic mappings outside of symbol charmaps.
> **\<Oneric>** But the commit titles messages can admittedly be better formatted.
> **\<Oneric>** Personally i’d also do the refactor before adding new logic to keep the summed+up diff smaller, but ig this is nitpicking

I was going to just polish the commit title and message myself and then merge given lack of further remarks, but i see you signed the commits. If you want to keep the signature you’ll need to do this yourself. My suggestions are below.
It’s mainly just cosmetics, most notably breaking lines in the message after 72 characters and using lower-case after the colon _(both styles with lower- and uppercase for complete sentences after a colon exist and can be valid, but in libass we usually use lowercase)_.  

```
ass_font: improve symbol font base offset

Previously we always used an 0xF000 offset for symbol-encoded MS fonts.
However, while this is the most common offset, different values are
possible too and it is furthermore affected by some specific,
non-obvious conditions.
For example, the "Baloloc" font did not work with unconditional 0xF000.

The new logic matches our current understanding of GDI behaviour
_except_ for ignoring the system codepage/language.

Note, due to this omission https://github.com/libass/libass/issues/319
remains unresolved.

```

```
cosmetic/ass_font: split out symbol-cmap logic into a new function

It accumulated a non-trivial amount of logic
and is set to gain even more complexity.
```
*(note: more complexity above refers to legacy hebrew mappings + emulating system codepages based on in-file headers. If moved before the other commit, also the additions of this PR. It’s correct either way)*


```
ass_font: handle legacy Arabic charsets on Unicode cmaps

Turns out, GDI performs the legacy charset mapping not only on
symbol cmaps but Microsoft Unicode (but not UCS-4!) cmaps too.

Due to the mapping being specific to Microsoft Unicode cmaps,
we need to switch from pre-normalised FreeType identifiers to
raw TrueType charmap IDs in our main switch-case.
```

moi15moi:
> If you want to keep the signature you’ll need to do this yourself.

I don't care, but I did change the commit title/description with your suggestion.

> most notably breaking lines in the message after 72 characters and using lower-case after the colon

You might wanna write that in a ``CONTRIBUTE`` file in the repos. I did try to have good commit description, but this isn't the type of detail I check, so I missed that. My bad, but thank you for your suggestion. I really appreciate your feedback and the time you took to revise this PR.

TheOneric:
i had to strip the signatures afterall since it turns out commits weren’t rebased on current master yet, but since you said you don’t mind it should be fine `^^

> You might wanna write that in a CONTRIBUTE file in the repos. 

Incidentally, the draft `CONTRIBUTE.md` in another PR already mentions this, so it should be clearly soon (But in absence of signatures it’s also not much effort to just fix it up as part of merging)

PR #949, opened 2026-08-20
drawing: reject overflowed vector coordinates

### LINKED ISSUE #931: Crash: UBSan: signed integer overflow in ass_render.c when processing \p vector drawing commands
### libass version

0.17.4 (git master, commit c425f6d) 

### Is it a regression?

Unknown — not tested on older versions, discovered via fuzzing.

### ASS Sample

[Script Info]                                                                                                                                                                                                                                
ScriptType: v4.00+                                                                                                                                                                                                                         
PlayResX: 1920
PlayResY: 1080                             
                                           
[V4+ Styles]                                                                                                                                                                                                                                 
Style: Default,Arial,40,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,2,2,2,10,10,10,1                                                                                                                                   
                                                                                                                                                                                                                                             
[Events]                                                                                                                                                                                                                                     
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0,0,0,,{\p1}m -2147483648 -2147483648 l 2147483647 2147483647{\p0} 

### Special Fonts

### System Information

OS:  Ubuntu 20.04.6 LTS (x86_64, kernel 4.18)                                                                                                                                                                                                
CPU: 64-bit x86 (Intel), multi-core                                                                                                                                                                                                        
Compiler: clang-10 (libFuzzer), clang-14 / afl-clang-fast (AFL++)
Sanitizers: AddressSanitizer + UndefinedBehaviorSanitizer

### Log and Stacktrace

To reproduce:                                                                                                                                                                                                                                
                                                                                                                                                                                                                                           
1. Build libass with UBSan:                                                                                                                                                                                                                  
   ./configure CC=clang CFLAGS="-g -O1 -fsanitize=address,undefined,fuzzer-no-link" \                                                                                                                                                        
               LDFLAGS="-fsanitize=address,undefined" --disable-shared --enable-static                                                                                                                                                       
   make -j$(nproc)                                                                                                                                                                                                                           
                                                                                                                                                                                                                                             
2. Build fuzzer:                                                                                                                                                                                                                             
   clang -g -O1 -fsanitize=address,undefined,fuzzer \                                                                                                                                                                                        
       target.c -I. .libs/libass.a \                                                                                                                                                                                                         
       -lfreetype -lfontconfig -lfribidi -lharfbuzz -lm -o fuzzer_ubsan                                                                                                                                                                      
                                                                                                                                                                                                                                           
3. Run with the PoC sample above (saved as poc.ass):                                                                                                                                                                                         
   ./fuzzer_ubsan poc.ass                                                                                                                                                                                                                    
                                                                                                                                                                                                                                             
UBSan output:                                                                                                                                                                                                                                
                                                                                                                                                                                                                                             
libass/ass_render.c:1264:37: runtime error:                                                                                                                                                                                                
  signed integer overflow: -2147483648 - 2147483647                                                                                                                                                                                          
  cannot be represented in type 'int'                                                                                                                                                                                                        
                                           
libass/ass_render.c:1265:33: runtime error:                                                                                                                                                                                                  
  signed integer overflow: -2147483648 - 2147483647                                                                                                                                                                                          
  cannot be represented in type 'int'                                                                                                                                                                                                        
                                                                                                                                                                                                                                             
libass/ass_render.c:3026:9: runtime error:                                                                                                                                                                                                   
  6.66667e+36 is outside the range of representable values of type 'int'                                                                                                                                                                     
  SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior libass/ass_render.c:3026:9 

### Additional info

Root cause analysis:                                                                                                                                                                                                                         
                                                                                                                                                                                                                                           
1. Integer overflow at ass_render.c:1264-1265                                                                                                                                                                                                
                                                                                                                                                                                                                                             
   Source code:                                                                                                                                                                                                                              
       v->advance = bbox.x_max - bbox.x_min;   // line 1264                                                                                                                                                                                  
       v->asc     = bbox.y_max - bbox.y_min;   // line 1265                                                                                                                                                                                  
                                           
   The bbox fields (x_max, x_min, y_max, y_min) are declared as int32_t                                                                                                                                                                      
   (see ass_outline.h:39). When a \p vector drawing command specifies                                                                                                                                                                        
   extreme coordinates (e.g., x_max near INT32_MAX and x_min near                                                                                                                                                                            
   INT32_MIN), the subtraction overflows signed 32-bit integer, which                                                                                                                                                                        
   is undefined behavior in C.                                                                                                                                                                                                               
                                                                                                                                                                                                                                             
2. Downstream UB at ass_render.c:3026                                                                                                                                                                                                        
                                                                                                                                                                                                                                             
   Source code:                                                                                                                                                                                                                              
       event_images->left =                                                                                                                                                                                                                  
           (device_x + bbox.x_min) * render_priv->par_scale_x                                                                                                                                                                                
           - text_info->border_x + 0.5;    
                                                                                                                                                                                                                                             
   Due to the cascading effect of the above overflow, the floating-point                                                                                                                                                                     
   intermediate result reaches ~6.67e36, which exceeds INT_MAX (~2.1e9),                                                                                                                                                                     
   causing undefined behavior when implicitly cast to int.                                                                                                                                                                                   
                                                                                                                                                                                                                                             
Suggested fix:                                                                                                                                                                                                                               
                                                                                                                                                                                                                                             
   Clamp bbox coordinates before subtraction, or use int64_t for the                                                                                                                                                                         
   intermediate values:                                                                                                                                                                                                                      
                                                                                                                                                                                                                                             
       v->advance = (int64_t)bbox.x_max - bbox.x_min;  // then clamp to int32                                                                                                                                                              
       v->asc     = (int64_t)bbox.y_max - bbox.y_min;                                                                                                                                                                                        
                                            
   Or add a guard in ass_drawing_parse() to reject coordinates outside                                                                                                                                                                       
   a safe range before they propagate into rendering.                                                                                                                                                                                        
                                                                                                                                                                                                                                             
Discovery method:                                                                                                                                                                                                                            
                                                                                                                                                                                                                                             
   Found via coverage-guided fuzzing (libFuzzer + AFL++ 4.41a) with                                                                                                                                                                          
   UBSan instrumentation on libass 0.17.4. The fuzzer ran for ~12 hours                                                                                                                                                                      
   with a structure-aware custom mutator targeting \p drawing commands.                                                                                                                                                                      
   Two separate UBSan runs both reproduced the same three violations                                                                                                                                                                       
   at the same source locations.

Alb3e3:
Fixes #931.

The parser currently accepts impossible `\\p` drawing coordinates, converts them into D6 integers, and lets the invalid values flow into cached drawing bounds and later render math. Instead of continuing with wrapped/poisoned coordinates, reject them at the parser boundary.

This change:

- range-checks drawing coordinates in D6 space before conversion,
- propagates invalid-coordinate failures through tokenization,
- makes `ass_drawing_parse()` fail when tokenization rejected the drawing,
- adds a focused regression executable for the reported PoC shape.

Local verification:

- `build/test/drawing_overflow`
- `build/test/libass_test /hosttmp/libass-931.png /hosttmp/libass-931.ass 0`

Both were run under a GCC UBSan/float-cast-overflow Docker build; the PoC is now rejected and the render path completes with `0 images blended` and no sanitizer abort.

### DIFF
--- a/libass/ass_drawing.c
+++ b/libass/ass_drawing.c
@@ -58,13 +58,22 @@ static inline bool add_node(ASS_DrawingToken **tail, ASS_TokenType type, ASS_Vec
     return true;
 }
 
+static inline bool get_coord(const char **str, int32_t *coord)
+{
+    double val;
+    if (!mystrtod((char **) str, &val))
+        return false;
+    val *= 64;  // to D6
+    // Out of range coordinates (and NaN) count as invalid numbers
+    if (!(fabs(val) <= OUTLINE_MAX))
+        return false;
+    *coord = ass_lrint(val);
+    return true;
+}
+
 static inline bool get_point(const char **str, ASS_Vector *point)
 {
-    double x, y;
-    if (!mystrtod((char **) str, &x) || !mystrtod((char **) str, &y))
-        return false;
-    *point = (ASS_Vector) {double_to_d6(x), double_to_d6(y)};
-    return true;
+    return get_coord(str, &point->x) && get_coord(str, &point->y);
 }
 
 


MrSmile:
Overflow handling looks overcomplicated (AI generated?) and wrong.
Overflows should be either treated as invalid number (and reuse existing error code path) or clamped, not introduce some special logic. First method is preferred as outline functions already use it.

Alb3e3:
You're right on both counts. Rewritten in `83af97e` (rebased on b2fe9d8): the whole PR is now +13/-4 in `ass_drawing.c`, no status enum, no error propagation.

An out of range coordinate now just fails like an unparsable number, so the existing invalid number path handles it:

```c
static inline bool get_coord(const char **str, int32_t *coord)
{
    double val;
    if (!mystrtod((char **) str, &val))
        return false;
    val *= 64;  // to D6
    // Out of range coordinates (and NaN) count as invalid numbers
    if (!(fabs(val) <= OUTLINE_MAX))
        return false;
    *coord = ass_lrint(val);
    return true;
}

static inline bool get_point(const char **str, ASS_Vector *point)
{
    return get_coord(str, &point->x) && get_coord(str, &point->y);
}
```

`get_point()` keeps its `bool` signature, so every caller is untouched. `get_point()` is the only place a coordinate enters from the string — everything else copies from a token — so bounding it here covers the parser.

I also dropped the separate test executable; libass has no unit test harness for the parser and adding one isn't this PR's job. Happy to add coverage wherever you'd prefer it.

Verification: I couldn't build the full library here (no freetype/fribidi), so I extracted `get_coord`/`get_point` verbatim from the file and linked them against the real `ass_strtod.c`, under `-fsanitize=undefined,float-cast-overflow`:

| input | master | this branch |
| --- | --- | --- |
| `0 0` | accept (0, 0) | accept (0, 0) |
| `1.5 -2.25` | accept (96, -144) | accept (96, -144) |
| `268435455 268435455` | accept **(-64, -64)** | reject |
| `-2147483648 -2147483648` | accept **(0, 0)** | reject |
| `2147483647 2147483647` | accept **(-64, -64)** | reject |
| `1e300 1e300` | accept **(0, 0)** | reject |
| `nan nan` / `inf -inf` | reject | reject |

That is the actual defect behind #931: `OUTLINE_MAX * 64` exceeds `int`, so the values wrapped into small in-range numbers, passed `valid_point()`, and overflowed later in the render math instead of being caught at the parser.

On your question — yes, I use AI assistance, and on the first version I clearly let it run past what the problem needed. Fair hit. Flag anything else that reads that way.

PR #950, opened 2026-08-21
Minor perf fixes 

kasper93:


libass/ass_shaper.c
```diff
@@ -77,9 +77,17 @@ struct ass_shaper {
...
+// fixes (font, face, size), so entries are keyed by glyph index alone and
+// the full metrics cache lookup is skipped for repeated glyphs.
+#define METRICS_MEMO_SIZE 128
+
 struct ass_shaper_metrics_data {
     Cache *metrics_cache;
     FaceSizeMetricsHashKey hash_key;
+    hb_codepoint_t memo_glyph[METRICS_MEMO_SIZE];
+    FT_Glyph_Metrics *memo_metrics[METRICS_MEMO_SIZE];
+    bool memo_valid[METRICS_MEMO_SIZE];
```
rcombs: Could be an array of an anonymous struct instead of 3 separate arrays?

kasper93: Fixed, thanks.

libass/ass_shaper.c
```diff
@@ -77,9 +77,19 @@ struct ass_shaper {
...
+// the full metrics cache lookup is skipped for repeated glyphs.
+#define METRICS_MEMO_SIZE 128
+
 struct ass_shaper_metrics_data {
     Cache *metrics_cache;
     FaceSizeMetricsHashKey hash_key;
+    struct {
+        FT_Glyph_Metrics *metrics;
+        hb_codepoint_t glyph;
+        bool valid;
```
rcombs: Isn't this redundant? Couldn't you just use metrics being (non)null?

kasper93: removed

MrSmile: `NULL` metric in itself can be returned from `ass_cache_get()`, better to use `HB_CODEPOINT_INVALID` for `glyph` instead (also simpler condition).

kasper93: Ok, done.

libass/ass_font.c
```diff
@@ -479,6 +479,8 @@ static int add_face(ASS_FontSelector *fontsel, ASS_Font *font, uint32_t ch)
 
     ret = font->n_faces++;
 
+    memset(font->cmap_cache, 0, sizeof(font->cmap_cache));
```
rcombs, line 482: Gotta also invalidate this after the `FT_Set_Charmap` call up in `ass_font_get_index`.

kasper93: Done. I've also added restore charmap instead of keeping last if there is no better candidate.

libass/ass_font.c
```diff
@@ -690,8 +700,12 @@ int ass_font_get_index(ASS_FontSelector *fontsel, ASS_Font *font,
                         index = FT_Get_Char_Index(face, index);
                     if (index) break;
                 }
-                if (index == 0 && charmap)
+                if (index == 0 && charmap) {
                     FT_Set_Charmap(face, charmap);
+                } else {
```
MrSmile, line 705: I think we don't need `{}` for this case by current libass style.

libass/ass_render.c
```diff
@@ -3188,12 +3188,20 @@ static int fit_rect(Rect *s, Rect *fixed, int *cnt, int dir)
...
-    fixed[*cnt].y0 = s->y0 + shift;
-    fixed[*cnt].y1 = s->y1 + shift;
-    fixed[*cnt].x0 = s->x0;
-    fixed[*cnt].x1 = s->x1;
+    Rect r = { s->x0, s->y0 + shift, s->x1, s->y1 + shift };
+
+    // insert into the y0-sorted position instead of re-sorting the array
+    int lo = 0, hi = *cnt;
+    while (lo < hi) {
+        int mid = (lo + hi) / 2;
```
MrSmile, line 3196: `(lo + (unsigned) hi) / 2` is faster and overflow-free.

MrSmile (commented):
Not sure about dedicated caches for glyphs and metrics: small performance improvement for a specially constructed case hardly justifies increased complexity. Are there real world cases with a noticeable performance gain?

kasper93:
> Not sure about dedicated caches for glyphs and metrics: small performance improvement for a specially constructed case hardly justifies increased complexity. Are there real world cases with a noticeable performance gain?

Fair question. For ordinary Latin or CJK dialogue the gain is within noise, you are right about that. The real case is text containing codepoints no installed font covers, emoji, symbols, CJK under a Latin style. Every such character currently reruns the whole fontselect fallback on every frame, 22k provider queries over a 5 minute here, and the cmap cache reduces that to zero, 0.152 to 0.027 ms per frame. The original motivation was mpv-player/mpv#17963 where these lookups were 15% of a very large render, but the fallback case is the one that matters for normal files.

libass/ass_render.c
kasper93: Done.

libass/ass_font.c
kasper93: Done.

PR #951, opened 2026-08-22
render: cache rendered events

kasper93:
Cache the finished image list of time-invariant events, keyed by the event text, style, margins and a hash of all renderer settings and track state the output depends on. Events using time-dependent tags (\t, \move, \fad, karaoke) or Effect bypass the cache. Cache hits return newly allocated ASS_Image clones referencing the cached bitmaps and holding a reference on the cache entry, so per-frame collision shifting never modifies cached data and eviction cannot free bitmaps still in use. Entries are evicted LRU at a 32 MiB budget.

Re-rendering an unchanged event with 512 KiB of text: 161 -> 0.4 ms.

### DIFF
--- a/libass/ass_cache.c
+++ b/libass/ass_cache.c
@@ -29,6 +29,7 @@
 #include "ass_font.h"
 #include "ass_outline.h"
 #include "ass_cache.h"
+#include "ass_render.h"
 
 // Always enable native-endian mode, since we don't care about cross-platform consistency of the hash
 #define WYHASH_LITTLE_ENDIAN 1
@@ -325,6 +326,38 @@ const CacheDesc glyph_metrics_cache_desc = {
 };
 
 
+// event render cache
+static bool event_render_key_move(void *dst, void *src)
+{
+    EventRenderHashKey *d = dst, *s = src;
+    if (!d)
+        return true;
+
+    *d = *s;
+    d->text.str = ass_copy_string(s->text);
+    return d->text.str;
+}
+
+static void event_render_destruct(void *key, void *value)
+{
+    EventRenderHashKey *k = key;
+    EventRenderHashValue *v = value;
+    ass_frame_unref(v->imgs);
+    free((char *) k->text.str);
+}
+
+size_t ass_event_render_construct(void *key, void *value, void *priv);
+
+const CacheDesc event_render_cache_desc = {
+    .hash_func = event_render_hash,
+    .compare_func = event_render_compare,
+    .key_move_func = event_render_key_move,
+    .construct_func = ass_event_render_construct,
+    .destruct_func = event_render_destruct,
+    .key_size = sizeof(EventRenderHashKey),
+    .value_size = sizeof(EventRenderHashValue)
+};
+
 
 // Cache data
 typedef struct cache_item {
@@ -573,4 +606,9 @@ Cache *ass_bitmap_cache_create(void)
 Cache *ass_composite_cache_create(void)
 {
     return ass_cache_create(&composite_cache_desc);
+}
+
+Cache *ass_event_render_cache_create(void)
+{
+    return ass_cache_create(&event_render_cache_desc);
 }
--- a/libass/ass_cache.h
+++ b/libass/ass_cache.h
@@ -42,6 +42,15 @@ typedef struct {
     int asc, desc;  // ascender/descender
 } OutlineHashValue;
 
+typedef struct {
+    bool valid;             // false: the event renders to nothing
+    bool time_dependent;    // never cached, always rendered fresh
+    ASS_Image *imgs;        // ref-held rendered image list (pre-collision)
+    int top, height, left, width;
+    int detect_collisions;
+    int shift_direction;
+} EventRenderHashValue;
+
 // Create definitions for bitmap, outline and composite hash keys
 #define CREATE_STRUCT_DEFINITIONS
 #include "ass_cache_template.h"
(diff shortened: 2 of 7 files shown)


MrSmile:
That certainly negatively impact performance for parsing-heavy cases. Time-invariance should be detected during the main parsing pass in `ass_parse_tags()` (`ass_event_has_hard_overrides()` has the same problem).

Also I have mixed feelings about using hash values as keys: I consider it acceptable only for cryptographic hash functions and ours is definitely not.

kasper93:
> That certainly negatively impact performance for parsing-heavy cases. Time-invariance should be detected during the main parsing pass in `ass_parse_tags()` (`ass_event_has_hard_overrides()` has the same problem).
> 
> Also I have mixed feelings about using hash values as keys: I consider it acceptable only for cryptographic hash functions and ours is definitely not.

Reworked. Time-invariance is now detected in ass_parse_tags() during the normal parse. The first render of an event goes through the cache construct and time-dependent events become small negative entries that always render fresh, so there is no separate scan over the event text anymore. ass_event_has_hard_overrides() could be converted the same way as a follow-up.

The state hash is gone too. Renderer and track state is snapshotted and compared exactly with memcmp and the snapshot maps to a generation id used in the key, so every key component compares exactly now.

For context, this came out of profiling mpv-player/mpv#17963. Re-rendering unchanged events every frame dominates OSD and static subtitle playback cost. An unchanged 512 KiB event goes from 161 ms to 0.1 ms here and typical dialogue playback loses about 30% of total player CPU. mpv call ass_render_frame() every frame and libass repeated the full parse, shaping and layout for every visible event, including the overwhelmingly common case where the event is unchanged since the last frame. The existing caches cannot cover this because their keys are only known after that work is done.

rcombs:
Out of curiosity, what's the dominant consumer of CPU time in this situation? Shaping?

kasper93:
> Out of curiosity, what's the dominant consumer of CPU time in this situation? Shaping?

I profiled it two ways, an ETW capture of mpv while it re-renders the OSD from mpv-player/mpv#17963, and with timers inside `ass_render_event()` on stock libass for three workloads. Everything below is the share of `ass_render_event()` time.

In mpv, 95% of the process CPU was inside `ass_render_event()`. The renderer already reused the existing bitmap and composite caches for every glyph, so rasterization was essentially free. What remained was everything above the bitmap level.

**The mpv#17963 event (one 1 MiB line, re-rendered unchanged)**

| stage | |
|---|---|
| shaping (`ass_shaper_shape`, HarfBuzz) | 29% |
| `retrieve_glyphs` (outline cache probes) | 11% |
| `render_and_combine_glyphs` (bitmap and composite cache probes, combine bookkeeping) | 10% |
| `ass_shaper_find_runs` | 8% |
| `wrap_lines_smart` | 7% |
| `align_lines` | 7% |
| `parse_events` | 6% |
| `reorder_text` | 5% |
| remaining per glyph passes | 17% |

Inside shaping the trace `hb_shape` 20 samples, of which FreeType charmap lookups from the nominal glyph callback were 6 samples, GPOS kerning 4, advance queries 3. Generic `ass_cache_get` probes across all caches added up to 11%.

**A typical two line dialogue event, re-rendered unchanged per frame**

| stage | |
|---|---|
| `render_and_combine_glyphs` | 45% |
| shaping | 26% |
| `retrieve_glyphs` | 8% |
| `init_render_context` | 4.5% |
| `ass_shaper_find_runs` | 4% |
| `parse_events` | 4% |
| everything else | 8.5% |

A 40 line OSD page as per line events looks the same, 49% combine, 24% shaping, 8.5% retrieve.

So for real sized events the dominant consumer is not shaping but `render_and_combine_glyphs`, which on a repeated render does no rasterization at all. The existing caches hit, but the work to discover that they hit is half the frame. Shaping is the other big part and cannot use those caches at all, HarfBuzz is called on the full run every time.

That's why we can try caching the events, as the repeated render of the same text has constant cost, that can be avoided. Everything above scales with the glyph count.

kasper93:
Also I forgot to mention, currently mpv renders big parts of stats as a single event, this can be split into but without event caching, there is no really a win doing that currently. However, with caching it gives nice boost, because stats are mostly a static text with some lines changing, but most stay static.

clsid2:
I assume the OSD constantly generates new events? Suppose there was a new API to adjust the start/end time of an existing event. That could be used for static parts of the OSD. You basically constantly adjust the lifetime of an event while OSD is active. And when toggling OSD on or after a seek you send a fresh set of events.

Then caching could be much simpler. No need for hashes or comparing contents of events. Just pair each static event with its rendered bitmap and re-use that. Clear cache when any settings change. Clear cached bitmaps when render time goes outside their time range.

kasper93:
Yes, mpv's OSD throws away all events and adds them again on every update. Subtitle tracks are different, their events live for the whole file.

The problem with caching by event identity is that libass can't tell whether an event changed. The fields are public and applications do change Text, Style or margins in place. So the cache would need a promise from every application to never touch an event after it was rendered, and if someone does, they get stale bitmaps with nothing to warn them. Comparing the content avoids all of that, it just works for every existing application. It would also only help applications that are rewritten to keep their events alive and adjust them. For subtitle tracks, which is the common case, the events are already stable, so both approaches do exactly the same work there.

The cost of the content check is very small, one strlen, one hash and one memcmp of the event text per frame, far below a microsecond, while a render of a typical line is tens of microseconds. The hash is only used to pick the bucket, the actual comparison is exact. If the lookup ever shows up in a profile, a stable event could remember its last cache entry in ASS_RenderPriv and skip the lookup entirely. That doesn't need a new API, so it can be a follow-up rather than a change to how the cache is keyed.

PR #952, opened 2026-08-22
ci/gha/vm: disable per-step file sync

kasper93:
The sync_files input only controls the synchronization performed by the VM start step itself. The cpa.sh shell does not consult it and defaults to syncing both directions around every step. Pass --sync-files none to the shell to get the intended behavior of a single initial sync.

### DIFF
--- a/.github/workflows/vm.yml
+++ b/.github/workflows/vm.yml
@@ -80,7 +80,8 @@ jobs:
 
     defaults:
       run:
-        shell: cpa.sh {0}
+        # Avoid syncing after each step, we only need an initial sync for repos
+        shell: cpa.sh {0} --sync-files none
 
     steps:
       # Available core count changes on runner type and over time, but GHA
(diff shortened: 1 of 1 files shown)


kasper93:
@TheOneric: I recently ported the job in mpv to new syntax, and noticed that the sync is backward in libass.

TheOneric:
hmm, i was assuming the start step sets all the defaults which will automatically be used by `cpa.sh`, since it appears to work like this for forwarded environment vars. Or is their value too only copied once on VM boot, and then never refreshed to match potentially new host values?

I see an explicit note about sync in the output of the removed syncing step, but not in any other steps using `cpa.sh`. However, such a note also is absent from mpv’s job logs eventhough they should sync.

Anyway, I assume you did your testing for mpv’s jobs and with an explicit `cpa.sh` arg it definitely should work either way.  
Thanks!

PR #953, opened 2026-08-22
Support legacy hebrew charmap

moi15moi:
Turn out that there is also legacy hebrew charmap.
Microsoft still haven't documented this special charmap, but I requested it here (https://github.com/MicrosoftDocs/typography-issues/issues/814#issuecomment-5286919064).
Note that Harfbuzz still doesn't support properly this special charmap, but they have an opened PR for it (https://github.com/harfbuzz/harfbuzz/pull/6161).


MrSmile:
Hmm... x4 hash map size. Maybe we should try different hash functions.

PR #955, opened 2026-09-12
render: avoid UB on extreme numeric input (#947)

### LINKED ISSUE #947: Crash: Signed integer subtraction overflow in ass_outline_construct (and float-to-int overflow in ass_render_event) on extreme numeric subtitle input -> UBSAN SIGILL
### libass version

libass 0.17.5 (commit 3087d2b2ffda76602a17f9b09d25cb8addc8d313, 2026-08-10, master).

### Is it a regression?

Unknown — I did not test older releases. The trigger uses only numeric extremes that libass has always accepted at parse time; the crash occurs in the renderer's integer arithmetic, so I suspect it is not a recent regression, but I have not bisected.

## ASS Sample

Two minimal, fully functional ASS files reproduce the crash deterministically. Both trigger UBSAN undefined behaviour in the renderer; with `-fno-sanitize-recover=all` (as used by OSS-Fuzz / fuzzing builds) the process dies with SIGILL.

### Sample A — triggers `ass_outline_construct` signed integer subtraction overflow (8 of 13 crashes)

```ini
[Script Info]
ScriptType: v4.00+
PlayResX: 2147483647
PlayResY: 1
LayoutResX: -2147483648
Timer: 1e308
WrapStyle: -1
ScaledBorderAndShadow: yes

[V4+ Styles]
Style: X, Arial, 2147483647, &H00FFFFFF, &H000000FF, &H00000000, &H00000000, 0, 0, 0, 0, 100, 100, 0, 0, -1, 2, 1, 0, 2147483647, 2147483647, 2147483647, -1

[Events]
Dialogue: 0,0:00:01.00,0:00:04.00,Default,,2147483647,2147483647,2147483647,,extreme numbers
```

Key ingredients: `PlayResX: 2147483647`, `LayoutResX: -2147483648`, `Fontsize: 2147483647`, `MarginL/R/V: 2147483647`. The extreme style values flow into the outline bounding box and make `bbox.x_max - bbox.x_min` / `bbox.y_max - bbox.y_min` overflow a signed 32-bit int.

### Sample B — triggers `ass_render_event` float-to-int conversion overflow (5 of 13 crashes)

```ini
[Script Info]
ScriptType: v4.00+
PlayResX: 854
PlayResY: 480

[V4+ Styles]
Style: Default, Arial, 20, &H00FFFFFF, &H000000FF, &H00000000, &H00000000, 0, 0, 0, 0, 100, 100, 0, 0, 1, 2, 1, 2, 10, 10, 10, 1

[Events]
Dialogue: 0,0:00:01.00,0:00:03.00,Default,,0,0,0,,{\frx(1000000)}{\fry(1000000)}{\frz(1000000)}rot 1e6
Dialogue: 0,0:00:01.00,0:00:03.00,Default,,0,0,0,,{\frx(2147483647)}{\fry(-2147483648)}{\frz(999999999)}rot extremes
Dialogue: 0,0:00:01.00,0:00:03.00,Default,,0,0,0,,{\org(5000,5000)}{\frz(1000000000)}{\frx(1000000000)}org big rot
Dialogue: 0,0:00:01.00,0:00:03.00,Default,,0,0,0,,{\fax(1000000000)}{\fay(1000000000)}{\fr(1000000000)}shear 1e9
Dialogue: 0,0:00:01.00,0:00:03.00,Default,,0,0,0,,{\frx(-1000000000)}{\fry(1000000000)}neg rot
```

Key ingredients: extreme `\frx/\fry/\frz/\fax/\fay/\fr` override values (up to `2147483647` / `1000000000`). These are parsed as doubles and later converted back to integers in `ass_render_event` (via `ass_lrint`/`lround`), overflowing the integer conversion.

### Special Fonts

### System Information

OS: Debian (x86-64 Linux)
CPU: 64-bit x86-64

### Log and Stacktrace

With -fsanitize=address,undefined -fno-sanitize-recover=all, UBSAN traps as a ud1 instruction (SIGILL); ASAN reports DEADLYSIGNAL ILL. Representative stack for Sample A (RC-A):

#0 __clang_trap_msg$UndefinedBehaviorSanitizer$Integer subtraction overflowed /src/libass/ass_render.c
#1 ass_outline_construct /src/libass/ass_render.c
#2 ass_cache_get /src/libass/ass_cache.c:424:18
#3 get_outline_glyph /src/libass/ass_render.c:1184:15
#4 retrieve_glyphs /src/libass/ass_render.c:2188:13
#5 ass_render_event /src/libass/ass_render.c:2838:5
#6 ass_render_frame /src/libass/ass_render.c:3392:17
#7 consume_track /src/harness.c:128:23
#8 LLVMFuzzerTestOneInput /src/harness.c:168:13
Affected code (Sample A), libass/ass_render.c, ass_outline_construct, OUTLINE_DRAWING branch:

v->advance = bbox.x_max - bbox.x_min;   /* signed int overflow with extreme input */
v->asc     = bbox.y_max - bbox.y_min;
Affected code (Sample B), ass_render_event: ass_lrint/lround on extreme float override values (e.g. \frx, \frz, \pos) overflows the integer conversion.

Reproduction:

# build the harness (ass_read_memory + ass_render_frame) with clang ASan+UBSan
clang -O1 -g -fsanitize=fuzzer,address,undefined -fno-sanitize-recover=all \
      harness.c -o libass_read_memory_fuzzer $(pkg-config --cflags --libs libass)

# save Sample A as sample_a.ass, then:
./libass_read_memory_fuzzer sample_a.ass
# -> exit 132 (SIGILL) from UBSAN trap
Or with ffmpeg (per report_agent suggestion):

ffmpeg -f lavfi -i 'color=black:s=854x480' -vf 'ass=sample_a.ass' -frames:v 1 out.png
Both reproduction paths crash deterministically (13/13 crash inputs, 2 unique root causes, all in libass/ass_render.c). The crash is target code, not harness misuse: ass_read_memory + ass_render_frame are the documented public API sequence, and it reproduces under three different FONTCONFIG configurations.

Image

### Additional info

_No response_

1820893135-pixel:
Fixes the two undefined-behaviour crashes reported in #947, plus the further instances of the same two classes that the same inputs reach once the first ones are fixed.

**Input:** malformed `.ass`/`.ssa` subtitles with extreme numeric values — huge `PlayResX`/`PlayResY`/`LayoutResX`, `Fontsize`, `MarginL/R/V`, or `\frx/\fry/\frz/\fax/\fay/\fr` override parameters. Under `-fsanitize=address,undefined -fno-sanitize-recover=all` (the OSS-Fuzz configuration) the renderer aborts with SIGILL; in a normal build the arithmetic silently wraps to garbage coordinates and sizes.

Following @rcombs' review, both classes now keep the practical x86/VSFilter result rather than saturating:

## float → int conversion of out-of-range values

- `ass_render_event()`: the event image rectangle (`top`, `height`, `left`, `width`) was built by implicitly converting `double` expressions to `int`; extreme `device_x/y`, line ascents/descents and border values fall outside `int` range.
- `apply_baseline_shear()`: accumulated a `double` into an `int32_t` shear.

These now go through the existing `dtoi32()` helper, which returns `INT32_MIN` for non-representable values, i.e. the same result as `cvttsd2si` on x86. `dtoi32()` was moved from `ass_parse.c` into `ass_utils.h` so it can be shared.

## signed integer overflow in 32-bit arithmetic

- `ass_outline_construct()`: `bbox.x_max - bbox.x_min` / `y_max - y_min` on `int32_t` extrema.
- `preliminary_layout()`, `reorder_text()`, `align_lines()`, `apply_baseline_shear()`, `calculate_rotation_params()`: glyph advance / cluster-advance / offset / shift accumulations in `ASS_Vector` (`int32_t`).
- `wrap_lines_naive()` / `wrap_lines_rebalance()`: `bbox.x_max + pos.x` style sums.
- `ass_render_event()`: `2 * text_info->border_x`.
- collision handling: `imgs[i].top + imgs[i].height` / `left + width`.

These now compute in 64-bit and reduce back to `int32_t` by wrapping via a new `wrap_i32()` helper (`(int32_t)(uint32_t)val`), reproducing what the equivalent 32-bit arithmetic does on x86 — the same approach as `d1140ac` for the shift operands.

All of these values are coordinates or collision-rectangle bounds; none is used directly as an allocation size or array index, so wrapping cannot introduce out-of-bounds accesses (the bitmap allocator also has its own bound check).

## Verification

Built at `3087d2b` with clang `-fsanitize=address,undefined -fno-sanitize-recover=all` and the harness from the report (`ass_read_memory` + `ass_render_frame`). Before: 13/13 crash inputs abort (2 reported root causes, 6+ distinct source locations). After: all crash inputs, plus a 1715-file corpus/seed sweep, run clean with **zero** UBSan reports.

To confirm the wraparound doesn't change rendered output, I also built the **unpatched** revision with *recoverable* UBSan (so it reports and then continues with native x86 wraparound) and diffed it against this patch: a harness hashing every bitmap's full contents plus all image geometry (`w/h/stride/dst_x/dst_y/color`) over the same 1715 inputs gives **byte-identical** results on every input.

Fixes #947.


rcombs:
I guess these cases are unlikely to come up organically in real input, but I'll ask nevertheless: is saturating actually appropriate here? In most of these overflow cases, we try to match vsfilter's behavior, which is usually wraparound (assuming the wraparound doesn't lead to, say, out-of-bounds accesses). I'd hesitate to introduce a saturation helper if we don't have good reason to prefer that over wrap here.

1820893135-pixel:
Good point, and you're right — I've reworked it to wrap instead of saturate.

I went with saturation initially only because `dtoi32()` was already in the tree doing exactly that for double→int (INT32_MIN, matching `cvttsd2si`), so it looked like the established convention. But that's specific to the float→int case; for plain 32-bit integer overflow the existing precedent (e.g. `d1140ac` for the shift operands) is to reproduce the practical x86/VSFilter result, which is wraparound. So I've switched all the integer cases to a `wrap_i32(int64_t)` helper (`(int32_t)(uint32_t)val`) and kept `dtoi32()` only for the float→int sites, where it already is the wrap-equivalent.

On your OOB caveat: I checked each site, and all of them are coordinates or collision-rectangle bounds — none is used directly as an allocation size or array index (the bitmap path has its own bound check in `ass_alloc_bitmap` anyway), so wrapping is safe here.

To be sure this doesn't change any actual output, I built the unpatched revision with *recoverable* UBSan (so it reports and then continues with native x86 wraparound) and diffed it against the patched build, hashing every bitmap's full contents plus geometry (`w/h/stride/dst_x/dst_y/color`): byte-identical on all 1715 corpus/seed inputs, and no UBSan reports from the patched build. So the only thing that changes is the UB.

